PR #60, opened 2013-10-23
Remove `@Nullable` annotation from `escapeJavaType` method

dpursehouse:
Summary:
The `type` parameter is annotated with `@Nullable`, but the parameter
should not actually be null because it is being dereferenced within
the method.

Remove the annotation.

Test Plan:
Run some test builds.

Change-Id: I0605bea537a60525921c4d9f5d162e8702d2ed0c


authorizatio:
Gitback

Issue #69, opened 2013-11-22
Buck website makes unnecessary apologies about ABI fingerprinting.


cakoose:
Under "What Makes Buck so Fast?" -> "A Java rule computes its ABI when it is built".

JLS ABI compatibility is to make sure your ".class" files will still work with a newer version of some library as long as that library remains ABI compatible.  That's useful, but that's not what you want for a build tool.

If you add a method to some class in a library, that could cause Java builds to fail, for example, if the name conflicts with a method in a subclass.  If Buck only checked JLS ABI compatibility, it could produce incorrect output.  It would say "nothing to do" when in fact it should be showing you a compile error because your build is broken -- doing a clean build would yield a compile error.

Very few API changes can guarantee no effect on code that uses it.  Even adding a new class can cause name conflicts if someone is doing wildcard import.

What Buck is doing might be simply be termed "checking for visible API changes".  I think it's worth updating the docs to make this clearer.

PR #115, opened 2014-05-19
docs: Fix ABI/API confusion on "what makes buck so fast" page.

cakoose:
https://github.com/facebook/buck/issues/69

Also added missing `<p>` tags.  The paragraphs without a leading`<p>`
were smooshed together.

Visually checked result with "soyweb-local.sh" server and Chrome.

### DIFF
--- a/docs/concept/what_makes_buck_so_fast.soy
+++ b/docs/concept/what_makes_buck_so_fast.soy
@@ -10,6 +10,8 @@
 
 <h2>A build rule knows all of the inputs that can affect its output</h2>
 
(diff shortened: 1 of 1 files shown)


cakoose:
facebook-github-bot might be malfunctioning: https://www.dropbox.com/s/derhi3j4l2d8vqf/Screenshot%202016-03-23%2017.01.30.png

sdwilsh:
I think it's my fault.  I've been playing with the code backing it, and 
I think I made it too aggressive...

PR #121, opened 2014-05-28
Introduce buck_java_test() macro.

bolinfest:
The goal is to help standardize the structure of the test
rules for the Buck project itself. This makes it easier to
run integration vs. unit tests:

```
buck test --all --include unit
buck test --all --include integration
```

We could also experiment with further parallelizing these tests
in the future. For example, it may make sense to have one `java_test()`
rule for each integration test, but not for each unit test.

This diff is an RFC, so it only converts two of the `BUCK` files
over to `buck_java_test()`.


bolinfest:
I'll abandon this because this is so far behind master that it seems like too much of a pain to rebase.

PR #205, opened 2014-10-14
Error in aapt_package when prebuilt AAR does not include any `res` files

jhansche:
If a prebuild AAR file contains an *empty* `res/` directory, the
directory does not get created in the `__unpack_*#aar_unzip__/res/`
location.  This causes the `aapt` tool to fail with error:

    ERROR: resource directory 'buck-out/bin/__unpack_<prebuilt_aar_name>#aar_unzip__/res' does not exist

In the `AndroidPrebuiltAarGraphEnhancer` class, there is a step
that accounts for a similar error on the `assets` directory:

    steps.add(new MkdirStep(getAssetsDirectory()));

Adding a similar step for the `res` directory resolves the error.


sdwilsh:
Closing this in favor of fixing #234

PR #217, opened 2014-11-10
add sub project symbols to main project

zhchang:
this change is to append symbol files generated by resource-dependencies to main-resource symbol file, so that main project could use resources define in sub projects (android library project)


natthu:
We decided to leave out symbols from transitive dependencies in the resulting R.txt for performance.

sdwilsh:
You can still use them in your main project; you just have to declare all of your dependencies and your dependencies dependencies.

PR #224, opened 2014-12-16
Improve the messages printed when buck self-updates

dpursehouse:
Summary:

(1.) When the .buckversion file specifies a revision that is not present
in the current repository, print a message stating that Buck is fetching
updates.

(2.) When the git fetch fails, only display a "failed" message.  Move the
bit about how the user can disable the update into the informational
message that was already displayed as described in (1.).

(3.) If the revision specified in .buckversion does not exist on the
remote, either because it's plain wrong or because the revision is on a
remote that is not configured in the local repository, the git checkout
fails. Catch the exception and display a message rather than a Python
Traceback.

Test Plan:

Set .buckversion to a revision that does not exist. Run `buck clean` and
observe that it prints the message in (1.), then runs git fetch, and then
fails to check out the revision and fails with an error message.

### DIFF
--- a/programs/buck_repo.py
+++ b/programs/buck_repo.py
@@ -116,6 +116,13 @@ class BuckRepo(BuckTool):
     def _checkout_and_clean(self, revision, branch):
         with Tracing('BuckRepo._checkout_and_clean'):
             if not self._revision_exists(revision):
+                print(textwrap.dedent("""\
+                    Required revision {0} is not
+                    available in the local repository.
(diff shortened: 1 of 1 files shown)


dpursehouse:
I can't see any way to restart the travis build.

sdwilsh:
Don't worry about it; I verified it locally and things look OK.  This should get merged soon.

sdwilsh:
Thanks for the pull request!

PR #264, opened 2015-03-02
Add ability to use a .src.zip as a python_library source.

mikekap:
The intention is to allow genrules to generate > 1 file worth of code (or some directory structure) that can be used in a python_library. For example, you could write a small wrapper around genrule to take a PyPI package name & version, download it & repackage it to include certain directories as a python_library (similar to, but slightly more involved than maven_jar in bucklets).

### DIFF
--- a/src/com/facebook/buck/cli/BUCK
+++ b/src/com/facebook/buck/cli/BUCK
@@ -55,6 +55,7 @@ java_library(
     '//src/com/facebook/buck/parser:config',
     '//src/com/facebook/buck/parser:parser',
     '//src/com/facebook/buck/python:config',
+    '//src/com/facebook/buck/python:rules',
     '//src/com/facebook/buck/rules:types',
     '//src/com/facebook/buck/thrift:rules',
     '//test/com/facebook/buck/...',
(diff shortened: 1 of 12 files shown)


k21:
Thanks for your contribution! Can you please share your use-case for this feature? I would like to understand in what context it will be used.

k21:
I have found a slight issue with your tests. We would like to be able to run Buck's tests with older Python versions (2.6). Your test code uses `argparse` and `with` on a zipfile, which are not supported in that Python version. Can you please fix the new test to work with Python 2.6?

mikekap:
Fix py2.6 issues & added a description that should give you an idea of where I'm going with this.

k21:
> added a description that should give you an idea of where I'm going with this.

I might be looking in the wrong places, but where did you add the description? I don't see it anywhere.

mikekap:
Oh sorry - I added it to the pull request. It's a bit weird that github doesn't e-mail when that happens. The main purpose is to use genrule() to generate > 1 python file.

sdwilsh:
Thanks for your contribution!

PR #273, opened 2015-03-13
Prebuilt libraries should default to the native extension for the platform

mikekap:
Also allows you to do something like (albeit ugly, but working):

```python
import subprocess
import collections

# A small amount of magic to get cxx_python_extension working.
python_ldflags = subprocess.check_output('python-config --ldflags', shell=True).split()
prebuilt_cxx_library(
    name='pylib',
    provided=True,
    lib_name='python2.7',
    lib_dir=[v[len('-L'):] for v in python_ldflags if v.startswith('-L')][0],
    include_dirs=[v[len('-I'):] for v in
                  subprocess.check_output('python-config --includes', shell=True).split()],
    linker_flags=python_ldflags,
    visibility=['PUBLIC']
)
```


sdwilsh:
Thanks for your contribution!

PR #276, opened 2015-03-18
Update indicated location of maven_repo

tejohnso:
Current suggestion to use tools -> maven_repo is not correct.

see https://github.com/facebook/buck/blob/master/src/com/facebook/buck/rules/KnownBuildRuleTypes.java#L375


Coneko:
Thanks!

sdwilsh:
Thanks for your contribution!

PR #278, opened 2015-03-31
Add zip_safe flag to all python targets & set it to False on cxx_python_extension

mikekap:
Native libraries are not zipsafe (see https://docs.python.org/2/library/zipimport.html "only files .py and .py[co] are available for import. ZIP import of dynamic modules (.pyd, .so) is disallowed"). This change introduces the zip_safe flag on all python targets that can potentially be not-zip-safe, and marks cxx_python_extension as definitely-not zip safe.

I made it a tristate in PythonPackageComponents mostly to keep the default from being hardcoded in every python rule.

### DIFF
--- a/src/com/facebook/buck/cxx/CxxLibrary.java
+++ b/src/com/facebook/buck/cxx/CxxLibrary.java
@@ -195,7 +195,8 @@ public class CxxLibrary extends AbstractCxxLibrary {
     return PythonPackageComponents.of(
         /* modules */ ImmutableMap.<Path, SourcePath>of(),
         /* resources */ ImmutableMap.<Path, SourcePath>of(),
-        /* nativeLibraries */ libs.build());
+        /* nativeLibraries */ libs.build(),
+        /* zipSafe */ Optional.<Boolean>absent());
   }
 
   @Override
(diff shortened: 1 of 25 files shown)


sdwilsh:
I totally missed the fact that you updated this since github didn't send mail.  I'm running this through our own internal CI now and it should go out Wednesday next week assuming it all checks out :)

mikekap:
All done - separately, is there any interest in adding the ability to bundle prebuilt eggs into pexs? (similar - in spirit - to https://github.com/mikekap/buck/commit/56c3507511d1c350ff867b93404c9701b2e79486)

sdwilsh:
Yeah, we'd take something like that, but I think we'd want to go with `prebuilt_python_library` or to match what we do with other platforms, and then just pull it from the list of `deps` instead of adding a new `requirements`.  Feel free to send a pull request our way (I'll ask for docs to be added too if you do add that).

sdwilsh:
Note that your change introduced some lint errors (identified with `ant lint`).  Travis keeps dying because of the NDK, so I'm going to remove that, sadly.

I'll fix those before merging in your change.

mikekap:
Thanks and sorry about that (I didn't know there was an ant lint :X)

sdwilsh:
No problem.  I'll be adding it to Travis tomorrow so you won't have to know until it fails :)

sdwilsh:
Thanks for your contribution!

sdwilsh:
Thanks again for your fix!  This made it into our weekly update post: https://www.facebook.com/buckbuildtool/posts/813592978710413

PR #284, opened 2015-04-21
Support pyrun when querying available Python version

mread:
We bundle pyrun in all environments where we use buck. This pull request ensures that pyrun is recognised as valid python.


sdwilsh:
I apparently missed the update here.  I'll try to get this merged in for our sync on Wednesday.

sdwilsh:
Thanks again!

PR #285, opened 2015-04-21
Pass system property to test runner containing the base dir of the test module

mread:
This helps with legacy code migrations that open files in locations relative to the root of the test module.


sdwilsh:
Thanks for the fix!

PR #292, opened 2015-04-29
Remove second definition of buf

gsaraf:
[exec] third-party/nailgun/nailgun-client/ng.c:115:17: error: redefinition of 'char buf [2048]'
     [exec]  char buf[BUFSIZE];
     [exec]                  ^
     [exec] third-party/nailgun/nailgun-client/ng.c:58:7: note: 'char buf [2048]' previously declared here
     [exec]   char buf[BUFSIZE];
     [exec]        ^

### DIFF
--- a/third-party/nailgun/nailgun-client/ng.c
+++ b/third-party/nailgun/nailgun-client/ng.c
@@ -54,8 +54,6 @@
 	#define FILE_SEPARATOR '/'
 	typedef int HANDLE;
 	typedef unsigned int SOCKET;
-	/* buffer used for reading an writing chunk data */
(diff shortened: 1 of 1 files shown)


shs96c:
Thank you for the PR!

nailgun is a dependency we've pulled into the tree. It would be a good idea to also send this patch upstream: https://github.com/martylamb/nailgun

Will merge this in, since it appears to be trivial.

gsaraf:
I have actually already submitted a pull request for nailgun, and it was accepted. I will update the internal nailgun source, so this pull request is no longer relevant.

PR #296, opened 2015-05-01
Add a ruletypes command that prints the list of known rule types

luser:
I got frustrated with the fact that the docs don't describe everything in the source, so I hacked this up so I could more easily see what rule types were available.

@sdwilsh suggests that this would be better implemented as `buck audit rules`, and also that it needs a test.

### DIFF
--- a/src/com/facebook/buck/cli/Command.java
+++ b/src/com/facebook/buck/cli/Command.java
@@ -59,6 +59,10 @@ public enum Command {
       "generates a default project directory",
       QuickstartCommand.class,
       false),
+  RULETYPES(
+      "list built-in rule types",
+      RuleTypesCommand.class,
+      true),
   RUN(
       "runs a target as a command",
       RunCommand.class,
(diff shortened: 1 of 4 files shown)


luser:
`buck audit rules` already exists, so maybe `buck audit known-build-rules`.

luser:
(The test failures in the Travis build are because I didn't update the unit tests that check the usage output.)

sdwilsh:
You'll need to rebase your changes for us to merge this.  I commented on your change about a few minor things too.  It'd also be nice to add docs for this command (under docs/command).

@shs96c, do you have any opinions on what to call this command?  I think it's useful.

shs96c:
buck audit rule-types

sdwilsh:
@luser - any chance you might be up for rebasing and addressing the comment from @shs96c?

luser:
Probably not right now, although I haven't forgotten about it. If you would like to do it yourself I won't be offended.

nayrmi:
Closing the pull request if this is abandoned for now. We'll happily take a PR in the future, though.

Thanks!

PR #301, opened 2015-05-06
Test and fix copy_resources issue - facebook#286

mread:
As discussed on facebook#286, if a resource path contained a filename that matched the name of the resource root directory then the wrong symlink target would be constructed causing the copy_resources step to fail.


Coneko:
Thanks!

I've merged it internally, so it should go out soon.

sdwilsh:
Thanks for the pull request!

PR #307, opened 2015-05-16
force usage of python2

tgummerer:
Summary:
Buck fails to run under python3.  Test the version of the default python
interpreter before launching it, and use the python2 binary if the
default version on the system is python3.

Test plan: buck test, run buck on system with python3 as default

### DIFF
--- a/bin/buck
+++ b/bin/buck
@@ -9,4 +9,9 @@ while [ -h "$BUCK_PATH" ]; do
 done
 BUCK_DIR=$(dirname "$BUCK_PATH")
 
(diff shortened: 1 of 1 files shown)


sdwilsh:
Thanks for your contribution.  After some internal discussion, we modified the patch a bit to not invoke `python`.

tgummerer:
Thanks, that looks much nicer!

sdwilsh:
Thanks again for your fix!  This made it into our weekly update post: https://www.facebook.com/buckbuildtool/posts/813592978710413

Issue #262, opened 2015-02-26
ng client can heartbeat on a closed socket


benhyland:
See #164 for symptoms.
This is not fixed on latest trunk (so may be a regression?), and occurs intermittently in our build when using buckd.
Workarounds are to repeat the build (sometimes several times) or to disable buckd.
I believe (from some quick strace'ing) that the client https://github.com/facebook/buck/blob/master/third-party/nailgun/nailgun-client/ng.c can heartbeat in the stream forwarding loop after the server has closed the socket.
Does this sound reasonable? I don't know much about nailgun.

Coneko:
Is the error reproducible in the same way as #164? "simply by running buck over and over without making any changes"? Or is there a different way of reproducing it?

benhyland:
Different team members seem to have somewhat different experiences, but in my case I can repro fairly reliably by 
a) first running 'buck test --no-results-cache' with a filter that selects our usual complement of 'fast' unit tests, and then 
b) running 'buck test' repeatedly with no changes will repro for at least the first few builds.

If instead in b) I run 'strace -f -e connect,sendto,execve buck test' I can reliably see buck get an EPIPE in response to a sendto which was sending "\0\0\0\0H" - a heartbeat. Because (I think) the signal handler and socket are not configured from the default, we also get a SIGPIPE raised.

benhyland:
Since pulling in changes from the upstream nailgun client, natthu's changes (e.g. https://github.com/facebook/buck/commit/371f502640527492eda2046f0dcc4afb2d5216bf#diff-56a419ae8baaa423da656f7379077866) have been lost, and the above-referenced patch no longer applies cleanly (by the way, we have had this change on our fork for a couple weeks and it does fix the issue I described above).
You should probably decide how you wish to manage buck-local changes to the nailgun client to avoid losing work in future - perhaps make it clear that patches should be applied upstream and stop accepting pull requests in this area, or else maintain the ability to merge local changes with upstream changes when you choose to resync.

sdwilsh:
(╯°□°）╯︵ ┻━┻

Yeah, we are moving to a world where we get things upstreamed first.  @bgertzfield and I both missed that in the review of picking in upstream.

benhyland:
I don't think this is fixed until https://github.com/martylamb/nailgun/pull/57 is merged upstream and buck's copy of the client updated.

Coneko:
Thanks for pointing that out. Would martylamb/nailgun#49 fix it as well?

[...]

PR #310, opened 2015-05-19
don't SIGPIPE, and don't EPIPE on sending heartbeats

benhyland:
Fixes https://github.com/facebook/buck/issues/262
We've had this change on our fork for a couple weeks and can no longer reproduce the issue.


benhyland:
Sorry for the noise, didn't notice the merge conflicts

sdwilsh:
Note that we just (last week) updated our nailgun to what upstream has.  If your fix was there, we have it, and if it wasn't, you should get it there first :)
________________________________

PR #311, opened 2015-05-21
Add prebuilt python library.

mikekap:
This adds the ability to include eggs & wheels (well not yet; needs a pex upgrade) in pexs generated by buck.
This can be extremely useful when trying to include third-party libraries that don't easily lend themselves to buck-ification (e.g. gevent).

The syntax looks like

```python
prebuilt_python_library(
    name='gevent'
    srcs=['gevent-1.0.1-py2.7.egg'],
)

python_binary(
    name='my_module'
    main='main.py',
    deps=[
        ':gevent'
    ],
)
```


sdwilsh:
Nice!  One thing that I see missing from this is documentation.  Can you follow the `README.md` under `docs/` to update those and also add docs for the new rule?

sdwilsh:
Alright, I've had @shs96c and @andrewjcg look this over, as well as myself now.  I've made sure all of our comments are on your commit.  If you address all that, this should be good to go :)

mikekap:
PTAL - I believe I addressed all the comments (those I didn't reply to are just "Done")

mikekap:
Ah shoot you seem to have quite an old version of python/virtualenv. Would replacing setuptools with `distutils.core` work?

sdwilsh:
I think it's whatever comes with centos6, but if we need to use a library, we should be packaging it in with Buck.  `distutils.core` looks like it is shipped with python, so that would be okay too.

mikekap:
I think buck actually includes setuptools, but I doubt you could use setuptools from within a pex. That test makes an egg via a genrule, which is why there's a dependency on distutils.core. I can instead check in the egg itself, but this seems easier to inspect in case of failure. I think using distutils.core should work fine here.

Would you like me to rebase along with the changes you made?

sdwilsh:
That or just add a commit to the PR to make the change to use `distutils.core` and I'll patch it in.

mikekap:
Unfortunately I couldn't get distutils.core to make eggs (seems it was only "built-in" to setuptools). Adding setuptools just for a test seemed wasteful (and I doubt I could get it to work from within a pex). Instead I just checked in the egg - it's cross-platform (since there's no binary modules), and it's pretty unlikely to change.

Let me know if you'd like me to check in setuptools instead.

mikekap:
Thanks and sorry for how hard this seemingly trivial change was.

sdwilsh:
It's software; most things aren't trivial.  Thanks for hanging in there and helping to get it in!

sdwilsh:
All merged in and the docs are live: http://buckbuild.com/rule/prebuilt_python_library.html

Thank you so much!

sdwilsh:
Looking at the commit, it looks like the email address you used to commit it isn't attached to your github account (wanted to make sure you were aware of it so you can get credit for it).

sdwilsh:
Also, this made our weekly update notes.  Thanks again!

https://www.facebook.com/buckbuildtool/posts/823959861007058

Issue #313, opened 2015-05-25
"MaxPermSize=256m; support was removed in 8.0" warning on Java 8


davido:
Running `buck` on OpenJDK 1.8.40 reports the `MaxPermSize ignore` warning (twice):

```
$ buck build buck
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=512m; support was removed in 8.0
```

Java version is:

```
$ java -version
openjdk version "1.8.0_40"
OpenJDK Runtime Environment (build 1.8.0_40-b22)
OpenJDK 64-Bit Server VM (build 25.40-b25, mixed mode)
```

davido:
The issue seems to be OpenJDK specific, as the version reportd here is not matched by this regex:

```
re.compile('java version "1\.8\..*').match(version_line)
```

as the first line is:

```
openjdk version "1.8.0_40"
```

To reproduce, buckd should be shut down:

```
$ NO_BUCKD=1 buck build buck
```

The fix is:

```
diff --git a/programs/buck_tool.py b/programs/buck_tool.py
index ea844fe..ef33334 100644
--- a/programs/buck_tool.py
+++ b/programs/buck_tool.py
@@ -457,4 +457,4 @@ def which(cmd, mode=os.F_OK | os.X_OK, path=None):
 def is_java8():
     output = check_output(['java', '-version'], stderr=subprocess.STDOUT)
     version_line = output.strip().splitlines()[0]
-    return re.compile('java version "1\.8\..*').match(version_line)
+    return re.compile('(openjdk|java) version "1\.8\..*').match(version_line)
```

PR #316, opened 2015-05-26
Prevent passing of MaxPermSize on OpenJDK 1.8

davido:
This addresses an issue raised in #313.

TEST PLAN:

Set up OpenJDK 1.8 and run:

  $ buck build buck

and confirm that "MaxPermSize=256m" is not passed to Java process.


sdwilsh:
Thanks for the fix!

Issue #312, opened 2015-05-22
Buck doesn't compile on Java 8


davido:
On most recent master (b7030deb681193b4c531dca892f9f10ed6e42cf2) `.\bin\buck build buck-win` is failing: [1].

Environment: Windows 8.1, JDK 8.0.45, ant 1.9.4, Python 27.

[1] http://paste.openstack.org/show/231454/

sdwilsh:
Hmm, I'm not seeing this with Java 1.7

davido:
Interestingly, the error messages is saying:

```
BUILD FAILED: //src/com/facebook/buck/rules:types failed with exit code 1:
javac
src\com\facebook\buck\rules\BUCK (:types) is missing deps:
    '//src/com/facebook/buck/step:step',
```

And indeed, this diff fixed it:

```
diff --git a/src/com/facebook/buck/rules/BUCK b/src/com/facebook/buck/rules/BUCK
index 3bce5a1..30113d9 100644
--- a/src/com/facebook/buck/rules/BUCK
+++ b/src/com/facebook/buck/rules/BUCK
@@ -34,6 +34,7 @@ java_immutables_library(
     '//src/com/facebook/buck/cxx:rules',
     '//src/com/facebook/buck/rules/macros:macros',
     '//src/com/facebook/buck/shell:rules',
+    '//src/com/facebook/buck/step:step',
     '//src/com/facebook/buck/thrift:rules',
     '//src/com/facebook/buck/util:constants',
     '//src/com/facebook/buck/util:exceptions',
``` 

Can confirm the same failure on Linux (on most recent master: 1cc4747f5a8fa93458cfbd98d8c01c09a9a19950):

```
$ javac -version
javac 1.8.0_40

$ java -version
openjdk version "1.8.0_40"
OpenJDK Runtime Environment (build 1.8.0_40-b22)
OpenJDK 64-Bit Server VM (build 25.40-b25, mixed mode)

$ buck build buck
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=512m; support was removed in 8.0
[-] PROCESSING BUCK FILES...FINISHED 0,6s
[+] BUILDING...10,7s (167/178 JOBS, 167 UPDATED, 0,0% CACHE HITS)
 |=> IDLE
 |=> IDLE
[...]

PR #317, opened 2015-05-26
Fix buck build on Java 8

davido:
Buck, by default, builds java_library targets with first order deps
only, so that transitive dependencies don't considered.

So the target '//src/com/facebook/buck/rules:types' is using
SmartDexingStep class that implements Step class. Step class is not
used directly by this target and Java 7 compiler is happy about it.
However Java8 compiler needs to touch Step class because of default
methods and cannot reach it, even though it's contained in the
transitive dependency chain:

  rules:types => rules:build_rule => step:step

To verify this, the following command can be used

  $ buck build  --build-dependencies=TRANSITIVE \
//src/com/facebook/buck/rules:types

To fix the problem for both compilers, add step:step as direct
dependency to the rules:types target.

TEST PLAN:

  $ buck build //src/com/facebook/buck/rules:types

on both Java 7 and java 8 should succeed.

This addresses an issue raised in #312.

### DIFF
--- a/src/com/facebook/buck/rules/BUCK
+++ b/src/com/facebook/buck/rules/BUCK
@@ -34,6 +34,7 @@ java_immutables_library(
     '//src/com/facebook/buck/cxx:rules',
     '//src/com/facebook/buck/rules/macros:macros',
     '//src/com/facebook/buck/shell:rules',
+    '//src/com/facebook/buck/step:step',
     '//src/com/facebook/buck/thrift:rules',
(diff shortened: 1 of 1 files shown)


shs96c:
Thank you!

sdwilsh:
Thanks for the fix.  After chatting with @andrewjcg, we went with a slightly different solution (using `exporterd_deps`, but I still attributed you for the fix.

PR #322, opened 2015-05-29
Improvements to the Quickstart command

sschuberth:


k21:
Thanks for your contribution! Can you please create separate pull requests for each commit? It will work better with our internal systems and we can review the commits independently. For example the typo fixes can be merged fairly quickly, but the Android SDK directory discovery might require some discussion.

sschuberth:
Done, I've split out the typo fixes to PRs #323 and #324.

src/com/facebook/buck/cli/QuickstartCommand.java
```diff
@@ -52,7 +58,7 @@
   private String destDir = "";
 
   @Option(name = "--android-sdk", usage = "Android SDK directory")
-  private String androidSdkDir = "";
+  private String androidSdkDir;
```
k21: Please annotate this field with `@Nullable`.

src/com/facebook/buck/cli/QuickstartCommand.java
```diff
@@ -71,6 +77,19 @@ public String getDestDir() {
   }
 
   public String getAndroidSdkDir() {
+    if (androidSdkDir == null) {
+      ProjectFilesystem projectFilesystem = new ProjectFilesystem(Paths.get("."));
+
+      AndroidDirectoryResolver androidDirectoryResolver =
+          new DefaultAndroidDirectoryResolver(projectFilesystem,
+              Optional.<String>absent(),
+              new DefaultPropertyFinder(projectFilesystem, ImmutableMap.copyOf(System.getenv())));
```
k21: I am afraid that when we run with the daemon, `System.getenv()` will give you the environment of the daemon process. We should make this use the client's environment.

sschuberth: As long as the daemon is inheriting the client's environment that should not be a problem, or?

Other than that, any hints on how to explicitly get the client's environment?

src/com/facebook/buck/cli/QuickstartCommand.java
sschuberth: Done.

src/com/facebook/buck/cli/QuickstartCommand.java
k21: No, the daemon's `System.getenv()` will stay the same even if the client's environment changes. In this case I think you can use `params.getRepository().getAndroidDirectoryResolver()` to get the resolver and avoid creating a new instance.

sschuberth: Done.

k21:
I have merged your changes internally. I think @sdwilsh is going to export them to the Github repo later today. Thanks again!

sdwilsh:
Thanks for your contribution!

PR #344, opened 2015-06-13
Get devices function

tgummerer:
This pull request depends on the Refactor AdbHelper pull request I sent earlier (I was under the impression that one pull request per commit is preferred).

@sdwilsh I saw that you commented, that you'd like me to add a test for this, but I don't know how to test this, as the return value depends on the devices that are connected to the machine where the tests are run, and I didn't find a way to get the AndroidDebugBridge to return a predefined list of devices, as all it's constructors are private.

### DIFF
--- a/src/com/facebook/buck/android/AdbHelper.java
+++ b/src/com/facebook/buck/android/AdbHelper.java
@@ -232,6 +232,28 @@ public class AdbHelper {
     return isAdbInitialized(adb) ? adb : null;
   }
 
+  public List<IDevice> getDevices() throws InterruptedException {
+    List<IDevice> devices;
+
+    // Initialize adb connection.
+    AndroidDebugBridge adb = createAdb(context);
+    if (adb == null) {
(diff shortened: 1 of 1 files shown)


sdwilsh:
Hmm, this ended up pulling in both of your changes.

tgummerer:
Yeah, sorry I didn't notice.  I tried, but GitHub doesn't seem to let me just add one commit to the pull request, as the second one depends on the first one.  I could make it a separate commit, not depending on the first one, but that will result in a merge conflict, so I guess it's still simpler for you this way?

sdwilsh:
Oh GitHub.  I'll merge your change shortly, so it should be easy to 
rebase and make that issue go away.

tgummerer:
Thanks, I saw you merged the other pull request now, so I updated this one to just include the right commit.

sdwilsh:
Thanks for pulling this out!

PR #346, opened 2015-06-15
Minor TestNG fixes.

liuyang-li:
@jiangty-addepar

### DIFF
--- a/src/com/facebook/buck/junit/TestNGRunner.java
+++ b/src/com/facebook/buck/junit/TestNGRunner.java
@@ -90,7 +90,7 @@ public final class TestNGRunner extends BaseRunner {
   private XmlSuite createXmlSuite(Class<?> c) {
     XmlSuite xmlSuite = new XmlSuite();
     xmlSuite.setName("TmpSuite");
-    xmlSuite.setTimeOut("1000");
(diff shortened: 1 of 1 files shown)


liuyang-li:
@sdwilsh, separate pull-requests have been submitted for each of the logical changes.

sdwilsh:
Thanks!

liuyang-li:
I agree. Let me try and add SKIPPED ResultType.

sdwilsh:
Any luck adding the skipped result type @liuyang-li?

liuyang-li:
@sdwilsh , a SKIPPED result type already exists, which I was able to use.

I realized that: We need to pass on the timeout value from `.buckconfig` to testng, so that it will terminate for tests that run indefinitely. Let me separate out my changes for the test time out into another PR, and update this one.

liuyang-li:
@sdwilsh , I just updated the change set to set the default time out value correctly. Let us do the changes for testng in phases.

I will submit another PR to update the default behavor for skipped tests in TestNGRunner. After that, I will refactor things a bit to move TestNG to its own package (it is under junit now, which is not correct). There are a couple of other fixes that I plan to do after that, e.g., correctly pass the filters to testng to select test classes.

In all, let us push this small change first.

PR #347, opened 2015-06-15
Fix PythonBuckConfigTest.

liuyang-li:
Summary:
  Update the executable name to 'python2', which has precedence over
  'python'.

Test Plan:
  `buck test`.

@sdwilsh


sdwilsh:
Thanks for pulling this out and fixing it!

PR #348, opened 2015-06-15
Fix GenruleIntegrationTest.

liuyang-li:
Summary:
  Making the regex matching more robust.

Test Plan:
  `buck test`.

@sdwilsh


sdwilsh:
Thanks for pulling this out and fixing it!

PR #349, opened 2015-06-15
Intellij Buck Support Update

liuyang-li:
@sdwilsh

### DIFF
--- a/docs/rule/project_config.soy
+++ b/docs/rule/project_config.soy
@@ -15,10 +15,10 @@
 {param status: 'UNFROZEN' /}
 {param overview}
 A <code>project_config()</code> rule is used to specify the information needed
-to generate the configuration files for an IDE so the primary build target 
+to generate the configuration files for an IDE so the primary build target
 (and optionally its tests) of a BUCK file can be built using that IDE.
 
-<p>Note that there may only be one <code>project_config()</code> rule per 
+<p>Note that there may only be one <code>project_config()</code> rule per
 BUCK file.</p>
 {/param}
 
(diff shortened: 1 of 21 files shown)


marcinkosiba:
Hey!
 We've been concentrating efforts behind the experimental IntelliJ project generator which doesn't use project_config at all. Regardless of that I don't think forcing the user to specify this as a parameter on project_config is the right approach - we should be able to extract the information from the java_library targets, right?
 The other two patches (5a7da9f and 2e49735) look good so I'll merge those in.

liuyang-li:
Great. Thanks for the feedback.

Taking things directly from the java_library might be problematic, especially when there are multiple java_library s defined in the same module: we actually do this to build thin/fat jars in different ways. We might be able to specify which java_library to use in the project_config. Is the long term goal to remove project_config entirely? Please let us know what's the preferred way of adding the source/resource roots, and we'll try to update the changeset.

marcinkosiba:
The long-term goal is to remove project_config entirely. The idea is to use IntelliJ to index the code, but build with Buck.
You can take a look at IjModuleFactory on how I plan to handle multiple targets in the same file. I actually have a diff to add resources support to it, but haven't had the time to push on it lately.

liuyang-li:
That sounds good. That could also solve some dependency management issues, e.g., Intellij doesn't know about jars genereated/downloaded via genrule, and users have to run the command line build before being able to build in Intellij. I was considering adding dependencies to `project_config`, but agree that having Intellij build through buck is cleaner. It would be great if we'd be able to still allow users to build the same repo via Intellij and command line at the same time though.

marcinkosiba:
>  It would be great if we'd be able to still allow users to build the same repo via Intellij and command line at the same time though.

The current plan for that is to have an IntelliJ plugin that will delegate building the project to Buck. Is there some other benefit to building with IntelliJ other than the convenience of being able to do this from the UI? If not then I don't think there is any reason for us to burn time on trying to support that.

liuyang-li:
I agree with the plan. My comment was a bit confusing.

It's great to have Intellij build via Buck. By "It would be great if we'd be able to still allow users to build the same repo via Intellij and command line at the same time though.", I meant that when Intellij builds through Buck, hopefully, we don't have some kind of lock such that building on the command line with `buck build` is not possible until the Intelilj build is finished. Maybe detaching from the nailgun server would be sufficient here.

marcinkosiba:
OK, we've merged in 5a7da9f and 2e49735, the plan for the IntelliJ plugin is unchanged. I don't think there is anything left to do here, so closing.

PR #353, opened 2015-06-18
Set grep to --color=none in autocomplete script

jiangty-addepar:
Autocomplete breaks when grep colors are set to "always." This forces grep to
use `--color=none`.


sdwilsh:
Hey, thanks for this!  We can't pull it in until you sign our CLA though: https://code.facebook.com/cla (there's a bot that is supposed to tell you that, but it appears to be busted)

jiangty-addepar:
@sdwilsh ok, I signed it!

sdwilsh:
Thanks for the PR!

PR #358, opened 2015-06-21
Refactor exopackage installer

tgummerer:
This is another pull request that will mainly benefit the work I'm currently doing for the android_instrumentation_test rule.  Thought I'd get this out early, as it makes sense on it's own, and the end game will be simpler without this included.


sdwilsh:
Okay, I got your first commit in.  I had to fix up some lint and java doc issues.  Can you rebase this so I can get the second one in too?

tgummerer:
Sorry about that.  Is there any way to catch those issues automatically? I rebased the second commit now.

sdwilsh:
If you run `ant lint verify-javadoc`, you should be good.

tgummerer:
Thanks, will do for my future commits.

Issue #305, opened 2015-05-15
Android Instrumentation Tests


tgummerer:
Hi,

I've recently started exploring buck for android applications.  I've tried searching through the documentation, but couldn't find a mention of it anywhere.  Is there any support to run Android Instrumentation Tests from Buck?

Regards,
Thomas

sdwilsh:
There's nothing preventing you from using them today, but no explicit support currently either.  It probably wouldn't be too much work to add support.

tgummerer:
Thanks for your quick reply.  I'm interested in adding support for this, do you have any pointers on where to start?

sdwilsh:
This is where having #256 done would be super helpful.  There is a braindump in there that would be helpful, however.

If you have more questions, let me know!

tdrhq:
It's still possible to create an instrumentation test, just that you need to add some boilerplate. I'll explain the easier version where you're creating an instrumentation test for a library, which means it's mostly just an android_binary() that includes all your sources and your test code.

The difference is you need to specify the <instrumentation> tag in your AndroidManifest for this binary, make the instrumented package to point to the itself. (i.e. if the AndroidManifest's package is com.foo.bar, the android:targetPackage in the <instrumentation> should point to com.foo.bar):
<code>
  &lt;instrumentation
      android:name="android.test.InstrumentationTestRunner"
      android:targetPackage="com.foo.bar" />
</code>

You can then install it using buck-install, and run it manually using:
<code>
  adb shell am instrument -w com.foo.bar/android.test.InstrumentationRunner
</code>

Obviously, this isn't ideal, and we should definitely make this flow better.

tgummerer:
Thanks!  I also found ```android_instrumentation_apk()```, which seems to work.

I did however run into a few issues trying to run the tests.  I'm trying to use the android test kit, and get the following exception when trying to run the tests via adb:
```
java.lang.NoClassDefFoundError: android.support.test.espresso.matcher.ViewMatchers$12
	at android.support.test.espresso.matcher.ViewMatchers.withId(ViewMatchers.java:393)
	at com.wueventapp.apso.activity.MainActivityTest.testSearchButtonIsShown(MainActivityTest.java:47)
	at java.lang.reflect.Method.invokeNative(Native Method)
	at android.test.InstrumentationTestCase.runMethod(InstrumentationTestCase.java:214)
	at android.test.InstrumentationTestCase.runTest(InstrumentationTestCase.java:199)
[...]

PR #368, opened 2015-07-01
Android instrumentation test

tgummerer:
This is the first version of the android instrumentation test rule, discussed in #305.

The first commit (786c1ab) is the same as in #358, and is included here, because removing it would cause some conflicts.

I'm not sure about the 2nd and 3rd commit, (ba95887 and 207fcca).  Those changes are so `android_instrumentation_test` doesn't post any output, but I'm not sure whether that is required or not for the tests.

The pull request is also still missing a documentation update, which I'm still working on.  There are also no new tests, but I don't know how to add them, as all parts have different behaviour with attached devices vs. no attached devices.


src/com/facebook/buck/android/AndroidInstrumentationTestDescription.java
```diff
@@ -0,0 +1,125 @@
...
+    return new AndroidInstrumentationTest(
+        params.appendExtraDeps(
+                BuildRules.getExportedRules(
+                    Iterables.concat(
+                        params.getDeclaredDeps(),
+                        resolver.getAllRules(args.deps.get())))),
+        new SourcePathResolver(resolver),
+        (InstallableApk) baseApk,
+        (AndroidInstrumentationApk) instrumentationApk,
+        args.testRunner.or("android.test.InstrumentationTestRunner"),
```
tdrhq: it's a little bit of work but this can be detected from the apk without having to manually specify it. Here's for example some python code that does the same (this might not be the best way of doing it, but it does the job):

def get_instrumentation_runner(apk):
    output = _check_output(['aapt', 'dump', 'xmltree', apk, 'AndroidManifest.xml'], stderr=os.devnull)
    name_regex = re.compile('E: instrumentation.*?A: android:name.*?="(.*?)".*?Raw', flags=re.DOTALL)

    m = name_regex.search(output)
    if m:
        return m.group(1)
    raise RuntimeError("Could not find instrumentation")

tdrhq:
Also, can you add an example target? I think all the examples are in test/com/facebook/buck/android/testdata (@sdwilsh would know better). That way I can apply your patch and play around with running the tests locally.

The example target would also help us iron out issues with how the API looks like.

src/com/facebook/buck/android/AndroidInstrumentationTestDescription.java
tgummerer: @tdrhq Thanks for the suggestion, that makes sense.  I modeled it after tryToExtractPackageName that is already in the AdbHelper.

tgummerer:
@tdrhq thanks for the review, I updated the pull request according to your suggestions.  The example is very basic, as it includes no actual tests, I can expand that if needed.

There's still no documentation, I'll add that once we agree how the API will look like.

src/com/facebook/buck/android/AdbHelper.java
```diff
@@ -902,4 +956,23 @@ public static String tryToExtractPackageNameFromManifest(
       throw new HumanReadableException("Could not extract package name from %s", pathToManifest);
     }
   }
+
+  public static String tryToExtractInstrumentationTestRunnerFromManifest(
+      InstallableApk androidBinaryRule,
+      ExecutionContext context) {
+    Path pathToManifest = androidBinaryRule.getManifestPath();
```
tdrhq: does this also work with the manifest points to an android_manifest rule?

sdwilsh: It should.  By the time the `androidBinaryRule` has run, all of its deps should have run, which means this `Path` should exist.

src/com/facebook/buck/android/ApkInstallStep.java
```diff
@@ -0,0 +1,92 @@
...
+    final AdbHelper adbHelper = new AdbHelper(
+        new AdbOptions(0, true),
+        new TargetDeviceOptions(),
+        context,
+        Optional.<Console>absent(),
+        Optional.<BuckEventBus>absent(),
+        false);
+
+    if (adbHelper.getDevices().isEmpty()) {
+      return 0;
```
tdrhq: this should a failure, right?

src/com/facebook/buck/android/AndroidInstrumentationTest.java
```diff
@@ -0,0 +1,252 @@
...
+      ExecutionContext executionContext,
+      boolean isDryRun,
+      boolean isShufflingTests,
+      TestSelectorList testSelectorList,
+      TestRule.TestReportingCallback testReportingCallback) {
+    ImmutableList.Builder<Step> steps = ImmutableList.builder();
+
+    Path pathToTestOutput = getPathToTestOutputDirectory();
+    steps.add(new MakeCleanDirectoryStep(pathToTestOutput));
+    steps.add(new ApkInstallStep(baseApk));
```
tdrhq: I would also love to support the case where the APK is instrumenting itself. (that is the case for most "library" tests, you only have to install the APK which bundles both the library and the tests together)

It doesn't look like that would be too much more work to support, correct?

src/com/facebook/buck/android/ApkInstallStep.java
```diff
@@ -0,0 +1,92 @@
...
+import com.facebook.buck.event.BuckEventBus;
+import com.facebook.buck.rules.InstallableApk;
+import com.facebook.buck.step.ExecutionContext;
+import com.facebook.buck.step.Step;
+import com.facebook.buck.util.Console;
+import com.google.common.base.Optional;
+
+import com.android.ddmlib.IDevice;
+
+public class ApkInstallStep implements Step {
```
tdrhq: @sdwilsh please look at this file

I don't know the caching semantics of Steps, so I don't know if this is a reasonable thing to do in buck steps.

Also, should this be renamed to something specific for instrumentation tests? I know there's nothing instrumentation tests specific in this, but it could be confusing for n00b looking through buck's code.

src/com/facebook/buck/android/ApkInstallStep.java
```diff
@@ -0,0 +1,92 @@
...
+        context,
+        Optional.<Console>absent(),
+        Optional.<BuckEventBus>absent(),
+        false);
+
+    if (adbHelper.getDevices().isEmpty()) {
+      return 0;
+    }
+
+    if (!adbHelper.installApk(installableApk, false)) {
```
tdrhq: just to verify, this works will work with exopackage, right?

test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/BUCK
```diff
@@ -0,0 +1,33 @@
...
+)
+
+keystore(
+  name = 'keystore',
+  store = 'debug.keystore',
+  properties = 'debug.keystore.properties',
+  visibility = ['PUBLIC'],
+)
+
+android_instrumentation_apk(
```
tdrhq: thanks for the sample buck files!

src/com/facebook/buck/android/ApkInstallStep.java
sdwilsh: If yes, an integration test would be great.  If not, we should raise a `HumanReadableException` if that is what we are given.

src/com/facebook/buck/android/ApkInstallStep.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private final InstallableApk installableApk;
+
+  public ApkInstallStep(InstallableApk apk) {
+    this.installableApk = apk;
+  }
+
+  public ApkInstallStep(AndroidInstrumentationApk apk) {
+    this.installableApk = apk;
+  }
```
sdwilsh: I think you could replace both of these constructors that just take an `AndroidBinary`, right?

src/com/facebook/buck/android/AdbHelper.java
tgummerer: Good point, I didn't think about it first, but it still works, I just tested it.

src/com/facebook/buck/android/ApkInstallStep.java
tgummerer: No I don't think so.  If this is a failure, we get a test failure, we can have the test in the `ASSUME` state otherwise, which I think is more appropriate when no devices are attached.  I think that's in line with the other tests which return the assume state when some prerequisite is missing. Or would you prefer to fail outright when no devices are attached?

src/com/facebook/buck/android/AndroidInstrumentationTest.java
tgummerer: That's a good idea, thanks!  I'll try to do that for the next version.

src/com/facebook/buck/android/ApkInstallStep.java
tgummerer: Yeah, this should work with exopackages since 806e7ccaa09c1ea5928bccdcdf993f299caef8e1.  I'm not sure how exopackages interact with instrumentation tests however.  I'll test it and modify it/add a test as appropriate.

src/com/facebook/buck/android/ApkInstallStep.java
tgummerer: Heh yes, I think just removing the second constructor would work.  Not sure what I was thinking there.  I think I still prefer `InstallableApk` over `AndroidBinary`, as that will cover an `apk_genrule()` as well.

tgummerer:
Thanks @sdwilsh and @tdrhq for your reviews.  I addressed the comments and updated the pull request.  Here's an overview of the changes since the last version:
- Add support for apk instrumenting itself
- Changed the test code in the testdata directory.  It now actually executes some tests and has an `android_instrumentation_test` that tests an exopackage (exopackages can't instrument test themselves.  If the user tries that a HumanReadableException is raised).
- Removed the extra constructor in ApkInstallStep
- Added documentation (It is in a separate commit, but I can squash it if that's preferred)
- Removed deps parameter.  I don't think it's needed, as all dependencies are needed in the apks.
- Fix a couple of infer eradicate errors (They were false positives, but simple to fix)
- Rebased on the latest master

src/com/facebook/buck/android/AndroidInstrumentationTest.java
```diff
@@ -0,0 +1,261 @@
...
+    if (instrumentationApk.isPresent()) {
```
tdrhq: I feel like it makes more semantic sense to make baseApk optional and instrumentation_apk as required.

The reason is that the instrumentation_apk is the one that defines <instrumentation> in the manifest, and then in the self-instrumenting case it somehow feels more right to still call that an instrumentation_apk.

src/com/facebook/buck/android/AndroidInstrumentationTestDescription.java
```diff
@@ -0,0 +1,134 @@
...
+          params.getBuildTarget(),
+          baseApk.getFullyQualifiedName(),
+          baseApk.getType());
+    }
+
+    if (((InstallableApk) baseApk).getExopackageInfo().isPresent() &&
+        !args.instrumentationApk.isPresent()) {
+      throw new HumanReadableException(
+          "In %s, apk='%s' can not be an exopackage if the instrumentation_apk parameter " +
+          "is not present.",
```
tdrhq: hmm, why is this? I don't see any obvious reason why this should be the case

tgummerer: I'm not 100% sure why, but when I tried to run the tests with just the exopackage as `apk` and no `instrumentation_apk`, no tests were run.  (Done by commenting out the `instrumentation_apk` line in apps/myapp_test/BUCK and adding `'//test/com/example/activity:activity_test',` in the exopackage `android_binary` rule in apps/myapp/BUCK.  Maybe I'm missing something here (I have very little experience with exopackages), but otherwise I'd prefer to just not allow it instead of having some users confused.

src/com/facebook/buck/android/AndroidInstrumentationTest.java
tgummerer: Yeah, I agree, will change for the next version, thanks!

src/com/facebook/buck/android/AndroidInstrumentationTestDescription.java
tgummerer: No wait, scratch that, I messed up my testing before.  I'll remove this block.

tgummerer:
@tdrhq thank you very much for the new review.  I addressed your comments in the new version.

tgummerer:
I somehow messed up the tests, and they failed to build the android_instrumentation_apk.  I've updated the pull request, and also rebased it onto the latest master branch.  This is the diff without the rebase:

```
diff --git a/test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp/BUCK b/test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp/BUCK
index 7fbfdc0..3c29cdb 100644
--- a/test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp/BUCK
+++ b/test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp/BUCK
@@ -28,7 +28,6 @@ android_binary(
     '//java/com/example/app:app',
     '//java/com/example/activity:activity',
     '//res/com/example/activity:res',
-    '//test/com/example/activity:activity_test',
   ],
   visibility = ['PUBLIC'],
 )
```

test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp_test/BUCK
```diff
@@ -0,0 +1,14 @@
+android_instrumentation_apk(
+  name = 'app_test',
+  manifest = 'AndroidManifest.xml',
+  apk = '//apps/myapp:app-exo',
```
tdrhq: suggestion for another version (we don't have to do this in this pull request):

we should be able to update the AndroidManifest to automatically generate the instrumentation/targetPackage field based on the properties here so we can remove that piece of redundancy. I think gradle does something like this, but let's leave this to another change.

test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp_test/BUCK
```diff
@@ -0,0 +1,14 @@
...
+  deps = [
+    '//test/com/example/activity:activity_test',
+  ],
+)
+
+android_instrumentation_test(
+  name = 'test',
+  apk = '//apps/myapp:app-exo',
```
tdrhq: this is redundant.. if you're supporting using android_instrumentation_apk() in the instrumentation_apk() target, you can automatically figure this out. So let's remove this altogether:

a. for apks instrumenting itself, we just need to point to an android_binary()
b. for apks instrumenting another apk, we point to an android_instrumentation_apk(). 

Is this easy to do?

test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp_test/BUCK
tgummerer: That definitely sounds like a nice to have.  But I do agree that's kind of unrelated to this pull request.  I'll try to do it if I find some time.

test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp_test/BUCK
tgummerer: hmm I'm not sure I follow.  You mean just removing the `apk` target here and finding the main apk from the android_instrumentation_apk() target if necessary?

If that's what you mean it shouldn't be too hard, we could just store the apkUnderTest in the AndroidInstrumentationApk class, and then get it if we have an `android_instrumentation_apk()` as `instrumentation_apk`.

Or did you mean something else?

tdrhq: Yep that's what I meant. Shouldn't need changes to android_instrumentation_apk, right?

tgummerer: Great! No, only a small change in the internal implementation is needed.

tgummerer:
I updated the pull request according to the latest comments.  The end result looks much nicer and is simpler to use with the `apk` target removed. Thanks!

A diff between the two versions is available at https://gist.github.com/tgummerer/71bd19b2f142b149071c.

test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/apps/myapp_test/BUCK
```diff
@@ -0,0 +1,13 @@
...
+  manifest = 'AndroidManifest.xml',
+  apk = '//apps/myapp:app-exo',
+  deps = [
+    '//test/com/example/activity:activity_test',
+  ],
+)
+
+android_instrumentation_test(
+  name = 'test',
+  instrumentation_apk = ':app_test',
```
tdrhq: perhaps we can just call this 'apk' now. Aside from that the API looks great

tgummerer: Sounds good.  I updated the pull request.  Thanks a lot for your help!

marcinkosiba:
Are tests the only reason you're making the bus and console optional? I know making it optional was a lot of work but if the only reason for the change are tests then could I ask you to try using a nice easymock mock instead (http://easymock.org/user-guide.html#mocking-nice) instead? Adding the bus/console optional makes the code a bit harder to read and suggests that missing values are OK in production code.

tgummerer:
@marcinkosiba yes, they were the only reason.  I didn't particularly like these changes in the first place, I just didn't know of a better alternative to avoid the output.  Using easymock does indeed look much nicer, I'll try to change it to that.

tgummerer:
I updated the pull request to use easymock.  Using `createNiceMock()` doesn't work, because it returns null, but using `createMock()` works well.

tgummerer:
@marcinkosiba oh wait, I think I might have misread the message.  The Console and BuckEventBus are optional, so that there is no unnecessary output when running `buck test` in a project with an `android_instrumentation_apk`, not when running it on buck itself.  I still think this is nicer, but I'm not sure easymock should be used in the production code.  I uploaded the version with `Console` and `BuckEventBus` as Optional to https://github.com/tgummerer/buck/tree/instrumentation-test-optional for comparison.

sdwilsh:
@marcinkosiba should finish the review soon.  @tdrhq told me he was happy with this diff already.

tgummerer:
Thanks :) Is there anything like a Helped-by: line or something similar for buck, as especially @tdrhq helped a lot getting this code in shape.

sdwilsh:
You can put whatever you want in the commit message.  I promise to keep 
it :)

tgummerer:
Thanks, I updated the pull request (just the commit message in the `Add android_instrumentation_test` commit).

tgummerer:
I updated the pull request, with most of @marcinkosiba's comments addressed.  The only one's that are still not addressed are adding the AdbHelper to the ExecutionContext (see comment https://github.com/tgummerer/buck/commit/e52918311db8be5ef940ef2a08c895a19be51e03#commitcomment-12375376), and moving the code to `AndroidInstrumentationApk`, see comment (https://github.com/tgummerer/buck/commit/e52918311db8be5ef940ef2a08c895a19be51e03#commitcomment-12374549)

tgummerer:
@marcinkosiba thanks for your latest comments, I updated the pull request accordingly.

marcinkosiba:
Looks good, just a couple of questions.

tgummerer:
@marcinkosiba Thanks a lot for the help! I addressed the questions in the update.  Interdiff:

```diff
diff --git a/build.xml b/build.xml
index 5effe72..74d147f 100644
--- a/build.xml
+++ b/build.xml
@@ -104,7 +104,6 @@
     <include name="jetty/jetty-all-9.2.10.v20150310.jar" />
     <include name="jimfs/jimfs-1.0.jar" />
     <include name="jsr/javax.inject-1.jar" />
-    <include name="kxml2/kxml2-2.3.0.jar" />
     <include name="okhttp/okhttp-2.2.0.jar" />
     <include name="okio/okio-1.2.0.jar" />
     <include name="servlet-api/javax.servlet-api-3.1.0.jar" />
diff --git a/programs/buck_repo.py b/programs/buck_repo.py
index 05ae450..c72391c 100644
--- a/programs/buck_repo.py
+++ b/programs/buck_repo.py
@@ -21,7 +21,6 @@ JAVA_CLASSPATHS = [
     "third-party/java/aopalliance/aopalliance.jar",
     "third-party/java/args4j/args4j-2.0.30.jar",
     "third-party/java/asm/asm-debug-all-5.0.3.jar",
-    "third-party/java/cglib/cglib-nodep-2.2.jar",
     "third-party/java/closure-templates/soy-excluding-deps.jar",
     "third-party/java/commons-compress/commons-compress-1.8.1.jar",
     "third-party/java/dd-plist/dd-plist.jar",
@@ -52,7 +51,6 @@ JAVA_CLASSPATHS = [
     "third-party/java/jsr/jsr305.jar",
     "third-party/java/kxml2/kxml2-2.3.0.jar",
     "third-party/java/nailgun/nailgun-server-0.9.2-SNAPSHOT.jar",
-    "third-party/java/objenesis/objenesis-1.2.jar",
     "third-party/java/okhttp/okhttp-2.2.0.jar",
     "third-party/java/okio/okio-1.2.0.jar",
     "third-party/java/servlet-api/javax.servlet-api-3.1.0.jar",
```

sdwilsh:
Hmm, I'm getting a build failure (this could be due to an older base revision on your end).  Also possible that the commit that added this isn't in open source yet (I'm doing a push right now):

    [mkdir] Created dir: /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/build/classes
     [copy] Copying 1 file to /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/build/classes/com/facebook/buck/json
    [javac] Compiling 27 source files to /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/build/classes
    [javac] Compiling 17 source files to /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/build/classes
    [javac] Compiling 1095 source files to /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/build/classes
    [javac] [compiled 175671 lines in 6707 ms: 26192.1 lines/s]
    [javac] [2077 .class files generated]
    [javac] ----------
    [javac] 1. ERROR in /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/src/com/facebook/buck/android/AndroidInstrumentationTestDescription.java (at line 36)
    [javac] 	public class AndroidInstrumentationTestDescription
    [javac] 	             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    [javac] The type AndroidInstrumentationTestDescription must implement the inherited abstract method Description<AndroidInstrumentationTestDescription.Arg>.createBuildRule(TargetGraph, BuildRuleParams, BuildRuleResolver, A)
    [javac] ----------
    [javac] 2. ERROR in /data/sandcastle/boxes/instance-buck-git-buck-verification-linux/src/com/facebook/buck/android/AndroidInstrumentationTestDescription.java (at line 58)
    [javac] 	public <A extends Arg> AndroidInstrumentationTest createBuildRule(
    [javac]       BuildRuleParams params,
    [javac]       BuildRuleResolver resolver,
    [javac]       A args) {
    [javac] 	                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    [javac] The method createBuildRule(BuildRuleParams, BuildRuleResolver, A) of type AndroidInstrumentationTestDescription must override or implement a supertype method
    [javac] ----------
    [javac] 2 problems (2 errors)

tgummerer:
Thanks for all the help getting this in, especially @tdrhq and @marcinkosiba for the reviews :)

PR #377, opened 2015-07-14
Use try-with-resources to prevent resource leaks

dpursehouse:
Summary:
In several places resources are opened but potentially not closed
if an exception occurs. In these places, use try-with-resource so
that the resource is properly closed on exception.

Test plan:

* `buck test`

### DIFF
--- a/third-party/java/dx-from-kitkat/src/com/android/dex/Dex.java
+++ b/third-party/java/dx-from-kitkat/src/com/android/dex/Dex.java
@@ -101,13 +101,13 @@ public final class Dex {
      */
     public Dex(File file) throws IOException {
         if (FileUtils.hasArchiveSuffix(file.getName())) {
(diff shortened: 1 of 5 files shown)


sdwilsh:
The first two are OK, but the last three changes create a change in behaviour by putting JarOutputStream & IndentingWriter in a t-w-r the underlying stream gets closed when the t-w-r block is left, which will cause problems for other parts of the code (e.g. the call to closeOutput in the case of the JarOutputStream change)

dpursehouse:
@sdwilsh thanks for the feedback.  I think t-w-r is OK for the first `IndentingWriter` because the underlying stream is also created in the same method and only used there.  For the other two, I've undone the change in the reworked version and added suppressions instead.

sdwilsh:
Thanks for fixing this up!

PR #379, opened 2015-07-14
AdbHelper: only restart adb when no device is attached

tgummerer:
Summary:
Currently the adb server is restarted both when there is no device
attached, or when there is more than one device attached, but the user
hasn't enabled multi install mode.  Restarting the adb server helps when
adb is flaky and shows devices as offline or not attached, but it
doesn't help when there are too many devices are attached and the user
forgets to enable multi install mode.  This is usually a user error.

Disable the server restart and save a little time for the user.

Test plan:
Test buck install with different device configurations and show it does
the right thing about restarting the adb server.

### DIFF
--- a/src/com/facebook/buck/android/AdbHelper.java
+++ b/src/com/facebook/buck/android/AdbHelper.java
@@ -183,19 +183,6 @@ public class AdbHelper {
       return null;
     }
 
-    // Found multiple devices but multi-install mode is not enabled.
-    if (!options.isMultiInstallModeEnabled() && devices.size() > 1) {
-      console.printBuildFailure(
-          String.format("%d device(s) matches specified device filter (1 expected).\n" +
-                        "Either disconnect other devices or enable multi-install mode (%s).",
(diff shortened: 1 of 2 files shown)


marcinkosiba:
I don't think the change is entirely correct - will it not cause us to not restart ADB in the case that the user has 2 devices but neither of them are online?
From a quick look it seems the code is awkward because filterDevices is doing too much (it's both filtering and trying to decide how to react to various scenarios). How about if we moved the logic around printing errors into getDevices (and updated AdbHelperTest accordingly) and just had filterDevices do the filtering?

tgummerer:
Yes, you're right, I didn't consider that.

Makes sense, I'll change it to do that.  I probably won't get to it before tomorrow though.

tgummerer:
The code deciding how to react in the various scenarios is more closely tied to the filtering code than I thought.  I don't think it's worth to factor out all code for deciding what to do, as that will just add unnecessary complexity.

I decided to just take out the code for deciding whether there are more than two devices, when multi install mode is not enabled.

The pull request is updated with the changes.

sdwilsh:
Thanks for fixing this up and getting it in :)

PR #380, opened 2015-07-15
Correctly handle ':' with the project command.

liuyang-li:
Summary:
The current project command does not support ':' wildcard specification.
For example, `buck project folder:` will throw an error. The wildcard
expansion is useful in cases where one buck file has targets with disjoint
sets of dependencies.

Test:
`buck test`

### DIFF
--- a/src/com/facebook/buck/cli/ProjectCommand.java
+++ b/src/com/facebook/buck/cli/ProjectCommand.java
@@ -44,6 +44,7 @@ import com.facebook.buck.model.BuildTarget;
 import com.facebook.buck.model.BuildTargetException;
 import com.facebook.buck.model.FilesystemBackedBuildFileTree;
 import com.facebook.buck.model.HasBuildTarget;
+import com.facebook.buck.model.Pair;
 import com.facebook.buck.parser.BuildFileSpec;
 import com.facebook.buck.parser.BuildTargetSpec;
 import com.facebook.buck.parser.ParserConfig;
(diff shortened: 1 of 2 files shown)


liuyang-li:
@jiangty-addepar

liuyang-li:
Thanks for the feedback, @marcinkosiba.

I have removed the event posts.

liuyang-li:
@marcinkosiba , shall we push this while we work on the long term goal to remove `project_config`, as this is pretty much just a bug fix?

src/com/facebook/buck/cli/ProjectCommand.java
```diff
@@ -307,18 +308,31 @@ public boolean isExperimentalIntelliJProjectGenerationEnabled() {
...
-
   @Override
   public int runWithoutHelp(CommandRunnerParams params) throws IOException, InterruptedException {
     if (getIde(params.getBuckConfig()) == ProjectCommand.Ide.XCODE) {
       checkForAndKillXcodeIfRunning(params, getIdePrompt(params.getBuckConfig()));
     }
 
-    ImmutableSet<BuildTarget> passedInTargetsSet = getBuildTargets(
-        getArgumentsFormattedAsBuildTargets(params.getBuckConfig()));
+    Pair<ImmutableSet<BuildTarget>, TargetGraph> result = null;
```
sdwilsh: @Coneko left this comment on our internal tracker but forgot to put it here so you can address it.

The TargetGraph from here should be used to generate the project as well instead of creating a new one using the targets below.

liuyang-li:
@Coneko it looks like we won't be able to reuse the TargetGraph here for the project generation. The target graph here contains only the root targets as we do not traverse the tree recursively when parsing the root level targets. To generate the project, we need the root targets and (potentially) their siblings, together with root targets children.

Please let me know how we can avoid the extra traversal, and I'd be happy to give it a try.

Coneko:
It should be possible by changing `getTargetNodeSpecsForIde` to take `Collection<TargetNodeSpec>` instead of `Collection<BuildTarget>` as parameter.

sdwilsh:
@Coneko - not sure why you removed the needs-revision tag when you want changes still :)

liuyang-li:
@Coneko , I gave that a try, and was not able to make it work. I feel that we need to do two traversals here: 1st one to get all the targets in the module when `:` is used, and 2nd one to get all the dependencies of the targets.

Coneko:
Ok, sorry for back and forth, it's ok then.

src/com/facebook/buck/cli/ProjectCommand.java
sdwilsh: Also, this doesn't compile.  On line 411 below we have `int result`, which introduces a duplicate variable name.

liuyang-li: This was a bad rebase. It is fixed, and all tests pass now.

test/com/facebook/buck/java/intellij/ProjectIntegrationTest.java
```diff
@@ -443,19 +443,32 @@ public void testBuckProjectWithMultipleLibrariesInOneBuildFile() throws IOExcept
...
   @Test
-  public void testNonexistentTarget() throws IOException {
+  public void testProjectWithColon() throws IOException {
     ProjectWorkspace workspace = TestDataHelper.createProjectWorkspaceForScenario(
         this,
         "project1",
         temporaryFolder);
     workspace.setUp();
 
-    expectedException.expect(HumanReadableException.class);
```
sdwilsh: Fun fact: with this, and the change below, we now have an unused import of this class which fails `ant lint`.

liuyang-li: The PMD error is fixed. Would you please let us know the official build command for building buck? I typically do: `buck test`, which does not run static analysis I think.

sdwilsh: I should have said that I fixed it locally; it should go out later today.

Our internal CI roughly does this to verify: `ant clean default lint && ./bin/buck build buck && ./bin/buck test`

liuyang-li: That's good to know. Sorry about the back and forth, and thanks for pushing this through.

sdwilsh: It's okay; we don't have great docs on this and we had to stop using Travis because we'd just get OOMs in random places of the build/setup.  Normally I try to fix things up on my own, but I've been a bit swamped today.

PR #381, opened 2015-07-16
Select only *enabled* activity aliases when looking for a default act…

TheoLassonder:
…ivity to launch (avoid 'Default activity is ambiguous' error).

This is useful for apps that create multiple <activity-alias> elements, but only enable one of them. This is populate with apps that want to offer the user a choice of different icons for their app.

Thanks,

Theo

### DIFF
--- a/src/com/facebook/buck/android/DefaultAndroidManifestReader.java
+++ b/src/com/facebook/buck/android/DefaultAndroidManifestReader.java
@@ -47,7 +47,8 @@ public class DefaultAndroidManifestReader implements AndroidManifestReader {
    * to show up in the launcher.
    */
   private static final String XPATH_LAUNCHER_ACTIVITIES =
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/android/DefaultAndroidManifestReader.java
```diff
@@ -47,7 +47,7 @@
    * to show up in the launcher.
    */
   private static final String XPATH_LAUNCHER_ACTIVITIES =
-      "/manifest/application/*[self::activity or self::activity-alias]" +
+      "/manifest/application/*[self::activity or self::activity-alias[not(@android:enabled) or @android:enabled='true']]" +
```
sdwilsh: This line is over 100 characters (can be spotted with `ant lint`).  Can you please fix that?

sdwilsh:
Thanks for updating your pull request!

sdwilsh:
Thanks again; this made our weekly update post!

https://www.facebook.com/buckbuildtool/posts/846856528717391

PR #384, opened 2015-07-22
Adding batch file for Windows users

jimevans:
This allows Windows users to simply call bin\buck and have it work
correctly.

### DIFF
--- a/bin/buck.bat
+++ b/bin/buck.bat
@@ -1,0 +1,2 @@
+@echo off
+python %~dp0..\programs\buck.py %*


sdwilsh:
This is what we wrote https://github.com/facebook/buck/blob/master/bin/buck.ps1 for.  Any reason you are using `cmd.exe` instead of Power Shell?

jimevans:
@sdwilsh While I agree that PowerShell is in most ways superior to a simple command prompt, at the moment, it's impossible to integrate Buck into a build process that uses Visual Studio for part of the build, as that product requires specific environment variables to be set, and the product provides command-prompt shortcuts preinstalled for developers to use. Adding a batch file option allows seamless integration with existing Visual Studio-centric build processes.

sdwilsh:
Today I learned.  Thanks!

shs96c:
Reviewing this now. Thank you!

PR #386, opened 2015-07-31
Fix buck build buck

tgummerer:
Summary:
buck build buck was broken by b50c6d26b8 ("Load
TranslatingJavacPhaseTracer from resources"), with the following error
message:
/home/tommy/work/buck/src/com/facebook/buck/java/Jsr199Javac.java:171: error: cannot access com.facebook.buck.java.tracing.JavacPhaseTracer
      TranslatingJavacPhaseTracer.setupTracing(
                                 ^
  class file for com.facebook.buck.java.tracing.JavacPhaseTracer not found
Errors: 1. Warnings: 1.

This is because of a missing dependency.  Add the dependency to fix the
build.

Test plan:
Check that buck build buck compiles buck successfully.

### DIFF
--- a/src/com/facebook/buck/java/BUCK
+++ b/src/com/facebook/buck/java/BUCK
@@ -93,6 +93,7 @@ java_immutables_library(
     '//src/com/facebook/buck/event:event',
     '//src/com/facebook/buck/io:io',
     '//src/com/facebook/buck/java/abi:protocol',
+    '//src/com/facebook/buck/java/tracing:api',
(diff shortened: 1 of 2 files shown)


sdwilsh:
We've got an internal change that I need to push out for this too.  This is a Java 8-only bug.

sdwilsh:
Thanks for reporting this, however :)

tgummerer:
@sdwilsh Ah ok thanks, I suspected it was something about my system.

PR #389, opened 2015-08-10
Fix AndroidBinaryIntegrationTest with python3 as default

tgummerer:
Summary:
On systems on which python refers to python3, the
AndroidBinaryIntegrationTest fails in the testPreprocessorForcesReDex
test, because of a python script that is not compatible with python3.
PEP 0394 suggests that python2 should always refer to a valid python2
interpreter.  Adjust the shebang line to make the script use python2,
which fixes the test.

Test plan: buck test

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -177,7 +177,7 @@
   <target name="python-get-prefix">
     <exec executable="python" failonerror="true" outputproperty="python.prefix" logError="true">
       <arg value="-c" />
-      <arg value="import distutils.sysconfig; print distutils.sysconfig.PREFIX" />
+      <arg value="import distutils.sysconfig; print(distutils.sysconfig.PREFIX)" />
     </exec>
   </target>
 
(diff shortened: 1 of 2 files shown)


tgummerer:
I just noticed another error with python referring to python3 on my system, let me add that to this pull request.

tgummerer:
Sorry I submitted the pull request a bit too fast.  This one addresses all failures on my system.

sdwilsh:
The problem with this change is that `python2` doesn't exist on OS X

tgummerer:
Aargh, I didn't think about that.  I guess the best way is to just make the things that failed `python3` compatible, which is what I did in the updated pull request.  Tests pass with both `python2` and `python3` as default.

sdwilsh:
Thanks!

tgummerer:
Thank you for pointing out my initial mistakes!

Not a big deal, but just wondering: you always seem to pull in the first message of the pull request as the commit message, which in this case doesn't really fit what the commit does anymore.  When updating the commit message would it be appropriate to edit the first message and sync it with the edited commit message, or do you pull it out some other way?

sdwilsh:
I believe our automation pulls the text from the first comment (which is editable), so updating that is best.
________________________________

Issue #328, opened 2015-06-03
android_prebuilt_aar should actually export its deps like prebuilt_jar


edisonw:
Now that Gradle supports transitive aars, there are libraries out there that requires a prebuilt aar to be depended on another prebuilt aar, there is no way to pass that in right now. 

Can we add support for the following pattern? 

android_prebuilt_aar {
      deps = [
          other prebuilt aars. 
      ]
}

sdwilsh:
deps should work today.  What are you seeing when doing this?

edisonw:
it was complaining about R resources not found.

edisonw:
I'm using a buck from few weeks ago, let me update to latest and give it another try.

edisonw:
Yea, with master I'm still getting BUILD FAILED: //:depdendant#aar_android_resource failed with exit code 1:
generate_resource_ids 


Setup: 

android_prebuilt_aar(
    name = 'depdendant',
    aar = 'depdendant/depdendant.aar',
    deps = [
      ':depdendant-core'
    ]
)

sdwilsh:
I'm guessing you need `:dependant-core` also in whatever libraries need `:dependant`, yeah?  If so, we'll need to add `exported_deps` to the rule and you'd want to use that.

edisonw:
^yea, correct, there are other libraries that need :dependent-core as well.

sdwilsh:
Can you apply https://gist.github.com/sdwilsh/3d4645edc5d4f3bbf0a1 locally and see if it fixes it for you?  Writing a test is going to take a while for this, but if this unblocks you, I can just get it in.

edisonw:
Umm...it won't get through the process stage now:

BUILD FAILED: //:dependent: exported dep //:dependent#aar_android_resource (android_resource) must be a type of java library.

sdwilsh:
Alright, the naive fix won't do it.  Hopefully I can get this resolved next week.  For now, just include both in the deps of dependent libraries (which sucks, I know).

sdwilsh:
Also, do you happen to have an example AAR that does this?

rallat:
This [aar](https://twittersdk.artifactoryonline.com/twittersdk/public/com/digits/sdk/android/digits/1.5.0/) has transitive dependencies on another aars

sdwilsh:
@rallat - looking over the `pom`, that seems to depend on twitter-core, but I'm not seeing that available for download on that site.  Any pointers?

[...]

PR #391, opened 2015-08-11
Include resource dependencies from other aars

tgummerer:
Summary:
Aars can depend on resources that are defined in another aar, most
notably the google design library depending on resources defined in the
appcompat library.  This can cause build errors or app crashes, if the
resources are not included in the build.  Add tests to make sure we
don't regress the recent fix for such aars.
    
Test Plan:
Added tests
Checked a project using the design library now works


src/com/facebook/buck/android/aapt/MiniAapt.java
```diff
@@ -140,7 +140,7 @@ public int execute(ExecutionContext context) throws InterruptedException {
       collectResources(filesystem, context.getBuckEventBus());
       processXmlFilesForIds(filesystem, references);
     } catch (IOException | XPathExpressionException | ResourceParseException e) {
-      context.logError(e, "Error parsing resources to generate resource IDs for %s.", resDirectory);
+      context.logError(e, "Error parsing resources to generate resource IDs for %s.", resDirectories);
```
sdwilsh: nit: this line is now too long and fails `ant lint`

test/com/facebook/buck/android/AndroidResourceIntegrationTest.java
```diff
@@ -48,13 +48,9 @@ public void setUp() throws IOException {
   }
 
   @Test
-  public void testOnlyUsesFirstOrderResources() throws IOException {
+  public void testUsesResourcesFromTransitiveDependencies() throws IOException {
     ProjectWorkspace.ProcessResult result = workspace.runBuckBuild("//res1:res");
-    result.assertFailure();
-    assertTrue(result.getStderr().contains("The following resources were not found"));
-    assertTrue(result.getStderr().contains("another_name"));
```
sdwilsh: These were the last two uses of `assertTrue`, so `ant lint` fails here.

src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java
```diff
@@ -106,7 +106,11 @@ public PrebuiltJar apply(AndroidPrebuiltAar input) {
             pathResolver,
             ImmutableSortedSet.<BuildRule>copyOf(javaDeps)));
     AndroidResource androidResource = buildRuleResolver.addToIndex(
-        createAndroidResource(unzipAar, params, pathResolver));
+        createAndroidResource(
+            unzipAar,
+            params,
+            pathResolver,
+            buildRuleResolver.getAllRules(args.deps.get())));
```
sdwilsh: Should this only be `android_resource` deps like we filter out `javaDeps` above?

src/com/facebook/buck/android/AndroidResource.java
```diff
@@ -157,6 +158,16 @@
...
+  private final Function<HasAndroidResourceDeps, Path> resourceDirectoryFunction =
+      new Function<HasAndroidResourceDeps, Path>() {
+        @Override
+        public Path apply(HasAndroidResourceDeps input) {
+          return Optional.fromNullable(input.getRes())
+              .transform(getResolver().getPathFunction())
+              .get();
+        }
+      };
+
```
sdwilsh: I'd be fine with this being inline below

sdwilsh: With that said, this undoes an optimization @natthu did to require first-order deps for android resources.  The issue you are trying to address is the last part of #328 that I haven't gotten to yet, and it's tricky.

I think the right way to do this is in `AndroidLibraryGraphEnhancer.getBuildableForAndroidResources`, look for any AARs as a dep, and then add the `aar_android_resource` flavor as a dep and all of its deps (because those should be considered as exported deps) so it actually gets packaged.  That should then package up what you need to prevent that compiler error you were seeing.

sdwilsh:
Now that I have the context pulled in, and I'm not sick, review turnaround time should be quicker.

src/com/facebook/buck/android/AndroidResource.java
tgummerer: Thanks for the advice.  I'll have a look at `AndroidLibraryGraphEnhancer.getBuildableForAndroidResources` and see if I can cook something up over the next few days!

sdwilsh: It occurred to me this morning that you'll also need to hook up any deps that are `android_resource` rules to the `android_resource` of the AAR (including reach inside any other AARs).  The easiest way to do that is likely to expose it on AndroidPrebuiltAAR.

tgummerer: Thanks, I'll have a look!

tgummerer:
Looking into this a bit deeper, I think the change I did is actually needed.  The resources on which the design library depends on have to go through the MiniAapt step, because the design library expects the resources to be in the the android.support.design package.

Adding the `aar_android_resource` flavor as a dep in the `AndroidLibraryGraphEnhancer` will package the resources into the dex, but in the `android.support.v7.appcompat` package, so when the design library tries to access them, the app will still crash.  Or did I overlook something?

In order not to break @natthu's optimization, I added a flag to the AndroidResource class, so the transitive dependencies are only pulled in if we have a `android_prebuilt_aar` in which case we can't do anything about how the resources are accessed.

The rest of your review comments are also addressed in the updated pull request.

src/com/facebook/buck/android/AndroidAarDescription.java
```diff
@@ -179,7 +179,8 @@ public Arg createUnpopulatedConstructorArg() {
         new BuildTargetSourcePath(assembleAssetsDirectories.getBuildTarget()),
         /* assetsSrcs */ ImmutableSortedSet.<Path>of(),
         new BuildTargetSourcePath(manifest.getBuildTarget()),
-        /* hasWhitelistedStrings */ false);
+        /* hasWhitelistedStrings */ false,
+        /* includeTransitiveDeps */ false);
```
sdwilsh: Let's take a `AndroidLibraryGraphEnhancer.ResourceDependencyMode` instead of a boolean here

sdwilsh:
Can you also create an integration test for this so we don't regress it?

tgummerer:
Thanks for the review! The code is using `ResourceDependencyMode` instead of the boolean now, and I added a test.  It's also rebased to the latest master.

sdwilsh:
I've poked @natthu to take a look at this too and I've kicked off our internal CI to make sure there aren't any lingering issues.

test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
```diff
@@ -16,6 +16,10 @@
 
 package com.facebook.buck.android;
 
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.not;
```
sdwilsh: You don't need to fix this unless @natthu has any suggested changes, but this is unused.

tgummerer:
Thanks!

test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
tgummerer: Oops, sorry.  I really need to find a way to remember to run `ant lint` :(

sdwilsh: Well, I really just need to get some decent CI up here so machines can tell us when it's broken :)

tgummerer: That would be neat! In the meantime I set up a pre-push hook for myself, so this shouldn't happen anymore in the future.

sdwilsh:
@natthu has had a few comments, but I fixed them locally (our importer doesn't handle binary changes right, and it would be more to have fix them).  This should be in *soon*.

tgummerer:
Thanks!

sdwilsh:
So I was sick Wednesday through Friday last week, so this didn't get very far.  @natthu rightly pointed out two issues, which I'll translate onto here in a moment.  However, fixing them, I've run into an issue with `robolectric_test`.

src/com/facebook/buck/android/AndroidResource.java
```diff
@@ -387,6 +396,29 @@ public Sha1HashCode getAbiKeyForDeps() {
...
+      case TRANSITIVE:
+        return FluentIterable.from(getNonEmptyResourceDeps())
+            .append(this)
+            .transform(
+                new Function<HasAndroidResourceDeps, Path>() {
+                  @Override
+                  public Path apply(HasAndroidResourceDeps input) {
+                    return Optional.fromNullable(input.getRes())
+                        .transform(getResolver().getPathFunction())
+                        .get();
```
sdwilsh: You have an `get()` call without checking presence.  I did this instead:

```diff
-  .get();
+  .orNull();
```

And then...

src/com/facebook/buck/android/AndroidResource.java
```diff
@@ -387,6 +396,29 @@ public Sha1HashCode getAbiKeyForDeps() {
...
+                  }
+                })
+            .toList();
```
sdwilsh:
```diff
+ .filter(Predicates.notNull())
```

sdwilsh:
TIL you cannot add comments to parts of a file that haven't been touched.  `getAbiKeyForDeps` also needs to handle this special case.  I did a change in `getNonEmptyResourceDeps` to use `UnsortedAndroidResourceDeps` here, but I'm finding that some of our internal tests in our Android repository die trying to build a `robolectric_test` rule:
```
BUILD FAILED: //javatests/path/to/some/test:test#dummy_r_dot_java failed on step android-res-merge with an exception:
Multiple entries with same key: buck-out/gen/third-party/java/card-io/__card-io#aar_android_resource_text_symbols__/R.txt=io.card.payment and buck-out/gen/third-party/java/card-io/__card-io#aar_android_resource_text_symbols__/R.txt=io.card.payment
```

I'm not sure I'll have time to dive into this this week, sadly.

tgummerer:
Thanks for the review, I'll have a look.  Though I'm on vacation this week, so I won't get to it before next week either.

sdwilsh:
I'll try to apply your change later today (it's going to be fun because I need to merge your changes with some of the nits I've fixed) and let you know.  Thanks!

tgummerer:
Thanks! I fixed all the nits you posted here.

src/com/facebook/buck/android/AndroidResource.java
```diff
@@ -387,6 +412,30 @@ public Sha1HashCode getAbiKeyForDeps() {
...
+    switch (resourceDependencyMode) {
+      case FIRST_ORDER:
+        return ImmutableSet.of(getResolver().getPath(Preconditions.checkNotNull(getRes())));
+      case TRANSITIVE:
+        return FluentIterable.from(getNonEmptyResourceDeps())
+            .append(this)
+            .transform(
+                new Function<HasAndroidResourceDeps, Path>() {
+                  @Override
+                  public Path apply(HasAndroidResourceDeps input) {
```
sdwilsh: This should also be annotated with `@Nullable`.  Infer caught this internally.

src/com/facebook/buck/android/AndroidResource.java
```diff
@@ -367,8 +377,23 @@ public BuildableProperties getProperties() {
...
+    ImmutableSet<HasAndroidResourceDeps> resourceDeps;
+    switch (resourceDependencyMode) {
+      case FIRST_ORDER:
+        resourceDeps = ImmutableSet.copyOf(getNonEmptyResourceDeps());
+        break;
+      case TRANSITIVE:
+        resourceDeps = UnsortedAndroidResourceDeps.createFrom(
+            getDeps(),
+            Optional.<UnsortedAndroidResourceDeps.Callback>absent())
+                       .getResourceDeps();
```
sdwilsh: @natthu spotted that we don't make sure these are available with top-down-building (the default).  I've got a fix locally, but @natthu wants to try to fix something in the graph enhancement that will make my fix less awful, I think.  We should be able to get this in real soon.

sdwilsh:
(in case it isn't clear, you don't need to update this diff, I just wanted to keep you updated)

natthu:
Hi @tgummerer - sorry for the back and forth on this issue, and thanks for bearing with us.

FWIW, with just these local changes, I got the first half of your integration test to pass:
```
diff --git a/src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java b/src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java
index d5c1cb5..7b30af1 100644
--- a/src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java
+++ b/src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java
@@ -106,7 +106,8 @@ public class AndroidPrebuiltAarDescription
             pathResolver,
             ImmutableSortedSet.<BuildRule>copyOf(javaDeps)));
     AndroidResource androidResource = buildRuleResolver.addToIndex(
-        createAndroidResource(unzipAar, params, pathResolver));
+        createAndroidResource(unzipAar, params, pathResolver,
+            AndroidResourceHelper.androidResOnly(buildRuleResolver.getAllRules(args.deps.get()))));
     return buildRuleResolver.addToIndex(new AndroidPrebuiltAar(
         /* androidLibraryParams */ params.copyWithDeps(
             /* declaredDeps */ Suppliers.ofInstance(ImmutableSortedSet.<BuildRule>of(
@@ -154,18 +155,18 @@ public class AndroidPrebuiltAarDescription
   private AndroidResource createAndroidResource(
       AndroidPrebuiltAarGraphEnhancer.UnzipAar unzipAar,
       BuildRuleParams params,
-      SourcePathResolver resolver) {
+      SourcePathResolver resolver, ImmutableSortedSet<BuildRule> buildRules) {
     BuildRuleParams buildRuleParams = params.copyWithChanges(
         /* buildTarget */ BuildTargets.createFlavoredBuildTarget(
             params.getBuildTarget().checkUnflavored(),
             ImmutableFlavor.of("aar_android_resource")),
-        /* declaredDeps */ Suppliers.ofInstance(ImmutableSortedSet.<BuildRule>of()),
+        /* declaredDeps */ Suppliers.ofInstance(buildRules),
         /* extraDeps */ Suppliers.ofInstance(ImmutableSortedSet.<BuildRule>of(unzipAar)));

     return new AndroidResource(
         /* buildRuleParams */ buildRuleParams,
         /* resolver */ resolver,
-        /* deps */ ImmutableSortedSet.<BuildRule>of(),
+        /* deps */ buildRules,
         /* res */ new BuildTargetSourcePath(
             unzipAar.getBuildTarget(),
             unzipAar.getResDirectory()),
diff --git a/test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java b/test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
index 228a191..31fc5c7 100644
--- a/test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
+++ b/test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
@@ -16,6 +16,9 @@

 package com.facebook.buck.android;

+import static org.hamcrest.Matchers.containsString;
+import static org.junit.Assert.assertThat;
+
 import com.facebook.buck.testutil.integration.DebuggableTemporaryFolder;
 import com.facebook.buck.testutil.integration.ProjectWorkspace;
 import com.facebook.buck.testutil.integration.TestDataHelper;
@@ -68,4 +71,18 @@ public class AndroidPrebuiltAarIntegrationTest {
   public void testAndroidPrebuiltAarInDepsIsExported() throws IOException {
     workspace.runBuckBuild("//android_prebuilt_aar-dep:lib").assertSuccess();
   }
+
+  @Test
+  public void testPrebuiltRDotTxtContainsTransitiveDependencies() throws IOException {
+    workspace.runBuckBuild("//third-party/design-library:design-library").assertSuccess();
+
+    String appCompatResource = "TextAppearance_AppCompat_Body2";
+
+    String rDotTxt = workspace.getFileContents("buck-out/gen/third-party/design-library/" +
+        "__design-library#aar_android_resource_text_symbols__/R.txt");
+    assertThat(
+        "R.txt contains transitive dependencies",
+        rDotTxt,
+        containsString(appCompatResource));
+  }
 }
diff --git a/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design-library/BUCK b/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design
new file mode 100644
index 0000000..19fd727
--- /dev/null
+++ b/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design-library/BUCK
@@ -0,0 +1,14 @@
+android_prebuilt_aar(
+  name = 'design-library',
+  aar = 'design-22.2.1.aar',
+  visibility = ['PUBLIC'],
+  deps = [
+    ':appcompat',
+  ],
+)
+
+android_prebuilt_aar(
+  name = 'appcompat',
+  aar = 'appcompat-v7-22.2.0.aar',
+  visibility = ['PUBLIC'],
+)
diff --git a/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design-library/appcompat-v7-22.2.0.aar b/test/com/facebook/buck/android/testdata/android_prebuilt_aar
new file mode 100644
index 0000000..c2f9094
Binary files /dev/null and b/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design-library/appcompat-v7-22.2.0.aar differ
diff --git a/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design-library/design-22.2.1.aar b/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third
new file mode 100644
index 0000000..1fe0a82
Binary files /dev/null and b/test/com/facebook/buck/android/testdata/android_prebuilt_aar/third-party/design-library/design-22.2.1.aar differ
```

i.e. "//third-party/design-library:design-library" builds fine. 

I still don't understand the "app crashing" part. Why should `R.java` for the `design-library` include symbols from the `appcompat` library? IIUC, if some code in `design-library` is referencing to resources in the `appcompat` library, it should use the fully qualified name of those resources (i.e. `<package name of `appcompat` library>.string.resource_name`) - is my assumption incorrect?

tgummerer:
@sdwilsh thanks for keeping me updated!

@natthu thank you for the help getting this into good shape! I'm not sure about the exact reasoning, why `R.java` for the `design-library` needs to include those resources, but it is the way it works.  The design library tries to access the resources from the `android.support.design` package, e.g. as shown in the exception below `android.support.design.R$attr.colorPrimary`

```
E/AndroidRuntime( 1278): java.lang.RuntimeException: Unable to start activity ComponentInfo{com.example/com.example.activity.MainActivity}: android.view.InflateException: Binary XML file line #110: Error inflating class android.support.design.widget.NavigationView
E/AndroidRuntime( 1278): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:1956)
E/AndroidRuntime( 1278): 	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:1981)
E/AndroidRuntime( 1278): 	at android.app.ActivityThread.access$600(ActivityThread.java:123)
E/AndroidRuntime( 1278): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1147)
E/AndroidRuntime( 1278): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime( 1278): 	at android.os.Looper.loop(Looper.java:137)
E/AndroidRuntime( 1278): 	at android.app.ActivityThread.main(ActivityThread.java:4424)
E/AndroidRuntime( 1278): 	at java.lang.reflect.Method.invokeNative(Native Method)
E/AndroidRuntime( 1278): 	at java.lang.reflect.Method.invoke(Method.java:511)
E/AndroidRuntime( 1278): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:784)
E/AndroidRuntime( 1278): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:551)
E/AndroidRuntime( 1278): 	at dalvik.system.NativeStart.main(Native Method)
E/AndroidRuntime( 1278): Caused by: android.view.InflateException: Binary XML file line #110: Error inflating class android.support.design.widget.NavigationView
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.createView(LayoutInflater.java:606)
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:680)
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.rInflate(LayoutInflater.java:739)
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.inflate(LayoutInflater.java:489)
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.inflate(LayoutInflater.java:396)
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.inflate(LayoutInflater.java:352)
E/AndroidRuntime( 1278): 	at android.support.v7.app.AppCompatDelegateImplV7.setContentView(AppCompatDelegateImplV7.java:249)
E/AndroidRuntime( 1278): 	at android.support.v7.app.AppCompatActivity.setContentView(AppCompatActivity.java:106)
E/AndroidRuntime( 1278): 	at com.example.activity.MainActivity.onCreate(MainActivity.java:104)
E/AndroidRuntime( 1278): 	at android.app.Activity.performCreate(Activity.java:4466)
E/AndroidRuntime( 1278): 	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1049)
E/AndroidRuntime( 1278): 	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:1920)
E/AndroidRuntime( 1278): 	... 11 more
E/AndroidRuntime( 1278): Caused by: java.lang.reflect.InvocationTargetException
E/AndroidRuntime( 1278): 	at java.lang.reflect.Constructor.constructNative(Native Method)
E/AndroidRuntime( 1278): 	at java.lang.reflect.Constructor.newInstance(Constructor.java:417)
E/AndroidRuntime( 1278): 	at android.view.LayoutInflater.createView(LayoutInflater.java:586)
E/AndroidRuntime( 1278): 	... 22 more
E/AndroidRuntime( 1278): Caused by: java.lang.NoSuchFieldError: android.support.design.R$attr.colorPrimary
E/AndroidRuntime( 1278): 	at android.support.design.widget.NavigationView.createDefaultColorStateList(NavigationView.java:346)
E/AndroidRuntime( 1278): 	at android.support.design.widget.NavigationView.<init>(NavigationView.java:117)
E/AndroidRuntime( 1278): 	at android.support.design.widget.NavigationView.<init>(NavigationView.java:88)
E/AndroidRuntime( 1278): 	... 25 more
W/ActivityManager(   88):   Force finishing activity com.example/.activity.MainActivity
```

Looking into the design library, still using `colorPrimary` as example we can find it used like this, without using the whole package name:
```
tommy at hank in test $ gg --no-index colorPrimary
R.txt
84:int attr colorPrimary 0x7f0100d1
85:int attr colorPrimaryDark 0x7f0100d2
1296:int styleable Theme_colorPrimary 81
1297:int styleable Theme_colorPrimaryDark 82

res/values/values.xml
99:        <item name="android:background">?attr/colorPrimary</item>
105:        <item name="statusBarScrim">?attr/colorPrimaryDark</item>
108:        <item name="statusBarBackground">?attr/colorPrimaryDark</item>
```

tgummerer:
@natthu I took another look at this over the weekend, and set up a robolectric test, and just having your changes the robolectric tests fail for me as well.  It happens because Robolectric tests use transitive resources in the `AndroidLibraryGraphEnhancer`.

When adding the resources from `AndroidResourceHelper.androidResOnly` as dependency, we add the `android_prebuilt_aar#aar_android_resource` flavor as deps to the android resource, but `UnsortedAndroidResourceDeps` uses the `android_prebuilt_aar` directly as dependency.  Therefore we have both the PrebuiltAar, as well as the AndroidResource as dep and get a conflict as they both have the same R.txt file.

Using the following diff fixes the `robolectric_test`, by getting the `android_prebuilt_aar` directly from `androidResOnly`, as it is a wrapper around the AndroidPrebuiltAar is a wrapper around the AndroidResource.

I haven't updated the pull request, as I'm not sure what fix @sdwilsh meant in [his comment](https://github.com/facebook/buck/pull/391#discussion-diff-38474757).  

```diff
diff --git a/src/com/facebook/buck/android/AndroidResourceHelper.java b/src/com/facebook/buck/android/AndroidResourceHelper.java
index c3e1e62..2265d7d 100644
--- a/src/com/facebook/buck/android/AndroidResourceHelper.java
+++ b/src/com/facebook/buck/android/AndroidResourceHelper.java
@@ -21,7 +21,6 @@ import com.google.common.base.Function;
 import com.google.common.base.Predicates;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableSortedSet;
-import com.google.common.collect.Iterables;
 
 import javax.annotation.Nullable;
 
@@ -30,8 +29,8 @@ public class AndroidResourceHelper {
   private AndroidResourceHelper() {}
   /**
    * Filters out the set of {@code android_resource()} dependencies from {@code deps}. As a special
-   * case, if an {@code android_prebuilt_aar()} appears in the deps, the {@code android_resource()}
-   * that corresponds to the AAR will also be included in the output.
+   * case, if an {@code android_prebuilt_aar()} appears in the deps, it will also be included in
+   * the output, as it is basically a wrapper around the {@code android_resource()} it depends on.
    * <p>
    */
   public static ImmutableSortedSet<BuildRule> androidResOnly(ImmutableSortedSet<BuildRule> deps) {
@@ -41,17 +40,9 @@ public class AndroidResourceHelper {
           @Override
           @Nullable
           public BuildRule apply(BuildRule buildRule) {
-            if (buildRule instanceof AndroidResource) {
+            if (buildRule instanceof AndroidResource ||
+                buildRule instanceof AndroidPrebuiltAar) {
               return buildRule;
-            } else if (buildRule instanceof AndroidLibrary &&
-                ((AndroidLibrary) buildRule).isPrebuiltAar()) {
-              // An AndroidLibrary that is created via graph enhancement from an
-              // android_prebuilt_aar() should always have exactly one dependency that is an
-              // AndroidResource.
-              return Iterables.getOnlyElement(
-                  FluentIterable.from(buildRule.getDeps())
-                      .filter(Predicates.instanceOf(AndroidResource.class))
-                      .toList());
            }
            return null;
          }
```

natthu:
Sorry about the delay again. I've been working on cleaning up how we deal with resources inside prebuilt android AARs, which should hopefully make all of this obsolete. I hope to land it soon.

tgummerer:
@natthu there's no rush, thanks for working on this :)

sdwilsh:
4cf0eb7a744c9583cc3f45f7952ef3aa6ae8afae is the commit from @natthu.  If you can rebase past that, I think we just need to make sure your integration tests pass (so this PR can turn into just adding tests).

tgummerer:
@natthu @sdwilsh thanks for fixing this :) I updated the pull request to just include the tests.

The app using the design-library no longer crashes, but when building it a lot of messages like the following are printed by buck (this is only a few of them, totally about 200-300 lines are printed).  I have no idea where this is coming from yet, they aren't defined in the design library, nor in the appcompat library.  I'll try to investigate this a bit more tomorrow, unless you have an idea :)

```
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=adSize, idValue=0x7f010000}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=adSizes, idValue=0x7f010001}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=adUnitId, idValue=0x7f010002}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=appTheme, idValue=0x7f010017}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=buyButtonAppearance, idValue=0x7f01001e}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=buyButtonHeight, idValue=0x7f01001b}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=buyButtonText, idValue=0x7f01001d}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=buyButtonWidth, idValue=0x7f01001c}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=cameraBearing, idValue=0x7f010008}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=cameraTargetLat, idValue=0x7f010009}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=cameraTargetLng, idValue=0x7f01000a}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=cameraTilt, idValue=0x7f01000b}' in the uber R.txt.
Cannot find resource 'RDotTxtEntry{idType=int, type=attr, name=cameraZoom, idValue=0x7f01000c}' in the uber R.txt.
```

natthu:
It's coming from https://github.com/facebook/buck/blob/42d1e6c3757e6430cf687a0db71684dcedba066a/src/com/facebook/buck/android/MergeAndroidResourcesStep.java#L291

I think you can suppress that by setting a `.buckconfig` property.

natthu:
I was wrong. Currently, you cannot suppress these warnings with any override.

tgummerer:
@natthu The missing resources come from the google play libraries.  These libraries are split into multiple different libraries, but every aar seems to contain all the RDotTxtEntries for all of the different libraries, even if the entry doesn't exist in the aar.  I initially thought that there might be a deeper problem, but it seems to be only a UI issue.

The messages can be suppressed by adding a locale argument to the `android_binary`, but that definitely isn't the right solution.

I can see three possible solutions:
 - Add a config option to suppress the warning
 - Suppress the warnings all together
 - Check if the RDotTxtEntry comes from a prebuilt aar, and don't show the warning if it does

I think the first two solutions are mostly peppering over the actual issue, while the third might be too invasive and slow, but I haven't looked into that too much yet.  If nobody gets to it first I'll see if I can find some time over the weekend to look into this.

tgummerer:
I submitted another pull request to get rid of the warning messages (#420).

@sdwilsh I think this pull request and #420 have a conflict, let me know if you want me to rebase either of them.

PR #394, opened 2015-08-12
Elaborate that a genrule could output a folder

rami-khalil:
The docs fail to elaborate that an entire folder full of files could be generated from a genrule.


sdwilsh:
Thanks; the docs should go live soon.

Issue #385, opened 2015-07-23
Exopackage with guava


tgummerer:
I recently tried to add exopackage support to my project, but I'm getting the following error when trying to add guava as a dependency:

```
[2015-07-23 16:09:17.413][error][command:1cfc5dc9-b91f-442e-912c-5c137342d6e0][tid:15][com.facebook.buck.event.listener.LoggingBuildListener] DexWithClasses buck-out/gen/lib/guava#dex.dex.jar with cost 4892294 exceeds the max cost 4194304 for a secondary dex file.
BUILD FAILED: Secondary dex exceeds linear alloc limit.
```

It seems the following diff fixes the error and the application works just fine, but I'm not sure it's the right thing to do to just change the limit.

```diff
diff --git a/src/com/facebook/buck/android/AndroidBinaryDescription.java b/src/com/facebook/buck/android/AndroidBinaryDescription.java
index f892555..0668b80 100644
--- a/src/com/facebook/buck/android/AndroidBinaryDescription.java
+++ b/src/com/facebook/buck/android/AndroidBinaryDescription.java
@@ -73,7 +73,7 @@ public class AndroidBinaryDescription
    * By default, assume we have 5MB of linear alloc,
    * 1MB of which is taken up by the framework, so that leaves 4MB.
    */
-  private static final long DEFAULT_LINEAR_ALLOC_HARD_LIMIT = 4 * 1024 * 1024;
+  private static final long DEFAULT_LINEAR_ALLOC_HARD_LIMIT = 8 * 1024 * 1024;
 
   private static final MacroHandler MACRO_HANDLER =
       new MacroHandler(
```

sdwilsh:
I believe that older versions of Android have only 4MB, which is the problem (I want to say Gingerbread, although @dreiss would likely know best).

tgummerer:
Ah yeah, thanks for clarifying!  A little digging shows you were right about Gingerbread, found this article by @dreiss: https://www.facebook.com/notes/facebook-engineering/under-the-hood-dalvik-patch-for-facebook-for-android/10151345597798920 ;)

Digging a little further, I found the `linear_alloc_hard_limit` option for the `android_binary` rule, which actually does just what I need.

PR #397, opened 2015-08-21
Document linear_hard_alloc_limit

tgummerer:
Test Plan: docs/soyweb-local.sh


tgummerer:
I was wondering about how this worked exactly, as I needed it to solve #385.  Hopefully documenting the argument might help someone else with similar problems in the future.

PR #400, opened 2015-08-27
Some cleanups

fkorotkov:
While checking sources for the first time I found a few minor issues.

### DIFF
--- a/scripts/autofix_lint.py
+++ b/scripts/autofix_lint.py
@@ -277,8 +277,7 @@ class FixJunit3Use(SingleLineFixer):
                 start_index = translate_metadata['start_index']
                 end_index = translate_metadata['end_index']
                 offset_array = translate_metadata['offset_array']
-                if (problem.line >= start_index and
-                        problem.line < end_index):
+                if start_index <= problem.line < end_index:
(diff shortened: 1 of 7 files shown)


fkorotkov:
@facebook-github-bot signed.

scripts/autofix_lint.py
```diff
@@ -277,8 +277,7 @@ def translateFunc(problem):
                 start_index = translate_metadata['start_index']
                 end_index = translate_metadata['end_index']
                 offset_array = translate_metadata['offset_array']
-                if (problem.line >= start_index and
-                        problem.line < end_index):
+                if start_index <= problem.line < end_index:
```
sdwilsh: Today I learned this was valid in Python.  Fancy.

sdwilsh:
I'm just curious if you used a tool to spot these or came across them by hand?

fkorotkov:
I was checking source in IntelliJ and saw some warning in the editor like one you mentioned above. And then I ran additional inspections on all sources and picked a few obvious fixes. There were way more warnings and possible bugs but I don't have so much context on Buck right now to evaluate them.

PR #401, opened 2015-08-28
Fix NullPointerException when traversing the filesystem.

liuyang-li:
Summary:
When we traverse the file system to the root, directory.getFileName()
returns null and we do not want to get toString() on it.

Test:
`buck test`

### DIFF
--- a/src/com/facebook/buck/cli/TestRunning.java
+++ b/src/com/facebook/buck/cli/TestRunning.java
@@ -789,41 +789,50 @@ public class TestRunning {
     Set<String> srcFolders = Sets.newHashSet();
     loopThroughSourcePath:
     for (Path javaSrcPath : javaSrcs) {
-      if (!MorePaths.isGeneratedFile(javaSrcPath)) {
-        // If the source path is already under a known source folder, then we can skip this
-        // source path.
-        for (String srcFolder : srcFolders) {
-          if (javaSrcPath.startsWith(srcFolder)) {
-            continue loopThroughSourcePath;
-          }
-        }
+      if (MorePaths.isGeneratedFile(javaSrcPath)) {
+        continue;
+      }
 
-        // If the source path is under one of the source roots, then we can just add the source
-        // root.
-        ImmutableSortedSet<String> pathsFromRoot = defaultJavaPackageFinder.getPathsFromRoot();
-        for (String root : pathsFromRoot) {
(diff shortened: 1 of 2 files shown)


src/com/facebook/buck/cli/TestRunning.java
```diff
@@ -788,41 +788,50 @@ private static Step getReportCommand(
...
+      if (pathElements.isEmpty()) {
+        continue;
+      }
+
+      while (directory != null && directory.getFileName() != null &&
+          !pathElements.contains(directory.getFileName().toString())) {
+        directory = directory.getParent();
+      }
+
+      if (directory == null && directory.getFileName() == null) {
```
sdwilsh: This is a compilation error since `directory` must be `null` when you reference it the second time (after the `&&`)

liuyang-li: My bad. Should have been an `||` instead. It's updated.

test/com/facebook/buck/cli/TestRunningTest.java
```diff
@@ -165,6 +165,28 @@ public void testNonGeneratedSourceFile() {
...
+
+    DefaultJavaPackageFinder defaultJavaPackageFinder =
+        createMock(DefaultJavaPackageFinder.class);
+    expect(defaultJavaPackageFinder.getPathsFromRoot()).andReturn(pathsFromRoot);
+    expect(defaultJavaPackageFinder.getPathElements()).andReturn(ImmutableSet.<String>of("/"));
+
+    replay(defaultJavaPackageFinder);
+
+    ImmutableSet<String> result = TestRunning.getPathToSourceFolders(
+        javaLibrary, Optional.of(defaultJavaPackageFinder), new FakeProjectFilesystem());
```
sdwilsh: You have an `ant lint` error here because you don't use `result`.  I'll clean this up before landing.

sdwilsh: Hmm, actually, I'm not sure what you wanted to happen here.  That set appears to be empty, which I don't think is the right behavior.

liuyang-li: All the paths are filtered, so the result is empty. I removed variable assignment, as it's not used. What we want to make sure in this test is that the path filtering in `getPathToSourceFolders` works without NPEs, so invoking it is enough here.

PR #402, opened 2015-09-01
Fix build when only empty res directories are present

tgummerer:
Summary:
When an android_binary or android_instrumentation_apk depend on
libraries or android_resources which have empty res directories, aapt
doesn't generate an R.txt file.  Therefore the MergeAndroidResourcesStep
fails with a NoSuchFileException.

Check if the R.txt file exists before trying to read it, and return an
empty list if the file doesn't exist.

Test Plan:
Added integration tests


tgummerer:
I doubt this will happen in `android_binary` rules, but might be fairly common in `android_instrumentation_apk` rules, where the resources from the app are excluded, but things like the `android_prebuilt_aar` for the espresso library are included, which contains an empty res directory.

natthu:
`mini-aapt` generates an empty `R.txt` file if there are no resources.

I'd prefer if we instead of this, add a `TouchStep` after `AaptStep` in `AaptPackageResources`, so that we don't have to special case this, and keep it consistent.

tgummerer:
@natthu that makes sense, thanks!  I updated the pull request.

src/com/facebook/buck/android/AaptPackageResources.java
```diff
@@ -282,6 +287,8 @@ public String getDescription(ExecutionContext context) {
              */
             !skipCrunchPngs /* && packageType.isCrunchPngFiles() */));
 
+    steps.add(new TouchStep(getPathToRDotTxtFile()));
```
natthu: Need a comment here to explain why we need the `TouchStep`.

test/com/facebook/buck/android/AndroidBinaryIntegrationTest.java
```diff
@@ -347,8 +347,22 @@ public void testLibraryMetadataChecksum() throws IOException {
     }
     file.close();
     contents.close();
+  }
+
+  @Test
+  public void testApkEmptyResDirectoriesBuildsCorrectly() throws IOException {
+    workspace.runBuckCommand("build", "//apps/sample:app_no_res").assertSuccess();
```
natthu: nit: you can use `runBuckBuild` instead.

test/com/facebook/buck/android/AndroidBinaryIntegrationTest.java
```diff
@@ -347,8 +347,22 @@ public void testLibraryMetadataChecksum() throws IOException {
...
+    BuckBuildLog buildLog = workspace.getBuildLog();
 
+    buildLog.assertTargetBuiltLocally("//apps/sample:app_no_res");
```
natthu: Since this is a fresh build, this is bound to be true. This isn't testing anything. Please remove.

test/com/facebook/buck/android/AndroidBinaryIntegrationTest.java
```diff
@@ -347,8 +347,22 @@ public void testLibraryMetadataChecksum() throws IOException {
...
+  }
+
+  @Test
+  public void testInstrumentationApkWithEmptyResDepBuildsCorrectly() throws IOException {
+    workspace.runBuckCommand("build", "//apps/sample:instrumentation_apk").assertSuccess();
+    BuckBuildLog buildLog = workspace.getBuildLog();
 
+    buildLog.assertTargetBuiltLocally("//apps/sample:instrumentation_apk");
```
natthu: Same.

sdwilsh:
I'll fix those so I don't have to import again.

tgummerer:
Thanks @natthu for the review and @sdwilsh for cleaning up after me :)

PR #411, opened 2015-09-10
Don't enforce a maven URL has 6 parts since one is optional

rowillia:


src/com/facebook/buck/file/MavenUrlDecoder.java
```diff
@@ -66,7 +66,7 @@ public static URI toHttpUrl(Optional<String> mavenRepo, URI uri) {
 
     Matcher matcher = URL_PATTERN.matcher(uri.getSchemeSpecificPart());
 
-    if (!matcher.matches() || matcher.groupCount() != 6) {
+    if (!matcher.matches()) {
```
sdwilsh: I guess writing a test for this would be rather difficult, wouldn't it?

sdwilsh:
@shs96c and I would like to see a test for this

rowillia:
Fiiiiiiiiiiine....added a test

src/com/facebook/buck/file/MavenUrlDecoder.java
rowillia: Added a test

PR #412, opened 2015-09-10
Send missing symbols to stderr

rowillia:
It makes sense to send this to stderr instead of stdout since it's an error condition.  This makes parsing this output easier.

### DIFF
--- a/src/com/facebook/buck/cli/MissingSymbolsHandler.java
+++ b/src/com/facebook/buck/cli/MissingSymbolsHandler.java
@@ -221,6 +221,6 @@ public class MissingSymbolsHandler {
   }
 
   private void print(String line) {
-    console.getStdOut().println(console.getAnsi().asWarningText(line));
+    console.getStdErr().println(console.getAnsi().asWarningText(line));
   }
 }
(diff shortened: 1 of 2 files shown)


sdwilsh:
`com.facebook.buck.java.MissingSymbolsHandlerIntegrationTest` now fails with this change.  Can you patch it up to deal with the change in output?

src/com/facebook/buck/cli/MissingSymbolsHandler.java
```diff
@@ -221,6 +221,6 @@ private String formatTarget(BuildTarget buildTarget) {
   }
 
   private void print(String line) {
-    console.getStdOut().println(console.getAnsi().asWarningText(line));
+    console.getStdErr().println(console.getAnsi().asWarningText(line));
```
andrewjcg: Thanks for the patch!  However, the new way to report stdout/stderr is via the `BuckEventBus` object (so that it interacts properly with the super console):

  buckEventBus.post(ConsoleEvent.warning(line));

The constructor already gets access to the event bus object, so this should be an easy change.  Also, it's probably worth batching all the output together into a single message, rather than sending each line individually.

rowillia: Unfortunately this doesn't work :/

SuperConsole does a dirty check before every render, and the fact that the build failed has already dirtied the console.  We need SuperConsole to keep printing `ConsoleEvents` even after stderr is dirty.

rowillia:
@sdwilsh fixed tests.

PR #413, opened 2015-09-10
Support specifying multiple Maven repositories in a .buckconfig

rowillia:
Some android repositories support more than one one Maven repository at a time.  This allows us to declare them all in one place.


shs96c:
I've pinged Roy with an alternative proposal.

PR #426, opened 2015-09-21
Make python extension that depend on cxx_libraries work.

mikekap:
We do this by adding a "-rpath @loader_path/" to the link flags of a cxx_python_extension.


Coneko:
@andrewjcg: You might want to take a look at this.

PR #439, opened 2015-09-30
Support for groovy_library and groovy_test

mread:
A somewhat speculative pull request to introduce groovy_library and groovy_test rules. This is an unpleasant cut'n'paste of the Java rules with a handful of strings changed to .groovy and a load of Java-centric features removed.

What would it take to get this merged into upstream?


.idea/codeStyleSettings.xml
```diff
@@ -21,6 +21,10 @@
           </value>
         </option>
         <option name="RIGHT_MARGIN" value="100" />
+        <Python>
+          <option name="SPACE_AROUND_EQ_IN_NAMED_PARAMETER" value="true" />
+          <option name="SPACE_AROUND_EQ_IN_KEYWORD_ARGUMENT" value="true" />
+        </Python>
```
sdwilsh: Nice, but unrelated to this change.  Want to pull it out?

mread: Done.

sdwilsh:
Is it possible for us to have some concrete base class that both Java and this could use?  This might make it easier for someone like @bkase to support Kotlin.

mread:
I'm working on pulling some of the functionality up to base class. What did you mean by concrete base class? As in not abstract? I don't quite understand when you'd actually want to use it rather than one of its implementations?

sdwilsh:
An `abstract` class would be correct.  I work in too many languages sometimes and used a horribly misleading word there.

mread:
@sdwilsh, I keep hitting deadends trying to break out some common "jvm-language" functionality so that Java, Groovy, Kotlin etc can be supported with minimal code duplication. I think the code needs a bit of re-organisation first. There's a few issues that I keep on butting up against. E.g. JavaLibrary...
image
Things like HasClasspathEntries and HasSources don't really seem to fit. HasClasspathEntries is used indirectly by almost everything. When I make its methods return something more abstract it ends badly deep within AndroidBinaryGraphEnhancer.

I'm happy to have a go at this but only if you're sure a pull-request is the right way to go about it. Changes would probably cause major merge headaches. A small price to pay for the merge hell I go through keeping this Groovy support working in a fork though.

Coneko:
Pull request is the right way. If there is some substantial cleanup to do create smaller pull request we can merge in early in the process so you don't have to keep remerging the changes.

If you have some particular issues with how the code is set up and have some high level idea on how to fix it we can discuss it.

The java code is not the best as far reusability goes because it wasn't designed with other JVM languages in mind.

mikekap:
+1 from me. As soon as adding a JVM language is easy, you can expect a PR for Scala support. I'm currently maintaining https://github.com/mikekap/buck/commit/334783ab1b80525e2601caea167073140ff57902 as a copy-paste of java because the refactoring work is a bit daunting.

mread:
> If there is some substantial cleanup to do create smaller pull request we can merge in early in the process so you don't have to keep remerging the changes.

Here's a structure which breaks up the `com.facebook.buck.java` package to make it easier to add support for additional JVM languages

* c.f.b.java - perhaps renamed to jvmlang
  * lang-specific, e.g. java, groovy, scala, kotlin
    * lang-specific build rule code, e.g. java_library, java_binary
    * lang-specific shared implementation, e.g. javac related code
  * supporting rules relevant for several languages, e.g. jar packaging, source jars
  * supporting shared implementation relevant for several languages, e.g. classpath and dependency wrangling

If you think this is ok then we could do this as a single pull request although our preference would be several smaller increments to ensure that we don't cause merge problems.

What do you think?

Coneko:
Sounds reasonable, something like:

`c.f.b.jvm.core` (including shared classes for rules, classpath and dependencies)
`c.f.b.jvm.java` (java specific)
`c.f.b.jvm.groovy` (groovy specific)

and so forth.

Buck doesn't allow circular dependencies, so care must be take to avoid them: `core` shouldn't reference the other packages in `jvm`.

We would also prefer smaller increments, it's easier to review.

@shs96c, @marcinkosiba You guys know more about jvm languages, how does that sound at a high level?

shs96c:
The code for handling java and android has grown somewhat (umm...) organically. Cleaning it up is an excellent idea, and the suggested structures make sense to me. Would be happy to hang out on IRC to help coordinate this, since there are a few of us who are interested. Or I can just review PRs :)

Coneko:
Ok, back to you @mread.

mread:
IRC sounds like a good idea. Sorry if I'm missing the obvious - which server/channel?

Coneko:
#buck on freenode

Coneko:
I assume this PR is being superseded by the smaller PRs coming in, can we close it?

mread:
Agreed. Closing.

PR #441, opened 2015-09-30
Add output for `prebuilt_android_aar`

rowillia:
--show_output currently returns null for AARs.  This fixes it.


src/com/facebook/buck/android/UnzipAar.java
```diff
@@ -203,7 +213,7 @@ public BuildOutput initializeFromDisk(OnDiskBuildInfo onDiskBuildInfo) throws IO
   @Override
   @Nullable
```
sdwilsh: This is no longer `Nullable`

test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
```diff
@@ -73,6 +74,16 @@ public void testAndroidPrebuiltAarInDepsIsExported() throws IOException {
   }
 
   @Test
+  public void testAndroidPrebuiltAarHasOutput() throws IOException {
+    ProjectWorkspace.ProcessResult targetsResult = workspace.runBuckCommand(
+        "targets",
+        "--show_output",
+        "//:aar");
+    targetsResult.assertSuccess();
+    assertEquals("//:aar buck-out/gen/aar#aar_unzip.aar", targetsResult.getStdout().trim());
```
sdwilsh: You hate Windows, don't you? :P

src/com/facebook/buck/android/AndroidPrebuiltAar.java
```diff
@@ -119,6 +121,12 @@ public Path getBinaryJar() {
     return prebuiltJar.getPathToOutput();
   }
 
+  @Nullable
```
sdwilsh: This shouldn't be `Nullable`

sdwilsh:
I'll fix those comments up before merging.

test/com/facebook/buck/android/AndroidPrebuiltAarIntegrationTest.java
rowillia: With a burning passion.

src/com/facebook/buck/android/UnzipAar.java
```diff
@@ -203,7 +213,7 @@ public BuildOutput initializeFromDisk(OnDiskBuildInfo onDiskBuildInfo) throws IO
...
   public Path getPathToOutput() {
-    return null;
+    return copiedAar;
```
andrewjcg: It seems odd that we're listing essentially a copy of the input file as the output.  Shouldn't the output of `UnzipAar` be `unpackDirectory`?

src/com/facebook/buck/android/AndroidPrebuiltAar.java
```diff
@@ -119,6 +121,12 @@ public Path getBinaryJar() {
...
+  @Override
+  public Path getPathToOutput() {
+    return unzipAar.getPathToOutput();
```
andrewjcg: It feels a bit round-about to use the `UnzipAar` rule to thread through the original AAR (but it's likely I'm missing something).  Can we just have the constructor of this rule take the orignal AAR `SourcePath` and list that as the output without touching `UnzipAar` (and also propagate the appropriate runtime dep on any rule that generates the original AAR)?

rowillia: The reason I went down this path is `JavaBinary` followed a similar pattern of first copying the input jar to `buck-out` before doing anything with it.

src/com/facebook/buck/android/UnzipAar.java
rowillia: Yeah, it does feel a little janky.  I guess we could graph enhance another step in there to copy the input first and then pass that to the prebuilt jar and aar?

sdwilsh: That would probably be best.

Coneko:
Back to @andrewjcg for @rowillia's comment.

facebook-github-bot-6:

styurin:
@rowillia this PR is pretty old. Please, rebase and resubmit it if you want to merge these changes in.

PR #446, opened 2015-10-02
Avoid running ant when buck is already built

tgummerer:
Summary:
When building buck using ant, no "successful-build" file is created in
the build directory.  Because buck checks this file to determine whether
to invoke ant or not to build buck from source when it is executed, ant
is unnecessarily invoked again when buck is run, when buck was build
directly using ant.

Instead of writing the file after the python wrapper executes ant, write
the file when ant compiles buck successfully.

Test Plan:
Build buck using ant, and then build a project using buck

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -162,7 +162,7 @@
 
   <!-- Default target so that a command line build can
        do more than one thing. -->
-  <target name="default" depends="compile, dx, client, javac-tracing, report-generator-jar" />
+  <target name="default" depends="start-build, compile, dx, client, javac-tracing, report-generator-jar, build-successful" />
 
   <target name="checkversion">
     <!-- The nio Files class was only introduced in Java 7. -->
@@ -992,4 +992,16 @@
 
   <target name="pre-checkin" depends="clean, default, compile-tests, lint"/>
   <target name="all" depends="jar, test, javadoc, client" />
+
+  <target name="start-build"
+          description="Remove build successful marker before building">
+    <delete file="${build.dir}/successful-build" />
+  </target>
+
+  <target name="build-successful"
+          depends="compile"
+          description="Mark the build as successful">
+    <touch file="${build.dir}/successful-build" />
+  </target>
+
 </project>
--- a/programs/buck_repo.py
+++ b/programs/buck_repo.py
@@ -260,7 +260,6 @@ class BuckRepo(BuckTool):
                 ant = self._check_for_ant()
                 self._run_ant_clean(ant)
                 self._run_ant(ant)
-                open(self._build_success_file, 'w').close()
                 print("All done, continuing with build.", file=sys.stderr)
 
     def _has_resource(self, resource):


build.xml
```diff
@@ -988,4 +988,11 @@
 
   <target name="pre-checkin" depends="clean, default, compile-tests, lint"/>
   <target name="all" depends="jar, test, javadoc, client" />
+
+  <target name="build-successful"
+          depends="compile"
+          description="Mark the build as successful">
+    <touch file="${build.dir}/successful-build" />
```
sdwilsh: Don't we need to remove this file when we start a build in case it fails?

Coneko:
@bhamiltoncx tried to do this once, but I think there were complications with the launcher script or something.

tgummerer:
@Coneko I tried digging in the history a bit, and @natthu basically did the reverse thing I did in 501b31.  Shouldn't the `success` target in that commit, or the `build-successful` target in my commit only be run in case all the other dependencies in the `default` target are run before?  Am I missing something here?

build.xml
tgummerer: We don't need to remove it to match current behaviour.  After running buck in any project, the build-successful file is there, and when buck is rebuilt with ant, it stays there even if the build fails.  It is only removed when buck is updated to a new revision from git.

However I think it would make sense to remove it unconditionally before trying to build buck with ant anyway, so when the build fails a more sensible error message is shown when running buck.

sdwilsh:
Yeah, @natthu did the reverse in 501b311bd361f72394aae575a7935385996dc000.  Looking at internal conversations about this (bleh), the concern of not doing this was that `ant clean` could fail, and we wouldn't remove this (in part because sometimes users would run `buck` as root, which messes up everything.  @ttsugriy has tried to fix that a few times but ran into issues.  I think we can't take this until we get a resolution on that.

tgummerer:
Ah ok, thanks for the explanation.

sdwilsh:
I think @ttsugriy is taking one more stab at it, so all you'd have to do is update this to do the right thing when `ant clean` is run.

tgummerer:
Hrm not sure I follow, I'd have to update once @ttsugriy's change lands, or do you mean your comment about deleting the `build-successful` file before starting the build from above?

sdwilsh:
The latter.  @ttsugriy's change just checks to make sure Buck isn't running as some user it shouldn't (with some smarts).

ttsugriy:
Sigh... Yeah I tried preventing running buck by users with elevated privileges, but first attempt had this check only in Main.java which was too late to prevent temporary files from being created, which, unfortunately, also result in inability to invoke buck as "normal" user afterwards. My second attempt was to add this check in buck (bash-only) entry point, but due to some internal issues we had to roll it back... Hopefully the change will find its way to github repo by the end of this week.

tgummerer:
Thanks for the explanation!

In the meantime I updated the pull request, according to the comments.

Coneko:
This was done in 707646a84826e59c45525e2a1db72837d42f99ca.

PR #449, opened 2015-10-05
Add some support for running go tests.

mikekap:
This isn't as complete as the `go test` command, mostly because `go test` "magically" recompiles the package under test. The closest
you can get to `go test`-like functionality is via duplication, as mentioned in the docs. It might be nice to make this easier via
param or different test rule (`go_internal_test`) that just does the src/dep copying for you. We may be able to make this even more
amazing via some sort of auto-discovery (we can easily tell if the test is external or internal at test main generation time) and
create two different libs. But I digress; this is a v1.

This also fixes support for custom package names (was previously somewhat broken).

The slightly more controversial feature here is the test main generator. Particularly, the source (in go) is a) mostly ripped from
https://github.com/golang/go/blob/master/src/cmd/go/test.go & b) checked in as the source. In order to actually use it at run-time, it
creates a "fake" test target //__internal_buck:go_test_main_gen. This is mostly to allow you to build buck without having go
installed. I'm not sure of a better way to do this, so any ideas are appreciated.

Tested via the integration tests - they work :)


mikekap:
I've also looked at making code coverage support work for go. This isn't particularly hard on the Go side; it just requires running `go tool cover` on the source files before compiling them, and running the test main generator with some options. Unfortunately I couldn't figure out how to get buck to build a different target when code coverage is enabled. Does anyone have a hint on how I might do that?

sdwilsh:
You'll want to run `ant lint` on this as it currently fails.

As for code coverage, it's set on the `ExecutionContext`; can you use that to do the right thing?

mikekap:
Done. Also seems like some cell stuff changed, which made the test main gen target even uglier :X

I saw the code coverage switch in ExecutionContext - but correct me if I'm wrong - I think it's impossible to change the flavor of dependencies at that point (i.e. the only way to change deps/flavors is at `createBuildRule` time).

sdwilsh:
Oh, you need to change things while building the action graph.  We'd have to change that code to feed it through to the descriptions, which probably isn't awful, but it'd be a large change.  I wouldn't do it as a part of this PR.

mikekap:
Friendly ping :). I don't intend to implement coverage support in this one.

sdwilsh:
I've been on vacation this whole week and haven't been around to nag people for reviews.  I'm only doing a half day today and I'm back 100% next week.  I'll try to poke the reviewers now though.

sdwilsh:
I can tell you two things though:
* Your new file doesn't pass the license test (run `buck test -f LicenseCheckTest`)
* Your new test doesn't have an assumption check to make sure the needed go binary is available, so if it isn't, it fails.  Rust and D both have code to check for the needed binaries before running tests so they don't fail on systems that don't have them.

Can you get those two things fixed up while I prod for the rest of the code to be reviewed?

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -33,9 +41,24 @@
 import com.google.common.collect.ImmutableSortedSet;
 
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 
 abstract class GoDescriptors {
+  private static final Path PATH_TO_TEST_GEN =
```
Coneko: This is how we used to add resources in the past, now we specify them in `resources` in the `java_library` in the BUCK file.

See `build-with-buck.st` for example.

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -127,4 +147,84 @@ static GoLibrary createGoLibraryRule(BuildRuleParams params, BuildRuleResolver r
...
+  static synchronized Tool getTestMainGenerator(
+      GoBuckConfig goBuckConfig, BuildRuleResolver resolver, CxxPlatform cxxPlatform,
+      ProjectFilesystem projectFilesystem, Function<Optional<String>, Path> cellRoots,
+      RuleKeyBuilderFactory ruleKeyBuilderFactory) {
+    Optional<Tool> configTool = goBuckConfig.getGoTestMainGenerator(resolver);
+    if (configTool.isPresent()) {
+      return configTool.get();
+    }
+
+    BuildTarget testGenTarget = BuildTarget.builder(
```
Coneko: When we need generated build targets we use the existing target and add flavours to it.

You should not change the base path or name, as there is no way to guarantee the user hasn't named a directory or a target with that name.

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -127,4 +147,84 @@ static GoLibrary createGoLibraryRule(BuildRuleParams params, BuildRuleResolver r
...
+        library,
+        goBuckConfig.getGoLinker().get(),
+        ImmutableList.<String>builder()
+            .addAll(goBuckConfig.getLinkerFlags())
+            .addAll(linkerFlags)
+            .build());
+  }
+
+  static synchronized Tool getTestMainGenerator(
+      GoBuckConfig goBuckConfig, BuildRuleResolver resolver, CxxPlatform cxxPlatform,
```
Coneko: One argument per line.

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -127,4 +147,84 @@ static GoLibrary createGoLibraryRule(BuildRuleParams params, BuildRuleResolver r
         goBuckConfig.getGoCompiler().get());
 
   }
+
+  static GoBinary createGoBinaryRule(BuildRuleParams params, BuildRuleResolver resolver,
+      GoBuckConfig goBuckConfig, CxxPlatform cxxPlatform, ImmutableSet<SourcePath> srcs,
```
Coneko: One argument per line.

src/com/facebook/buck/go/testmaingen.go
```diff
@@ -0,0 +1,437 @@
...
+ * under the License.
+ */
+
+// Generates testmain.go files from a group of test files.
+//
+// Regular go tests (ones run with `go test`) don't actually define a main
+// package. Moreover, Go's reflection does not have the ability to inspect
+// packages (e.g. list functions). This script generates a main.go that
+// runs some set of tests passed in on the CLI. The code liberally borrows from
+// the `go test` implementation at https://github.com/golang/go/blob/master/src/cmd/go/test.go
```
Coneko: Can we use the upstream file? That makes it easier to merge in updates from upstream.

src/com/facebook/buck/go/GoTestDescription.java
```diff
@@ -0,0 +1,172 @@
...
+  @SuppressFieldNotInitialized
+  public static class Arg implements HasSourceUnderTest {
+    public ImmutableSet<SourcePath> srcs;
+    public Optional<String> packageName;
+    public Optional<List<String>> compilerFlags;
+    public Optional<List<String>> linkerFlags;
+    public Optional<ImmutableSortedSet<BuildTarget>> deps;
+    public Optional<ImmutableSet<String>> contacts;
+    public Optional<ImmutableSet<Label>> labels;
+    @Hint(isDep = false) public Optional<ImmutableSortedSet<BuildTarget>> sourceUnderTest;
```
Coneko: Don't add `sourceUnderTest`, instead add `tests` to `go_library` and `go_binary`.

src/com/facebook/buck/go/GoTestDescription.java
```diff
@@ -0,0 +1,172 @@
...
+        new SourcePathResolver(resolver),
+        testMain,
+        args.labels.get(),
+        args.contacts.get(),
+        resolver.getAllRules(args.sourceUnderTest.get()),
+        args.runTestSeparately.or(false));
+  }
+
+  @SuppressFieldNotInitialized
+  public static class Arg implements HasSourceUnderTest {
```
Coneko: Don't implement `HasSourceUnderTest`.

Coneko:
Looks ok, some parts of the code are based on patterns we are deprecating, mainly using system properties for resources and source under test.

Back to you.

src/com/facebook/buck/go/GoDescriptors.java
mikekap: So here's my issue - the test main generator does not vary per test. It felt weird to compile N test main generators when I only need one.

I can obviously compile N, so if you'd like that, I can easily change it.

src/com/facebook/buck/go/testmaingen.go
mikekap: Unfortunately not - the file itself depends on almost all other files in that directory. Most of those aren't needed for the test main generation, but they live in the same file. Bazel did something very similar, but removed some features (coverage & test case matching). See https://github.com/bazelbuild/bazel/blob/master/tools/build_rules/go/tools/generate_test_main.go .

I can alternatively try to not copy that code and write something buck-specific.

src/com/facebook/buck/go/GoTestDescription.java
sdwilsh: (see the `HasTests` interface for those)

src/com/facebook/buck/go/GoDescriptors.java
Coneko: Ah, I see, we don't have a namespace reserved for generated targets, but maybe we should.

One thing is that it's a bit confusing to have a target show up out of nowhere.

@shs96c Any opinions on this?

src/com/facebook/buck/go/testmaingen.go
Coneko: No that's fine. Just copy the license header from there, add it after the Facebook Apache 2 license:

```
// Copyright 2011 The Go Authors.  All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.
```

And copy the `LICENSE` file into the same directory, and add it as a resource to the `java_library`.

mikekap:
I addressed all comments but the test main generator pseudo-rule. Would it just be easier to switch to a generator per rule for now and leave a TODO to fix it at some point?

sdwilsh:
Yeah, that's safer to do, so let's do that.

Coneko:
@facebook-github-bot import

src/com/facebook/buck/go/GoSymlinkTree.java
```diff
@@ -39,10 +40,9 @@
 
   protected GoSymlinkTree(
```
andrewjcg: Obviously, a but outside the scope of this diff, but why can't we just reuse `SymlinkTree` here?  The only different seems that we want to create the links via walking the deps, but this looks like it could easily be done in a method that does the walk to build up the link map, then just news up a `SymlinkTree`.

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -16,26 +16,50 @@
...
+import java.net.URL;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 
 abstract class GoDescriptors {
+
+  private static final String TEST_MAIN_GEN_PATH = "testmaingen.go";
+
+  public static final Function<String, Path> TO_GO_PACKAGE_NAME = new Function<String, Path>() {
```
andrewjcg: We already have this as `MorePaths.TO_PATH`.

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -48,14 +72,13 @@ static BuildTarget createTransitiveSymlinkTreeTarget(BuildTarget source) {
         .build();
   }
 
-  static GoSymlinkTree requireDirectSymlinkTreeRule(BuildRuleParams sourceParams,
```
andrewjcg: We use the "require" method naming convention for methods which only create the build if it doesn't already exist.  Since this one (and others below) just unconditionally always creates a new build rules, use "create" instead.

src/com/facebook/buck/go/GoSymlinkTree.java
mikekap: SymlinkTree used SourcePaths, which I thought were generally from source files. In this case all the entries are build rule outputs. Is it still encouraged to use SymlinkTree with BuildTargetSourcePath in that case?

mikekap:
Addressed all of the above, sans the symlink tree.

scripts/travisci_run.sh
```diff
@@ -15,7 +15,7 @@ export TERM=dumb
 # Go tests in Travis
 ./bin/buck test \
   --num-threads=3 \
-  --test-selectors '!GoBinaryIntegrationTest'
+  --test-selectors '!GoBinaryIntegrationTest,!GoTestIntegrationTest'
```
sdwilsh: This did not work as expected (the build failed with both of these failing).  Based on the help text, I think you can just pass these both like so: `--test-selectors '!GoBinaryIntegrationTest' '!GoTestIntegrationTest'`

sdwilsh: Alternatively, we can just add a new file like `scripts/travisci_test_filters.txt` and have one test per line on there and pass `--test-selectors @scripts/travisci_test_filters.txt` here.

sdwilsh:
Should be able to get this merged in today if you get the CI passing :)

You might also want to rebase as I just got a few more changes to the CI out (shouldn't actually impact your diff, but we cover more things now).

Coneko:
> Error: Unable to access jarfile /home/travis/android-sdk-linux/tools/proguard/lib/proguard.jar

That doesn't seem related to the PR, let me import it and run it on out internal CI.

Coneko:
@facebook-github-bot import

sdwilsh:
@facebook-github-bot shipit

sdwilsh:
Whoops; I forgot to fixup the commit message before pushing the changes.  Let me fix that...

PR #463, opened 2015-10-14
Refactor: remove getAbiKey() method

mread:
Commited on top of #461, I'll rebase this once that's merged.

This removes some code that I think is now obsolete as input-based rule keys invented a new mechanism for depending on the ABI of a rule. This is a small step towards decoupling the JavaLibrary interface from everything as part of the refactoring in #439.


Coneko:
Looks good, I'll try to cherry pick it.

PR #471, opened 2015-10-16
Docs: remove deps parameter for genrules

tgummerer:
Summary:
The deps parameter for genrules is not respected, and should therefore
not be documented, as it might confuse users otherwise.  The deps are
instead pulled out of the cmd, cmd_exe or bash parameters, if location,
exe or classpath macros are given.

Test Plan: ./docs/soyweb-local.sh

### DIFF
--- a/docs/__genrule_common.soy
+++ b/docs/__genrule_common.soy
@@ -103,16 +103,3 @@
   {/param}
 {/call}
 {/template}
-
-/***/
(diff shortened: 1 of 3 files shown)


Coneko:
Thanks! I removed the `deps` from the example as well.

tgummerer:
Thanks, totally forgot that.

PR #472, opened 2015-10-19
[Publisher/PMD] Refactor publisher to workaround a PMD issue.

liuyang-li:
Summary:
The version of PMD that we use dose not handle varargs properly, and hit
errors while processing them. We update the varargs to Lists. This also
avoids arrays and array-list conversions. There is no functional changes
in the change set.

Test:
ant clean lint && ant && buck test


sdwilsh:
Ha, is this why lint always dies on this file?  I've been meaning to look into this, but haven't had the cycles.

liuyang-li:
Yes. I think PMD went into infinite recursion trying to deal with the varargs. This fixes it.

PR #478, opened 2015-10-24
Include PrebuiltJar in transitive deps to allow running inside inttelij

strangemonad:
NOTE: this is just a sketch of the fix but has issues. There are no tests and the generated IJModule includes the transitive jars (IjLibraries) as compile time dependencies rather than runtime dependencies. I've tested that this allows running main() and junit tests from inside Ij
on a moderate size personal project (~70 buck targets, ~140 mvn dependencies, ~60k lines of .java excluding comments and 1k lines of .proto)

Summary: The classpath Ij generates when trying to run a module is incorrect because the IjModule doesn't include runtime dependencies of the libraries it depends on.

Prebuilt jars are often not fat (shaded) jars. The build targets listed in the 'deps' option of the 'prebuilt_jar' must be jars included on the runtime classpath. For example, imagine a java_app 'myapp' that depends on jetcd (https://github.com/justinsb/jetcd) a java / netty client for etcd. The jar might look something like:

```
    prebuilt_jar(
      name = 'jetcd',
      deps = [
        '//lib:jackson-core',
        '//lib:jackson-databind',
        '//lib:netty-all',
        '//lib:slf4j-api',
      ],
      binary_jar = ':jetcd_lib',
      visibility = ['PUBLIC'],
    )
```

Both `buck run` and `buck audit classpath` produce the correct result, including jackson, netty and slf4j on the classpath.

Bug: ExportedDepsClosureResolver#getExportedDepsClosure called by IjModuleGraph.from doesn't handle the PrebuildJar node type.

### DIFF
--- a/src/com/facebook/buck/jvm/java/intellij/ExportedDepsClosureResolver.java
+++ b/src/com/facebook/buck/jvm/java/intellij/ExportedDepsClosureResolver.java
@@ -18,6 +18,7 @@ package com.facebook.buck.jvm.java.intellij;
 
 import com.facebook.buck.android.AndroidLibraryDescription;
 import com.facebook.buck.jvm.java.JavaLibraryDescription;
+import com.facebook.buck.jvm.java.PrebuiltJarDescription;
 import com.facebook.buck.model.BuildTarget;
 import com.facebook.buck.rules.TargetGraph;
 import com.facebook.buck.rules.TargetNode;
@@ -63,6 +64,9 @@ public class ExportedDepsClosureResolver {
       AndroidLibraryDescription.Arg arg =
           (AndroidLibraryDescription.Arg) targetNode.getConstructorArg();
       exportedDeps = arg.exportedDeps.get();
+    } else if (targetNode.getType().equals(PrebuiltJarDescription.TYPE)) {
+      PrebuiltJarDescription.Arg arg = (PrebuiltJarDescription.Arg) targetNode.getConstructorArg();
+      exportedDeps = arg.deps.get();
     }
 
     ImmutableSet<BuildTarget> exportedDepsClosure = FluentIterable.from(exportedDeps)


Coneko:
@facebook-github-bot import

marcinkosiba:
Hi there! I see the problem you're trying to address, but just adding the classes to the exported deps makes them also available in the auto-completion index (so you end up auto-completing classes that are missing from your deps and you get a compile error when you build with buck). Luckily IJ has the concept of 'add this to classpath but don't add it to the autocomplete index' (Runtime dep type in the 'Dependencies' tab). Currently the project generator doesn't use this type, but it shouldn't be hard to add (it would be a new IjModuleGraph.DependencyType)

An alternative would be to use the buck plugin (the one under `src/com/facebook/buck/intellij/plugin/`) to run tests via buck.

strangemonad:
@marcinkosiba yup, I think we agree mostly. I only want the transitive lib to be included in the IjModule's runtime classpath, this patch doesn't do that but I'm wondering how hard it would be to do that? I'm basically trying to get the IJ classpath when running a target to match the 'buck run' classpath.

Thanks for pointing out the the intellij plugin in the sources. I only knew about this 3rd party library: https://github.com/wangyanxing/buck_idea_plugin. I spent about 30 mintues trying to get `src/com/facebook/buck/intellij/plugin/` working but the state of the plugin's build seems to be questionable?

petrumarius:
Hey @strangemonad,

Building the plugin with buck is kinda broken at the moment. However, in $buck_src_dir/src/.../intellij/plugin there is a full IntelliJ project (you don't need to buck project that, just use the way it is) and you should be able to build the plugin with IntelliJ (eg, Build -> Prepare plugin module ...). 

Running tests with the plugin is not yet supported, there was some work on this, but is definitely not ready.

strangemonad:
@petrumarius. thanks. I'll take a look at that approach for the plugin

marcinkosiba:
no updates for over a year, closing

PR #479, opened 2015-10-25
Allow using go rules when the Go root is outside the "normal" location.

mikekap:
"normal" in this case means /usr/local/

Introduce the concept of a GoTool - just a Tool that needs GOROOT set. The go compiler/linker are these kinds of tools.

This should let travis tests work for go.

### DIFF
--- a/scripts/travisci_run.sh
+++ b/scripts/travisci_run.sh
@@ -14,8 +14,7 @@ export TERM=dumb
 # Until we get https://github.com/facebook/buck/pull/479, we cannot run the
 # Go tests in Travis
 ./bin/buck test \
-  --num-threads=3 \
-  --test-selectors '!GoBinaryIntegrationTest|GoTestIntegrationTest'
+  --num-threads=3
 
 # Run all the other checks with ant.
 ant travis
--- a/src/com/facebook/buck/go/GoTool.java
+++ b/src/com/facebook/buck/go/GoTool.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2015-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.go;
+
+import com.facebook.buck.rules.BuildRule;
+import com.facebook.buck.rules.RuleKeyBuilder;
+import com.facebook.buck.rules.SourcePath;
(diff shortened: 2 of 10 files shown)


sdwilsh:
I really wish github handled stacked diffs better...

src/com/facebook/buck/go/GoTool.java
```diff
@@ -0,0 +1,61 @@
...
+import com.facebook.buck.rules.RuleKeyBuilder;
+import com.facebook.buck.rules.SourcePath;
+import com.facebook.buck.rules.SourcePathResolver;
+import com.facebook.buck.rules.Tool;
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+
+import java.nio.file.Path;
+
+public class GoTool implements Tool {
```
andrewjcg: So the only reason we need this is to set the `GOROOT` environment variable?  Env vars seem general enough that it feels we should support those in the `Tool` abstraction (via a `getEnvironment()` getter or something), and avoid adding this class to support it.

sdwilsh:
Good news: we've got CI again!
Bad news: I had to disable the Go integration tests to get it passing because we need this PR (and its dep).  I referenced this PR so you can see the commit above.  When you rebase this (because you'll have to), could you re-enable the test too?  That'll be a good way to verify that the fix works too :)

src/com/facebook/buck/go/GoTool.java
mikekap: Agreed. I can do a separate CR to add that, but since Tool is so widely used, I have a few choices:
 - Attempt to fix all the places that call getCommandPrefix to also pass-through getEnvironment. There's about 90 places where it's used, and this will likely be merge hell for you guys (or me).
 - Add the abstraction, but leave all the places that use them in a buggy state. Not great, but leaves a bunch of surprising tech debt.
 - Add environment variables as a new Tool builder that basically defers to `/usr/bin/env X=Y ...` to set the env variables. Easiest, but not particularly clean.

Thoughts?

mikekap:
Looks like travis does indeed work with this change :)

scripts/travisci_run.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -eux
 
+# Install go 1.5
+eval "$(gimme 1.5)"
```
sdwilsh: It would be prefered if you did this in the `.travis.yml` file where we also setup Android: https://github.com/facebook/buck/blob/master/.travis.yml#L18

sdwilsh:
Can you rebase this so that it is easier to review please?  Your other PR is in :)

src/com/facebook/buck/go/GoTool.java
sdwilsh: I think the first option is best.  We cannot rely on `/usr/bin/env` since that doesn't exist on Windows.  I'll let you do that in a follow-up, and we'll be sure to get it in quick so we don't have to worry about merge hell.

sdwilsh:
I'm going to wait for the Travis results, but I suspect it'll be fine.

sdwilsh:
I don't know why the JDK7 tests fails in Travis.  It doesn't in my test repo.  Clearly I should just remove it.

sdwilsh:
@facebook-github-bot shipit

PR #480, opened 2015-10-26
[Test] Skip JNI test for MacOS.

liuyang-li:
Summary:
The assumption for the JNI folder at java home only works reliably for
linux systems. For example, for MacOS, some of the headers are under
`$JAVA_HOME/include/darwin/` instead of `$JAVA_HOME/include/linux`. This
change set change the test to only run for Linux to make the MacOS test
pass. We need to find a reliable way to find the headers on different OSes.

Test:
`ant clean lint && ant && buck test`

### DIFF
--- a/test/com/facebook/buck/jvm/java/JavaTestIntegrationTest.java
+++ b/test/com/facebook/buck/jvm/java/JavaTestIntegrationTest.java
@@ -189,7 +189,7 @@ public class JavaTestIntegrationTest {
 
   @Test
   public void testWithJni() throws IOException {
-    assumeTrue(Platform.detect() != Platform.WINDOWS);
+    assumeTrue(Platform.detect() == Platform.LINUX);
     ProjectWorkspace workspace = TestDataHelper.createProjectWorkspaceForScenario(
         this,
         "test_with_jni",


Coneko:
@facebook-github-bot import

Coneko:
Does it fail for you when running in ant, Buck or IntelliJ?

It seems to be failing with different configurations when running locally, but it always passes both with Buck and ant when running on CI.

liuyang-li:
It failed for me: in Ant and Buck on the command line, and in Intellij. I am using a MacBook.

I was able to make it work by tweak my $JAVA_HOME (I have multiple versions of JDK installed), and add `/include/darwin` to the parameters to start C/C++ compiler instead of `/include/linux`. I was not sure how to get it working for all Mac users reliably, so I disabled it.

Coneko:
> add `/include/darwin` to the parameters

I did that and it fixed the building on some of the configurations.

The final issue I think we cannot solve: it has to do with El Capitan stripping `DYLD_LIBRARY_PATH`, which is necessary for running tests with JNI as we do it right now, if System Integrity Protection is enabled:

http://www.oracle.com/technetwork/java/javase/8u66-relnotes-2692847.html

> When running on OSX 10.11 "El Capitan", when SIP is enabled, certain environment variables intended for debugging applications, such as `DYLD_LIBRARY_PATH`, may be stripped from the environment when running Java from the command line or when double-clicking a JAR file. Applications should not rely on these variables in a production environment, they are only intended for debugging during development.

I think after adding the header search path the test should pass when run with Buck, but not in ant or IntelliJ.

Let me know if it doesn't.

liuyang-li:
Adding the additional include path works reliably with `buck test`. Let me submit a PR to do that for now. Thanks for the explanation.

PR #481, opened 2015-10-26
[TestNG] Rename folder junit to testrunner as it contains testng as w…

liuyang-li:
…ell.

Summary:
Rename the junit folder for src and test to testrunner.

Test:
`ant clean lint && ant && buck test`


Coneko:
@facebook-github-bot  import

PR #488, opened 2015-10-28
Scope IjModule COMPILED_SHADOW library dep to the 'compile' scope.

strangemonad:
related to https://github.com/facebook/buck/pull/478

Note: this patch is a sketch of an approach and actually introduces a regression to
the correctness of 'provided_deps'. See the last comment as we figure out the correct approach.

Summary: a java_library IJModule's exported_deps become dependency graph nodes
of type COMPILED_SHADOW and are then included in the IJModule as a 'provided' library rather
than a 'compile' library.

It's unclear how IJ would ever get that provided dep on its classpath. An example of a 'provided'
dep would be acrypto provide like Bouncy castle that's installed into the jdk on the machine and
automatically included in the classpath of the jvm or slf4j that's included at runtime via another lib.

Here's an example to make it more concrete. Consider the following buck dependency chain:

my_app:java_binary -> my_lib:java_library -> my_lib_compiled_java_protos:java_libary -> my_lib_java_protos:genrule

In this setup, my_lib.exported_deps = [':my_lib_compiled_java_protos'] where the leaf
library is a buck 'java_library' in the module graph that was built simply by defining a
'java_library' who's sources are the java files output by protoc in the ':my_lib_java_protos'
target.

By the time AbstractIjLibrary#addAsDependency is reached, we fall into the
`if (dependencyType.equals(IjModuleGraph.DependencyType.COMPILED_SHADOW))` branch which
calls IjDependencyListBuilder#addCompiledShadow (its only caller).

Additionally, I think there might be 2 other bugs hiding in here

1) we unconditionally add the lib as 'exported', this should depend on whether the compiled
shadow lib was included in 'deps' or 'exported_deps'

2) We should properly handle 'provided_deps' for both COMPILED_SHADOW and PROD/TEST types
as well. This patch breaks the first but the later dep types are always included as 'compile'.

### DIFF
--- a/src/com/facebook/buck/jvm/java/intellij/AbstractIjLibrary.java
+++ b/src/com/facebook/buck/jvm/java/intellij/AbstractIjLibrary.java
@@ -70,6 +70,7 @@ abstract class AbstractIjLibrary implements IjProjectElement {
     if (dependencyType.equals(IjModuleGraph.DependencyType.COMPILED_SHADOW)) {
       dependencyListBuilder.addCompiledShadow(getName());
     } else {
+      // XXX how to properly handle 'provided_deps' here.
       IjDependencyListBuilder.Scope scope = IjDependencyListBuilder.Scope.COMPILE;
       if (dependencyType.equals(IjModuleGraph.DependencyType.TEST)) {
         scope = IjDependencyListBuilder.Scope.TEST;
--- a/src/com/facebook/buck/jvm/java/intellij/IjDependencyListBuilder.java
+++ b/src/com/facebook/buck/jvm/java/intellij/IjDependencyListBuilder.java
@@ -178,7 +178,7 @@ public class IjDependencyListBuilder {
             .setSortOrder(SortOrder.COMPILED_SHADOW)
             .setData(DependencyEntryData.builder()
                     .setName(name)
-                    .setScope(Scope.PROVIDED)
+                    .setScope(Scope.COMPILE)
                     .setExported(true)
                     .build())
             .build());


marcinkosiba:
I haven't looked at the diff yet, but it looks like you're trying to get the generated projects to work with IntelliJ so I wanted to set expectations: getting the generated project to compile and run in IntelliJ is not a goal at the moment. The reason for the buck project command is to enable you to edit the code in IntelliJ and we plan to delegate build+run functionality to buck via the IntelliJ plugin.
Your description makes sense and I'll look over the diff when I have a moment, however I don't think it's worthwhile to invest much effort into fixing the generated project.

marcinkosiba:
got some feedback that this sounds differently to what I intended to say, so let me rephrase: I think the approach is correct and we will most likely take the diff, it's just that making the IntelliJ project build with IntelliJ is not the best thing to focus on.

marcinkosiba:
> 1) we unconditionally add the lib as 'exported', this should depend on whether the compiled
shadow lib was included in 'deps' or 'exported_deps'

the compiled_shadow is a hack to work around issues around annotation processors (for big projects you don't want to have to "build" with IJ to be able to reference classes created by annotation processors) and sourcejars (buck magically supports zip files containing .java files as inputs, IntelliJ doesn't). The compiled_shadow jar contains all of the classes in the original module plus any code-gen and sourcejar-related stuff. Because the generated project already has exported deps resolved (any exported deps appear as regular deps in the project) it's fine to 'cheat' and mark the compiled-shadow as exported (which has the effect that anything that depends on the target also depends on its shadow, so you get to access the classes coming from code-gen without adding an explicit dependency on this library in the other module). Ultimately this is just a trick to get IntelliJ's indexing to pick up code-gen'd classes.

> 2) We should properly handle 'provided_deps' for both COMPILED_SHADOW and PROD/TEST types
as well. This patch breaks the first but the later dep types are always included as 'compile'.

I believe IJ just has either provided (for prod) or test deps, no provided_for_test, so you'd probably need to promote any provided deps coming from tests to prod. But yes, it would be nice to support that properly as well.

If you get all of your compiled_shadow deps as scope=compile the only thing that's making sure the duplicated classes are resolved correctly is classpath order. The result is a bit of a franken-build that will break in weird ways if you forget to `buck build` after changing anything that affects code-gen. I think the real issue here is that buck project is not handling either annotation processors or src.zip genrule inputs (or both) in your project and I don't know of a good way to make that happen:
 - I've found that our internal annotation processors work with Buck, but don't want to work with IntelliJ, so I never invested a lot of effort in making the generated project expose annotation processors to IJ, you're welcome to try though,
 - I don't have a good idea for src.zip - using an intermediate folder means this breaks with caching, un-zipping the file as part of `buck project` means it doesn't get updated when you `buck build`.

marcinkosiba:
no activity for over a year, closing

PR #499, opened 2015-11-09
Add getEnvironment to Tool

mikekap:
This is quite incomplete - I wanted to get first impressions on the approach. It's basically adding an `ImmutableMap<String, String> environment` to all the steps that use `getCommandPrefix`. To some extent I wonder if passing Tool to the steps instead would be better - although I suppose that's a bit harder to test with.

Separately, if you could, let me know whether you prefer a commit per directory or one giant squash. I'm hoping one-per-directory makes it easier to review, but it's definitely a PITA to bisect.

Either way - let me know if this looks good. I'll finish the other ~70 uses once I get the ok :)


Coneko:
Looks good to me, and changing all the instances in one PR is fine too.

mikekap:
Alright - should be ready I think.

sdwilsh:
I'll let @Coneko merge this in since he already started looking at it.

Coneko:
@facebook-github-bot shipit

PR #506, opened 2015-11-12
Use absolute paths when dealing with apks and test output

tgummerer:
Summary:
Using relative paths when dealing with installing apks or exopackages or
in test outputs causes problems when buck is not run from the right
location, such as in integration tests.  Fix this by using absolute
paths where appropriate.

Test Plan: buck test


sdwilsh:
@facebook-github-bot shipit

Issue #501, opened 2015-11-10
Empty jar committed to repository


copumpkin:
If you look at https://github.com/facebook/buck/blob/master/test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/buck-android-support.jar, the jar file is empty. Is that intentional? It seems to be crashing my build. I can look further into the build failure if the empty file is intentional, but it seemed iffy.

sdwilsh:
*sigh*  No, that's probably a bug somewhere.  I'm sure it wasn't meant to be empty.

copumpkin:
@sdwilsh thanks, glad it's not just me :smile:. Is it expected that the broken file would crash my ant build, or is something weird also going on on my side?

sdwilsh:
No, it shouldn't crash it (it doesn't for me or anyone else that I'm aware of).

copumpkin:
Yeah, I figured out that that was a post-processing step I was running inadvertently, sorry!

sdwilsh:
Still a valid issue; we'll get to the bottom of it soon.  Our importer seems to drop binary artifacts, so we never pulled the right file in from @tgummerer's PR.  Although, the fact that the test passes despite the empty file is a bit concerning...

copumpkin:
Yeah. You seem to have a few others, too:

```bash
$ find . -name "*.jar" -size 0
./test/com/facebook/buck/android/testdata/android_instrumentation_test_integration_test/buck-android-support.jar
./test/com/facebook/buck/cli/testdata/project_slice/libs/guava.jar
./test/com/facebook/buck/cli/testdata/project_slice/libs/jsr305.jar
./test/com/facebook/buck/cli/testdata/project_slice/libs/junit.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/experimental_project1/libs/guava.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/experimental_project1/libs/jsr305.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/experimental_project1/libs/junit.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/project1/libs/guava.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/project1/libs/jsr305.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/project1/libs/junit.jar
./test/com/facebook/buck/jvm/java/intellij/testdata/project2/libs/guava.jar
[...]

PR #507, opened 2015-11-12
Hook up android_instrumentation_test to test infrastructure

tgummerer:
Summary:
When android_instrumentation_test was introduced, an integration test
was added, but I forgot to hook it up to the test infrastructure.

This caused us to miss that some binary files broke when importing the
pull request from GitHub.

Hook up the integration test to the test infrastructure and fix the
broken binary files.

Test Plan: buck test with and without android devices attached


tgummerer:
This needs to be applied on top of #506, otherwise the tests don't pass.  Addresses #501, at least for the empty buck-android-support.jar, I haven't looked at the other empty jars @copumpkin mentioned.

sdwilsh:
#506 will be landing soon.  Note to any other Facebook people: we can't use the normal import flow for this because it's broken for binary jars.

sdwilsh:
I've imported this, but making sure it passes CI before landing it.

test/com/facebook/buck/android/AndroidInstrumentationTestIntegrationTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Test
+  public void testAndroidInstrumentationTest() throws IOException {
+    tmpFolder.doNotDeleteOnExit();
+    ProjectWorkspace workspace = TestDataHelper.createProjectWorkspaceForScenario(
+        new AndroidBinaryIntegrationTest(),
+        "android_instrumentation_test_integration_test",
+        tmpFolder);
+    workspace.setUp();
+    workspace.runBuckCommand("test").assertSuccess();
```
sdwilsh: We need to only run when the android sdk is present.  I'll add the appropriate assume statement here.

tgummerer:
Thank you very much for fixing this up and getting it in :)

PR #508, opened 2015-11-13
Decouple JavaLibrary from AnnotationProcessingParams

grumpyjames:
Summary:
We only need to expose the generated source directory (which may not
exist). As AnnotationProcessingParams is java specific (rather than
jvmlang agnostic) we'd prefer JavaLibrary (soon to become JvmLibrary)
to not know about it.

Context: I'm another LMAX person. We really want to make some progress towards the work described in https://github.com/facebook/buck/pull/439

Tested with ant && ./bin/buck test (advice on whether this is enough gratefully received!)


sdwilsh:
@facebook-github-bot shipit

PR #510, opened 2015-11-15
Move jvmlang agnostic interfaces to jvm.core

grumpyjames:
Move jvmlang agnostic interfaces to jvm.core

Summary:
While implementations of these interfaces may be language specific,
the interfaces themselves are not, so move them to jvm.core.

Test plan: CI


grumpyjames:
Continuing on the #439 theme (probably safe to assume anything else from me will be along such lines; will make any exceptions explicit).

If this looks OK, will move on to JavaLibrary and friends. Once that's out, the way forward will hopefully be a little clearer.

grumpyjames:
I _think_ that is broken in travis because of the current state of upstream, rather than this change.

Coneko:
@facebook-github-bot shipit

sdwilsh:
I don't understand the build failure at all.  It looks like Travis is (once again) killing our tests (exit code 173 is `SIGKILL`).  The only android-related changes that went in were @tgummerer's tests, and those passed when the PR was there.

grumpyjames:
Regarding travis: could it be oomkiller?

sdwilsh:
Almost certainly is.  I'm going to try and split up that target.

PR #512, opened 2015-11-17
Make DefaultJavaLibrary a little more language agnostic

grumpyjames:
Summary:

The main goal: to decouple JavacOptions from DefaultJavaLibrary.
We end up passing an instance of JavacStepFactory (which will probably
sprout an interface named 'CompileStepFactory' shortly). This means
the DefaultJavaLibrary::getOptions is now only called from
Project. Android extensions to DefaultJavaLibrary extend their
javacOptions at step creation time by providing a custom
JavacOptionsAmender.

Lots of other small changes along the way. We found one or two places 
where we had to encapsulate better in order to make progress - one or
two tests disappear or move to different places as a result.

Test Plan: CI


grumpyjames:
One thing we are not sure about is the implementations of RuleKeyAppendable in JavacStepFactory and the two JavacOptionsAmender implementations. If what we've done there is wrong or unusual, some further guidance would be much appreciated :-)

Coneko:
There's a failure in the tests: setting the source level and target level isn't reflected in the key correctly.

grumpyjames:
You are quite right. That's interesting, because I remember changing that test having seen it fail (the test also acts as a detector for _any_ part of the key changing because of how it's written), and ./bin/buck test passes for me locally (indeed, that test passes for me in intellij). I will take a closer look to see what is missing here.

grumpyjames:
So, it's good that this failed, as it made me wonder why it wasn't failing _anyway_, and it's because we are incorrectly implementing RuleKeyAppendable in JavacStepFactory (we omit the call to JavacOptionsAmender's impl). Further changes to follow.

I am still somewhat puzzled as to why it fails in travis and not locally, investigations continue.

Coneko:
Sounds good, I'm looking at the rest of the changes in the meantime.

Coneko:
Ok, high level the changes seem ok, will review in depth after the tests are fixed.

I would like you to split some of the commits out into their own PR though, so that they have their own commit after they're imported.

grumpyjames:
Thanks for the speedy review. Will have to level up my github fu to 'understand how to create dependent pull requests', but that sounds like a worthy quest if it creates a cleaner commit log in the real master.

Regarding travis: if I perform the same steps as it does (from a fresh checkout) on this machine, the build passes.

The diff it finds (//:java_lib_1#abi versus //:java_lib_1) is hard for me to explain, particularly given that the previous version (before my change) also contains the "#abi" in the expectations, and the key changes ought to be below that level. I will continue to add print statements to diff_rulekeys.py locally in the hope of understanding, but any more specific hints would be greatly appreciated.

Coneko:
> The changes you mention actually look pleasantly orthogonal

Yes, that is why I asked them to be pulled out: it's not really about the size of the changes in a PR, it's more about whether the changes are part of a single unit or not.

grumpyjames:
Understood. Closing this, will reconstruct the agnostification (definitely a word) changes in a separate PR later.

PR #514, opened 2015-11-17
Rename TestSourcePath to FakeSourcePath

grumpyjames:
Summary:
We found the original name to be confusing in the context of a build
system. The new name is less ambiguous and matches naming of other
classes in the same package.

Test plan: CI


Coneko:
@facebook-github-bot shipit

PR #516, opened 2015-11-17
Make buck work under PyPy

mikekap:
Particularly: pypy outputs a two-line version string like:
   Python 2.7.10 (850edf14b2c75573720f59e95767335fb1affe55, Oct 30 2015, 00:18:28)
   [PyPy 4.0.0 with GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)]

Also functools is implemented in python so shows up on the stack.

### DIFF
--- a/test/com/facebook/buck/python/PythonBuckConfigTest.java
+++ b/test/com/facebook/buck/python/PythonBuckConfigTest.java
@@ -266,6 +266,32 @@ public class PythonBuckConfigTest {
   }
 
   @Test
+  public void testGetPypyVersion() throws Exception {
+    String pypyOutput =
+        "Python 2.7.10 (850edf14b2c75573720f59e95767335fb1affe55, Oct 30 2015, 00:18:28)\n" +
+        "[PyPy 4.0.0 with GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)]\n";
+
+    PythonVersion version =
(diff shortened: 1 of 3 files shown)


test/com/facebook/buck/python/PythonBuckConfigTest.java
```diff
@@ -266,6 +266,32 @@ public void testGetPyrunVersion() throws Exception {
...
+  public void testGetPypyVersion() throws Exception {
+    String pypyOutput =
+        "Python 2.7.10 (850edf14b2c75573720f59e95767335fb1affe55, Oct 30 2015, 00:18:28)\n" +
+        "[PyPy 4.0.0 with GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.1.76)]\n";
+
+    PythonVersion version =
+        PythonBuckConfig.extractPythonVersion(
+            Paths.get("non", "important", "path"),
+            new ProcessExecutor.Result(0, "", pypyOutput));
+    assertEquals("Python 2.7", version.toString());
```
sdwilsh: We generally prefer `assertThat` with a `Matcher` as the test below the two new ones you've added.

sdwilsh: I should have flagged that in your last iteration, so let me import this and fix it up myself.

sdwilsh:
@facebook-github-bot import

PR #517, opened 2015-11-17
Update python pkg_resources to latest (18.5)

mikekap:
This lets you use resource_stream in __enter__ and __exit__, at least.

Fun note: turns out the python in-place config doesn't actually include the vendored pkg_resources in the PYTHONPATH. I didn't fix that in this change. It might be worth
considering looking into in-place pex running instead of re-implementing pex, e.g. - in tp/py/.../pex/bin/pex.py:

```python
    log('Running PEX file at %s with args %s' % (pex_builder.path(), args), v=options.verbosity)
    pex = PEX(pex_builder.path(), interpreter=pex_builder.interpreter)
    return pex.run(args=list(args))
```

Separately, as part of (a hope!) to import the new pex, I've renamed pex.py to make_pex.py to fix the name collision.


sdwilsh:
Oof.  I'd like that (it might be easier to update the PEX code first/with it too), but if it ends up being a lot of work, I'm okay with just a patch for now.

mikekap:
Done - I've tested with a buck.pex and it seems to work.

sdwilsh:
@andrewjcg or @bhamiltoncx - you two are probably the most qualified to review this.

bhamiltoncx:
Looks OK to me, but @andrewjcg is the PEX master, I'll defer to him.

sdwilsh:
@andrewjcg is on PTO for a bit.  Since @bhamiltoncx is okay with this, let's ship it.

@facebook-github-bot shipit

sdwilsh:
I ran this in our internal CI which is a bit more exhaustive than Travis.  We hit this error:

```
  File "/path/to/repo/buck-out/tmp/buck_run.QGbq97/resources/f2342654476907498e0de7165ecdc2af81415e93/path_to_pex/src/com/facebook/buck/python/make_pex.py", line 42, in <module>
  File "/path/to/repo/buck-out/tmp/buck_run.QGbq97/resources/f2342654476907498e0de7165ecdc2af81415e93/path_to_pex/.bootstrap/pkg_resources/__init__.py", line 1174, in resource_stream
    self, resource_name
  File "/path/to/repo/buck-out/tmp/buck_run.QGbq97/resources/f2342654476907498e0de7165ecdc2af81415e93/path_to_pex/.bootstrap/pkg_resources/__init__.py", line 1617, in get_resource_stream
    return io.BytesIO(self.get_resource_string(manager, resource_name))
  File "/path/to/repo/buck-out/tmp/buck_run.QGbq97/resources/f2342654476907498e0de7165ecdc2af81415e93/path_to_pex/.bootstrap/pkg_resources/__init__.py", line 1620, in get_resource_string
    return self._get(self._fn(self.module_path, resource_name))
  File "/path/to/repo/buck-out/tmp/buck_run.QGbq97/resources/f2342654476907498e0de7165ecdc2af81415e93/path_to_pex/.bootstrap/pkg_resources/__init__.py", line 1698, in _get
    return self.loader.get_data(path)
IOError: [Errno 11] Resource temporarily unavailable: 'pkg_resources.zip'
```

Any idea on what's up?  I don't think I'm missing any files for your change internally (and not all builds are failing this way).

mikekap:
Hmm it's hard to say - is pkg_resources.zip in /path/to/repo/buck-out/tmp/buck_run.QGbq97/resources/f2342654476907498e0de7165ecdc2af81415e93/path_to_pex/src/com/facebook/buck/python/ at least? If not, is it anywhere in that pex?

sdwilsh:
I couldn't tell you; the machines take a new job which nukes state once the job is done.

mikekap:
So without more debugging info, I assumed that the .zip file was missing. I rewrote the make_pex script to pull in the pkg_resource that it's using into the result pex (it's a bit meta).

Unfortunately this makes make_pex a bigger part of buck.pex bootstrap - it pulls in the correct pkg_resources at runtime via path inspection (i.e. the dependency isn't declared anywhere in BUCK files). Let me know if we should declare it - it would only be superficial afaict.

sdwilsh:
I was hoping to dig into this more yesterday or today, but I'm oncall for my team, and it's been a bit crazy.  I might not get to this until next week due to the US holiday, unfortunately.  Maybe @Coneko can help you out if you need this in sooner.

mikekap:
No rush. I'm a bit more confident in this version working in an arbitrary environment that the old version worked in. The downside is it's a bit more magical.

mikekap:
Ping? :)

sdwilsh:
Last week I got totally derailed, alas.  I wonder if @andrewjcg has some insight on this failure.

mikekap:
Any chance you could give the new version a spin on your internal CI? I have a feeling it works a lot better (there's no .zip file anymore).

sdwilsh:
I thought I had!  Silly brain misleading me.  @facebook-github-bot import

sdwilsh:
Bah.  The importer had some errors importing the diff.  I'll have to try to import it by hand later today.

sdwilsh:
Alright, I imported this locally this morning, but trying `buck build buck` failed for me with this:

```
Traceback (most recent call last):
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 155, in <module>
    sys.exit(main())
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 116, in main
    copy_resources(pex_builder, 'pkg_resources', prefix=pex_builder.BOOTSTRAP_DIR)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 55, in copy_resources
    os.path.join(resource, entry), prefix)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 55, in copy_resources
    os.path.join(resource, entry), prefix)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 55, in copy_resources
    os.path.join(resource, entry), prefix)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 69, in copy_resources
    os.path.join(prefix, target_path))
  File "C:\Users\sdwilsh\code\buck\third-party/py/twitter-commons/src/python\twitter\common\python\pex_builder.py", line 104, in add_source
    self._chroot.link(filename, env_filename, "source")
  File "C:\Users\sdwilsh\code\buck\third-party/py/twitter-commons/src/python\twitter\common\python\common.py", line 283, in link
    shutil.copyfile(abs_src, abs_dst)
  File "C:\Python27\lib\shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 13] Permission denied: 'c:\\users\\sdwilsh\\appdata\\local\\temp\\tmpy8gqvw'
Traceback (most recent call last):
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 155, in <module>
    sys.exit(main())
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 116, in main
    copy_resources(pex_builder, 'pkg_resources', prefix=pex_builder.BOOTSTRAP_DIR)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 55, in copy_resources
    os.path.join(resource, entry), prefix)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 55, in copy_resources
    os.path.join(resource, entry), prefix)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 55, in copy_resources
    os.path.join(resource, entry), prefix)
  File "C:\Users\sdwilsh\code\buck\src\com\facebook\buck\python\make_pex.py", line 69, in copy_resources
    os.path.join(prefix, target_path))
  File "C:\Users\sdwilsh\code\buck\third-party/py/twitter-commons/src/python\twitter\common\python\pex_builder.py", line 104, in add_source
    self._chroot.link(filename, env_filename, "source")
  File "C:\Users\sdwilsh\code\buck\third-party/py/twitter-commons/src/python\twitter\common\python\common.py", line 283, in link
    shutil.copyfile(abs_src, abs_dst)
  File "C:\Python27\lib\shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 13] Permission denied: 'c:\\users\\sdwilsh\\appdata\\local\\temp\\tmpijmklt'
```

sdwilsh:
I'm not sure what's up with that error.  The directory exists there, and I'm not sure why it's getting permission denied there, but it looks like that code path is new, so I'm pretty sure it's that change.

mikekap:
Seems like a bug in python for windows: you have to call close() on a temporary file for it to be readable: https://bugs.python.org/issue14243 . I tested a workaround on my windows box and it seems to work. I'll have a patch in a few minutes since I don't have enough development tools on windows :P

sdwilsh:
Wow, that is special.  I'll import this, make sure it builds on my Windows machine, and then run it through our internal CI.

sdwilsh:
*sigh*

So, I wasn't able to make Travis use python2.6, and our internal CI actually does check that (Buck supports it).  You have some things that make python2.6 unhappy too:
```
  File "/data/sandcastle/boxes/instance-buck-git-buck-verification-linux/src/com/facebook/buck/python/make_pex.py", line 80
    with closable_named_temporary_file() as fp, \
                                              ^
SyntaxError: invalid syntax
  File "/data/sandcastle/boxes/instance-buck-git-buck-verification-linux/src/com/facebook/buck/python/make_pex.py", line 80
    with closable_named_temporary_file() as fp, \
                                              ^
SyntaxError: invalid syntax
  File "/data/sandcastle/boxes/instance-buck-git-buck-verification-linux/src/com/facebook/buck/python/make_pex.py", line 80
    with closable_named_temporary_file() as fp, \
                                              ^
SyntaxError: invalid syntax
```

I'm pretty sure this is the "can't have more than one thing in a with statement" issue.

mikekap:
Two with statements it is then. Hopefully this is the only python 2.6 incompatibility - I have to go find a way to install python 2.6 on my mac.

mikekap:
Confirmed: I got a buck.pex to build (and build itself again) under python 2.6.

sdwilsh:
I've kicked off internal CI again with your changes.

third-party/py/twitter-commons/src/python/BUCK
```diff
@@ -2,9 +2,6 @@ python_library(
   name = 'twitter-commons',
   base_module = '',
   srcs = glob(['**/*.py']),
-  deps = [
-    '//third-party/py/setuptools:pkg_resources',
-  ],
```
andrewjcg: Why remove this dep?  I think the pex sources import `pkg_resources`, which means there should be a dep here.

third-party/py/setuptools/BUCK
```diff
@@ -1,17 +1,2 @@
...
-)
-
-export_file(
-  name = 'pkg_resources.py',
-  visibility = [
-    'PUBLIC',
-  ],
-)
+# Note: you shouldn't depend on pkg_resources directly,
+# as it's included in every pex by make_pex.
```
andrewjcg: But the pex builder uses it right?  So the pex-builder binary we build (to build the actual PEXs) should depend on this, no?

src/com/facebook/buck/python/make_pex.py
```diff
@@ -27,20 +28,8 @@
...
-else:
-    import pkg_resources
-    pkg_resources_py_tmp = tempfile.NamedTemporaryFile(
-        prefix='pkg_resources.py')
-    pkg_resources_py_tmp.write(
-        pkg_resources.resource_string(__name__, 'pkg_resources.py'))
-    pkg_resources_py_tmp.flush()
-    pkg_resources_py = pkg_resources_py_tmp.name
+
+import pkg_resources
```
andrewjcg: I might just be getting confused here, but does this mean we're pulling from a the system `pkg_resources`?  What if the system doesn't have this installed?

mikekap: See the sys.path stuff above - we're trying to ensure this imports the bundled one. I don't know how successful it is, but it's certainly as successful as it was before - pex uses pkg_resources internally.

third-party/py/twitter-commons/src/python/BUCK
mikekap: pkg_resources is part of a "pex stdlib" (i.e. every pex made by buck will include pkg_resources) so it's an implicit dependency of every python_library. It just creates two copies of pkg_resources in the resulting python_binary which doesn't end well.

src/com/facebook/buck/python/make_pex.py
andrewjcg: Yup, just tested this locally on a system without pkg_resources and it works when run from both the in-place and PEX buck.  In the former case, it's pulling from the sys.path above, in the latter it's using `pkg_resources` from the pex builder's own PEX.

third-party/py/twitter-commons/src/python/BUCK
andrewjcg: I'll add a comment before committing this about this anti-dep, since it appears initially counter-intuitive.

third-party/py/setuptools/BUCK
mikekap: So this is where it gets a bit...hand-wavy. There are two explanations for each mode in which pex.py runs:
 - repo mode - In this case, pex.py changes sys.path to ensure it imports pkg_resources in third-party at runtime, so it doesn't need the dependency explicitly.
 - pex mode - in this case pex.py does NOT use the pkg_resources that exist in the repo (i.e. it does not use the dependency) - it uses whatever it was compiled with. That is, building buck.pex (say A) from a buck.pex (say B) will include pkg_resources(B) in buck.pex(A). This was already the case before, so it's not new behavior, but the intent was to clarify it.

andrewjcg:
Looks good!  Sorry for the delay and thanks for walking me through it.  Will start importing this.

src/com/facebook/buck/python/make_pex.py
mikekap: One thing I'm not sure about (although I guess there's no change here) is how good the standalone pex.py isolates itself from the enclosing environment - i.e. if the local system has a pkg_resources, do we actually do enough to make sure that the correct one is loaded? I think some site-packages magic might cause pkg_resources to be imported with sys so we'd always pick up the system one.

mikekap:
Thinking about it more, adding a dependency between pex.py & pkg_resources might be superficial in terms of behavior, but buck might need it to properly do cache invalidation (i.e. if you want buck.pex to rebuild when pkg_resources changes - as it is now). Am I understanding that right? (i.e. when you build buck.pex in repo mode, you'd want to do a rebuild when pkg_resources changes; though I *think* all of buck currently gets rebuilt if you change any file in the repo.)

An easy fix could be to make a dummy genrule that references pkg_resources sources but outputs nothing. Would that be a good idea here?

andrewjcg:
Yeah, it might be nice to have that explicit dep, but currently, I think Buck will rebuild it's own world when any source in the tree changes (e.g. even a change to `third-party/java/immutables/README.facebook` will cause Buck to rebuild the buck.pex).

PR #519, opened 2015-11-18
Add go.prefix for better interoperability with the `go` tool.

mikekap:
Inspiration from https://github.com/bazelbuild/bazel/tree/master/tools/build_rules/go#go_prefix .


nemith:
Should this be a buck wide config setting?   Bazel looks like it may be per BUILD file.

mikekap:
Bazel only allows it at top-level (note the "globally unique target"). However, this is only for defaults - any `go_library` can override the absolute import path via `package_name`. In that sense, you could implement this feature as a monkey-patch of go_library, but I figured this might be friendlier.

sdwilsh:
@facebook-github-bot shipit

PR #520, opened 2015-11-18
Don't create a copy of defaultJavacOptions in KnownBuildRuleTypes

grumpyjames:
Summary:

We create a copy of the defaultJavacOptions only to give it a
different, android related name; this is unnecessary given the
immutable nature of JavacOptions - and those android tasks still use
javac; it's ok for them to use the defaultJavacOptions.

Test plan: CI


Coneko:
@facebook-github-bot shipit

PR #521, opened 2015-11-18
Encapsulate javacOptions within DefaultJavaLibrary

grumpyjames:
Summary: Work towards decoupling the two completely. We don't let
javacOptions escape from DefaultJavaLibrary, and we begin to move the
main point of use into the newly renamed JavacStepFactory, which at
some point will sprout a CompileStepFactory interface.

Test-plan: CI


Coneko:
If the final result of this PR is to remove `DefaultJavaLibrary.getJavacOptions()` 8bcd2cc and a9c3087 don't seem to be related.

grumpyjames:
Good spot; will split those out and resubmit.

PR #522, opened 2015-11-18
Encapsulate javacOptions within DefaultJavaLibrary

grumpyjames:
Summary: Work towards decoupling the two completely. We don't let
javacOptions escape from DefaultJavaLibrary, and we begin to move the
main point of use into the newly renamed JavacStepFactory, which at
some point will sprout a CompileStepFactory interface.

Test-plan: CI


Coneko:
@facebook-github-bot shipit

PR #524, opened 2015-11-19
Fix iteration order of structure dicts within diff_rulekeys.py

grumpyjames:
Summary:
Different Python versions have differently-deterministic iteration
order for defaultdicts. We change to using an OrderedDict to avoid
ambiguity. We discovered this tracking down tests that failed in
travis and not locally (travis is on 2.7.3, we are on 2.7.5)

Test Plan: CI


sdwilsh:
Good catch!

PR #525, opened 2015-11-19
Decouple JavacOptions from DefaultJavaLibrary

grumpyjames:
Summary:
Instead, pass an instance of JavacStepFactory. This means
that DefaultJavaLibrary::getProperties is now called only from
Project. Android extensions to DefaultJavaLibrary extend their
javacOptions at step creation time by providing a custom
JavacOptionsAmender.

Test Plan: CI


grumpyjames:
Is there anything we can do to help get this reviewed? We don't want to move too far from tip, at the same time though, we're really keen to make some progress. Let me know where to send the cake ;-)

src/com/facebook/buck/jvm/java/JavacOptionsAmender.java
```diff
@@ -0,0 +1,39 @@
...
+
+public interface JavacOptionsAmender extends RuleKeyAppendable {
+  JavacOptions amend(JavacOptions original, BuildContext context);
+
+  String RULE_KEY_NAME = "javacOptionsAmender";
+
+  JavacOptionsAmender IDENTITY = new JavacOptionsAmender() {
+    @Override
+    public RuleKeyBuilder appendToRuleKey(RuleKeyBuilder builder) {
+      return builder.setReflectively(RULE_KEY_NAME, "id");
```
Coneko: I think this shouldn't append anything to the rule key.

src/com/facebook/buck/android/BootClasspathAppender.java
```diff
@@ -0,0 +1,37 @@
...
+import com.facebook.buck.jvm.java.JavacOptions;
+import com.facebook.buck.jvm.java.JavacOptionsAmender;
+import com.facebook.buck.rules.BuildContext;
+import com.facebook.buck.rules.RuleKeyBuilder;
+
+
+class BootClasspathAppender implements JavacOptionsAmender {
+  @Override
+  public RuleKeyBuilder appendToRuleKey(RuleKeyBuilder builder) {
+    return builder.setReflectively(JavacOptionsAmender.RULE_KEY_NAME, "android");
```
Coneko: This shouldn't use `JavacOptionsAmender.RULE_KEY_NAME`, it should use its own key.

Coneko:
@facebook-github-bot import

Coneko:
Ok, just those two notes, the rest looks good.

Coneko:
I'll fix them up then ship this, you don't need to update this unless you have the changes ready.

grumpyjames:
That sounds great, thanks again.

PR #526, opened 2015-11-22
Improve performance of get_dirty_buck_version

rhencke:
When running Buck from a dirty working copy, a large amount of startup
time was spent calculating the version in get_dirty_buck_version.  git
read-tree was reading a fresh index into a temporary file, which
contained no stat() information about the working copy's current stat.
When git add -u then ran, it then had to rebuild this information for
all tracked files, which was time-consuming.

The new version re-uses the current index's stat() information when
writing the index to a temporary file.  It also shaves off one call to
git, taking advantage of the fact that HEAD counts as a tree-ish object
that can be passed to git read-tree.

### DIFF
--- a/programs/buck_version.py
+++ b/programs/buck_version.py
@@ -80,20 +80,16 @@ def get_clean_buck_version(dirpath, allow_dirty=False):
 
 
 def get_dirty_buck_version(dirpath):
-    git_tree_in = check_output(
-        ['git', 'log', '-n1', '--pretty=format:%T', 'HEAD', '--'],
-        cwd=dirpath).strip()
-
     with EmptyTempFile(prefix='buck-git-index') as index_file:
+        subprocess.check_call(
+            ['git', 'read-tree', '--index-output', index_file.name, '--reset', 'HEAD'],
+            cwd=dirpath)
+
         new_environ = os.environ.copy()
         new_environ['GIT_INDEX_FILE'] = index_file.name
-        subprocess.check_call(
-            ['git', 'read-tree', git_tree_in],
-            cwd=dirpath,
-            env=new_environ)
 
         subprocess.check_call(
-            ['git', 'add', '-u'],
+            ['git', 'add', '-A'], 
             cwd=dirpath,
             env=new_environ)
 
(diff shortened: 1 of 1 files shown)


programs/buck_version.py
```diff
@@ -80,17 +80,13 @@ def get_clean_buck_version(dirpath, allow_dirty=False):
 
 
 def get_dirty_buck_version(dirpath):
-    git_tree_in = check_output(
-        ['git', 'log', '-n1', '--pretty=format:%T', 'HEAD', '--'],
-        cwd=dirpath).strip()
-
     with EmptyTempFile(prefix='buck-git-index') as index_file:
+        subprocess.check_call(
+            ['git', 'read-tree', '--index-output', index_file.name, '--reset', 'HEAD'],
```
sdwilsh: I realize this doesn't change the checked out files, but should we pass `--dry-run` here as well?  Not 100% clear to me from the man page.

sdwilsh:
We'd be happy to take a PR to clean up the `is_dirty` stuff too.  Thanks!

programs/buck_version.py
rhencke: In this particular case, running with ```--dry-run``` would actually prevent the updated index from being written to ```index_file.name```.  (```--index-output``` prevents our modifications from being written back to the original index file)

(I got to the same point on the man page... had to just poke at it a lot!  If it's worth some explanatory comments in the code, please feel free to request they be added)

sdwilsh: I glossed right over the part about writing over the file.  That's what I get for looking at changes right when I get in :)

This is fine as-is.

sdwilsh:
@facebook-github-bot shipit

rhencke:
@sdwilsh Thank you. :) What are the intended semantics on 'dirty'?  ```get_buck_dirty_version```'s definition is 'only tracked files affect dirty state', but ```get_clean_buck_version```'s is 'tracked and untracked (except ignored) files affect dirty state'.  (From what I can piece together anyway.)

One way you can see the effect of this split definition: in a dirty Buck directory, notice that changes to tracked files will cause a re-launch of the buck daemon, but changes to untracked files will not.  However, the presence of untracked files will cause the Buck directory to be considered dirty.

sdwilsh:
I think the latter case is the one we care about.  The former feels potentially error-prone still.

PR #533, opened 2015-11-30
Use the -test.timeout flag for go testing rather than a subprocess timeout

mikekap:
This gives a slightly nicer output which includes all the active goroutine stacks.

### DIFF
--- a/src/com/facebook/buck/go/GoTest.java
+++ b/src/com/facebook/buck/go/GoTest.java
@@ -63,6 +63,8 @@ public class GoTest extends NoopBuildRule implements TestRule, HasRuntimeDeps,
       "^=== RUN\\s+(?<name>.*)$");
   private static final Pattern TEST_FINISHED_PATTERN = Pattern.compile(
       "^--- (?<status>PASS|FAIL|SKIP): (?<name>.+) \\((?<duration>\\d+\\.\\d+)(?: seconds|s)\\)$");
+  // Extra time to wait for the process to exit on top of the test timeout
+  private static final int PROCESS_TIMEOUT_EXTRA_MS = 5000;
 
   private final GoBinary testMain;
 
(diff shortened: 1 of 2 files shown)


src/com/facebook/buck/go/GoTest.java
```diff
@@ -107,11 +107,12 @@ public boolean hasTestResultFiles(ExecutionContext executionContext) {
...
                 .addAll(testMain.getExecutableCommand().getCommandPrefix(getResolver()))
                 .add("-test.v")
-                .add("-test.timeout", executionContext.getDefaultTestTimeoutMillis() + "ms")
+                .add(
+                    "-test.timeout",
+                    testRuleTimeoutMs.or(executionContext.getDefaultTestTimeoutMillis()) + "ms")
                 .build(),
             testMain.getExecutableCommand().getEnvironment(getResolver()),
             getPathToTestExitCode(),
-            testRuleTimeoutMs,
```
sdwilsh: What about adding some padding here and using both (in case the executable does hang)?

sdwilsh:
@facebook-github-bot shipit

src/com/facebook/buck/go/GoTest.java
```diff
@@ -37,6 +37,7 @@
 import com.facebook.buck.test.TestRunningOptions;
 import com.facebook.buck.test.result.type.ResultType;
 import com.google.common.base.Charsets;
+import com.google.common.base.Function;
```
sdwilsh: I should have looked at the travis build.  This is an unused import, so `ant lint` fails.

sdwilsh:
@facebook-github-bot shipit

test/com/facebook/buck/go/GoTestIntegrationTest.java
```diff
@@ -66,9 +65,7 @@ public void testGoTest() throws IOException {
   @Test
   public void testGoTestTimeout() throws IOException {
     ProjectWorkspace.ProcessResult result = workspace.runBuckCommand("test", "//:test-spinning");
-    result.assertSpecialExitCode("test should fail", 42);
-    String stderr = result.getStderr();
-    assertTrue(stderr, stderr.contains("Timed out running test command"));
+    result.assertTestFailure("test timed out");
```
sdwilsh: @dreiss just pushed 7c2311282079c9e06117974b1d025b0079452d19 which conflicts with this, and I want to make sure that the end result should still be the same here.

mikekap: Updated to include the timeout in the assert (go's testing does this by default)

PR #535, opened 2015-12-01
Correct duplicate warning in JarDirectoryStepHelper

grumpyjames:
Summary:

JarDirectoryStepHelper emits warnings when duplicates are added to
jars. Unfortunately, the warning currently outputs the path to the
source where it should output the target. Amend this to show the 
duplicate entry name, the target path, and the source path.

Test-plan: CI


sdwilsh:
@facebook-github-bot shipit

PR #541, opened 2015-12-02
Break out DefaultSuggestBuildRules from DefaultJavaLibrary

grumpyjames:
Summary:

It feels like it would be nice to test it without having to rely on
DefaultJavaLibrary, so we extract out the previously anonymous
SuggestBuildRules implementation, rework its inputs to make it look a
little more like a pure function. We also split the existing test into
two cases and substantially rework the tests to make adding further
cases a little simpler.

Test-plan: CI


Coneko:
@facebook-github-bot shipit

PR #546, opened 2015-12-04
Opportunity for discussion: expand any zipped java sources in one place

grumpyjames:
Don't merge me, I don't work!

### DIFF
--- a/src/com/facebook/buck/android/GenAidl.java
+++ b/src/com/facebook/buck/android/GenAidl.java
@@ -16,7 +16,7 @@
 
 package com.facebook.buck.android;
 
-import static com.facebook.buck.jvm.java.Javac.SRC_ZIP;
+import static com.facebook.buck.jvm.java.JavaLibrary.SOURCE_ZIP;
 import static com.facebook.buck.rules.BuildableProperties.Kind.ANDROID;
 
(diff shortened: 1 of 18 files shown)


grumpyjames:
So, this

marcinkosiba:
Thanks for the great writeup! 

I think the Java rules support this as a workaround for the fact that genrules didn't support directories at the time this was added. I don't think it is important to have source .zip support baked in for all jvm-based languages and it's fine for this to stay a java-only feature.

If you wanted to support this later on then we could inject the unzip step based on whether the compiler supports reading directly from .zip files (we could do it by having the compilestepfactory return this information along with the step, for example). We could even pull that into a separate rule (the benefit there is that you could potentially run that 'unzip' rule earlier in the build), though I don't think it's worth it unless we somehow find ourselves flinging a lot of code around in .zip files.

grumpyjames:
Thanks for taking the time to read it - we'll get on with our first go at groovy_library.

shs96c:
I added the `SrcZipAwareFileBundler` to make it easier to work with src zips, which makes option 1 easier to implement. The main thing with the `Jsr199Javac` is that it avoids the need to start a separate compiler process, which reduces the overhead of dealing with things.

Issue #545, opened 2015-12-04
Problem with consuming cross-cell dependencies


davido:
When replacing a cell with command line config option, dependencies are not seen in primary cell:

```
 $ buck build --config repositories.jgit=../jgit gerrit-pgm:http
com/google/gerrit/pgm/http/jetty/JettyServer.java:58: error:
package org.eclipse.jgit.lib does not exist
import org.eclipse.jgit.lib.Config;
```

This is strange, as the dependency is set correctly and is wokring, when internal jgit cell (inside gerrit) is used, without routing the jgit cell through standalone JGit development tree:

```
$ buck targets --json gerrit-pgm:http
[+] PROCESSING BUCK FILES...0,6s
[
{
  "annotationProcessorDeps" : [ ],
  "annotationProcessorOnly" : null,
  "annotationProcessorParams" : [ ],
  "annotationProcessors" : [ ],
  "buck.base_path" : "gerrit-pgm",
  "buck.direct_dependencies" : [ "//:jgit", "//gerrit-common:server", "//gerrit-extension-api:api", "//gerrit-gwtexpui:linker_server", "//gerrit-gwtexpui:server", "//gerrit-httpd:httpd", "//gerrit-reviewdb:server", "//gerrit-server:server", "//gerrit-sshd:sshd", "//lib:guava", "//lib/guice:guice", "//lib/guice:guice-assistedinject", "//lib/guice:guice-servlet", "//lib/jetty:jmx", "//lib/jetty:server", "//lib/jetty:servlet", "//lib/log:api", "//lib/log:jsonevent-layout", "//lib/log:log4j", "//gerrit-pgm:http-jetty" ],
  "buck.type" : "java_library",
  "compiler" : null,
  "deps" : [ "//gerrit-common:server", "//gerrit-extension-api:api", "//gerrit-gwtexpui:linker_server", "//gerrit-gwtexpui:server", "//gerrit-httpd:httpd", "//gerrit-server:server", "//gerrit-sshd:sshd", "//lib:guava", "//lib/guice:guice", "//lib/guice:guice-assistedinject", "//lib/guice:guice-servlet", "//lib/log:api", "//lib/log:log4j", "@jgit//:jgit", "//gerrit-reviewdb:server", "//lib/log:jsonevent-layout", "//lib/jetty:jmx", "//lib/jetty:server", "//lib/jetty:servlet" ],
  "exportedDeps" : [ ":http-jetty" ],
[...]

PR #548, opened 2015-12-05
Honor cell root during classpath entry construction

davido:
Rule can be provided from different cell and thus output file name
must be rellocated according to cell root. Otherwise classpath would
contain non existent entries from foreign project file system.

TEST PLAN:

1. Clone JGit with this patch: [1].
2. Clone Gerrit Code Review with this patch: [2].
3. Replace JGit cell during Gerrit build, with:

  $ buck build --config repositories.jgit=../jgit gerrit

Observe, that without this diff, the classpath contains invalid
entries: non rellocated jgit output file. This diff rellocates it to
jgit cell.

[1] https://git.eclipse.org/r/61938
[2] https://gerrit-review.googlesource.com/73000

Fixes: https://github.com/facebook/buck/issues/545

### DIFF
--- a/src/com/facebook/buck/jvm/java/JavaLibraryClasspathProvider.java
+++ b/src/com/facebook/buck/jvm/java/JavaLibraryClasspathProvider.java
@@ -55,7 +55,8 @@ public class JavaLibraryClasspathProvider {
     }
 
     if (outputJar.isPresent()) {
-      outputClasspathBuilder.put(javaLibraryRule, outputJar.get());
+      outputClasspathBuilder.put(javaLibraryRule,
+          javaLibraryRule.getProjectFilesystem().resolve(outputJar.get()));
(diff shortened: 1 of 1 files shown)


Coneko:
Thanks for finding the root cause of the issue.

To make sure this doesn't happen again, change all the methods on `JavaLibraryClasspathProvider` that take `Optional<Path> outputJar` to take `Optional<SourcePath> outputJar` instead.

`SourcePaths` are much safer because they encapsulate the project filesystem. That way we know they can always be resolved correctly.

See https://github.com/facebook/buck/blob/master/src/com/facebook/buck/android/AndroidPackageableCollector.java#L193 for example.

davido:
@Coneko I addressed your comments in https://github.com/facebook/buck/pull/552

Issue #544, opened 2015-12-03
Poblem with $(location :foo) macro expansion in cross-cell context


davido:
I'm trying to isolate third party library in its own cell inside my project:

```
repositories.jgit = lib/jgit
```

The `jgit_src` is defined as:

```
$ buck targets --json //lib/jgit:jgit_src
[
{
  "binaryJar" : ":jgit__download_src",
  "buck.base_path" : "",
  "buck.direct_dependencies" : [ "//lib:LICENSE-jgit", "//lib/jgit:jgit__download_src" ],
  "buck.type" : "prebuilt_jar",
  "deps" : [ "//lib:LICENSE-jgit" ],
  "gwtJar" : null,
  "javadocUrl" : null,
  "mavenCoords" : null,
  "name" : "jgit_src",
  "sourceJar" : null,
  "visibility" : [ "PUBLIC" ]
}
]
```

And the target output is located:

```
$ buck targets --show_output //lib/jgit:jgit_src
/lib/jgit:jgit_src buck-out/gen/lib/jgit/jgit_src.jar

$ ls -all buck-out/gen/lib/jgit/jgit_src.jar
-rw-r--r-- 1 davido users 2025387 Dec  4 00:23 buck-out/gen/lib/jgit/jgit_src.jar
```

Now, the consumer is saying this:

```
genrule(
  name = 'jgit_edit_src',
  cmd = 'unzip -qd $TMP $(location @jgit//:jgit_src) ' +
    'org/eclipse/jgit/diff/Edit.java;' +
    'cd $TMP;' +
    'zip -Dq $OUT org/eclipse/jgit/diff/Edit.java',
  out = 'edit.src.zip',
)
```

But the `location` macros expansion is relocated in root context of cell container:

```
(cd /home/davido/projects/gerrit/buck-out/gen/gerrit-patch-jgit/jgit_edit_src__srcs && OUT=/home/davido/projects/gerrit/buck-out/gen/gerrit-patch-jgit/jgit_edit_src/edit.src.zip TMP=/home/davido/projects/gerrit/buck-out/gen/gerrit-patch-jgit/jgit_edit_src__tmp /bin/bash -e -c 'unzip -qd $TMP /home/davido/projects/gerrit/buck-out/gen/jgit_src.jar org/eclipse/jgit/diff/Edit.java;cd $TMP;zip -Dq $OUT org/eclipse/jgit/diff/Edit.java')
```

[...]

PR #549, opened 2015-12-05
Cross-cell: Fix file system mismatch in macro expansion machinery

davido:
Owning rule must not necessary belong to the same cell as the macros
it contains:

  genrule(
    name = 'foo',
    cmd = 'echo "$(location @bar//:baz)" > $OUT',
    out = 'qux',
  )

Cell can be located in the same project, but can be hijacked by the
command line parameter. Consider the following configuration in
.buckconfig for the example above:

  [repositories]
    bar = lib/bar

It can optionally be relocated with --config CLI parameter:

  $ buck build //:foo
  $ buck build --config repositories.bar=/path/to/hijacked/bar //:foo

IOW it is always wrong to assume that the owning rule and all macros
it contains belong to the same cell. Remove filesystem parameter from
the macro expansion machinery and rely on the rule that has access
to its own filesystem to perform correct path resolution.

TEST PLAN:

1. Clone JGit with this patch: [1].
2. Clone Gerrit Code Review with this patch: [2].
3. Replace JGit cell during Gerrit build, with:

  $ buck build --config repositories.jgit=/path/to/jgit gerrit

Observe that without this diff, '$(location @jgit//:jgit_src)' is
expanded to <path-to-gerrit>/buck-out/gen/jgit_src.jar. With this diff,
file system mismatch is fixed and the expansion is correct:
/path/to/jgit/buck-out/gen/jgit_src.jar

[1] https://git.eclipse.org/r/61938
[2] https://gerrit-review.googlesource.com/73000

Fixes: https://github.com/facebook/buck/issues/544

### DIFF
--- a/src/com/facebook/buck/android/AndroidBinaryDescription.java
+++ b/src/com/facebook/buck/android/AndroidBinaryDescription.java
@@ -251,8 +251,7 @@ public class AndroidBinaryDescription
         MACRO_HANDLER.getExpander(
             params.getBuildTarget(),
             params.getCellRoots(),
-            resolver,
-            params.getProjectFilesystem()),
+            resolver),
         args.preprocessJavaClassesBash,
         rulesToExcludeFromDex,
         result,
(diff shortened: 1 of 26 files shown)


Coneko:
@facebook-github-bot shipit

Coneko:
@facebook-github-bot shipit

PR #563, opened 2015-12-09
IntelliJ modules inherit JDK versions from config

illuminace:
IntelliJ project JDK versions are configurable in the ```[intellij]``` section of ```.buckconfig``` via ```jdk_name``` and ```language_level```. These affect the project-wide versions configured in ```.idea/misc.xml```

JDK versions in .iml module configs, however, are hardcoded to JDK 1.7. This needs to be configurable from somewhere - I've taken it to be the ```[intellij]``` configuration section.

i.e. given the following configuration
```
[intellij]
  jdk_name = 1.8
  jdk_type = JavaSDK
  language_level = JDK_1_8
```
then the .iml dep

```<orderEntry type="jdk" jdkName="1.7" jdkType="JavaSDK" />```

should be 

```<orderEntry type="jdk" jdkName="1.8" jdkType="JavaSDK" />```

### DIFF
--- a/src/com/facebook/buck/jvm/java/intellij/Project.java
+++ b/src/com/facebook/buck/jvm/java/intellij/Project.java
@@ -474,7 +474,10 @@ public class Project {
       if (module.isIntelliJPlugin()) {
         jdkDependency = SerializableDependentModule.newIntelliJPluginJdk();
       } else {
-        jdkDependency = SerializableDependentModule.newStandardJdk();
+        jdkDependency = SerializableDependentModule.newStandardJdk(
+            intellijConfig.getJdkName(),
+            intellijConfig.getJdkType()
+        );
       }
     }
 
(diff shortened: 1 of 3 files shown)


sdwilsh:
cc @alsutton who might like this

alsutton:
I was wondering if you considered keeping newStandardJdk with no parameters and creating a new method newJdk which took the JDK information. newStandardJdk could call through to newJdk with Optional.absent parameters, and doing it that way would allow us to keep the newStandardJdk method signature, and create a newly named method which can create an object for any Jdk rather than using a method name which has "standard" as part of the method name which could be confusing.

illuminace:
Fair point. Perhaps I misjudged it, but I took the "standard" part of the name to relate to Buck's internal JDK type name as opposed to the IntelliJ type name. I.e. ```newStandardJdk``` sets the internal type to "jdk", whereas there's a similar method called ```newInheritedJdk``` which sets the internal type to "inheritedJdk". I felt like this distinction should be clearly reflected in the method signature, hence I opted to leave the name as-is. However, one could quite correctly argue that ```newJdk```, without any modifier, naturally reflects a "standard" or "default" state, so I'm perfectly open to the idea of renaming it. 

As for separating the methods, deferring to the original; we could do, but I don't see much benefit over simply using ```Optional```'s defaults? I.e. I can't envisage a case at this juncture where the added indirection would be useful - why would we need to call the original method, to _enforce_ use of JDK 1.7? I took that to be a Facebook-ism, addressed by the comments at the top of ```Project.java```: "Hopefully over time, the Facebook-specific logic will be removed". I've most probably overlooked something though, so I'm perfectly happy to adjust it if you'd prefer.

Cheers

PR #564, opened 2015-12-09
Run groovy related tests in travis

grumpyjames:
Summary:
We want to run the groovy tests in CI, so we install groovy in travis
and set GROOVY_HOME appropriately.

Test-plan: Look at the travis output, check that 
GroovyLibraryIntegrationTest and GroovyBuckConfigTest PASS 
rather than ASSUME


sdwilsh:
Memo to myself: fix the Android stuff that randomly broke.

@facebook-github-bot shipit

PR #566, opened 2015-12-10
Disable broken tests

davido:


davido:
Something ent here wrong.

PR #568, opened 2015-12-10
Extract population of JavacOptions from JavaLibraryDescription

grumpyjames:
Summary:

Shortly we will want to be able to pass a JavacOptions around in
groovy land to support cross compilation. In addition, many classes
currently depend on (and also duplicate an extra piece of)
JavaLibraryDescription's population of JavacOptions from its Arg. In
this change we extract a parent class, JavacArg, and move all
JavacOptions related behaviour to JavacArgInterpreter.

Test-plan: CI

### DIFF
--- a/src/com/facebook/buck/jvm/java/JavacArg.java
+++ b/src/com/facebook/buck/jvm/java/JavacArg.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2015-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


sdwilsh:
@facebook-github-bot import

src/com/facebook/buck/jvm/java/JavacArg.java
```diff
@@ -0,0 +1,71 @@
...
+import com.facebook.buck.rules.BuildRuleResolver;
+import com.facebook.buck.rules.SourcePath;
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSortedSet;
+
+import java.nio.file.Path;
+
+public class JavacArg {
```
sdwilsh: How would you feel about calling this `JvmLibraryArg

grumpyjames: Sounds good to me!

sdwilsh:
This should land later today.  I'm just running it through our internal CI again with the renames to make sure I got it all right.

PR #569, opened 2015-12-10
Abstract AbstractJavacOptions away from building the command line

grumpyjames:
Summary:

Instead of getting it to build the list of command line strings, get
JavacStep to visit AbstractJavacOptions with an OptionConsumer
instead. This will allow other JavacOptions users to know about the
command line flags without having to filter all the hyphens, for
example when attempting to pass key value pairs to groovyc's -J flag.

Test Plan: CI

### DIFF
--- a/src/com/facebook/buck/jvm/java/JavacStep.java
+++ b/src/com/facebook/buck/jvm/java/JavacStep.java
@@ -38,6 +38,7 @@ import com.google.common.collect.ImmutableSortedSet;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Collection;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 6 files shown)


grumpyjames:
Starting to worry that these somewhat abstract changes are a bit meaningless without the context; so I've been pushing the next pieces of work (quite spiky, not well tested, naming not well considered) to https://github.com/grumpyjames/buck/tree/spike/lmax-beta-2 .

Tip there at the moment shows one possible vision of a world where there is JvmLibraryDescription with Java/Groovy policies used to configure it. Groovy also gains java cross-compilation.

(It also has groovy tests, and a horrible hack that stops us hitting a classloader bug that we don't understand yet, but there you go)

src/com/facebook/buck/jvm/java/JavacStep.java
```diff
@@ -257,9 +258,24 @@ public String getShortName() {
...
+      @Override
+      public void addFlag(String flagName) {
+        builder.add("-" + flagName);
+      }
+
+      @Override
+      public void addExtras(Collection<String> extras) {
+        builder.addAll(extras);
+      }
+    }, filesystem.getAbsolutifier());
```
sdwilsh: nit: the formatting here is off.  `filesystem.getAbsolutifier());` should be on a line after `},`.

sdwilsh:
@facebook-github-bot shipit

PR #575, opened 2015-12-12
Update okhttp and okio.

LegNeato:
okhttp-2.4.0 is required to get this change mentioned in https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-240-rc1 :

"Interceptors lists are now deep-copied by OkHttpClient.clone(). Previously clones shared interceptors, which made it difficult to customize the interceptors on a request-by-request basis."

Upgrading lets a future diff use different headers for read and write OkHttp clients.

I upgraded to the latest of both (okio is a dep) because why not.

Test Plan:

* All tests passed.
* Built with ant.
* Built with buck.
* Built with buck via in-repo python script.
* Ran some builds with cache set to `http` and saw the requests worked


sdwilsh:
@facebook-github-bot import

sdwilsh:
Hmm, I do need the bot to approve this however...

sdwilsh:
@facebook-github-bot shipit

sdwilsh:
Whoops, this should have updated `.classpath` and `.idea/modules/buck_lib.iml` too.  I'll fix it myself.

PR #577, opened 2015-12-13
IntelliJ project support for resources

illuminace:
Summary:
Source and test roots are supported in generated IntelliJ projects, but there is no means at present to supply resource roots. This adds an optional `resource_roots` argument to the `project_config` build rule, which adds `java-resource` `sourceFolder` elements to the project's .iml config

Test Plan:
Add a `resource_roots` declaration to a project_config, run buck project, and verify both that the `java-resource` is present in the generated iml file, and that IntelliJ picks up the resource directory.

```
project_config(
  ...normal project config... ,
  resource_roots = [ 'res' ]
)
```

results in

```
<sourceFolder url="file://$MODULE_DIR$/res" type="java-resource" />
```

### DIFF
--- a/src/com/facebook/buck/command/intellij.py
+++ b/src/com/facebook/buck/command/intellij.py
@@ -285,6 +285,10 @@ def write_modules(modules, generate_minimum_project, android_auto_generation_dis
                 'is_test_source': str(source_folder['isTestSource']).lower(),
                 'package_prefix': package_prefix
             }
+        for resource_folder in module['resourceFolders']:
+            xml += '\n      <sourceFolder url="%(url)s" type="java-resource" />' % {
+                'url': resource_folder['url']
+            }
         for exclude_folder in module['excludeFolders']:
             xml += '\n      <excludeFolder url="%s" />' % exclude_folder['url']
         for exclude_folder in sorted(additional_excludes[module['pathToImlFile']]):
--- a/src/com/facebook/buck/jvm/java/intellij/Project.java
+++ b/src/com/facebook/buck/jvm/java/intellij/Project.java
@@ -374,6 +374,9 @@ public class Project {
         projectConfig.getSourceRoots(),
         false /* isTestSource */);
 
+    // resource folders
+    addResourceFolders(module, projectConfig.getResourceRoots());
+
     addRootExcludes(module, projectConfig.getSrcRule(), projectFilesystem);
 
     // At least one of src or tests should contribute a source folder unless this is an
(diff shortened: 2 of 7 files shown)


marcinkosiba:
Hey!
  Thanks for the pull request and apologies for not getting back to you sooner!
  I don't think it's necessary for resource roots to be manually specified: we already have that information on android_res() rules, so we don't need to force the user to type that in a second time.
  Al recently landed support for the Android plugin to the --experimental-ij-generation project mode in c2691f3c5ccdc162884c70ee39cddfe317ce551b, which should do what I just described.

illuminace:
Hi @marcinkosiba. No problem at all, thanks for following up!

I hadn't considered the android_resource rules. You make a really good point, it might be possible to aggregate project resources from existing configuration rules and avoid additional configuration. 

My use-case is a little different to your example though, as I'm dealing with plain old java projects as opposed to Android projects. I don't think java_library's resources section can help us achieve this - targets, bespoke entries for particular libs... it wouldn't give us what we need. We could perhaps attempt to infer based on resource_roots, which looks more viable, but it smacks of overloading and I suspect would cause problems in the real world.

I'm very open to suggestions - any superfluous configuration that can be removed/not added in the first place is a positive thing. Or alternatively if there's a more formal definition we can envisage that could be reused in other ways...?
 
Thanks

marcinkosiba:
> I don't think java_library's resources section can help us achieve this - targets, bespoke entries for particular libs... it wouldn't give us what we need.

I don't think I understand this bit. From the pull request it seems like you need to add a 
`<sourceFolder type="java-resource" />` line to the .iml file. You could start off simple and only handle resources that are SourcePaths (that way you side-step the issue of genrules, etc..). We already have most of the plumbing you'd need in place for android resources (look for AndroidResourceFolder), it looks like it should work out of the box if you have the JavaLibraryModuleRule add AndroidResourceFolder that correspond to its resources (might make sense to rename AndroidResourceFolder->ResourceFolder).

juddgaddie:
@illuminace I'm totally with you on what you are trying to do here, and also think adding the ability to add resources in project_config is the way to go however the patch is missing the ability to distinguish between test resources and app resources.

From the module.iml:
```xml
<sourceFolder url="file://$MODULE_DIR$/src/main/resources" type="java-resource" />
<sourceFolder url="file://$MODULE_DIR$/src/test/resources" type="java-test-resource" />
```

marcinkosiba:
FYI the default for `buck project` changed (it took a while), so, unless you run with the --deprecated-ij-generation option, `project_config` entries aren't used.

Marking as needs revision because I imagine the diff doesn't apply cleanly anymore.

dxxchen:
As far as I can tell, this is not yet resolved?

```python
java_library(
  name = 'foo',
  resources = ...,
  resources_root = "src/main/resources",
  ...
)
```

creates an .iml file that contains

```xml
<sourceFolder url="file://$MODULE_DIR$/src/main/resources" isTestSource="false" .../>
```

I would expect that resource folders have the attribute `type="java-resource"`.

styurin:
Most of this code was removed. `project_config` is long dead

PR #582, opened 2015-12-17
Fix not properly initialized test setting

davido:
Some of the test setting aren't initialized poroperly. This doesn't
matter atm, but break Java library tests once classpath entry resolution
is switched to absolute paths and rely on rule resolver to be properly
initialized.

Test Plan:

  buck test


sdwilsh:
@facebook-github-bot shipit

PR #584, opened 2015-12-21
RFC: pex-based in-place packages.

mikekap:
This adds support for another python package style - pex_inplace. It uses the regular pex code to make the chroot, but then just doesn't zip it. This is a use case supported by pex.

Also included are some optimizations to pex generation, but the speedup is really only seen with the in-place version. Generating buck.pex via regular pex takes ~ 4-5s on my mac; with pex_inplace it's down to ~300ms. This is still a little slower than the regular inplace mode, but supports more features (e.g. prebuilt packages, proper isolation, etc).

The optimizations to pexing are:
 - Don't produce a code hash in PEX-INFO in inplace mode. This is only used to check if we need to extract the contents of a pex from a zip; if it's already a directory, it's useless. (3-4s)
 - Pass the python version information from buck to pex.py instead of querying it every time make_pex is run (400ms).

### DIFF
--- a/src/com/facebook/buck/python/AbstractPythonVersion.java
+++ b/src/com/facebook/buck/python/AbstractPythonVersion.java
@@ -25,11 +25,14 @@ import org.immutables.value.Value;
 abstract class AbstractPythonVersion {
 
   @Value.Parameter
-  public abstract String getVersionString();
+  public abstract String getInterpreterName();
+
+  @Value.Parameter
+  public abstract String getVersionString();  // X.Y.Z
 
   @Override
   public String toString() {
-    return getVersionString();
+    return getInterpreterName() + " " + getVersionString();
   }
 
 }
--- a/src/com/facebook/buck/python/PythonBuckConfig.java
+++ b/src/com/facebook/buck/python/PythonBuckConfig.java
@@ -25,6 +25,7 @@ import com.facebook.buck.model.BuildTarget;
 import com.facebook.buck.model.Flavor;
 import com.facebook.buck.model.ImmutableFlavor;
 import com.facebook.buck.rules.BuildRuleResolver;
+import com.facebook.buck.rules.CommandTool;
 import com.facebook.buck.rules.PathSourcePath;
 import com.facebook.buck.rules.SourcePath;
 import com.facebook.buck.rules.Tool;
@@ -36,6 +37,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
+import com.google.common.base.Splitter;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
@@ -46,8 +48,6 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.EnumSet;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.annotation.Nonnull;
 
@@ -57,9 +57,6 @@ public class PythonBuckConfig {
 
   private static final String SECTION = "python";
   private static final String PYTHON_PLATFORM_SECTION_PREFIX = "python#";
-
-  private static final Pattern PYTHON_VERSION_REGEX =
-      Pattern.compile(".*?(\\wy(thon|run) \\d+\\.\\d+).*");
 
   // Prefer "python2" where available (Linux), but fall back to "python" (Mac).
   private static final ImmutableList<String> PYTHON_INTERPRETER_NAMES =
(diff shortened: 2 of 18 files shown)


mikekap:
ping :)

sdwilsh:
I think @andrewjcg is on a plane today, but he might be jet lagged enough to look at this when he lands :)

mikekap:
Sorry I didn't mean to rush, just wanted to make sure this didn't get dropped given the pex commit/revert stuff.

mikekap:
Ping :)

mikekap:
If it'll help, I can split this up into two changes: one to add pex_inplace and the other to optimize the ~400ms interpreter lookup.

sdwilsh:
@facebook-github-bot import

andrewjcg:
So sorry for taking this long to get to this, but it's definitely on my radar now.  Generally it looks good, I just have a few concerns and suggestions about separating out the PEX-specific stuff from the more general python stuff which I'll comment on inline.

src/com/facebook/buck/python/AbstractPythonVersion.java
```diff
@@ -25,11 +25,22 @@
...
+  @Value.Parameter
+  public abstract String getVersionString();  // X.Y.Z
+
+  public String getPexCompatibilityVersion() {
+    StringBuilder sb = new StringBuilder(20);
+    sb.append(getInterpreterName());
+    sb.append(" ");
+    sb.append(getVersionString().replace('.', ' '));
+    return sb.toString();
+  }
```
andrewjcg: It'd be nice if we didn't have more PEX-specific stuff in the core python support.  Can we just have `PythonVersion` support the interpreter-name and version (as you've added above) and have the make_pex.py script do the necessary PEX-specific formatting?

src/com/facebook/buck/python/PythonBuckConfig.java
```diff
@@ -286,6 +301,7 @@ public PackageStyle getPackageStyle() {
   public enum PackageStyle {
     STANDALONE,
     INPLACE,
+    PEX_INPLACE,
```
andrewjcg: Long-term we'd like to move all the PEX-specific naming/code out of Buck, just supporting the general way to use arbitrary custom Python packagers via the `python:path_to_pex` config setting.

In this vein, it'd be nice to at least keep the zip-vs-directory choice separate from the inplace-vs-standalone choice. What do you think about having a separate PexStyle enum with `ZIP` and `DIRECTORY` choices (which would apply when `STANDALONE` is used with the default PEX builder)?

Another option might be to make the current in-place package style more general and provide a config option to choose between "simple" and "PEX" sub-styles, the latter of which implements the in-place PEX. The annoying thing about this is that the current code path for in-place assumes that the generated symlink trees with the modules will be used and that the in-place starter script won't do any of it's own packaging (even if it is just copying to a directory).

mikekap: Hmm I think if the goal is to remove most of the pex-specific parts of the code, I think just adding something like `python:pex_args` and using the existing python platform flavor machinery might be the right approach. At that point I think it would make sense to convert the `INPLACE` style to this scheme, by e.g. having a magic `path_to_pex` value that uses the existing inplace code. Alternatively, an option might be to extract the existing inplace code into a java binary, although that might be a bit more of a rewrite.

sdwilsh:
@facebook-github-bot import

test/com/facebook/buck/python/PythonBinaryIntegrationTest.java
```diff
@@ -126,6 +141,10 @@ public void commandLineArgs() throws IOException {
     assertThat(
         result.getStdout(),
         containsString("HELLO WORLD"));
+
+    if (pexDirectory) {
```
andrewjcg: It seems a bit weird to test for this in the `commandLineArgs` test.  Maybe move this to a separate `testOutput` test where we check for a directory for the `--directory` case and files for the others?

test/com/facebook/buck/python/PythonBinaryIntegrationTest.java
```diff
@@ -287,10 +306,10 @@ private PythonBuckConfig getPythonBuckConfig() throws IOException {
             new ProjectFilesystem(tmp.getRootPath()),
             Architecture.detect(),
             Platform.detect(),
-            ImmutableMap.<String, String>of());
+            ImmutableMap.copyOf(System.getenv()));
```
andrewjcg: Why was this necessary?

andrewjcg: Was this to find the `python` binary in the path?

andrewjcg:
Looks good thanks!  Just a few minor questions/comments inline.

test/com/facebook/buck/python/PythonBinaryIntegrationTest.java
mikekap: Yep. I think this might also help windows tests since you can't use shebangs.

mikekap:
Ping

Coneko:
I'm going to import since @andrewjcg said the remaining comments were minor and they seem to have been addressed.

Coneko:
@facebook-github-bot shipit

sdwilsh:
It looks like @andrewjcg reverted this in https://github.com/facebook/buck/commit/a03737b3e2e31e833e30e9013052dad7a936e5cb, but I don't see a comment here about it.

Issue #163, opened 2014-07-24
Assumptions aren't errors


anotherchrissmith:
I just saw this commit:

https://github.com/facebook/buck/commit/f5f748464ad827f5ff4f91fa78f5dffac5635c1a

and wondered why you'd ever want to treat an assumption failure as an error.  According to junit, assuptions are... "A set of methods useful for stating assumptions about the conditions in which a test is meaningful. A failed assumption does not mean the code is broken, but that the test provides no useful information.".  We use assumptions along with feature toggles, e.g. we add assumeTrue(Feature.X.isEnabled()) at the start of a test.  If the feature is disabled, we don't run the test.

Judging by your commit message, it looks like you're using assumptions where you should be using assertions.

I realise this is an option that you don't have to use, but it looks like it's based on a misinterpretation of what assumptions are for.

PR #589, opened 2015-12-28
Scala support

mikekap:
Adds barebones scala_library & scala_test (still junit) support. This is sort of experimental - Scala compiles a bit differently than java (i.e. dependencies can be executed as part of the compile), so I'm not sure this works in all cases. It does work for the simple cases though.

LMK if you'd like to document this support or leave it until it's a bit more solid.


sdwilsh:
I'm fine with not documenting this until we (well, you) feel it is more solid.

mikekap:
Particularly, zinc is almost a necessity with any size scala project. It works a bit differently than the normal compiler. It's an incremental compiler to the extent that it wants the previously compiled class files & uses a nailgun daemon itself.

I'm not sure how to integrate that into buck yet (other than "externally"), but certainly some scala support is better than no scala support :)

src/com/facebook/buck/jvm/scala/ScalaBuckConfig.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public ScalaBuckConfig(final BuckConfig delegate) {
+    this.delegate = delegate;
+  }
+
+  public Tool getScalac(BuildRuleResolver resolver) {
+    CommandTool.Builder scalac = new CommandTool.Builder(findScalac(resolver));
+
+    // Add some standard options.
+    scalac.addArg("-target:" + delegate.getValue("scala", "target_level").or("jvm-1.7"));
```
ruibm: Can you make "scala" a constant please?

src/com/facebook/buck/jvm/scala/ScalaBuckConfig.java
```diff
@@ -0,0 +1,91 @@
...
+
+    String scalaHome = delegate.getEnvironment().get("SCALA_HOME");
+    if (scalaHome != null) {
+      Path scalacInHomePath = Paths.get(scalaHome, "bin", "scalac");
+      if (scalacInHomePath.toFile().exists()) {
+        return new HashedFileTool(scalacInHomePath);
+      }
+    }
+
+    throw new HumanReadableException("Could not find scalac. Consider setting scala.compiler.");
```
ruibm: Could you please have two different messages for the case where SCALA_HOME is not defined and the case when it's defined in the env however the file does not exist.

src/com/facebook/buck/jvm/scala/ScalaLibraryDescription.java
```diff
@@ -0,0 +1,170 @@
...
+    // Note: scala does not have a exported_deps because scala needs the transitive closure of
+    // dependencies to compile. deps is effectively exported_deps.
+    public Optional<ImmutableSortedSet<BuildTarget>> providedDeps;
+    public Optional<ImmutableSortedSet<BuildTarget>> deps;
+
+    @Hint(isInput = false)
+    public Optional<Path> resourcesRoot;
+    public Optional<String> mavenCoords;
+
+    @Hint(isDep = false) public Optional<ImmutableSortedSet<BuildTarget>> tests;
```
ruibm: For consistency, either keep the @Hint annotation in the same line or breakline like in line #163.

ruibm:
Thanks for sending this PR. Looks good - I just added few minor comments.

ruibm:
Looks good! Thanks for the changes. Proceeding with the merge.

PR #593, opened 2016-01-03
Use immutables to simplify some code in the query package.

mogers:
Summary:
Unit tests for the QueryParser require the ability to compare QueryExpression's. Overriding equals() requires one to override hashCode() as well. 
This was done for FunctionExpression, SetExpression and Argument, but the code is a bit messy. This PR uses immutables to generate this code, making it less prone to errors.

Test plan: CI


src/com/facebook/buck/query/AbstractTargetLiteral.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public void collectTargetPatterns(Collection<String> literals) {
+    literals.add(getPattern());
+  }
+
+  @Override
+  public String toString() {
+    // Keep predicate consistent with Lexer.scanWord!
```
sdwilsh: I don't understand this comment, especially when looking at the code.  I realize that this is somewhat copypasta, but it should still make sense, and I don't think it does right now.

src/com/facebook/buck/query/AbstractFunctionExpression.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2015-present Facebook, Inc.
```
sdwilsh: nit: 2016 now

src/com/facebook/buck/query/AbstractTargetLiteral.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2015-present Facebook, Inc.
```
sdwilsh: nit: 2016

src/com/facebook/buck/query/AbstractTargetLiteral.java
mogers: You're right, this piece is still copy+paste, and the Lexer was imported from Bazel with minor changes.
IIRC, unit tests are the only place TargetLiteral#toString() is used. I think this comment is a warning to keep this method consistent with the rules for quoting arguments, in order to compare the expected values properly. 
However, I now realize that the Lexer does not enforce all the rules mentioned in the comments (e.g. an unquoted word starting by '*' or '$' does not trigger any error).

k21:
@mogers, are you interested in rebasing and updating this pull request or should I close it?

mogers:
Hi @k21 , I'm sorry, I completely forgot about this. I'll try to do it in the next couple of days.

facebook-github-bot-5:

mogers:
Hi @k21, I apologize again. I've been having too much on my plate. 
I tried to pick this again but my CLA expired and I'm having problems rebasing with the current version. 
I think this change is not important, so I will abandon this pr.

Issue #594, opened 2016-01-03
classpath macro resolution for dependent libraries is broken in cross-cell context


davido:
Given the following cell definition:

```
[repositories]
  jgit = lib/jgit
```

The calsspath is correctly resolved with `classpath` macro:

```
genrule(
  name = 'dump_correct_classpath',
  cmd = 'echo $(classpath @jgit//:jgit) > $OUT',
  out = 'dump_correct_classpath.txt',
)
```

The content of the ouput file is correct:

```
$ cat buck-out/gen/tools/dump_correct_classpath/dump_correct_classpath.txt
/home/davido/projects/gerrit/lib/jgit/buck-out/gen/ewah.jar:/home/davido/projects/gerrit/lib/jgit/buck-out/gen/jgit.jar
```

The `audit classpath` command is correct as well. However, when the same cell library is used in transitive dependency, cell path is erroneously re-located to buck root context:

```
java_library(
  name = 'broken',
  srcs = [],
  deps = ['@jgit//:jgit'],
  visibility = ['PUBLIC']
)

genrule(
  name = 'dump_broken_deps_classpath',
  cmd = 'echo $(classpath :broken) > $OUT',
  out = 'dump_broken_deps_classpath.txt',
)
```

The output file contains now wrong, non-existing path:

```
$ cat buck-out/gen/tools/dump_broken_classpath/dump_broken_classpath.txt
/home/davido/projects/gerrit/buck-out/gen/ewah.jar:/home/davido/projects/gerrit/buck-out/gen/jgit.jar
```

Interestingly, the `audit classpath` command still reports the correct classpath:

```
$ buck audit classpath tools:broken
:::
::: '.nobuckcheck' file is present. Not updating buck.
:::
[+] PROCESSING BUCK FILES...0.1s
/home/davido/projects/gerrit/lib/jgit/buck-out/gen/ewah.jar
/home/davido/projects/gerrit/lib/jgit/buck-out/gen/jgit.jar
```

Note that, I'm using the most recent Buck version with this PR on top: [1].

* [1] https://github.com/facebook/buck/pull/558

davido:
This diff fixed it:

```
diff --git a/src/com/facebook/buck/rules/macros/ClasspathMacroExpander.java b/src/com/facebook/buck/rules/macros/ClasspathMacroExpander.java
index 99b911d..a7570ad 100644
--- a/src/com/facebook/buck/rules/macros/ClasspathMacroExpander.java
[...]

PR #595, opened 2016-01-04
Cross-cell: Fix file system mismatch in classpath macro expansion

davido:
Fixes: #594.

Summary: The absolutifier, used in macro owning rule may be from
different cell.

That why it's always wrong to assume, that all classpath entries
belong to the same file system. The rule own file system must be
used to resolve every single classpath entry.

Test Plan:

Build Gerrit Code Review with this change applied: [1].

* [1] https://gerrit-review.googlesource.com/73000


sdwilsh:
@facebook-github-bot shipit

PR #596, opened 2016-01-04
Cross compile groovy and java where appropriate

grumpyjames:
Summary:

groovyc is able to collaborate with javac in order to compile .java
and .groovy files that express circular dependencies. No comment is
made as to whether this is a good idea, but it is now possible with
buck if one wished to.

By default, the inner javac will be parametrised in the same we it
would be as in a java library rule; these parameters can be overridden
in the same way they can be for a java_library rule.

Test plan: CI (there are new cases in GroovyLibraryIntegrationTest)

### DIFF
--- a/src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
+++ b/src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
@@ -112,8 +112,7 @@ public class GroovyLibraryDescription implements Description<GroovyLibraryDescri
                         params,
                         resolver,
                         pathResolver,
-                        args
-                    )),
+                        args)),
                 Optional.<Path>absent(),
                 Optional.<String>absent(),
                 ImmutableSortedSet.<BuildTarget>of()));
--- a/src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
+++ b/src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
@@ -20,6 +20,9 @@ import static com.facebook.buck.jvm.common.ResourceValidator.validateResources;
 
 import com.facebook.buck.jvm.java.CalculateAbi;
 import com.facebook.buck.jvm.java.DefaultJavaLibrary;
+import com.facebook.buck.jvm.java.JavacOptions;
+import com.facebook.buck.jvm.java.JavacOptionsFactory;
+import com.facebook.buck.jvm.java.JvmLibraryArg;
 import com.facebook.buck.model.BuildTarget;
 import com.facebook.buck.rules.BuildRule;
 import com.facebook.buck.rules.BuildRuleParams;
@@ -45,10 +48,14 @@ public class GroovyLibraryDescription implements Description<GroovyLibraryDescri
   public static final BuildRuleType TYPE = BuildRuleType.of("groovy_library");
 
   private final GroovyBuckConfig groovyBuckConfig;
+  // For cross compilation
+  private final JavacOptions defaultJavacOptions;
 
   public GroovyLibraryDescription(
-      GroovyBuckConfig groovyBuckConfig) {
+      GroovyBuckConfig groovyBuckConfig,
+      JavacOptions defaultJavacOptions) {
     this.groovyBuckConfig = groovyBuckConfig;
+    this.defaultJavacOptions = defaultJavacOptions;
   }
 
   @Override
@@ -99,7 +106,14 @@ public class GroovyLibraryDescription implements Description<GroovyLibraryDescri
                 /* additionalClasspathEntries */ ImmutableSet.<Path>of(),
                 new GroovycToJarStepFactory(
                     groovyBuckConfig.getGroovyCompiler().get(),
-                    args.extraArguments),
+                    args.extraGroovycArguments,
+                    JavacOptionsFactory.create(
+                        defaultJavacOptions,
+                        params,
+                        resolver,
+                        pathResolver,
+                        args
+                    )),
                 Optional.<Path>absent(),
                 Optional.<String>absent(),
                 ImmutableSortedSet.<BuildTarget>of()));
@@ -115,13 +129,12 @@ public class GroovyLibraryDescription implements Description<GroovyLibraryDescri
   }
 
   @SuppressFieldNotInitialized
-  public static class Arg {
+  public static class Arg extends JvmLibraryArg {
     public Optional<ImmutableSortedSet<SourcePath>> srcs;
     public Optional<ImmutableSortedSet<SourcePath>> resources;
-    public Optional<ImmutableList<String>> extraArguments;
+    public Optional<ImmutableList<String>> extraGroovycArguments;
     public Optional<ImmutableSortedSet<BuildTarget>> providedDeps;
     public Optional<ImmutableSortedSet<BuildTarget>> exportedDeps;
     public Optional<ImmutableSortedSet<BuildTarget>> deps;
   }
-
 }
--- a/src/com/facebook/buck/jvm/groovy/GroovycStep.java
+++ b/src/com/facebook/buck/jvm/groovy/GroovycStep.java
@@ -102,7 +102,8 @@ class GroovycStep implements Step {
 
     command.addAll(groovyc.getCommandPrefix(resolver));
 
-    String classpath = Joiner.on(File.pathSeparator).join(transform(declaredClasspathEntries, toStringFunction()));
+    String classpath =
+        Joiner.on(File.pathSeparator).join(transform(declaredClasspathEntries, toStringFunction()));
     command
         .add("-cp")
         .add(classpath.isEmpty() ? "''" : classpath)
@@ -125,9 +126,10 @@ class GroovycStep implements Step {
           // The implementation of `appendOptionsTo` provides a blank default, which
           // confuses the cross compilations step's javac (it won't find any class files
           // compiled by groovyc).
-          if (!option.equals("sourcepath")) {
-            command.add("-J" + String.format("%s=%s", option, value));
+          if (option.equals("sourcepath")) {
+            return;
           }
+          command.add("-J" + String.format("%s=%s", option, value));
         }
 
         @Override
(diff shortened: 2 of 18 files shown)


src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -61,8 +69,8 @@
 
   @Override
   public int execute(ExecutionContext context) throws IOException, InterruptedException {
-
-    ProcessBuilder processBuilder = new ProcessBuilder(createCommand());
+    ImmutableList<String> command = createCommand();
+    ProcessBuilder processBuilder = new ProcessBuilder(command);
```
sdwilsh: Two things:
* This change looks spurious
* We shouldn't be adding new uses of `ProcessBuilder` in the codebase.  You didn't do this here, but I'm letting you know.  We should be using `ProcessExecutor` or `ListeningProcessExecutor` instead.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -91,14 +98,61 @@ public String getDescription(ExecutionContext context) {
   }
 
   private ImmutableList<String> createCommand() {
-    ImmutableList.Builder<String> command = ImmutableList.builder();
-    command.addAll(groovyc.getCommandPrefix(resolver))
+    final ImmutableList.Builder<String> command = ImmutableList.builder();
+
+    command.addAll(groovyc.getCommandPrefix(resolver));
+
+    String classpath = Joiner.on(":").join(transform(declaredClasspathEntries, toStringFunction()));
```
sdwilsh: Again, this isn't wrong in this diff since you just moved the code, but we shouldn't hardcode `:` here, but instead use `File.pathSeparator` so this works on Windows too.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -91,14 +98,61 @@ public String getDescription(ExecutionContext context) {
...
     return command.build();
   }
+
+  private void addCrossCompilationOptions(final ImmutableList.Builder<String> command) {
+    if (shouldCrossCompile()) {
+      command.add("-j");
+      javacOptions.appendOptionsTo(new OptionsConsumer() {
+        @Override
+        public void addOptionValue(String option, String value) {
+          // javac won't find things compiled with groovyc otherwise.
```
sdwilsh: otherwise?  It seems like some text might be missing here or it is misleading me how it is worded.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -91,14 +98,61 @@ public String getDescription(ExecutionContext context) {
...
+          // alternatively, we could convert all the paths to be absolute.
```
sdwilsh: when you say "all paths", what do you mean?

test/com/facebook/buck/jvm/groovy/GroovyLibraryIntegrationTest.java
```diff
@@ -66,15 +64,38 @@ public void shouldCompileLibraryWithDependencyOnAnother() throws Exception {
...
-        workspace.runBuckCommand("build", "//com/example/bad:fail");
+        workspace.runBuckCommand("build", "//com/example/xcompile:xcompile");
+    buildResult.assertSuccess();
+
+    workspace.verify();
+  }
+
+  @Test
+  public void javaOptionsArePassedThroughToTheJavacCompiler() throws Exception {
+    // code requires java 7, source set to java 6
```
sdwilsh: nit: please use complete sentences, which means you need capital letters and punctuation at the end.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
LegNeato: Add a lint if that's the case? Not sure how external people would know and this is perfect for lint.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
LegNeato: Add a lint if that's the case? Not sure how external people would know and this is perfect for lint.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
sdwilsh: We totally should, but I'd rather spend that effort on getting Windows CI which would catch this in the first place (`ProcessBuilder` doesn't properly escape the arguments on Windows, so you might be okay, but you might also fail in a non-obvious way.  The other two classes work around that issue.)

src/com/facebook/buck/jvm/groovy/GroovycStep.java
sdwilsh: Same as above; Windows CI would have caught this for us, and I'm not sure how to write a lint rule that would be context-aware enough to know when `:` is safe to use and when it isn't.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
grumpyjames: Agreed on both counts!

src/com/facebook/buck/jvm/groovy/GroovycStep.java
grumpyjames: So, by default in AbstractJavacOptions, we have this:

    // Set the sourcepath to stop us reading source files out of jars by mistake.
    optionsConsumer.addOptionValue("sourcepath", "");

It turns out that this doesn't play well with cross compilation, to the point where passing empty sourcepath to the 'inner' javac means it cannot find the groovyc compiled classes.

We found two potential paths to avoiding this problem; the simpler one (just ignore the source path option) is presented here. The more complicated solution involves making every path we pass to the compile step absolute; we liked this less.

I'm not particularly happy with either solution and would welcome any further input. I'll go and think about the possible values of source path to see if we always want to drop it, whether we want to log a warning if it's set, etc...

test/com/facebook/buck/jvm/groovy/GroovyLibraryIntegrationTest.java
grumpyjames: Understood!

grumpyjames:
Thanks for reviewing. I'll take another look at the sourcepath option handling and add further commits here.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
grumpyjames: I'm going to undo the spurious part in this PR, but I think switching to ProcessExecutor or equivalent should probably be a separate commit.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
grumpyjames: Will fix this.

grumpyjames:
So, I think I addressed everything, apart from switching from ProcessBuilder to ProcessExecutor, which I'll look at in a separate commit.

I had a further look at sourcepath option; it's only ever set from within `AbstractJavacOptions`, and the setting it provides is incompatible with how cross compilation works. Quietly ignoring it sounds like the right thing to do, so I've updated the comment to explain.

sdwilsh:
You've got a lint failure that Travis caught:

    /home/travis/build/facebook/buck/src/com/facebook/buck/jvm/groovy/GroovycStep.java:105: Line is longer than 100

src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
```diff
@@ -99,7 +106,14 @@ public Arg createUnpopulatedConstructorArg() {
...
                     groovyBuckConfig.getGroovyCompiler().get(),
-                    args.extraArguments),
+                    args.extraGroovycArguments,
+                    JavacOptionsFactory.create(
+                        defaultJavacOptions,
+                        params,
+                        resolver,
+                        pathResolver,
+                        args
+                    )),
```
sdwilsh: small nit (that I'd love if we could lint for): we keep the closing paren after `args`

src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -61,7 +70,6 @@
 
   @Override
   public int execute(ExecutionContext context) throws IOException, InterruptedException {
-
```
sdwilsh: nit: a tiny bit of a spurious change still :)

src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -91,14 +98,63 @@ public String getDescription(ExecutionContext context) {
...
+      javacOptions.appendOptionsTo(new OptionsConsumer() {
+        @Override
+        public void addOptionValue(String option, String value) {
+          // Explicitly disallow the setting of sourcepath in a cross compilation context.
+          // The implementation of `appendOptionsTo` provides a blank default, which
+          // confuses the cross compilations step's javac (it won't find any class files
+          // compiled by groovyc).
+          if (!option.equals("sourcepath")) {
+            command.add("-J" + String.format("%s=%s", option, value));
+          }
```
sdwilsh: This logic might be a tiny bit clearer if you just did a return in the `if` block and afterwards had the common case.

sdwilsh:
I'm otherwise happy with this, and can merge it in once you fix the lint error.

src/com/facebook/buck/jvm/groovy/GroovycStep.java
grumpyjames: I'll fix the others in return for leaving this one ;-)

src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
grumpyjames: both of them in this case, I guess (from context that looks like the rule; will see if I can persuade intellij to do this for me in future)

src/com/facebook/buck/jvm/groovy/GroovycStep.java
sdwilsh: If you can figure out how to make that a project setting, we'd take a PR for that too :)

src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
sdwilsh: If you can figure that out, and it's something we can check into the project, we'd take a PR for that :)

sdwilsh:
@facebook-github-bot shipit

PR #598, opened 2016-01-04
Fix prebuilt package resolution in the new pex when pexs have '#' in the name.

mikekap:
Buck-based python tests tend to have these.

### DIFF
--- a/third-party/py/pex/pex/link.py
+++ b/third-party/py/pex/pex/link.py
@@ -73,7 +73,9 @@ class Link(object):
     """
     purl = urlparse.urlparse(url)
     if purl.scheme == '':
-      purl = urlparse.urlparse(self._normalize(url))
+      purl = urlparse.urlparse(self._normalize(url), allow_fragments=False)
+    elif purl.scheme == 'file':
+      purl = urlparse.urlparse(url, allow_fragments=False)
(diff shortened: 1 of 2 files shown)


third-party/py/pex/pex/link.py
```diff
@@ -73,7 +73,9 @@ def __init__(self, url):
     """
     purl = urlparse.urlparse(url)
     if purl.scheme == '':
-      purl = urlparse.urlparse(self._normalize(url))
+      purl = urlparse.urlparse(self._normalize(url), allow_fragments=False)
+    elif purl.scheme == 'file':
+      purl = urlparse.urlparse(url, allow_fragments=False)
```
sdwilsh: This also adds support for file, right?  Care to update the message and `README.facebook` to be more clear?

mikekap: file:// worked previously - see _normalize() - it just adds a file:// prefix. This just makes sure fragments are ignored when there's no scheme or the scheme is file://

sdwilsh: Yup.  I'm just blind.  I think I should stop reviewing code today :)

sdwilsh:
@facebook-github-bot shipit

PR #603, opened 2016-01-05
Disable the part of the pex builder that pulls in the local setuptools package

mikekap:
This messes with our ability to specify the path to the vendored pkg_resources.


sdwilsh:
@facebook-github-bot shipit

PR #605, opened 2016-01-06
Switch to the new pex (v1.1.1)

mikekap:
The new release properly does site-packages isolation, so pexs will no longer pull in locally installed packages on newer distros.

The old pex source isn't deleted here, although I could definitely do that.

### DIFF
--- a/src/com/facebook/buck/python/BUCK
+++ b/src/com/facebook/buck/python/BUCK
@@ -101,7 +101,7 @@ python_binary(
   name = 'pex',
   main = 'make_pex.py',
   deps = [
-    '//third-party/py/twitter-commons/src/python:twitter-commons',
+    '//third-party/py/pex:pex',
   ],
   visibility = [
     'PUBLIC',
--- a/src/com/facebook/buck/python/make_pex.py
+++ b/src/com/facebook/buck/python/make_pex.py
@@ -16,23 +16,17 @@ import zipfile
 # setuptools at runtime.  Also, locate the `pkg_resources` modules
 # via our local setuptools import.
 if not zipfile.is_zipfile(sys.argv[0]):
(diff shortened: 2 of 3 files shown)


sdwilsh:
@facebook-github-bot import

sdwilsh:
@facebook-github-bot shipit

sdwilsh:
I'll let you do the deletion in another PR.

sdwilsh:
I deeply regret not testing this on Windows first, and now I'm going to have to revert it.  I've got some changes that get it further along, but when I'm building Buck with Buck, python ends up crashing, which is a bit surprising.  I've spent about half my day on trying to resolve it, but at this point, I'm going to have to revert it.

The changes I've got are in https://gist.github.com/sdwilsh/ccb90fbc6a60d885a326 if you want to try to figure out why on Earth it crashes python.

mikekap:
Shoot sorry :( Unfortunately testing on windows isn't quite easy for me - is there a chance you guys might want to set up appveyor or similar?

I think I've got a patch for the crash - AFAICT pex has a bug where it removes sys.prefix from sys.path on windows, so allof the stdlib becomes inaccessible.

sdwilsh:
Setting up CI for Windows has been on my short list for a while 
(especially since I develop on Windows so breaking it means I'm stuck). 
  Sadly, it hasn't made it to the top of that list yet.

mikekap:
Would you like to try landing this again? Pex is still a bit buggy on windows, but at this point I don't think there are regressions.

sdwilsh:
Let me cherrypick the change and see how it goes for me and then I'll re-land it.

sdwilsh:
Bah, that didn't attribute to you.  I can fix that if you'd like; just let me know.

PR #612, opened 2016-01-09
Allow cxx_library targets without objects to export flags/libraries.

mikekap:
This allows using cxx_library targets to link to local libraries via e.g. exported_linker_flags=['-lpthreads'].


Coneko:
@facebook-github-bot shipit

PR #614, opened 2016-01-12
More pex fixes for windows.

mikekap:
Aside from the two issues @sdwilsh found, the last bit was convincing pex to not remove sys.prefix from sys.path.

### DIFF
--- a/third-party/py/pex/pex/common.py
+++ b/third-party/py/pex/pex/common.py
@@ -269,17 +269,8 @@ class Chroot(object):
     self._ensure_parent(dst)
     abs_src = src
     abs_dst = os.path.join(self.chroot, dst)
-    try:
-      os.link(abs_src, abs_dst)
-    except OSError as e:
-      if e.errno == errno.EEXIST:
-        # File already exists, skip XXX -- ensure target and dest are same?
-        pass
-      elif e.errno == errno.EXDEV:
-        # Hard link across devices, fall back on copying
-        shutil.copyfile(abs_src, abs_dst)
-      else:
-        raise
+    safe_copy(abs_src, abs_dst, overwrite=False)
+    # If the file already exists, skip XXX -- ensure target and dest are same?
 
   def write(self, data, dst, label=None, mode='wb'):
     """Write data to ``chroot/dst`` with optional label.
--- a/third-party/py/pex/pex/compiler.py
+++ b/third-party/py/pex/pex/compiler.py
@@ -3,6 +3,8 @@
 
 from __future__ import absolute_import
 
+import contextlib
+import os
 import subprocess
 import tempfile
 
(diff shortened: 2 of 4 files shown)


third-party/py/pex/pex/common.py
```diff
@@ -269,17 +269,8 @@ def link(self, src, dst, label=None):
...
-      if e.errno == errno.EEXIST:
-        # File already exists, skip XXX -- ensure target and dest are same?
-        pass
-      elif e.errno == errno.EXDEV:
-        # Hard link across devices, fall back on copying
-        shutil.copyfile(abs_src, abs_dst)
-      else:
-        raise
+    safe_copy(abs_src, abs_dst, overwrite=False)
+    # If the file already exists, skip XXX -- ensure target and dest are same?
```
sdwilsh: We can actually avoid this path entirely if we just modify the code in `src/com/facebook/buck/python/make_pex.py` to prefer to copy on Windows.  I had originally wrote something like this, but @andrewjcg suggested changing Buck's code so it was fewer changes to PEX itself.

third-party/py/pex/pex/compiler.py
```diff
@@ -58,6 +60,17 @@ def main(root, relpaths):
...
 
+@contextlib.contextmanager
+def named_temporary_file():
+    fp, path = tempfile.mkstemp()
+    os.close(fp)
+    try:
+        with open(path, 'w') as fp:
+            yield fp
+    finally:
+        os.remove(path)
```
sdwilsh: I'll have to try this when I get into the office, but when I was using something like, this, I was getting an exception with the call to `remove` because the `subprocess` call hadn't finished closing it yet.

third-party/py/pex/pex/pex.py
```diff
@@ -104,6 +104,9 @@ def _site_libs(cls):
       site_libs = set()
     site_libs.update([sysconfig.get_python_lib(plat_specific=False),
                       sysconfig.get_python_lib(plat_specific=True)])
+    # On windows getsitepackages() returns the python stdlib too.
+    if sys.prefix in site_libs:
+      site_libs.remove(sys.prefix)
```
sdwilsh: Good find.  I wasn't familiar with this issues, and I'm curious how you came across it; care to elaborate here (just on Github; the code is fine as-is).

sdwilsh:
Just want you to address the first comment.  Thanks!

third-party/py/pex/pex/common.py
mikekap: I'm actually going to try to upstream these in https://github.com/pantsbuild/pex/pull/198

third-party/py/pex/pex/compiler.py
mikekap: I hit that too, but it wasn't the subprocess call - it was the fp returned by mkstemp

third-party/py/pex/pex/pex.py
mikekap: I think the crash you saw was an extreme example of the bug - the windows box I used to test doesn't have any packages installed, so the actual thing I saw was an `ImportError` trying to import `json`. To trace it down further I set PEX_VERBOSE=5 and saw that the stdlib was being "scrubbed" from the path.

third-party/py/pex/pex/common.py
sdwilsh: You rock!

sdwilsh:
@facebook-github-bot shipit

PR #615, opened 2016-01-13
Extract commonality from (Java|Groovy)LibraryDescription

grumpyjames:
Summary:

These two classes are in fact more similar than their implementations
suggest. In this change we extract the common features to both Groovy
and Java into JvmLibraryDescription, and reduce the previous classes
to smaller configuration classes. This removes a lot of duplication
while also adding flavours and maven support to groovy modules
(documentation to follow).

Test-plan:

CI


grumpyjames:
I'm only semi-serious about this being a PR; there's quite a lot wrong with it that I'd probably want to address before expecting it to ship:

- the change feels big; it could perhaps be split into extracting `JvmLibraryDescription` and then porting `GroovyLibraryDescription` into it. This feels a little artificial though, as it's the existence of `GroovyLibraryDescription` that really drives the change.

- it doesn't feel that well tested on the groovy side; but then I'm not sure the features we gain from this change are that well tested for `JavaLibraryDescription` either

- some of the names are terrible (confession: I have an irrational hatred for concrete inheritance and avoid it at all costs, aspects of this change might be improved by using it, but I couldn't bring myself to do it yet)

- it might be too soon to do this change ; it's not clear how Kotlin/Scala library descriptions will want to vary, so this might be the wrong abstraction for them. That said, it shouldn't be too difficult to make more aspects of `JvmLibraryDescription` configurable (one only needs to move the behaviours to `JvmLibraryConfigurable`)

There's probably other stuff wrong with it that I haven't thought of, too.

Let me know what you think and we can discuss how we want to move forward!

Coneko:
Scala is over here: #589, so you can already have a look.

grumpyjames:
Well, that's jolly exciting; not least because scala_test related changes mirror what I have locally to get groovy_test to work :-). 

It looks like this duplicates as much as groovy library does, but has a few differences around the `deps` related keys contained in the `Arg`. It still has to call the giant `DefaultJavaLibrary` constructor though, and that's what I want to have happen in only a single place.

I'll wait for that to land and then rejig this.

sdwilsh:
That should be out later today :)

mikekap:
Thanks for the work James - it was very easy to add the Scala rules. Some thoughts about what other stuff needs to happen for scala:

Scala is definitely weirder compared to Java and I haven't found a good way to let the rules express that. Particularly, Scala has a concept of macros, which is similar to Java's annotation processors, except they don't look special when you're using them. Scala macro usage look like a regular function call. This makes it very hard to actually know when you're using a macro - it's considered an implementation detail of the function you're calling.

To make that work for now, I've basically forced deps to propagate up scala_library targets (i.e. all `deps` are `exported_deps`). However I can't tell if that works completely correctly in the presence of abi jars. I couldn't quite grasp when the abi jars are used. It also definitely doesn't work correctly with prebuilt_jars unless you do s/deps/exported_deps/ in your BUCK file.

Separately in order to make Scala development sane on buck (it currently takes ~10s to compile a 10 line file), using a cached compiler is pretty much a requirement. Unfortunately that means compiling `java_binary`/`scala_test` more like python - transitively gathering the sources & throwing them at the compiler in one go. It's a pretty terrible reality of developing Scala :(.

grumpyjames:
Thanks for the writeup; glad the work we've been doing was a help :-)

I'm hoping we can share even more code between java/groovy/scala. I worry that will preclude the scala side from being able to grow into something that approaches the usefulness of sbt/zinc, though. My temptation is to tie attempt to tie the three together for the time being despite that. It would make the job of Kotlin folks (who I don't think suffer the same compiler woes as scala folk, afaik) even easier, and I suspect that if we wanted to add friendlier scala support to buck, it would probably end up looking completely different to `DefaultJavaLibrary` anyway.

I'll see how things look when your change lands in master :-)

PR #617, opened 2016-01-15
Add groovy test description

grumpyjames:
Summary:

Add support for writing tests in groovy. Currently still only
via the junit runner, but that's fine, as almost everything supports
that anyway.

Test-plan: CI (GroovyTestIntegrationTest is added)


grumpyjames:
Further commits to dedupe between (Java/Groovy/Scala)TestDescription and add documentation to follow

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
```diff
@@ -0,0 +1,193 @@
+/*
+ * Copyright 2014-present Facebook, Inc.
```
sdwilsh: It is 2016 now!

test/com/facebook/buck/jvm/groovy/GroovyTestIntegrationTest.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright 2013-present Facebook, Inc.
```
sdwilsh: 2016!

sdwilsh:
Just some drive-bys there.  I'm going to let someone else actually review this.

grumpyjames:
Hey, it's still January, it takes me a while to get these things straight :P

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
```diff
@@ -0,0 +1,193 @@
...
+      sourceUnderTest.add(rule);
+    }
+    return sourceUnderTest.build();
+  }
+
+  @SuppressFieldNotInitialized
+  public static class Arg extends GroovyLibraryDescription.Arg {
+    public Optional<ImmutableSortedSet<String>> contacts;
+    public Optional<ImmutableSortedSet<Label>> labels;
+    @Hint(isDep = false) public Optional<ImmutableSortedSet<BuildTarget>> sourceUnderTest;
```
Coneko: Remove `sourceUnderTest`. I assume you mirrored `java_test`'s interface, but this particular part of it causes issues in practice so we don't want to add it to new types of test targets.

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
```diff
@@ -0,0 +1,193 @@
...
+        JavacOptionsFactory.create(
+            defaultJavacOptions,
+            params,
+            resolver,
+            pathResolver,
+            args
+        ));
+
+    JavaTest test =
+        resolver.addToIndex(
```
Coneko: You don't need to add `test` to the resolver since you're returning it from `createBuildRule`, it will get added by https://github.com/facebook/buck/blob/master/src/com/facebook/buck/rules/TargetGraphToActionGraph.java#L99.

Coneko: Just noticed `JavaTestDescription` does the same thing, maybe there's a reason why it's like that, maybe there isn't, try to not add it and if you see any issues that might be it.

Coneko:
I know nothing about groovy but the code looks fine.

Just two comments.

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
grumpyjames: I'm translating this as: 

- remove the part of the `Arg` that refers to it.
- destroy `validateAndGetSourcesUnderTest`
- pass an empty set to `JavaTest`'s constructor arg.

Is there a plan to remove it from the java side as well?

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
grumpyjames: I get 

`com.facebook.buck.util.HumanReadableException: Rule for target '//com/example/spock:passing' could not be resolved.`

The stack trace holds the answer:

<pre>
com.facebook.buck.util.HumanReadableException: Rule for target '//com/example/spock:passing' could not be resolved.

	at com.facebook.buck.rules.BuildRuleResolver.getRule(BuildRuleResolver.java:93)
	at com.facebook.buck.rules.SourcePathResolver$5.apply(SourcePathResolver.java:294)
	at com.facebook.buck.rules.SourcePathResolver$5.apply(SourcePathResolver.java:291)
	at com.google.common.collect.Iterators$8.transform(Iterators.java:799)
	at com.google.common.collect.TransformedIterator.next(TransformedIterator.java:48)
	at com.google.common.collect.ImmutableList.copyOf(ImmutableList.java:271)
	at com.google.common.collect.ImmutableList.copyOf(ImmutableList.java:226)
	at com.google.common.collect.FluentIterable.toList(FluentIterable.java:373)
	at com.facebook.buck.rules.SourcePathResolver.filterBuildRuleInputs(SourcePathResolver.java:297)
	at com.facebook.buck.rules.SourcePathResolver.filterBuildRuleInputs(SourcePathResolver.java:301)
	at com.facebook.buck.jvm.java.CalculateAbi.of(CalculateAbi.java:65)
	at com.facebook.buck.jvm.groovy.GroovyTestDescription.createBuildRule(GroovyTestDescription.java:139)
	at com.facebook.buck.jvm.groovy.GroovyTestDescription.createBuildRule(GroovyTestDescription.java:49)
</pre>

So we need it to be in the resolver by the time we call `CalculateAbi.of`. I'll see if there's a clever way around this that could perhaps be applied to JavaTestDescription also, but I might save it for a separate commit.

Coneko: Yeah sounds good. I figured there was a reason for that I just didn't know what.

Leave it like this for now.

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
Coneko: Yeah you can see the newer `apple_test` does not support it either: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/AppleTest.java#L202-L207

I'm not sure what the plan for `java_test` is but we definitely don't want to add it to new `*_test` rules.

Instead you should specify `tests` in `*_library` and `*_binary`.

Coneko:
@facebook-github-bot shipit

src/com/facebook/buck/jvm/groovy/GroovyTestDescription.java
grumpyjames: Looked at this further - there is a better way. We should port every caller of CalculateAbi.of over though (and there are 13 of them), so definitely a separate commit.

We can always avoid the resolver, because:

1. `SourcePathResolver::filterBuildRuleInputs(SourcePath... sourcePaths) is actually only ever called with a single `sourcePath` (still compiles when I remove the ellipsis)
2. All we're doing is seeing if that `sourcePath` is a `BuildTargetSourcePath`, and if it is, returning the build rule that builds it. We already know what the rule is from the context we are in, so we can probably either avoid `filterBuildRuleInputs` altogether or appropriately 'can' the BuildRule to return

tl;dr the only thing that resolver will be returning will be `[the rule we're currently building]`, or maybe `[]` in the `prebuilt_jar` case.

PR #622, opened 2016-01-19
Add ability to mark a prebuilt_jar as "provided" - i.e. it's always a provided_dep

mikekap:
This can be useful for e.g. hadoop-client jars which should never be packaged into a fat jar. This also moves getDeclaredClasspathEntries out of the JavaLibrary interface since it's only used in DefaultJavaLibrary.

### DIFF
--- a/src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java
+++ b/src/com/facebook/buck/android/AndroidPrebuiltAarDescription.java
@@ -187,7 +187,8 @@ public class AndroidPrebuiltAarDescription
         /* sourceJar */ Optional.<SourcePath>absent(),
         /* gwtJar */ Optional.<SourcePath>absent(),
         /* javadocUrl */ Optional.<String>absent(),
-        /* mavenCoords */ Optional.<String>absent());
+        /* mavenCoords */ Optional.<String>absent(),
+        /* provided */ false);
 
   }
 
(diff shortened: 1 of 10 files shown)


sdwilsh:
Why would we need this when we have `provided_deps` on `java_library`?

mikekap:
I think in some cases the decision of whether this library is provided should be in the library itself. We have several (~7) libraries that depend on hadoop-client in our repo - we could manually go through and set it as provided_deps on each one, but it seemed more logical to put that decision in the library itself, since nobody should ever make a fat jar with hadoop-client in it (that's just not how hadoop works).

This is sort of mirroring the sbt "provided" scope; my goal is to make the maven resolver use this to support "provided" deps.

dreiss:
The way we've done this in Facebook for Android is to define the real jar as rule like "hadoop-client-private" that has restricted visibility, then a wrapper java_library ("hadoop-client") that lists hadoop-client-private as a provided_dep and exported_dep.

I agree that this is a bit cumbersome and it would be convenient to just specify it on the jar.

mikekap:
Hmm are you sure that works? From reading the code, I don't think provided deps don't propagate, while exported deps do. Assuming `lib1 -> hadoop-client -> hadoop-client-private`, buck will rewrite the graph to have `lib1 -> hadoop-client-private` directly, but wouldn't put it in as a provided dep, so it would end up in the jar.

dreiss:
Hrm.  I'm sure it works for building Android apps, but it's possible that java_binary works differently.

sdwilsh:
@shs96c should probably make a call on this

Coneko:
We've been discussing something similar to support system libraries in `prebuilt_cxx_library`, so I think it makes sense.

cc @andrewjcg

mikekap:
Ping

Coneko:
@facebook-github-bot shipit

PR #624, opened 2016-01-20
Add the ability to mark a genrule output as executable.

mikekap:
Similar to bazel, genrule(..., executable=True) will let you `buck run` it & uses `$(exe ...)` in macros.

### DIFF
--- a/docs/rule/genrule.soy
+++ b/docs/rule/genrule.soy
@@ -37,6 +37,16 @@
 
 {call genrule_common.out_arg /}
 
+{call buck.arg}
+  {param name: 'executable' /}
+  {param default : 'False' /}
+  {param desc}
+  Whether the output of the genrule is itself executable. Marking an output as
+  executable makes <code>buck run</code> and <code>$(exe ...)</code> macro
+  expansion work with this target.
+  {/param}
+{/call}
+
 {call buck.visibility_arg /}
 
 {/param} // args
--- a/src/com/facebook/buck/shell/BUCK
+++ b/src/com/facebook/buck/shell/BUCK
@@ -5,6 +5,7 @@ java_library(
     'ExportFile.java',
     'ExportFileDescription.java',
     'Genrule.java',
+    'GenruleBinary.java',
     'GenruleDescription.java',
     'ShBinary.java',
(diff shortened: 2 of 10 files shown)


andrewjcg:
Does wrapping the output in a `sh_binary` work?  That said, something like this does sound like a nice shorthand.

mikekap:
It definitely does, but if you also want to use said output as a resource (e.g. to execute at runtime), it becomes a little annoying (i.e. you have to have the target that you include as a resource & the target that you run be two separate X & X_run or something like that).

Coneko:
You can't have a sh_binary as a resource?

mikekap:
Hmm I'll try tomorrow, but from a glance, it looks like it would include the bootstrap .sh script as a resource, but not the actual main script.

mikekap:
Confirmed; here's what I tried:

```python
genrule(
    name='bin-rule',
    cmd='echo "echo hi" > $OUT && chmod +x $OUT',
    out='bin.sh',
)

sh_binary(
    name='bin',
    main=':bin-rule',
)

python_library(
    name='res',
    resources=[':bin'],
)

python_binary(
    name='test',
    main_module='json.tool',
    deps=[':res'],
)
```

```
$ buck build test && unzip buck-out/gen/test.pex
...
  inflating: PEX-INFO
  inflating: __main__.py
  inflating: __main__.pyc
  inflating: bin
$ cat bin
...
BUCK_PROJECT_ROOT=$BUCK_TMP_ROOT "$BUCK_TMP_ROOT/$SCRIPT_TO_RUN" "$@"
```

So the actual script wasn't included in the pex.

mikekap:
Ping :)

sdwilsh:
I thought @andrewjcg or @Coneko were looking at this.  I may have been wrong?

sdwilsh:
I am otherwise happy with this, but I'd like to see you add documentation for the feature.

Coneko:
Sorry I was supposed to, just a bit swamped.

mikekap:
Now with docs

sdwilsh:
@facebook-github-bot shipit

PR #627, opened 2016-01-21
Remove unnecessary 'workspace.verify()' calls from two test cases.

grumpyjames:
Summary:

These tests don't have any .expected files in their test data area;
there's no need to verify anything.

Test-plan:

CI


Coneko:
@facebook-github-bot shipit

PR #632, opened 2016-01-27
Remove redundant mkdir step from DefaultJavaLibrary

grumpyjames:
Summary: Presumably the gen path for a build target already exists by
the time we execute the steps. At least, not explicitly creating it in
DefaultJavaLibrary appears to have no ill effects.

Test-plan: CI

### DIFF
--- a/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java
+++ b/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java
@@ -43,7 +43,6 @@ import com.facebook.buck.rules.SourcePaths;
 import com.facebook.buck.rules.keys.SupportsInputBasedRuleKey;
 import com.facebook.buck.step.Step;
 import com.facebook.buck.step.fs.MakeCleanDirectoryStep;
-import com.facebook.buck.step.fs.MkdirStep;
 import com.facebook.buck.step.fs.TouchStep;
 import com.facebook.buck.util.HumanReadableException;
 import com.google.common.base.Function;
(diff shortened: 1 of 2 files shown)


grumpyjames:
I'm not actually convinced that it is redundant (this pattern turns up a lot elsewhere); I'd really like to understand the scenario in which it becomes necessary, though; advice welcome

grumpyjames:
Weak regex fu suggests there are ~95 other places that do roughly the same thing. If it isn't redundant, it perhaps ought to be made so...

grumpyjames:
Ok, took a closer look.

In fact, every `MkdirStep` does a mkdir -p, so anyone creating a folder in the gen area is going to end up creating it for us. That sort of leaves things in a working by coincidence state, so this probably shouldn't be merged.

We could remove lots of tedious step code by always having buck-out/gen (and, potentially, `bin`, `res`, and `annotation`) exist before build rules execute.

Thoughts?

k21:
We were thinking about standardizing the rules' output directories which would let us move all this set up / clean up code into a single common place and we could stop doing it ad-hoc in each rule (cc @Coneko). It would help make code like this more succinct and less error prone. Unfortunately we did not get to it yet. Unless there is a good reason to remove the step right now, I would prefer keeping it for now and cleaning it up together with everything else once we unify the logic in a single place.

Coneko:
> We could remove lots of tedious step code by always having buck-out/(gen|bin|res|annotation)/${buildTarget}/ already exist

As @k21 mentioned, I have a series of changes to do that. It's a bit harder than expected, and going towards it actually highlights the issue opposite to the one you noticed: while it's true we create more mkdir steps than strictly necessary, we also don't create some we should, because we assume the directories already exist.

I'll try to pick those changes back up.

I'll close this in the meantime, I agree not having to create these directories would be best, but we cannot remove the uses where it works by coincidence without setting up the infrastructure beforehand.

PR #633, opened 2016-01-27
Always provide a pathToSrcsList in jvm related rules

grumpyjames:
Summary: Previously a few rules had the option to avoid doing this. It
forces the downstream implementations to branch unnecessarily, so
port the offending usages over to providing the path and then stop it 
being Optional

Test-plan: CI


sdwilsh:
@facebook-github-bot shipit

PR #637, opened 2016-01-29
Use pathToSrcsList in GroovycStep (fixes #629)

### LINKED ISSUE #629: groovy_library use argfile for list of sources.
To avoid issues with `ARG_MAX` limits when running the `groovy_library` build rule, the `groovyc` step should work like that `javac` step by writing a list of sources to file then executing the compiler on those sources.

The `javac` build step writes a files to `buck-out/gen/x/__y__srcs` for a build target of `//x:y`

grumpyjames:
Summary: Previously we ignored the provided path, but it turns out
groovyc does support it, so let's be the same as everyone else.

Test-plan: CI

### DIFF
--- a/src/com/facebook/buck/jvm/groovy/GroovycStep.java
+++ b/src/com/facebook/buck/jvm/groovy/GroovycStep.java
@@ -33,7 +33,6 @@ import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


src/com/facebook/buck/jvm/groovy/GroovycStep.java
```diff
@@ -111,11 +118,32 @@ public String getDescription(ExecutionContext context) {
...
+            .transform(toStringFunction())
+            .transform(Javac.ARGFILES_ESCAPER),
+        pathToSrcsList);
+  }
+
+  private ImmutableList<Path> getExpandedSourcePaths(ImmutableSet<Path> javaSourceFilePaths) {
+    ImmutableList.Builder<Path> sources = ImmutableList.builder();
+    for (Path path : javaSourceFilePaths) {
+      String pathString = path.toString();
+      if (pathString.endsWith(".java") || pathString.endsWith(".groovy")) {
```
Coneko: We are the sources filtered at this stage? They should be sorted much earlier so we don't end up rebuilding stuff when sources that don't contribute change.

Why would it even be possible to specify sources that don't get compiled?

grumpyjames: This was inspired by `ExternalJavac` which needs the filtering to cope with the potential of expanding zipped sources. We totally don't need it here, I'll remove it.

sdwilsh:
@facebook-github-bot shipit

Issue #302, opened 2015-05-11
buck test --filter 'foo' executes 'bar'


davido:
On Buck master 8204fddf60b25a3c2090f3ef0742fca5d466d562 running the following command on Gerrit master (https://github.com/gerrit-review/gerrit):

```
$ buck test -f 'com.google.gerrit.acceptance.rest.change.SubmitByMergeAlwaysIT'
Using watchman.
Using buckd.
[-] PROCESSING BUCK FILES...FINISHED 2,2s
[-] BUILDING...FINISHED 43,8s (565/565 JOBS, 565 UPDATED, 72,9% CACHE HITS)
Log:
TESTING SELECTED TESTS
PASS    <100ms  4 Passed   0 Skipped   0 Failed   //tools:util_test
ASSUME   12,4s 13 Passed   1 Skipped   0 Failed   com.google.gerrit.acceptance.rest.change.SubmitByMergeAlwaysIT
TESTS PASSED (with some assumption violations)
```

I would expect that `//tools:util_test` isn't executed.

sdwilsh:
What kind of a test rule is that?

davido:
The test rule that is awlays executed is the only python test rules we have:


```
  python_test(
    name = 'util_test',
    srcs = ['util_test.py'],
    deps = [':util'],
    visibility = ['PUBLIC'],
  )
```

All other tests are Java test rules.

sdwilsh:
Yeah, the only tests that actually do any filtering is `java_test`.

mread:
Has behaviour changed here in last month or so? We are also seeing all sorts of other rules match when we specify a filter, e.g.:

```
buck test --dry-run --filter com.package.tests.SomeTest
```
Output:

```
MATCHING TEST RULES (715):
...
```

At first glance it looks like every sh_binary rule in our project, possibly some others too. Shouldn't this only apply to java_test rules?

In fact, if I type this:
```
./bebuck targets --type java_test
```
I get sh_binary rules matching too. That didn't used to happen did it? I might raise a separate issue for that.

sdwilsh:
This hasn't changed.  We've always only filtered for Java since filtering was implemented.

mread:
My mistake, we define a load of rules as java _test that having nothing to do with java tests so that we can run them using labels. Sorry for hijacking this issue.

sdwilsh:
Feel free to file issues (and/or pull requests) on getting label support on the things you need.

PR #646, opened 2016-02-05
Implement test selectors for python.

mikekap:
Same idea as Xctool-based tests. This runs the test binary once to get the
list of tests, filters them down and runs the selected tests with -r.

This approach is somewhat suboptimal in python since getting the list of tests
actually runs code in python (e.g. if you have module-level setup, it will
get run). However, since the double exec only happens with --test-selectors,
I think this is ok. Alternatively, it's pretty easy to re-implement
TestSelector(List) in python.

Addresses part of https://github.com/facebook/buck/issues/302

### DIFF
--- a/src/com/facebook/buck/python/PythonRunTestsStep.java
+++ b/src/com/facebook/buck/python/PythonRunTestsStep.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2016-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.python;
+
+import com.facebook.buck.shell.ShellStep;
+import com.facebook.buck.step.ExecutionContext;
+import com.facebook.buck.step.Step;
+import com.facebook.buck.test.selectors.TestDescription;
+import com.facebook.buck.test.selectors.TestSelectorList;
+import com.facebook.buck.util.HumanReadableException;
+import com.google.common.base.CharMatcher;
+import com.google.common.base.Function;
+import com.google.common.base.Joiner;
+import com.google.common.base.Optional;
+import com.google.common.base.Supplier;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.io.IOException;
+import java.nio.file.Path;
+
+public class PythonRunTestsStep implements Step {
+  private static final CharMatcher PYTHON_RE_REGULAR_CHARACTERS = CharMatcher.anyOf(
+      "_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890");
+
+  private final Path workingDirectory;
+  private final String testName;
+  private final ImmutableList<String> commandPrefix;
+  private final Supplier<ImmutableMap<String, String>> environment;
+  private final TestSelectorList testSelectorList;
+  private final Optional<Long> testRuleTimeoutMs;
+  private final Path resultsOutputPath;
+
+  private boolean timedOut;
+
+  public PythonRunTestsStep(
+      Path workingDirectory,
+      String testName,
+      ImmutableList<String> commandPrefix,
+      Supplier<ImmutableMap<String, String>> environment,
+      TestSelectorList testSelectorList,
+      Optional<Long> testRuleTimeoutMs,
+      Path resultsOutputPath) {
+    this.workingDirectory = workingDirectory;
+    this.testName = testName;
+    this.commandPrefix = commandPrefix;
+    this.environment = environment;
+    this.testSelectorList = testSelectorList;
+    this.testRuleTimeoutMs = testRuleTimeoutMs;
+    this.resultsOutputPath = resultsOutputPath;
+
+    this.timedOut = false;
+  }
+
+  @Override
+  public int execute(ExecutionContext context) throws IOException, InterruptedException {
+    int exitCode = doExecute(context);
+    if (timedOut) {
+      throw new HumanReadableException(
+          "Following test case timed out: " + testName + ", with exitCode: " + exitCode);
+    }
+    return exitCode;
+  }
+
+  private int doExecute(ExecutionContext context) throws IOException, InterruptedException {
+    if (testSelectorList.isEmpty()) {
+      return getShellStepWithArgs("-o", resultsOutputPath.toString()).execute(context);
+    }
+
+    ShellStep listStep = getShellStepWithArgs("-l", "-L", "buck");
+    int exitCode = listStep.execute(context);
+    if (timedOut || exitCode != 0) {
+      return exitCode;
+    }
+
+    String testsToRunRegex = getTestsToRunRegexFromListOutput(listStep.getStdout());
+
+    return getShellStepWithArgs(
+        "-o", resultsOutputPath.toString(), "-r", testsToRunRegex).execute(context);
+  }
+
+  private String getTestsToRunRegexFromListOutput(String listOutput) {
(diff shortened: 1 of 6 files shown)


sdwilsh:
@bhamiltoncx would probably be a good reviewer on this.

src/com/facebook/buck/python/PythonRunTestsStep.java
```diff
@@ -0,0 +1,160 @@
...
+      return getShellStepWithArgs("-o", resultsOutputPath.toString()).execute(context);
+    }
+
+    ShellStep listStep = getShellStepWithArgs("-l");
+    int exitCode = listStep.execute(context);
+    if (timedOut || exitCode != 0) {
+      return exitCode;
+    }
+
+    ImmutableList.Builder<String> testsToRunRe = ImmutableList.builder();
```
bhamiltoncx: Maybe `testsToRunPatternComponents`?

src/com/facebook/buck/python/PythonTest.java
```diff
@@ -162,7 +103,15 @@ public Void apply(Process process) {
       TestRule.TestReportingCallback testReportingCallback) {
     return ImmutableList.of(
         new MakeCleanDirectoryStep(getProjectFilesystem(), getPathToTestOutputDirectory()),
-        getRunTestStep());
+        new PythonRunTestsStep(
+            // TODO(shs96c): I'm not convinced this is the right root path
```
bhamiltoncx: I think it's what you want.

src/com/facebook/buck/python/PythonRunTestsStep.java
```diff
@@ -0,0 +1,160 @@
...
+
+    String output = listStep.getStdout();
+    for (String strTestCase : output.split("\n")) {
+      String[] testCase = strTestCase.split("#", 2);
+      if (testCase.length != 2) {
```
bhamiltoncx: I think it's valid to have test selectors `#method_name` and `class_name#`. Please add support for these and test them as well.

src/com/facebook/buck/python/PythonRunTestsStep.java
```diff
@@ -0,0 +1,160 @@
...
+    ShellStep listStep = getShellStepWithArgs("-l");
+    int exitCode = listStep.execute(context);
+    if (timedOut || exitCode != 0) {
+      return exitCode;
+    }
+
+    ImmutableList.Builder<String> testsToRunRe = ImmutableList.builder();
+
+    String output = listStep.getStdout();
+    for (String strTestCase : output.split("\n")) {
```
bhamiltoncx: Hmm. Why doesn't this fail on the last line of the output? Shouldn't `strTestCase` be empty on the last line?

src/com/facebook/buck/python/PythonRunTestsStep.java
```diff
@@ -0,0 +1,160 @@
...
+      TestDescription testDescription = new TestDescription(testCase[0], testCase[1]);
+      if (testSelectorList.isIncluded(testDescription)) {
+        testsToRunRe.add(Pattern.quote(testDescription.toString()));
+      }
+    }
+
+    String testsToRunRegex = "^" + Joiner.on('|').join(testsToRunRe.build()) + "$";
+
+    return getShellStepWithArgs(
+        "-o", resultsOutputPath.toString(), "-r", testsToRunRegex).execute(context);
```
bhamiltoncx: This argument could get really long really fast (especially with dynamically-generated tests).

Right now, `__test_main__.py` is using `optparse`. If you convert it to `argparse` (which should basically be a drop-in replacement), you can set `fromfile_prefix_chars='@'` and pass in the arguments via a temp file on disk.

You don't have to do that in this diff, but if you don't, please file a follow-up issue here and mention it in a comment.

test/com/facebook/buck/python/PythonTestIntegrationTest.java
```diff
@@ -73,6 +73,17 @@ public void testPythonTest() throws IOException {
   }
 
   @Test
+  public void testPythonTestSelectors() throws IOException {
+    ProcessResult result = workspace.runBuckCommand(
+        "test", "--test-selectors", "Test#test_that_passes", "//:test-failure");
```
bhamiltoncx: A test to filter a test class in a module would be nice.

bhamiltoncx:
Awesome stuff, great job! I left you some comments.

src/com/facebook/buck/python/PythonRunTestsStep.java
sdwilsh: Note that `argparse` is python2.7+, and Buck supports python2.6+.  If we did this, we'd have to import the package for 2.6.

bhamiltoncx: True. Anyway, we can leave that for another time.

src/com/facebook/buck/python/PythonRunTestsStep.java
mikekap: I have absolutely no idea why this worked. Added a trim.

src/com/facebook/buck/python/PythonRunTestsStep.java
mikekap: I'll leave it until next time & open an issue once this goes in.

test/com/facebook/buck/python/PythonTestIntegrationTest.java
mikekap: Great point - turns out that didn't work at all.

sdwilsh:
@facebook-github-bot shipit

sdwilsh:
Hmm, getting this error in CI (I'll look into it more later):
```
Traceback (most recent call last):
  File ".bootstrap/_pex/pex.py", line 320, in execute
  File "/usr/lib64/python2.6/contextlib.py", line 34, in __exit__
    self.gen.throw(type, value, traceback)
  File ".bootstrap/_pex/pex.py", line 223, in patch_pkg_resources
  File ".bootstrap/_pex/pex.py", line 320, in execute
  File ".bootstrap/_pex/pex.py", line 253, in _wrap_coverage
  File ".bootstrap/_pex/pex.py", line 285, in _wrap_profiling
  File ".bootstrap/_pex/pex.py", line 363, in _execute
  File ".bootstrap/_pex/pex.py", line 421, in execute_entry
  File ".bootstrap/_pex/pex.py", line 426, in execute_module
  File "/usr/lib64/python2.6/runpy.py", line 140, in run_module
    fname, loader, pkg_name)
  File "/usr/lib64/python2.6/runpy.py", line 34, in _run_code
    exec code in run_globals
  File "__test_main__.py", line 374, in <module>
  File "__test_main__.py", line 370, in main
  File "__test_main__.py", line 345, in run
  File "__test_main__.py", line 331, in get_test_names
  File "__test_main__.py", line 331, in get_test_names
  File "__test_main__.py", line 336, in get_test_names
ValueError: zero length field name in format
```

Without looking into the actual failure, I suspect it's a py2.6 issue given the machines it failed on.

mikekap:
Ah well that explains why there were indices everywhere else :X. Seems implicit positional .format() is 2.7-only. Fixed (& tested on 2.6).

sdwilsh:
Yeah, I couldn't get travis to use 2.6 for our CI unfortunately so it only gets flagged when I import it.

Here's hoping that the import won't have conflicts like it did last time...

sdwilsh:
@facebook-github-bot import

sdwilsh:
Of course there was a merge conflict.  I'll try to get it in today, but it might slip into tomorrow.

sdwilsh:
Ha.  I'm already working through the conflict (I'm actually not sure why it's conflicting at all to be honest).

mikekap:
I didn't actually touch any of my PRs today. You may want to check if your bot is becoming...proactive :D

sdwilsh:
Yeah, it spammed a lot of PRs.  This might actually land tonight, finally.

k21:
Unfortunately I had to revert this PR because it breaks some of our internal infrastructure that uses an external test runner.

mikekap:
@k21 is it related to the change in list output? That's really the only change that affects the external test runner. I could easily put that behind a different flag to keep the test runner working if it helps speed things along.

k21:
Yes, the issue was that the format in which tests were passed to the external test runner has changed. If the pull request either avoids changing the format or puts it behind a buckconfig option, I can reimport it without changing anything on our side, which would probably be easier and faster.

mikekap:
Updated the format of `--list-tests` to the original. The other change is fixing a bug in `-r` that caused test matching fail for class names due to an extra `__class__` (all the test cases had the class `__builtin__.type` when matching).

sdwilsh:
@facebook-github-bot import

k21:
@facebook-github-bot import

Issue #638, opened 2016-01-29
Buck Cli Not Working


tcyrus:
Installed buck using brew, when I try to run it I get this:

```
$ buck --help
Not using buckd because watchman isn't installed.
[2016-01-29 11:47:35.979][error][command:f3397a50-7979-4a62-8eea-8dd718719bda][tid:01][com.facebook.buck.cli.Main] Uncaught exception at top level
java.lang.IllegalAccessError: tried to access field com.sun.jna.Pointer.peer from class com.sun.jna.PointerTool
    at com.sun.jna.PointerTool.getPeer(PointerTool.java:13)
    at ca.weblite.objc.mappers.NSObjectMapping.cToJ(NSObjectMapping.java:38)
    at ca.weblite.objc.TypeMapper.cToJ(TypeMapper.java:122)
    at ca.weblite.objc.RuntimeUtils.msg(RuntimeUtils.java:619)
    at ca.weblite.objc.Client.send(Client.java:164)
    at ca.weblite.objc.Client.sendProxy(Client.java:344)
    at ca.weblite.objc.Client.sendProxy(Client.java:382)
    at com.facebook.buck.util.environment.MacWifiSsidFinder.getDefaultWifiInterface(MacWifiSsidFinder.java:77)
    at com.facebook.buck.util.environment.MacWifiSsidFinder.findCurrentSsid(MacWifiSsidFinder.java:61)
    at com.facebook.buck.util.environment.DefaultExecutionEnvironment.getWifiSsid(DefaultExecutionEnvironment.java:84)
    at com.facebook.buck.cli.Main.runMainWithExitCode(Main.java:859)
    at com.facebook.buck.cli.Main.tryRunMainWithExitCode(Main.java:1400)
    at com.facebook.buck.cli.Main.runMainThenExit(Main.java:1485)
    at com.facebook.buck.cli.Main.main(Main.java:1503)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at com.facebook.buck.cli.bootstrapper.ClassLoaderBootstrapper.main(ClassLoaderBootstrapper.java:62)
```

Same issue happens when using watchman

```
$ buck --help
[...]

PR #647, opened 2016-02-05
Add appveyor & windows test blacklist.

mikekap:
This allows tests to run on appveyor (https://ci.appveyor.com/project/mikekap/buck/build/1.0.54). Also generates a buck.pex for windows (although I don't think it varies by platform to start with). There's also a few easy fixes for windows tests in here.

This is currently rebased on the python test selector implementation, since there's a few failing python tests on windows that need to be excluded.


mikekap:
`FAIL //test/com/facebook/buck/parser:parser` - seems to be the same as what's happening in master, so likely unrelated to this change.

sdwilsh:
I had a hunch your python change was for this :)

sdwilsh:
> FAILURE //test/com/facebook/buck/parser:parser main: //test/com/facebook/buck/parser:parser failed with exit code 137

*sigh*.  Once again, Buck is getting killed.  Let me throw up a change so they aren't all in one binary and see if that helps.

DEFS
```diff
@@ -76,6 +76,9 @@ def java_test(
       # http://arihantwin.blogspot.com/2012/08/getting-error-illegal-local-variable.html
       '-XX:-UseSplitVerifier',
 
+      # Don't use the system-installed JNA; extract it from the local jar.
+      '-Djna.nosys=true',
+
```
sdwilsh: That's a little bit terrifying.  I wonder if this would also help fix https://github.com/facebook/buck/issues/638

sdwilsh: also cc @bhamiltoncx

appveyor.yml
```diff
@@ -0,0 +1,19 @@
+install:
+  - cinst -y jdk7
```
sdwilsh: We actually do jdk8 now for Travis, so we should do the same for Windows.

appveyor.yml
```diff
@@ -0,0 +1,19 @@
...
+  - cinst -y ant -ignoreDependencies
+  - set JAVA_HOME=C:\Program Files\Java\jdk1.7.0
+  - set PATH=%JAVA_HOME%\bin;%PATH%
+  - set BUCK_EXTRA_JAVA_ARGS=-Djna.nosys=true
+  - python --version
+  - ant -version
+
+
+build_script:
+  - "bin\\buck build buck"
```
sdwilsh: I guess this misses all the android stuff too, but it's better than nothing.  We can go in and add more things to this as needed.

appveyor.yml
```diff
@@ -0,0 +1,19 @@
...
+
+test_script:
+  - "bin\\buck test --all --test-selectors=@windows_failures.txt"
+
+artifacts:
+  - path: buck-out\gen\programs\buck.pex
```
sdwilsh: I actually made the changes recently to allow Buck to be built on Linux and be usable on Windows (and I suspect OS X).  I didn't test the other way around, however!

src/com/facebook/buck/go/GoBuckConfig.java
```diff
@@ -109,7 +110,11 @@ Path getDefaultPackageName(BuildTarget target) {
       public Tool get() {
         Optional<Path> toolPath = delegate.getPath("go", configName);
         if (!toolPath.isPresent()) {
-          toolPath = Optional.of(goToolDirSupplier.get().resolve(toolName));
+          String exeName = toolName;
+          if (delegate.getPlatform() == Platform.WINDOWS) {
+            exeName = toolName + ".exe";
+          }
```
sdwilsh: Hmm, if we used an `ExectuableFinder`, it would do this for us (and also handle the case where someone may have wrapped it in a `.bat` or some other script first).

windows_failures.txt
```diff
@@ -0,0 +1,180 @@
...
+!com.facebook.buck.util.PropertyFinderTest#testSdkPathFromPropertiesRaisesExceptionWhenDirectoryIsNotPresent
+!com.facebook.buck.util.WatchmanWatcherTest#watchmanQueryRelativizesExcludePaths
+!com.facebook.buck.util.WatchmanWatcherTest#watchmanQueryWithExcludePathsAddsExpressionToQuery
+!com.facebook.buck.util.WatchmanWatcherTest#watchmanQueryWithExcludePathsAddsMatchExpressionToQueryIfDirnameNotAvailable
+!com.facebook.buck.util.WatchmanWatcherTest#whenMultipleFilesThenMultipleEventsGenerated
+!com.facebook.buck.util.WatchmanWatcherTest#whenNameThenModifyEventIsGenerated
+!com.facebook.buck.zip.ZipRuleIntegrationTest#shouldUnpackContentsOfASrcJar
+!com.facebook.buck.zip.ZipRuleIntegrationTest#shouldZipSources
+!src.com.facebook.buck.json.buck_test.TestBuckPosix
+!src.com.facebook.buck.json.buck_test.TestBuckWindows
```
sdwilsh: *sigh*

This one hits home.

src/com/facebook/buck/go/GoBuckConfig.java
mikekap: Hmm AFAICT `ExecutableFinder` might need some changes to respect "in this path", since the binaries here are called something very generic (`compile` and `link`), and should not be looked up in $PATH.

I may just revert this change, since go tests still fail due to lack of C linker. Go technically doesn't need a C linker, but it doesn't seem like there's a way to check if the linker is "real" or not.

appveyor.yml
mikekap: I can try doing this, but this takes ~ 2-3m longer per CI run. appveyor machines have jdk7 pre-installed (this line is technically a noop), but jdk8 takes a while to install.

appveyor.yml
mikekap: Yea, I barely know how to set up the android sdk on linux, much less windows. Perhaps https://github.com/googlesamples/android-ndk/blob/master/appveyor.yml is a good place to start?

appveyor.yml
sdwilsh: That's a bummer, but I'd prefer if we kept to java8 since that's what we are testing elsewhere (and there are differences).

src/com/facebook/buck/go/GoBuckConfig.java
sdwilsh: `ExecutableFinder`'s constructor lets you pass in the env, so if you know where it should be, you can do something like this:
```
toolPath = new ExecutableFinder().getOptionalExectuable(
    Paths.get(toolname),
    goToolDirSupplier.get());
```

This has the added advantage of making sure it's executable on unix systems.  But I also agree that all this can be done in a different PR.

mikekap: I was looking at that, but I *think* the "fast path" (https://github.com/facebook/buck/blob/master/src/com/facebook/buck/io/ExecutableFinder.java#L121) makes it a bit weird. If you create a file called "compile" in the root of your project, it will prefer that even with an explicit directory passed in. Is that intentional?

sdwilsh: *groan*

That has to be a bug.  I can't imagine any reason why we'd want to do that.

mikekap:
This should be ready. There's a chance this will break the external test runner you guys use. I swapped what testName & testCaseName mean for python tests (it was reversed).

Just to make sure this doesn't get committed and immediately reverted, it would be nice to test that (and fix the runner if it depends on the incorrect behavior).

Coneko:
@facebook-github-bot import

Coneko:
Is the python testing change require to make the test selectors work? If not please separate it in another PR.

mikekap:
Done.

Coneko:
@facebook-github-bot shipit

PR #650, opened 2016-02-08
Add `library` to go_test that copies the srcs/package/deps of the library it's testing.

mikekap:
This saves the trouble of duplicating the srcs/deps on the test target if you're testing the same package.

### DIFF
--- a/docs/rule/go_test.soy
+++ b/docs/rule/go_test.soy
@@ -35,14 +35,24 @@
 {call go_common.srcs_arg /}
 
 {call buck.arg}
+  {param name: 'library' /}
+  {param default: 'None' /}
+  {param desc}
+  Specify the library that this internal test is testing. This will copy the <code>srcs</code>,
+  {sp}<code>package_name</code> and <code>deps</code> from the target specified so you don't have
+  to duplicate them.
+  {/param}
+{/call}
+
+{call buck.arg}
   {param name : 'package_name' /}
   {param default : 'go.prefix + path relative to the buck root + "_test"' /}
   {param desc}
   Sets the full name of the test package being compiled. This defaults to the path from the buck
   root with "_test" appended. (e.g. given a ./.buckconfig, a rule in ./a/b/BUCK defaults to package "a/b_test")
 
-  <p>Note: in order to test packages internally (i.e. same package name), you will need to re-declare
-  all the sources/dependencies on this test and set package_name appropriately.</p>
+  <p>Note: if you want to test packages internally (i.e. same package name), use the <code>library</code>
+  {sp}parameter instead of setting <code>package_name</code> to include the tested source files.</p>
   {/param}
 {/call}
 
@@ -96,6 +106,15 @@ go_test(
     ':join',
   ],
 )
+
+# Or
+
+go_test(
+  name='greeting-better-internal-test',
+  srcs=['greeting_test.go'],
+  library=':greeting',
+)
+
 </pre>{/literal}
 {/param}
 
--- a/src/com/facebook/buck/go/GoDescriptors.java
+++ b/src/com/facebook/buck/go/GoDescriptors.java
@@ -48,6 +48,7 @@ import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
+import java.util.Set;
 
 abstract class GoDescriptors {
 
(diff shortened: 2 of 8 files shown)


sdwilsh:
In the past we had `source_under_test` for references like this.  However, we have stopped doing that because it doesn't scale for large code bases.  This feels pretty similar to that.

I'm interested to hear what @Coneko thinks.

mikekap:
This idea isn't quite original: Bazel had it first https://github.com/bazelbuild/bazel/blob/master/tools/build_rules/go/def.bzl#L199 , although Bazel implemented it for all `go_library` targets (rather than just `go_test`), which doesn't quite make sense to me.

sdwilsh:
I would almost prefer to see something like a `$(package )` macro that you could use in the `package` field instead the more I think about this.  That could possible be useful elsewhere too (and not just for Go).  I don't use Go though, so I'm interested in your thoughts and still what @Coneko has to say.

mikekap:
My thoughts generally are to mirror bazel's go support, as described in https://docs.google.com/document/d/1_fNIFC-2gLETPBXWipNDaZEi1NCKDJRBqGN7lF0Eo7k/view . That is, if you're using Go, you probably shouldn't actually be writing BUCK files by hand.

That doesn't mean generated BUCK files should be cluttered, hence this change. It mostly lives to remove duplication between the go_library & go_test rules for the same package.

In terms of the frontend for it - I'm a little skeptical of macros. My mental model of macros is straightforward substitution (e.g. linker flags the linker command line). In this case the value of `package_name` would affect `srcs` & `deps`, which seems a bit strange. The internal test pattern is actually more common in go than the external test pattern (i.e. different package name in the tests), so I think making it not-too-special-looking would be nice.

Having said that, I'm not too attached to any particular approach, especially given you can already achieve this with some duplication. If this sets a bad precedent within Buck that you'd rather avoid, I'm ok with just dropping this change.

nemith:
This just came up internally, there should be a good way to not copy the same deps and srcs for tests in the same package so imho something would be nice but macros or another kwarg (like this diff) would be fine.  

The package name is less of an issue.

Coneko:
I think this is fine.

The big problem we have with `source_under_test` is that `buck test library` supports finding the tests that have `library` in `source_under_test`, which is really really slow, this doesn't support that so it's ok.

Alternatively if you want to make sure there is no confusion, add a check for `library.test` containing all the tests that specify `library` in `library`.

docs/rule/go_test.soy
```diff
@@ -35,14 +35,24 @@
...
   {param name : 'package_name' /}
   {param default : 'go.prefix + path relative to the buck root + "_test"' /}
   {param desc}
   Sets the full name of the test package being compiled. This defaults to the path from the buck
   root with "_test" appended. (e.g. given a ./.buckconfig, a rule in ./a/b/BUCK defaults to package "a/b_test")
 
-  <p>Note: in order to test packages internally (i.e. same package name), you will need to re-declare
-  all the sources/dependencies on this test and set package_name appropriately.</p>
+  <p>Note: if you want to test packages internally (i.e. same package name), use the <code>library</code>
+  {sp} parameter instead of setting <code>package_name</code> to include the tested source files.</p>
```
sdwilsh: no need for the space after `{sp}`

sdwilsh:
Back to you for the suggestion that @Coneko offered (since I think that's actually a good idea).  This looks good otherwise.

mikekap:
@Coneko Do you mean something like:
```py
if test_target.library:
    assert test_target in library_target.tests
```
or
```py
for test in library_target.tests:
    assert test.library == library_target
```
?

We can do the former, but the latter wouldn't work - you can have "external" tests that use a different package name and only exercise the public interface (the current `go_test` defaults to creating that kind of test by inferring `package_name` as `path + "_test"`).

Coneko:
Yes, I meant the former.

It's a bit clunky because it means that if you want to make an internal test you have to update the definition of the library as well, and you can't make internal tests for a library for which you cannot edit the `BUCK` file, but maybe that makes sense, as you should be creating an external test then?

mikekap:
Yea you really shouldn't be creating internal tests for different packages. I'd love to limit library to targets in the same file, but maybe that's too restrictive. I'll implement the assert.

nemith:
I don't think that would be too restrictive.  Internal packages should live in the same directory as the package they are testing.  We already use the location of BUCK file to determine the package name so there is a strong relation between BUCK file and Go package.  It would make sense to have a tight relationship with internal tests as well.

For external tests there shouldn't be any restrictions as they are really standalone packages.

docs/rule/go_test.soy
```diff
@@ -35,14 +35,24 @@
 {call go_common.srcs_arg /}
 
 {call buck.arg}
+  {param name: 'library' /}
+  {param default: 'None' /}
+  {param desc}
+  Specify the library that this internal test is testing. This will copy the <code>srcs</code>,
+  {sp} <code>package_name</code> and <code>deps</code> from the target specified so you don't have
```
sdwilsh: you don't need the space after `{sp}`.

nemith:
How would this work for testing go_binary() rules?  The same logic should apply right even if the parameter is called 'library'?

mikekap:
Hmm I don't believe that currently works (you can't depend on a go_binary from a go_test). It probably should though, but perhaps via special casing - I don't think there's a good reason to support go_library depending on go_binary generically.

To answer your question, we could use library or create a new property. Running tests that are part of the main package is a bit weird since I'm not sure how to distinguish which main function to use. This probably needs some more go build reverse engineering :(

sdwilsh:
@facebook-github-bot shipit

sdwilsh:
We can deal with the `go_binary` stuff in a future PR/change :)

PR #651, opened 2016-02-09
Make native libraries work in the pex package style.

mikekap:
This follows up (and fixes) using -rpath in python extensions to load the dependent libraries.

The one downside is that you can't simply load a library with `ctypes` by name - you have to
load it by full path (pkg_resources.resource_filename). I think this is acceptable since that's
"normal" in python (although definitely annoying).

### DIFF
--- a/src/com/facebook/buck/cxx/CxxLinkableEnhancer.java
+++ b/src/com/facebook/buck/cxx/CxxLinkableEnhancer.java
@@ -176,6 +176,14 @@ public class CxxLinkableEnhancer {
     // the soname.
     if (linkType == Linker.LinkType.SHARED) {
       argsBuilder.add(new StringArg("-shared"));
+
+      argsBuilder.addAll(
+          StringArg.from(
+              Linkers.iXlinker(
+                  "-rpath",
+                  String.format(
+                      "%s/",
+                      cxxPlatform.getLd().resolve(ruleResolver).libOrigin()))));
     } else if (linkType == Linker.LinkType.MACH_O_BUNDLE) {
       argsBuilder.add(new StringArg("-bundle"));
       // It's possible to build a Mach-O bundle without a bundle loader (logic tests, for example).
@@ -333,6 +341,12 @@ public class CxxLinkableEnhancer {
       ImmutableList<? extends Arg> args) {
     ImmutableList.Builder<Arg> linkArgsBuilder = ImmutableList.builder();
     linkArgsBuilder.add(new StringArg("-shared"));
+    linkArgsBuilder.addAll(StringArg.from(
+        Linkers.iXlinker(
+            "-rpath",
+            String.format(
+                "%s/",
+                cxxPlatform.getLd().resolve(ruleResolver).libOrigin()))));
     if (soname.isPresent()) {
       linkArgsBuilder.addAll(
           StringArg.from(cxxPlatform.getLd().resolve(ruleResolver).soname(soname.get())));
--- a/src/com/facebook/buck/python/CxxPythonExtensionDescription.java
+++ b/src/com/facebook/buck/python/CxxPythonExtensionDescription.java
@@ -36,6 +36,7 @@ import com.facebook.buck.cxx.NativeLinkable;
 import com.facebook.buck.cxx.NativeLinkableInput;
 import com.facebook.buck.cxx.SharedNativeLinkTarget;
 import com.facebook.buck.io.ProjectFilesystem;
+import com.facebook.buck.io.MorePaths;
 import com.facebook.buck.model.BuildTarget;
 import com.facebook.buck.model.BuildTargets;
 import com.facebook.buck.model.Flavor;
@@ -69,6 +70,7 @@ import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
 
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Map;
 
 public class CxxPythonExtensionDescription implements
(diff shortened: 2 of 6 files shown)


mikekap:
Seems the testrunner test is still angry: `//test/com/facebook/buck/testrunner:testrunner failed with exit code 137`

sdwilsh:
@facebook-github-bot import

mikekap:
small note: the hardcoded pkg_resources name in the test is required due to https://github.com/pantsbuild/pex/issues/210 .

mikekap:
Ping

src/com/facebook/buck/python/CxxPythonExtensionDescription.java
```diff
@@ -198,10 +200,16 @@ protected Path getExtensionPath(BuildTarget target, Flavor pythonPlatform, Flavo
                 args.platformLinkerFlags,
                 cxxPlatform)));
 
-    // Embed a origin-relative library path into the binary so it can find the shared libraries.
+    // Embed a extension-relative library path into the .so so it can find the shared libraries that
+    // it depends on.
+    String rootPath = MorePaths.relativize(
+        Paths.get(""), Paths.get(args.baseModule.or(""))).toString();
```
marcinkosiba: Paths.get("") is a no-no, use the ProjectFileSystem (from params.getProjectFileSystem()) to resolve paths.

src/com/facebook/buck/cxx/CxxLinkableEnhancer.java
```diff
@@ -163,6 +163,14 @@ public static CxxLink createCxxLinkableBuildRule(
     // the soname.
     if (linkType == Linker.LinkType.SHARED) {
       argsBuilder.add(new StringArg("-shared"));
+
```
marcinkosiba: won't this make all C++ libraries (even those built, say, for Android) reference that directory at runtime?

src/com/facebook/buck/python/CxxPythonExtensionDescription.java
mikekap: These aren't real paths - this attempts to get enough `../` so that a module placed in a directory specified in `args.baseModule` would find the "root".

src/com/facebook/buck/cxx/CxxLinkableEnhancer.java
mikekap: Yes this makes all shared libraries relocatable without using LD_LIBRARY_PATH. Although I think android ignores rpaths anyway.

mikekap:
Ping

src/com/facebook/buck/python/CxxPythonExtensionDescription.java
```diff
@@ -201,12 +203,16 @@ protected Path getExtensionPath(BuildTarget target, Flavor pythonPlatform, Flavo
...
+    // it depends on.
+    Path baseModule = PythonUtil.getBasePath(params.getBuildTarget(), args.baseModule);
+    String rootPath = MorePaths.relativize(Paths.get(""), baseModule).toString();
     argsBuilder.addAll(
         StringArg.from(
             Linkers.iXlinker(
                 "-rpath",
-                String.format("%s/", cxxPlatform.getLd().resolve(ruleResolver).libOrigin()))));
+                String.format("%s/%s", cxxPlatform.getLd().resolve(ruleResolver).libOrigin(),
+                rootPath))));
```
Coneko: The indentation makes it look like `rootPath` is the third argument of `Linkers.iXlinker` instead of `String.format`.

Move each argument to `String.format` on its own line.

test/com/facebook/buck/python/PythonBinaryIntegrationTest.java
```diff
@@ -248,7 +243,9 @@ public void binaryIsCachedProperly() throws IOException {
 
   @Test
   public void externalPexToolAffectsRuleKey() throws IOException {
-    assumeThat(packageStyle, equalTo(PythonBuckConfig.PackageStyle.STANDALONE));
+    if (packageStyle != PythonBuckConfig.PackageStyle.STANDALONE) {
+      return;
+    }
```
Coneko: Why this change?

test/com/facebook/buck/python/PythonBinaryIntegrationTest.java
```diff
@@ -275,7 +272,10 @@ public void externalPexToolAffectsRuleKey() throws IOException {
 
   @Test
   public void multiplePythonHomes() throws Exception {
-    assumeThat(Platform.detect(), not(Matchers.is(Platform.WINDOWS)));
+    if (Platform.detect() == Platform.WINDOWS) {
+      return;
+    }
+
```
Coneko: Why this change?

test/com/facebook/buck/python/PythonBinaryIntegrationTest.java
mikekap: This (and below) are mostly cosmetic - it's to get rid of guaranteed assumption violations. In this case the pex tool isn't used for anything but the `STANDALONE` package style - on any platform/installation. It's not an assumption violation - it's just an artifact of parametrized tests that the `externalPexToolAffectsRuleKey[INPLACE,*]` test even exists. It boils down to giving a green build rather than a yellow one.

facebook-github-bot-6:

mikekap:
ping

facebook-github-bot-2:

Coneko:
@facebook-github-bot shipit

mikekap:
ping

Coneko:
This was blocked internally because it changes all libraries, not just the ones used in pexes.

cc @andrewjcg

mikekap:
I could add a different `Linker.LinkType.RELOCATABLE_SHARED` instead. Would that be preferable?

mikekap:
@Coneko @andrewjcg any word on what's the path forward here? Would you guys prefer to keep this out of the mainline tree?

Coneko:
I'm not sure if @andrewjcg had any idea how to fix it, but the problem here is that the rpath is embedded in every DSO, regardless of whether we use it in a python binary with the pex package style or not.

styurin:
@andrewjcg Do we want to merge this in?

andrewjcg:
Yeah, I'm not completely comfortable about baking this in to *all* libraries.  Since the library deployment strategy is really binary/rule specific, ideally the top-level binary would set the RPATH, as is done with C/C++ binaries.  However, this is obviously tricky since the Python interpreter is prebuilt.  We had a similar issue with Lua binaries, so we ended up embedding the interpreter into a super simple C/C++ program and setting `-rpath` on that.  I wonder if we can do similar here (e.g. https://docs.python.org/3/extending/embedding.html#very-high-level-embedding).

Issue #649, opened 2016-02-08
RuntimeError when running fetch on Windows


janicduplessis:
I get this error trying to build react native with buck on Windows. It works fine on my Mac.

I'm using buck from master and python 2.7.11.

```
C:\Users\janic\Developer\react-native>buck fetch ReactAndroid/src/main/java/com/facebook/react
Not using buckd because watchman isn't installed.
Warning raised by BUCK file parser: Traceback (most recent call last):
Warning raised by BUCK file parser:   File "C:\Users\janic\Developer\react-native\buck-out\tmp\buck_run.unzln3\buck1494403354654771159.py", line 2279, in <module>
Warning raised by BUCK file parser:     main()
Warning raised by BUCK file parser:   File "C:\Users\janic\Developer\react-native\buck-out\tmp\buck_run.unzln3\buck1494403354654771159.py", line 851, in main
Warning raised by BUCK file parser:     for section, contents in bser.loads(f.read()).iteritems():
Warning raised by BUCK file parser:   File "C:/Users/janic/Developer/buck/third-party/py/pywatchman\pywatchman\pybser.py", line 292, in loads
Warning raised by BUCK file parser:     raise RuntimeError('bser data len != header len')
Warning raised by BUCK file parser: RuntimeError: bser data len != header len
Warning raised by BUCK file parser: Traceback (most recent call last):
Warning raised by BUCK file parser:   File "C:\Users\janic\Developer\react-native\buck-out\tmp\buck_run.unzln3\buck2963190285078485547.py", line 2279, in <module>
Warning raised by BUCK file parser:     main()
Warning raised by BUCK file parser:   File "C:\Users\janic\Developer\react-native\buck-out\tmp\buck_run.unzln3\buck2963190285078485547.py", line 851, in main
Warning raised by BUCK file parser:     for section, contents in bser.loads(f.read()).iteritems():
Warning raised by BUCK file parser:   File "C:/Users/janic/Developer/buck/third-party/py/pywatchman\pywatchman\pybser.py", line 292, in loads
Warning raised by BUCK file parser:     raise RuntimeError('bser data len != header len')
Warning raised by BUCK file parser: RuntimeError: bser data len != header len
Parse error for build file C:\Users\janic\Developer\react-native\ReactAndroid\src\main\java\com\facebook\react\BUCK: Parser exited unexpectedly
```

LegNeato:
Are you using `watchman`? If so, what version? If not, does the issue go away if you install it?

[...]

PR #652, opened 2016-02-12
Fix OnDiskMavenDownloader path not resolving properly on windows

janicduplessis:
OnDiskMavenDownloader#fetch used root.resolve to convert the uri returned from the MavenUrlDecoder to a local file scheme. This did not work on Windows because of the drive letter.
```
file:///C:/com/android/support/support-v4/23.0.1/support-v4-23.0.1.aar
```
on windows vs
```
file:///com/android/support/support-v4/23.0.1/support-v4-23.0.1.aar
```
on mac/linux

This removes the extra uri conversion that was breaking the path because it is only used to log the DownloadEvent which can use the actual target path instead.

Tested by downloading support-v4 while building react-native on both mac and windows.

Fixes the 2nd issue in #649

### DIFF
--- a/test/com/facebook/buck/file/OnDiskMavenDownloaderTest.java
+++ b/test/com/facebook/buck/file/OnDiskMavenDownloaderTest.java
@@ -17,7 +17,6 @@
 package com.facebook.buck.file;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
 
@@ -86,7 +85,7 @@ public class OnDiskMavenDownloaderTest {
   }
 
   @Test
-  public void shouldDownloadFileFromLocalMavenRepoOnWindows() throws URISyntaxException, IOException {
+  public void shouldDownloadFileFromLocalMavenRepoWindows() throws URISyntaxException, IOException {
     FileSystem filesystem = Jimfs.newFileSystem(Configuration.windows());
     Path root = filesystem.getPath("C:\\Users\\bob\\.m2");
     Files.createDirectories(root);
--- a/test/com/facebook/buck/file/OnDiskMavenDownloaderTest.java
+++ b/test/com/facebook/buck/file/OnDiskMavenDownloaderTest.java
@@ -19,11 +19,13 @@ package com.facebook.buck.file;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThat;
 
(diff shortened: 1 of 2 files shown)


janicduplessis:
cc @sdwilsh

sdwilsh:
Could you also add a test for this?  You can create a new class, `OnDiskMavenDownloaderIntegrationTest`, and write a simple test that downloads from the SDK.  I want to make sure we don't regress this again or on Unix systems now.

janicduplessis:
There's already `OnDiskMavenDownloaderTest` but it tests using a mock unix filesystem so it didn't catch this issue. I can add a test like `shouldDownloadFileFromLocalMavenRepo` that uses a windows mock filesystem instead or maybe run all the tests in this file on the 3 different file system mocks (unix, osx, windows).

janicduplessis:
Does something like this looks good? This new test breaks before my fix.

test/com/facebook/buck/file/OnDiskMavenDownloaderTest.java
```diff
@@ -84,6 +84,29 @@ public void shouldDownloadFileFromLocalMavenRepo() throws URISyntaxException, IO
   }
 
   @Test
+  public void shouldDownloadFileFromLocalMavenRepoOnWindows() throws URISyntaxException, IOException {
+    // Same test as `shouldDownloadFileFromLocalMavenRepo` but using a windows
+    // mock filesystem.
```
sdwilsh: I wouldn't bother with this comment; the name implies this.

test/com/facebook/buck/file/OnDiskMavenDownloaderTest.java
```diff
@@ -84,6 +84,29 @@ public void shouldDownloadFileFromLocalMavenRepo() throws URISyntaxException, IO
...
+    Path source = root.resolve("group/project/0.1/project-0.1.jar");
+    Files.createDirectories(source.getParent());
+    Files.write(source, "cake".getBytes(UTF_8));
+
+    Downloader downloader = new OnDiskMavenDownloader(root);
+    downloader.fetch(BuckEventBusFactory.newInstance(), uri, output);
+
+    String result = new String(Files.readAllBytes(output), UTF_8);
+
+    assertEquals("cake", result);
```
sdwilsh: nit: use `assertThat(result, Matchers.equals("cake"));`

sdwilsh:
That looks good; just fix the two comments please :)

janicduplessis:
There you go!

sdwilsh:
@facebook-github-bot shipit

sdwilsh:
Whoops.  You've got some lint failures that you can see if you run `ant lint`.  CI is busted right now, but @bolinfest is looking into getting that resolved.

janicduplessis:
Oupsi, fixed.

sdwilsh:
@facebook-github-bot import

PR #653, opened 2016-02-12
Add resource union options to AndroidResource.

### LINKED ISSUE #643: "error: cannot find symbol" When using dep android library resource.
It seems that we are not able to access a dependent android library resource from within another android library. 

In this scenario we have two android libraries in the `libs` folder:

* `lib1`
* `lib2`

`lib1` depends on `lib2` for a resource, which it accesses directly via R. For example ([excerpt from this file](https://github.com/polycoder/buck-scenario-libraryresourcemerging/blob/master/libs/lib1/src/main/java/com/buckbuild/scenario/lib1/subpackage1/LibOneHelper.java))

```
// lib_two_value comes from lib2 but is available via R in lib1 when using gradle
context.getString(R.string.lib_two_value)
```

When accessing a `lib2` resource like this and running `buck build //libs/lib1:lib`, we get the following output:

```
> buck build //libs/lib1:lib
[-] PROCESSING BUCK FILES...FINISHED 0.2s [100%]
[+] DOWNLOADING... (0.00 B/S, TOTAL: 0.00 B, 0 Artifacts)
[+] BUILDING...1.0s [50%] (4/5 JOBS, 0 UPDATED, 0.0% CACHE MISS)
 |=> //libs/lib1:lib...  0.5s (running javac[0.4s])
 |=> IDLE
 |=> IDLE
 |=> IDLE
 |=> IDLE
/Users/seank/repo/buck-tests/BuckScenarioLibraryResourceMerging/libs/lib1/src/main/java/com/buckbuild/scenario/lib1/subpackage1/LibOneHelper.java:14: error: cannot find symbol
        return context.getString(R.string.lib_two_value);
                                         ^
  symbol:   variable lib_two_value
  location: class com.buckbuild.scenario.lib1.R.string
Errors: 1. Warnings: 0.
```

To help as much as possible, I've created a test buck scenario with additional documentation on the problem [here](https://github.com/polycoder/buck-scenario-libraryresourcemerging)

Perhaps our buck configuration is not correct, or maybe buck does not provide the same lib1 -> lib2 resource merging as Gradle like we would normally expect.

polycoder:
Provide an option to union an android_resource with its deps. This results in a merged R.txt file and allows for integrating buck with libraries that rely heavily on resource merging to reference dependent library resources. 

test plan: Add unit tests to MiniAppt, RDotTxtEntry, and AaptResourceCollector
 
this commit resolves #643

### DIFF
--- a/src/com/facebook/buck/android/AndroidResource.java
+++ b/src/com/facebook/buck/android/AndroidResource.java
@@ -131,6 +131,9 @@ public class AndroidResource extends AbstractBuildRule
   @AddToRuleKey
   private final boolean hasWhitelistedStrings;
 
+  @AddToRuleKey
+  private final boolean resourceUnion;
+
   private final ImmutableSortedSet<BuildRule> deps;
 
   private final BuildOutputInitializer<BuildOutput> buildOutputInitializer;
@@ -171,7 +174,8 @@ public class AndroidResource extends AbstractBuildRule
       Optional<SourcePath> additionalAssetsKey,
       @Nullable SourcePath manifestFile,
       Supplier<ImmutableSortedSet<? extends SourcePath>> symbolFilesFromDeps,
-      boolean hasWhitelistedStrings) {
+      boolean hasWhitelistedStrings,
+      boolean resourceUnion) {
     super(
         buildRuleParams.appendExtraDeps(
             Suppliers.compose(resolver.filterBuildRuleInputsFunction(), symbolFilesFromDeps)),
@@ -192,6 +196,7 @@ public class AndroidResource extends AbstractBuildRule
     this.manifestFile = manifestFile;
     this.symbolsOfDeps = symbolFilesFromDeps;
     this.hasWhitelistedStrings = hasWhitelistedStrings;
+    this.resourceUnion = resourceUnion;
 
     BuildTarget buildTarget = buildRuleParams.getBuildTarget();
     if (res == null) {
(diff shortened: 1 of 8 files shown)


polycoder:
The merged resources should receive a new id since the id could be taken in the current set.

marcinkosiba:
@polycoder copyWithNewIdValue should be OK. Take a look at the comment in [MergeAndroidResourceStep.sortSymbols](https://github.com/facebook/buck/blob/e797c883f1f31d27b11f6d5ff37a11e8e167813e/src/com/facebook/buck/android/MergeAndroidResourcesStep.java#L322).

polycoder:
ok - will do @marcinkosiba.

marcinkosiba:
Also, docs for the new option and a test would be nice :)

sdwilsh:
Well, we didn't doc the other gradle-specific stuff that @dreiss added because we really don't want people to use it.  It's a stopgap to help people convert and nothing more as I understand it.

polycoder:
@marcinkosiba I've been experimenting a bit with alternative approaches via MergeAndroidResourceSources, but can't seem to find something that works as easily as replacing the R.txt values like what is done here. I'll clean this up a bit and add tests.

@sdwilsh I can see you not wanting to add docs for this, it worked really nicely for use to be able to build our app, but it's because we rely very much on the gradle library/binary resource merging techniques.

marcinkosiba:
Thanks for persisting @polycoder!

polycoder:
@marcinkosiba I changed the approach a bit based on your suggestions, and added unit tests.

polycoder:
Thank for the feedback @marcinkosiba I made some additional changes and pushed a new commit!

marcinkosiba:
Sweet! I'm going to start the import flow on this.

polycoder:
Great, let me know if you need anything else with this!

polycoder:
@marcinkosiba, I forgot to make sure this was linted properly, so I just pushed a change to make sure all the lint issues are fixed.

marcinkosiba:
@facebook-github-bot shipit

marcinkosiba:
internal tests are failing because RDotTxtEntryTest.java and AaptResourceCollectorTest.java doesn't have the correct license header. Could you please update the PR to have those? (I'd do it but I'm not sure if it's OK from a legal perspective).

polycoder:
@marcinkosiba :+1: on adding the license headers

marcinkosiba:
@facebook-github-bot shipit

PR #665, opened 2016-02-24
Fix python test case reporting.

mikekap:
Test case name and test case were inverted (found when using grep to construct the failure list for windows).


Coneko:
@facebook-github-bot shipit

PR #666, opened 2016-02-24
Add platform via flavor to Go

mikekap:
This restructures how go rules build to be a bit more similar to Cxx rules with a noop rule that picks the flavor depending on the platform. This will later help Go<->Cxx interop, but short-term this properly includes GOOS & GOARCH in the rule key.

GOOS & GOARCH values are also needed to properly implement build tag filtering & go asm compilation.


mikekap:
Travis just doesn't like me: ` //test/com/facebook/buck/testrunner:testrunner failed with exit code 137`

src/com/facebook/buck/go/BUCK
```diff
@@ -1,23 +1,21 @@
+IMMUTABLES = ['AbstractGoPlatform.java']
+
+java_immutables_library(
+  name = 'immutables',
+  srcs = IMMUTABLES,
```
Coneko: You need to specify these as `immutable_srcs` for autodeps to work correctly.

Coneko:
I'd rather not add a new set of targets that use the cxx style graph enhancement and a new set of no-op rules.

I've described the different types of graph enhancement we do in Buck here: https://www.facebook.com/notes/uri-baghin/architecture-of-buck-1-graphs/1660525917559134 .

We're trying to make new graph enhancements use `requireRule` and `requireMetadata`.

In this case `GoLibraryDescription` would return a `NoopBuildRule`, not a subclass of it, and return a list of `GoLibraryLinkable` from `createMetadata` (see `MetadataProvidingDescription`), then the code in `GoDescriptors` can call `requireMetadata` exactly like it's calling `requireRule` now.

Compatibility with the cxx targets can be implemented by returning native linkables from their `createMetadata` (they already implement that method, but it's only used for compilation databases.

Coneko:
The `createMetadata` implementations can be a bit annoying because of all the mess with the flavours, but their essentially the same as dealing with `createBuildRule` and flavours.

Let me know if you need more information.

mikekap:
The metadata approach seems to just remove the need for noop build rules that propagate information. That seems pretty good to me - I can use it for `go_test(...library=...)` propagation as well.

If I have multiple types of metadata would the class type be the right way to switch on what to return, or are flavors the preferred route?

Coneko:
I would prefer switching on class type, but sometime flavours are used as well, for distinguishing different metadata that have the same class.

I don't mind either way.

mikekap:
Done, I think? Let me know if this is what you had in mind.

src/com/facebook/buck/go/GoLibraryDescription.java
```diff
@@ -57,21 +69,59 @@ public Arg createUnpopulatedConstructorArg() {
...
+  }
+
+  @Override
+  public <A extends Arg, U> Optional<U> createMetadata(
+      BuildTarget buildTarget,
+      final BuildRuleResolver resolver,
+      A args,
+      Class<U> metadataClass) throws NoSuchBuildTargetException {
+    if (metadataClass.isAssignableFrom(Arg.class)) {
+      return Optional.of(metadataClass.cast(args));
```
Coneko: I love this, can you move it to https://github.com/facebook/buck/blob/master/src/com/facebook/buck/rules/BuildRuleResolver.java#L74 so we get this by default in all descriptions that implement `MetadataProvidingDescription`?

Actually let's go one step further: let's do it for all descriptions regardless of whether they implement `MetadataProvidingDescription` or not, as it's completely general.

Can you do it in a separate PR though?

Coneko: Hold it, I'm double checking something.

Coneko: Ok, yeah, go ahead.

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -252,40 +223,125 @@ private static Path extractTestMainGenerator() {
...
-            .build(),
-        Suppliers.ofInstance(ImmutableSortedSet.<BuildRule>of()),
-        Suppliers.ofInstance(ImmutableSortedSet.<BuildRule>of())
-    );
+  private static ImmutableSet<GoLinkable> requireTransitiveGoLinkables(
+      final BuildTarget sourceTarget, final BuildRuleResolver resolver, final GoPlatform platform,
+      ImmutableSet<BuildRule> rules)
+      throws NoSuchBuildTargetException {
+    final ImmutableSet.Builder<GoLinkable> linkables = ImmutableSet.builder();
+    new AbstractBreadthFirstTraversal<BuildRule>(rules) {
```
Coneko: Can you replace this traversal with `requireMetadata` as well? You might need to add a recursive component where every description returns the transitive set of `GoLinkable` that target depends on.

Coneko:
Yes, this is exactly it.

As an aside, what do you think of the `requireMetadata` API? Which parts of it are good and which parts of it are clunky? I've noticed the fact it returns an optional was a bit of an annoyance in some parts.

mikekap:
re: requireMetadata - it's pretty solid. Here's some of my thoughts:
 - it's a bit of a shame that there's no way to really "cache" work between `createBuildRule` and `createMetadata` without round-tripping through the metadata. e.g. the `GoTestDescription` needs the package name `GoLibraryDescription` would have been compiled with to copy it over to the test (in case `library=` is specified). I currently just duplicated the ~ 1 line of logic.
 - `Optional` isn't very useful in this case. It feels more of a programming error when you ask for metadata that the description can't provide. IMO that could easily be `@Nullable` and an exception in `requireMetadata`.
 - `metadataClass` is a bit clunky - it's supposed to help with type safety, but since you need to use `metadataClass.cast` everywhere, it doesn't help much. I ran into a runtime error when I forgot to call `.build()` in one of the return sites - easy to fix, annoying that it can happen.
 - In general I think there's a problem between how "macro"-like descriptions interact (or perhaps there aren't easy enough helpers). That is, I was looking into adding Go-thrift generation support. This wasn't terribly difficult until I realized `ThriftLibraryDescription` now needs to support `createMetadata` just the same, which it would propagate to the language-specific enhancers. It's a bit weird how this is a lot easier to do with macros and rule suffixes (e.g. %s__go_compile).

Coneko:
> `Optional` isn't very useful in this case.

It's not useful in most cases. The issue is `deps` really: sometimes you want a certain metadata from all your deps, but you have no idea which of your deps can provide that metadata because `deps` is heterogenous. If we had `linked_deps`, `resource_deps` and so on it would be easier to deal with them. We've noticed `deps` being heterogenous and being treated differently (`getDeclaredDeps()` vs `getExtraDeps()`) are at the root of various awkward parts of the APIs so we might revisit it at some point.

> In general I think there's a problem between how "macro"-like descriptions interact

I don't understand this part, can you expand on it?

mikekap:
When I say "macro"-like descriptions, it's generally any rule descriptions that uses other descriptions to create it's output. e.g. `thrift_library#lang` can be "expanded" as:
 - `genrule(cmd='$(thrift_compiler) --gen lang...')`
 - `lang_rule(srcs=[':compiler_output'])`.

As someone writing `ThriftLibraryDescription`, let's say I wanted to implement it exactly like that: using `GenruleDescription` and `$(Lang)LibraryDescription`. If you do that, you have to delegate `createBuildRule`/`createMetadata`/`findDepsForTargetFromConstructorArgs` to the proper class based on flavor. Compare it to writing these as macros in python: you don't have to do much other than list a few dynamic `srcs` entries. It would be nice if that was taken care of for you somehow - either via utilities or an ability to add `TargetNode`s with different descriptions to the graph.

For reference, there's a high chance I'm going to run into this problem in a worse way with the `go_test` rule. `go_test` is essentially a sequence of 4 rules:
 - `go_compile(name='test_lib', srcs=srcs)`
 - `go_binary(name='test_main_gen', srcs=['testmaingen.go'])`
 - `genrule(name='test_main', cmd='$(test_main_gen) --out $OUT $SRCS', srcs=srcs)`
 - `go_binary(name='test', srcs=[':test_main'], deps=[':test_lib'])`

Which expands even further since `go_binary` is a `GoCompile` + link step. If I wanted to expand `go_library` to support linking with cxx rules (the go linker can generate .a & .so files), I'm betting I'll run into some annoying flavoring problems when fixing up `go_test`.

A little more generally, it would be nice if descriptions weren't ever called again for the same target listed in the BUCK file - similar to the way macros work. That is, if `thrift_library(name='hi')`, appears as a dep `:hi#go` in `go_library`, then `createBuildRule` doesn't need to be invoked again for `:hi#go,linux_amd64,library`. I think the reentry becomes necessary because  `createBuildRule` gets called too late for `deps` - `BuildRuleParams` already resolves targets to rules by that point. Maybe the right answer is to add a hook between `findDepsForTargetFromConstructorArgs` and `createBuildRule` that would allow propagating/adding the flavors from `go_binary` to `go_library` before the deps are constructed? It might make it a lot easier to follow the code of how symlink trees, flag sets, etc. get created.

This is mostly a minor point though, since few rules actually need the full dynamism. You likely shouldn't be using `GenruleDescription` to implement `ThriftLibraryDescription` :)

Coneko:
True, every target type does that kind of thing differently.

https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/ApplePackageDescription.java#L93 does it like you're suggesting, by using `findDepsForTargetFromConstructorArgs`.

All the approaches have pros and cons, I'm mostly advocating `requireRule` and `requireMetadata` because ideally once everything uses that we can get rid of `BuildRuleParams` creating dependencies eagerly.

To make that easier we're planning on implementing coercing for flavour convertibles as well, like description's `Arg` types.

I'd really like to do something about that `metadataClass`, but I have no idea what.

Even `createBuildRule` and consequently `requireRule` currently return something you end up casting a lot of the time. It's so bad.

mikekap:
Ah that makes sense. If `BuildRuleParams` doesn't create dependencies eagerly a lot of the weirdness just goes away. I'll look forward to seeing it :)

As for `metadataClass` - although it's a bit different, I think the easiest way to get rid of it is to only allow one type of metadata supplier per rule. There's a lot to lose by doing that though - caching being the biggest. As it is now, you're effectively implementing typeless "get attribute" which is difficult to formalize in a type system. Bazel does this via effectively [duck typing](http://bazel.io/docs/skylark/cookbook.html#optional-providers), so this isn't too far off.

Coneko:
@facebook-github-bot shipit

mikekap:
There was a lint error in there; might want to re-import 😬. Sorry.

Coneko:
@facebook-github-bot shipit

Coneko:
Let's see if it works...

PR #669, opened 2016-02-26
Pex in-place packages, take 2

mikekap:
This time without the micro version (since CPython at least promises not to break ABI compat on micro upgrades). Is there any code you guys have that depends on the system python vs pyrun distinction? The ABI should be the same there, but that's the only change now.

### DIFF
--- a/src/com/facebook/buck/python/AbstractPythonVersion.java
+++ b/src/com/facebook/buck/python/AbstractPythonVersion.java
@@ -25,11 +25,14 @@ import org.immutables.value.Value;
 abstract class AbstractPythonVersion {
 
   @Value.Parameter
-  public abstract String getVersionString();
+  public abstract String getInterpreterName();
+
+  @Value.Parameter
+  public abstract String getVersionString();  // X.Y.Z
 
   @Override
   public String toString() {
-    return getVersionString();
+    return getInterpreterName() + " " + getVersionString();
   }
 
 }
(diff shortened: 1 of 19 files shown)


Coneko:
Wow that was fast, @andrewjcg was looking into reverting the revert with a fix, so I'll assign this to him.

mikekap:
I'm in between jobs until next week, so I have a lot of free time...

mikekap:
Ping

mikekap:
Ping

Coneko:
@facebook-github-bot shipit

mikekap:
This PR is approaching 2.5 months now. Is there any chance of this getting committed? All it does is introduce pex tool flags & change the python version fetching implementation. I can probably rip out the latter too if that's the issue, though it's more correct since it distinguishes interpreter types (PyPy/Jython/CPython).

sdwilsh:
Looking at the internal change, it looks like the diff failed to apply when @Coneko imported it.  It's possible rebasing would help, but it might also require more manual intervention which is why @Coneko hasn't landed it yet.

Coneko:
This is landed internally. There's some issues with our git hooks that is preventing the commit to be published to open source, but we hope to solve it soon.

Coneko:
Sorry it took so long, I kept getting pulled away by other stuff.

mikekap:
No worries; I was more wondering whether there was something off with the change. Thanks for pushing it through 👍

PR #672, opened 2016-02-28
requireMetadata can always return the constructor arg

mikekap:
as requested by @Coneko


Coneko:
@facebook-github-bot shipit

PR #673, opened 2016-02-29
Add support to compile Go gc assembly

mikekap:
Gc being the standard go compiler; not to be confused with the garbage collector...

Pretty easy - run `go tool asm` for each .s file & use `go tool pack` to shove them into the same archive.

### DIFF
--- a/src/com/facebook/buck/go/GoBuckConfig.java
+++ b/src/com/facebook/buck/go/GoBuckConfig.java
@@ -134,6 +134,12 @@ public class GoBuckConfig {
   Tool getCompiler() {
     return getGoTool("compiler", "compile", "compiler_flags");
   }
+  Tool getAssembler() {
+    return getGoTool("assembler", "asm", "asm_flags");
+  }
+  Tool getPacker() {
+    return getGoTool("packer", "pack", "");
+  }
   Tool getLinker() {
     return getGoTool("linker", "link", "linker_flags");
   }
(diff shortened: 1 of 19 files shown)


src/com/facebook/buck/go/GoBuckConfig.java
```diff
@@ -155,8 +166,10 @@ private Tool getGoTool(
     }
 
     CommandTool.Builder builder = new CommandTool.Builder(new HashedFileTool(toolPath.get()));
-    for (String arg : getFlags(extraFlagsConfigKey)) {
-      builder.addArg(arg);
+    if (!extraFlagsConfigKey.isEmpty()) {
```
Coneko: Why is this extra check necessary? If `getFlags` doesn't return an empty list for an empty key maybe it should return an optional, or have a precondition check that the key is not empty.

Coneko:
It seems like it should be possible to split the work into three rules, the compilation rule for go sources, the assembling rule for assembly and the packing rules to put them together.

That way you would get higher caching granularity and parallelism.

src/com/facebook/buck/go/GoBuckConfig.java
mikekap: This is a private helper to the functions above. The packer doesn't accept any flags, so it doesn't have an extra flags config key. I can make it an optional, but it seemed less verbose to use an empty string.

mikekap:
Go builds generally don't need the sub-package parallelism. Go packages are a single shared private namespace: people generally don't throw too much stuff in a single package. Most public go packages have ~2-3 .s files maximum and generally less than 20 .go files. I realize that parallelism is easy to achieve here, but I don't think there's a need for it. If we wanted sub-package parallelism, it would make a lot more sense to target .go files themselves - I *think* you can compile a single .go file to a single .o file, but I don't know if there's a performance improvement there.

Coneko:
Ok, sounds good.

Coneko:
@facebook-github-bot shipit

Coneko:
Hmm, the patch doesn't apply, can you rebase on master?

PR #675, opened 2016-03-01
Remove the AbstractDescriptionArg

shs96c:
There's no need to inherit from a base class for constructor
args. Doing so prevents Buck users from using domain classes
as an arg (which is, admittedly, a little odd but still nice
to have as a possibility)

Instead, add the parameter available on all args to the same
place as every other global parameter.


shs96c:
As an aside, when I wrote the original coercer and reflective magic, I very deliberately chose to avoid forcing people to inherit from a buck class for their parameters. Generally they fall into "everything has it", in which case it should just be handled transparently, or "specific to a rule, but perhaps incredibly common", in which case having it be on the arg is a Bad Idea.

yiding:
Why did we make AbstractDescriptionArg... all for `licenses`? that seems a bit silly.

I'd like this to eventually do something like immutables, so you can include a bunch of interfaces and then the actual record class gets codegenned from all the interface methods.

sdwilsh:
This was the solution @marcinkosiba and I came up with for licenses because putting it in `BuckPyFunction` is rather non-discoverable unless you happen to know Buck internals really well.

Coneko:
I'm confused, how are you supposed to access the licenses with this change? They're not a field on target node like visibility and they're not a field on the constructor arg like before this change.

I guess it's @sdwilsh's bad for not adding a test, but `buck query "labels('licenses', deps('buck'))"` should return the list of licenses used in Buck.

shs96c:
You do know the buck internals really well.

shs96c:
Add the test and I'll make sure it passes. Probably should end up as a field on the targetnode.

sdwilsh:
> You do know the buck internals really well.

I thought I did, but trying to add `licenses` proved that to not be totally true!

marcinkosiba:
Args being concrete classes with fields (rather than following the immutables abstract method pattern) and Java not allowing fields on interfaces (or mixins) make this hard to express in the type system. 
I'd love be able to say:
```
interface CanHaveLicense {
  Optional<ImmutableSet<String>> getLicenses()  
}

class Arg implements CanHaveLicense {
}
```

(so essentially same thing as @yiding said above).
Making this stuff magical in the parser is a bit gross, though we don't really have enough of these special magic things to make a strong case for putting in the work to do the immutables-like approach.

facebook-github-bot-8:

facebook-github-bot-9:

PR #679, opened 2016-03-02
Allow ScalaStep to expand .src.zip sources

mrkane27:
This is a bit exploratory. Basically, I'm trying to figure out how to get a `scala_library()` that comprises `.scala` files generated automatically by a `genrule()`.

My initial attempt was a genrule that generated a directory, but nothing really seemed to build when I passed off `:mygenrule` as a source to the `scala_library()` rule. I ended up slapping a `zip_file()` on top of the `genrule()` and then sending this PR. I reckon there's value in supporting zipped sources for Scala as well, but would like to know whether it's a reasonable thing to do for Buck.

### DIFF
--- a/src/com/facebook/buck/jvm/java/ExternalJavac.java
+++ b/src/com/facebook/buck/jvm/java/ExternalJavac.java
@@ -181,7 +181,8 @@ public class ExternalJavac implements Javac {
           filesystem,
           invokingRule,
           javaSourceFilePaths,
-          workingDirectory);
+          workingDirectory,
+          ".java");
     } catch (IOException e) {
       throw new HumanReadableException(
           "Unable to expand sources for %s into %s",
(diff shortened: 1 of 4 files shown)


src/com/facebook/buck/jvm/java/ExternalJavac.java
```diff
@@ -223,11 +224,12 @@ public int buildWithClasspath(
     return exitCode;
   }
 
-  private ImmutableList<Path> getExpandedSourcePaths(
+  public static ImmutableList<Path> getExpandedSourcePaths(
```
mrkane27: I would probably move this elsewhere if we decide to go with this PR.

src/com/facebook/buck/jvm/scala/ScalacStep.java
```diff
@@ -88,8 +97,24 @@ public String getShortName() {
...
+          sourceFilePaths,
+          Optional.of(workingDirectory),
+          ".scala");
+    } catch (IOException e) {
+      throw new HumanReadableException(
+          "Unable to expand sources for %s into %s",
+          invokingRule,
+          workingDirectory);
+    }
+
```
mrkane27: I saw that `ExternalJavac` supports passing things off via `@` listfiles. That might be something to support in Scala as well.

Coneko:
cc @grumpyjames

grumpyjames:
Historical context: https://github.com/facebook/buck/pull/546

It would be great if we could find a single place for this behaviour to live such that it supports each jvmlang implementation appropriately. That PR was a spike towards that; you might, if you were feeling particularly enthusiastic, feel like reviving it, or on the other hand perhaps not :-)

Having `getExpandedSourcePaths` be static and somewhere other than `ExternalJavac` doesn't sound like too bad a place to move to for the time being, in any case :-)

mrkane27:
@grumpyjames thanks for the feedback. It does look like this is indeed more involved than I assumed initially. I've added support for creating `@` files, but I would need to look at the other PR as well.

facebook-github-bot-6:

PR #680, opened 2016-03-02
Create classes.jar if prebuilt aar does not already contain one.

### LINKED ISSUE #417: android_prebuilt_aar fails if an aar is missing a classes.jar
For example, try creating an `android_prebuilt_aar` for $ANDROID_HOME/extras/google/m2repository/com/google/android/gms/play-services/7.5.0/play-services-7.5.0.aar

polycoder:
resolves #417

### DIFF
--- a/src/com/facebook/buck/android/UnzipAar.java
+++ b/src/com/facebook/buck/android/UnzipAar.java
@@ -128,6 +128,15 @@ public class UnzipAar extends AbstractBuildRule
         }
 
         Path classesJar = unpackDirectory.resolve("classes.jar");
+        if (!getProjectFilesystem().exists(classesJar)) {
+          try {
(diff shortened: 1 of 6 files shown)


sdwilsh:
I'm going to defer to @marcinkosiba on this one actually.

marcinkosiba:
Seems like a reasonable way to go about this. An integration test would be nice if it's not too much trouble to write one.

polycoder:
@marcinkosiba I'll see about moving the functionality to JarDirectoryStepHelper, and will def add an integration test.

polycoder:
@marcinkosiba I moved some of the logic to JarDirectoryStepHelper and added an integration test so this doesn't regress in the future.

marcinkosiba:
@facebook-github-bot shipit

PR #682, opened 2016-03-03
Add support for go 1.6-style vendoring.

mikekap:
I think longer-term it might be nice to remove the use of Path for go package names. It's definitely convenient but they're not really real paths and it confuses the rule key builder pretty badly.

### DIFF
--- a/src/com/facebook/buck/go/GoDescriptors.java
+++ b/src/com/facebook/buck/go/GoDescriptors.java
@@ -37,6 +37,7 @@ import com.facebook.buck.rules.SourcePathResolver;
 import com.facebook.buck.rules.SymlinkTree;
 import com.facebook.buck.rules.Tool;
 import com.facebook.buck.util.HumanReadableException;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
@@ -47,6 +48,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.Maps;
 import com.google.common.io.Files;
 import com.google.common.io.Resources;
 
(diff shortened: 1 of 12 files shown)


src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -142,6 +153,29 @@ static GoCompile createGoCompileRule(
...
   }
 
+  @VisibleForTesting
+  static ImmutableMap<Path, Path> getPackageImportMap(
+      Path basePackageName, Iterable<Path> packageNameIter) {
+    Map<Path, Path> importMapBuilder = Maps.newHashMap();
+    ImmutableSortedSet<Path> packageNames = ImmutableSortedSet.copyOf(packageNameIter);
+
+    Path prefix = Paths.get("");
+    for (Path component : FluentIterable.of(new Path[]{Paths.get("")}).append(basePackageName)) {
```
Coneko: `ImmutableList.of(Paths.get(""), basePackageName)` would work as well right?

src/com/facebook/buck/go/GoDescriptors.java
```diff
@@ -142,6 +153,29 @@ static GoCompile createGoCompileRule(
...
+      prefix = prefix.resolve(component);
+
+      Path vendorPrefix = prefix.resolve("vendor");
+      for (Path vendoredPackage : packageNames.tailSet(vendorPrefix)) {
+        if (!vendoredPackage.startsWith(vendorPrefix)) {
+          continue;
```
Coneko: I think you can `break` here: if they're ordered alphabetically if any string doesn't start with the prefix, any string following it won't start with the prefix either.

Coneko:
Seems good, only one comment and I'm not even sure I'm right on it.

I'll import it tomorrow if you haven't answered by then.

src/com/facebook/buck/go/GoDescriptors.java
mikekap: It's not quite the prefix. We're iterating through prefixes, but we're checking for `<prefix>/vendor` - i.e. for a package in `a/b/c` we'll check (& overwrite - longest wins here): `vendor/`, `a/vendor/`, `a/b/vendor/`, `a/b/c/vendor/`.

Coneko: I get that part, but that's the outer loop right? The inner loop is comparing a candidate, for example `a/vendor`, with all the paths.

I'm saying that if you start alphabetically by using `tailSet` you can also stop alphabetically by using `break`:

```
a/b
a/b/c
a/vendor <= tailSet makes you start here
a/vendor/t
a/vendor/u
a/x <= break makes you stop here
a/y
a/z
```

mikekap: There goes my brain. You're completely right of course.

Coneko:
@facebook-github-bot shipit

PR #695, opened 2016-03-09
Replace usages of static final fields on Guava's CharMatcher.

Dominator008:
This change replaces usages of static final fields on CharMatcher, eg. ChraMatcher.WHITESPAcE, with the equivalent static factory methods, eg. CharMatcher.whitespace().

Static final fields must be eagerly initialized, potentially causing many unnecessary implementation classes to be loaded. Using methods allow implementation classes to be loaded when actually needed.

These methods became available in Guava 19.0. Guava will eventually deprecate the static final fields.

See: https://github.com/google/guava/commit/d810fc4b42d80b64fd118c2cf1483fb7682d34a9.


Coneko:
@facebook-github-bot shipit

PR #696, opened 2016-03-10
Fix defect where a class that is abstract but has @Test annotations w…

ray-milkey:
…as attempted to be instantiated and threw java.lang.InstantiationException

If you have an abstract Java class that defines methods with @Test annotations, buck attempts to instantiate an object with the class type and fails with a java.lang.InstantiationException.

This patch skips abstract classes when selecting tests to run. We hit this error trying to port ONOS from maven to buck.

### DIFF
--- a/test/com/facebook/buck/testrunner/testdata/build_then_test/AbstractClassWithTests.java
+++ b/test/com/facebook/buck/testrunner/testdata/build_then_test/AbstractClassWithTests.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2016-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


test/com/facebook/buck/testrunner/testdata/build_then_test/AbstractClassWithTests.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2012-present Facebook, Inc.
```
sdwilsh: 2016, not 2012 :)

sdwilsh:
I'm otherwise happy with this; just the one little comment.

ray-milkey:
This is not a new file, I just added a Test method to an existing one, so the year in the comment should be correct.

sdwilsh:
You added a new test file which is where I commented.  I'm fine with the other files being the way they are.

ray-milkey:
Ah sorry, I was looking at the java test file, not the resource. I'll fix it.

ray-milkey:
OK the year in the copy right notice is updated.

sdwilsh:
@facebook-github-bot shipit

PR #705, opened 2016-04-01
Fix function docs always containing `visibility` and `licenses`

LegNeato:
Before:

screen shot 2016-03-31 at 5 10 58 pm

After:

screen shot 2016-03-31 at 5 11 07 pm


Coneko:
@facebook-github-bot shipit

PR #707, opened 2016-04-07
Adding MavenCoordinatesMacroExpander

bocon13:
This expander transforms build rule into maven coordinates associated
with that rule. The build rule must have maven coordinates.

Example:
$(maven_coords :target) --> groupId:artifactId:version


docs/__genrule_common.soy
```diff
@@ -60,6 +60,12 @@
       <dd>Expands to the location of the output of the build rule. This
       means that you can refer to these without needing to be aware of how
       Buck is storing data on the disk mid-build.</dd>
+
+      <dt><code>$(maven_coords /&#x2F;path/to:target)</code></dt>
+      <dd>Expands to the Maven coordinates for the build rule. This allows
+      you to access the Maven coordinates for Maven-aware build rules. The
+      format of the expansion is: <code>group_id:artifact_id:version</code>
```
sdwilsh: In the build files, we use a buildr format, so you'd have the artifact type before the version.  Why did you decide to drop that here?

bocon13: @sdwilsh The artifact type (extension) and classifier are not explicitly omitted from the expander. I can extend the documentation to contain the full format: 
<groupId>:<artifactId>[:<extension>[:<classifier>]]:<version>

docs/__genrule_common.soy
```diff
@@ -60,6 +60,14 @@
       <dd>Expands to the location of the output of the build rule. This
       means that you can refer to these without needing to be aware of how
       Buck is storing data on the disk mid-build.</dd>
+
+      <dt><code>$(maven_coords /&#x2F;path/to:target)</code></dt>
+      <dd>Expands to the Maven coordinates for the build rule. This allows
+      you to access the Maven coordinates for Maven-aware build rules. The
+      format of the expansion is: <code>
+      &lt;groupId&gt;:&lt;artifactId&gt;[:&lt;extension&gt;[:&lt;classifier&gt;]]:&lt;version&gt;
```
sdwilsh: Note: you can use the `{literal}` tag here to use actual `<` and `>` so this is more readable

sdwilsh:
Apologies for the delay.  This fell off my radar.

alisdair04:
@facebook-github-bot shipit

docs/__genrule_common.soy
bocon13: It seems like the `{literal}` tag blocks HTML escaping, and that `<` and `>` are not auto-escaped within the HTML template without `{literal}`. However, we can make things more readable with a `{let}` block, which should render/escape the angle brackets on replacement. I will push an update.

bocon13:
@sdwilsh No worries! Thanks

alisdair04:
@facebook-github-bot shipit

PR #728, opened 2016-04-27
Adding base_module_strip to Python targets

borgstrom:
Often times I want to organize my source tree so that my Python sources
start one or two directories down from the project root.  With a
.buckconfig at the top level these two directories must exist in the
module namespace unless you go and manually specify base_module for each
of the targets.  When moving sources into Buck this meant I would need
to refactor a large number of files just for imports.

This adds a new base_module_strip argument to Python targets, which if
specified strips that many elements when PythonUtil.getBasePath is
called using the subpath method on the path.  You can also specify the
value as a config value and have it apply to all targets.

### DIFF
--- a/docs/__python_common.soy
+++ b/docs/__python_common.soy
@@ -65,6 +65,21 @@
 {/template}
 
 /***/
+{template .base_module_strip_arg}
+{call buck.arg}
+  {param name: 'base_module_strip' /}
+  {param default : 'None' /}
+  {param desc}
+  An integer consisting of the number of path elements to strip from the beginning of the
+  `base_module`.  Often times you want to store your source under a directory named `src` or
+  something similar, but you want the Python namespace to be `import my.module` and not `import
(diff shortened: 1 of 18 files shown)


facebook-github-bot-4:
By analyzing the blame information on this pull request, we identified @andrewjcg, @michath and @ryu2 to be potential reviewers.

mikekap:
Not sure how the FB folks feel about this, though if you don't want to have a fork of buck, I've previously implemented something similar with a `DEFS`-like file. Basically you replace `python_library` with something that sets `base_module` based on `'.'.join(get_base_path().split('/')[1:])`.

borgstrom:
@mikekap -- Thanks for the info.  I was thinking that I could use something along those lines as a fallback, but in the sprit of Buck I'm trying to keep my BUCK files as simple as possible here.

I believe this is a fairly non-intrusive change that you fully need to opt-in to, so hopefully the FB folks are open to this change.  :)

PR #732, opened 2016-05-02
add `rt.jar` into javac classpath

Piasy:
Summary:
To let BUCK build java 8 sources correctly, we need add `rt.jar` into javac classpath, otherwise, `android_library` rule will fail to compile (but `java_library` compiles successfully).

Test Plan:
Build updated code, and tested in demo this repo: https://github.com/Piasy/BuckJava8RetroLambdaDemo, unit test code need to be updated.

### DIFF
--- a/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java
+++ b/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java
@@ -430,6 +430,7 @@ public class DefaultJavaLibrary extends AbstractBuildRule
         ImmutableSetMultimap.<JavaLibrary, Path>builder()
             .putAll(getDeclaredClasspathEntries())
             .putAll(this, additionalClasspathEntries)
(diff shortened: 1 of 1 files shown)


Piasy:
Without `rt.jar` in classpath, the compiler fails with:
```java
com.sun.tools.javac.code.Symbol$CompletionFailure: class file for java.lang.invoke.MethodType not found.
```
See https://github.com/Piasy/OkBuck/issues/32 and https://github.com/Piasy/BuckJava8RetroLambdaDemo

facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @jkeljo, @davido and @grumpyjames to be potential reviewers.

Piasy:
ping?

Piasy:
Since we could add `rt.jar` into `-bootclasspath` option via `extra_arguments` parameter, there is no need  for BUCK to do so.

PR #739, opened 2016-05-09
Make sh_test `buck run`-able

mikekap:


facebook-github-bot-0:
By analyzing the blame information on this pull request, we identified @andrewjcg, @marcinkosiba and @bhamiltoncx to be potential reviewers.

facebook-github-bot-9:

marcinkosiba:
While you could argue it makes sense for sh_test to be runable, it won't be the case for other tests (cxx, java). I would prefer to not introduce this kind of inconsistency. You then start getting into scenarios where you $(execute) a test as part of a genrule to build some production code..
If you want to use the same code in a binary and test why not create `sh_test` that runs the `sh_binary`?

mikekap:
Why wouldn't it make sense for other tests? Perhaps their output wouldn't be quite intelligible (or even considered stable), but it does make sense to run them. Go & Python tests are already this way.

The goal here is to make sh_test `buck run`-able. Ideally that would mean teaching the run command how to run tests. I don't have the expertise to do that. While having `sh_test` be executable isn't great, I'm really doubting anyone would ever use that. Making `sh_test` runnable for debugging purposes however, is extremely useful for debugging hangs and the like.

Making an `sh_binary` from an `sh_test` isn't always possible. It's not currently possible to have runtime deps for `sh_binary` (seems like a bug in `resources` handling), and moreover the values of some macros (e.g. `classpath`, `location`) should never be written to an artifact (they contain absolute paths and hence don't cache well). This makes an equivalent `sh_binary` for an `sh_test` with `args` & `env` containing macros impossible.

marcinkosiba:
It's mostly java/robolectric tests that get the special treatment, the invocation results get spat out into an .xml file that is interpreted by buck. If you were to just `run` those you'd not get any useful output, which is why I said it would not make sense (I did not consider the debugging hangs case).

From a quick look it does seem like `ShBinary` should also return its resources from `getRuntimeDeps` (rather than just the dep for `main`).

marcinkosiba:
@facebook-github-bot shipit

facebook-github-bot-6:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/983184065134600/int_phab) to review.

facebook-github-bot-6:

PR #742, opened 2016-05-12
Expose option to force final resource ids

### LINKED ISSUE #223: Support application code and libraries for android applications
Ant and Gradle have a concept of "application code" and "libraries", but Buck only has libraries. The application code gets one special privilege, however: its resource ids are constants, which means that they can be used in switch statements and annotations.

This would able things like this with ButterKnife:

```
@InjectView(R.id.button1) Button button1;
@InjectView(R.id.button2) Button button2;
```

kageiit:
Expose a `force_final_resource_ids` option on the `android_library` rule.

This is useful to set on library rules that are only depended on by `android_binary` rules. This enables libraries like Butterknife and android-annotations to work properly with buck when used in apps.

Partially resolves #223

### DIFF
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -96,7 +96,7 @@ public class AndroidLibraryDescription
             Suppliers.ofInstance(resolver.getAllRules(args.exportedDeps.get()))),
         javacOptions,
         DependencyMode.FIRST_ORDER,
-        /* forceFinalResourceIds */ false,
+        args.forceFinalResourceIds.or(false),
         args.resourceUnionPackage);
 
     boolean hasDummyRDotJavaFlavor =
@@ -175,5 +175,6 @@ public class AndroidLibraryDescription
   public static class Arg extends JavaLibraryDescription.Arg {
     public Optional<SourcePath> manifest;
     public Optional<String> resourceUnionPackage;
+    public Optional<String> forceFinalResourceIds;
   }
 }
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -175,6 +175,6 @@ public class AndroidLibraryDescription
   public static class Arg extends JavaLibraryDescription.Arg {
     public Optional<SourcePath> manifest;
     public Optional<String> resourceUnionPackage;
-    public Optional<String> forceFinalResourceIds;
+    public Optional<Boolean> forceFinalResourceIds;
   }
 }


facebook-github-bot-2:

facebook-github-bot-0:

kageiit:
@sdwilsh can this be reviewed please?

sdwilsh:
@marcinkosiba or @dreiss would be better.  I've changed teams and I'm not actively hacking on Buck these days.

dreiss:
Does this work?  It seems like there's no way to guarantee that the resource ids we generate in dummy_r_dot_java will be the same as what goes into the uber R.java and the resource bundle.  (In fact, it's practically guaranteed that they won't.)

kageiit:
@dreiss You are right. The R.java entries from the dummy and the uber ones differ, so the generated classes from butterknife have the wrong integer constant specified which crashes the app at runtime. One solution I can think of is to run the aapt step before javac/apt when force_final_resource_ids is true to ensure the ids do not mismatch. Let me know if that sounds ok, I will investigate how to do so

kageiit:
Running aapt before source code generation is how gradle behaves. Is there some reason why this is not the case in buck?

marcinkosiba:
> Running aapt before source code generation is how gradle behaves. Is there some reason why this is not the case in buck?

if we were to run appt beforehand and have the java code see the R.id values it would be impossible to share the results of compiling an individual java/android_library across >1 android_binary. Also, aapt is slow, so sticking it on the critical path is not a good idea

marcinkosiba:
It would be nice if butterknife could run in a mode where it doesn't inline the constants (you can still reclaim the perf benefits by running proguard on the final .jar and have it do the inlining).

kageiit:
@marcinkosiba this is only for the android library rule that corresponds to the android binary code. This will not be run on every android library rule. Only the R.java generated using the code of the app will be done this way. That code is not shared across binaries.

Butterknife is not the only tool to rely on resource IDs this way there are plenty of other libraries that do so like android-annotations, android-data-binding etc. We would be excluding an entire class of apps from being able to take advantage of buck without this.

Gradle runs resource processing before code gen already. Can you expand on what do you mean by it being slow? It still has to run atleast once during compilation. If we were to enable running aapt before only for the binary I think we can still share the result of compiling individual java/android libraries

sdwilsh:
I got pinged about this by someone else, so I'm chime in anyway.  In my 
comment on #223, what I was trying to say is that the best way to fix 
that is to have a `srcs` attribute on `android_binary` that would make 
the special library code with final ids.  I don't think exposing this UI 
in the build file is a good idea because it'll be easy to set as 
copy-pasta and you end up slowing your build way down.

However, we can take the `srcs` from an `android_binary` and graph 
enhance it to produce an `android_library` with `forceFinalResourceIds` 
set to `true`.  It's a bigger change, but it's the safer one and I think 
@marcinkosiba would be okay with that.  I'm pretty sure I've talked with 
him or @dreiss about that very solution (but never had time to code it up).

dreiss:
> It would be nice if butterknife could run in a mode where it doesn't inline the constants

I don't think this is possible.  Values in annotations need to be inlined constant values, not field references.

> Can you expand on what do you mean by it being slow?

It scales non-linearly and is extremely slow for large apps.

> It still has to run atleast once during compilation.

Not necessarily for incremental compilation.  But even when it is necessary, we want to compile java code while it's running.

> If we were to enable running aapt before only for the binary I think we can still share the result of compiling individual java/android libraries

I don't understand.  For normal android_library rules, we need to use non-final resource ids so the results can be shared between different apps.

I think the solution in Shawn's comment is the right one.

kageiit:
> However, we can take the `srcs` from an `android_binary` and graph
> enhance it to produce an `android_library` with `forceFinalResourceIds`
> set to `true`.  It's a bigger change, but it's the safer one and I think
> @marcinkosiba would be okay with that.  I'm pretty sure I've talked with
> him or @dreiss about that very solution (but never had time to code it up).

The issue is that aapt step rewrites the resource ids that get put into the resource bundle/apk. So it does not matter if we generate R.java with final values by setting `forceFinalResourceIds=true` , if it will get overwritten anyway.

sdwilsh:
Then we'd have to run `aapt` on the `srcs` provided in this library (it's been a bit since I've looked at how this works in Android, so I'm a bit fuzzy on it these days).

sdwilsh:
I just checked with @dreiss.  We'd have to graph-enhance an `android_library` from the `srcs` on the `android_binary` rule and have it depend on the `aapt_package` rule and it needs to compile against the final uber `R.java` file (and not the normal dummy `R.java` file).

kageiit:
@sdwilsh thats the same idea I had in mind as well. We can of course do this without exposing the forceFinalResourceIds as well. But it would still be nice to have this as an option since if an app does not use this, there is no need for it to suffer he burden of having aapt run on the critical path.

sdwilsh:
If an app doesn't specify any `srcs`, I don't think we'd have to do this.  Am I misunderstanding you?

kageiit:
> If an app doesn't specify any srcs, I don't think we'd have to do this. Am I misunderstanding you?

the `android_binary` rule does not expose anyway to set a `srcs` in buck. Are you talking about the sources that come from the various dependencies?

sdwilsh:
Sorry, it looks like I was ambiguous before.  We'd have to add `srcs` to `android_binary`, and only those would be subject to this.  Libraries would work as they do today.  Obviously this means that if you need to use butterknife, it needs to be a `src` of the binary, and not part of a library.

dreiss:
> Obviously this means that if you need to use butterknife, it needs to be a src of the binary, and not part of a library.

Note that this is the case with Gradle as well.

kageiit:
Sounds like a pretty major change :(

Google announced a new aapt at Google I/O yesterday that is not supposed to rewrite the IDs when packaging. Combined with the original PR and having resourceUnion = True I believe it might be able to solve the problem without this change (although it would be the right thing to do for the long term). I will try that out and update on what the result is.

kageiit:
I made a version of butterknife that outputs R.id.identifier in the generated code instead of inlining the integers. But then I realized that it does not matter even if the dummy R dot Java files are final, the resource will not be found at runtime since the final R dot Java has entirely different values. aapt seems to be still rewriting resource ids on the latest sdk tools version available.

Seems like the solution proposed by @sdwilsh is the proper one in the long term that will work for all cases.

dreiss:
> Google announced a new aapt at Google I/O yesterday that is not supposed to rewrite the IDs when packaging

Do you have a link for this?  I'll check it out and see if it might be able to solve this in a different way.

dreiss:
I think I found the reference.  It looks like they just improved its ability to produce compatible output from run to run.  That wouldn't help us with library projects, though.

marcinkosiba:
> I made a version of butterknife that outputs R.id.identifier in the generated code instead of inlining the integers. But then I realized that it does not matter even if the dummy R dot Java files are final, the resource will not be found at runtime since the final R dot Java has entirely different values. aapt seems to be still rewriting resource ids on the latest sdk tools version available.

I don't understand how appt changing the value defeats the butterknife changes.
Buck deals with aapt changing the values by using non-final fields in the generated R class (this way the java compiler doesn't inline R.id constants). This allows us to 'swap in' the correct final R.class right before creating the apk.

If you got butterknife to refer to fields then it should "just work" as the code generated by butterknife will continue to refer to the non-inlined fields (rather than their values), so when buck swaps in the R.class with the correct resource IDs the code generated by your version of butterknife should pick those up.

kageiit:
@marcinkosiba sorry, I forgot to mention that javac optimizations inline the constants so the class files will just have the integers (since R.id values are final and we cannot use non final values in annotations). I have however been able to come up with a system that will support libraries to work with butterknife and put the non final R values in the generated code. This has a side effect of making buck work with butterknife without any changes to buck. I plan to open a PR on butterknife soon.

dreiss:
> If you got butterknife to refer to fields

This is non-trivial.  See https://github.com/JakeWharton/butterknife/issues/100 .  I look forward to seeing  your PR, @kageiit .

kageiit:
@dreiss @sdwilsh @marcinkosiba I opted to make this work in butterknife instead. https://github.com/JakeWharton/butterknife/pull/613

marcinkosiba:
@kageiit - really glad to hear that you got it working that way! I'm assuming it's OK to close this PR. If there is any extra work needed for buck to work with the code you landed in ButterKnife, could you please open a separate PR?

PR #752, opened 2016-05-22
Remove warning message during dx that affects buck build progress printing

kageiit:
When using libraries like leakcanary dx prints warnings like
```
warning: Ignoring InnerClasses attribute for an anonymous inner class
(com.squareup.haha.guava.base.Joiner$1) that doesn't come with an
associated EnclosingMethod attribute. This class was probably produced by a
compiler that did not target the modern .class file format. The recommended
solution is to recompile the class from source, using an up-to-date compiler
and without specifying any "-target" type options. The consequence of ignoring
this warning is that reflective operations on this class will incorrectly
indicate that it is *not* an inner class.
```

That completely stalls the buck build progress from being output to stdout. The build will silently pass/fail without any indication whatsoever due to this after being stuck

```
[+] BUILDING...5.8s [28%] (227/495 JOBS, 227 UPDATED, 227 [45.9%] CACHE MISS)
```

This change prevents the warning from being print to the console.

### DIFF
--- a/src/com/facebook/buck/android/DxStep.java
+++ b/src/com/facebook/buck/android/DxStep.java
@@ -16,6 +16,7 @@
 
 package com.facebook.buck.android;
 
+import com.facebook.buck.event.ConsoleEvent;
 import com.facebook.buck.io.ProjectFilesystem;
 import com.facebook.buck.shell.ShellStep;
 import com.facebook.buck.step.ExecutionContext;
@@ -29,8 +30,10 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.PrintStream;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.EnumSet;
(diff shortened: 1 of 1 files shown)


kageiit:
@dreiss @marcinkosiba

dreiss:
Ignoring warnings seems kind of bad.  Maybe we can find a way to pipe this through Buck's console abstraction to avoid stalling the console.

Do you know where the source for that class is?  I don't see it in the haha repository.

marcinkosiba:
Agree with @dreiss, the correct way to fix this is to post a ConsoleEvent to the BuckMessageBus, doing that causes the warning to print without breaking the rendering

kageiit:
> Agree with @dreiss, the correct way to fix this is to post a ConsoleEvent to the BuckMessageBus, doing that causes the warning to print without breaking the rendering

I can just replace the warning call to instead post a ConsoleEvent to the BuckMessageBus. That sounds fine by me.

> Do you know where the source for that class is? I don't see it in the haha repository.

Maybe an older version of haha. This also happens on a lot of other sdks provided by baidu for example

kageiit:
@marcinkosiba @dreiss this code is in the third party dx code dump. Is there a way to get access to the BuckEventBus there?

marcinkosiba:
@kageiit take a look at DxStep instead. We give the dexer direct access to stdout/stderr, we would need to buffer the output and re-post as a consoleevent

facebook-github-bot-6:

kageiit:
@marcinkosiba updated with your suggestion

PR #753, opened 2016-05-22
Support for manifest placeholders

### LINKED ISSUE #743: Support for manifest placeholders
Does buck support placeholder values in manifest similar to http://tools.android.com/tech-docs/new-build-system/user-guide/manifest-merger#TOC-Placeholder-support ?

I cannot find anything in the docs that indicates so. Is there some undocumented api support for this?

kageiit:
This adds support for manifest placeholders:
 http://tools.android.com/tech-docs/new-build-system/user-guide/manifest-merger#TOC-Placeholder-support

Resolves #743 

Many components of libraries like Google Play Services etc. have an implicit `${applicationId}` placeholder which gets replaced correctly by the `android_binary`'s application id when building with gradle.

This PR enables buck to do the same and also support custom placeholders at the same time by adding an optional `placeholders` entry to the `manifest_entries` configuration.

### DIFF
--- a/src/com/facebook/buck/android/ReplaceManifestPlaceholdersStep.java
+++ b/src/com/facebook/buck/android/ReplaceManifestPlaceholdersStep.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2012-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.android;
+
+import com.facebook.buck.io.ProjectFilesystem;
+import com.facebook.buck.step.ExecutionContext;
+import com.facebook.buck.step.Step;
+import com.facebook.buck.step.StepExecutionResult;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Replaces placeholders in the android manifest.
+ */
+public class ReplaceManifestPlaceholdersStep implements Step {
+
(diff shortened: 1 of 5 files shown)


kageiit:
@marcinkosiba tagging you since you wrote the original `manifest_entries` logic

facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @asp2insp, @michath and @k21 to be potential reviewers.

facebook-github-bot-9:

kageiit:
@dreiss would you be able to take a look at this please?

dreiss:
Do you know if this is handled upstream by the manifest merger?  We've been talking about upgrading it for a while now.

kageiit:
> Do you know if this is handled upstream by the manifest merger? We've been talking about upgrading it for a while now.

The manifest merger does not handle placeholders afaik. we use it in okbuck (https://github.com/OkBuilds/OkBuck/blob/master/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy#L152) but the api has no way to specify the placeholders. These placeholders are filled in after the merging takes place at a separate step.

kageiit:
upgrading the manifest merger is out of scope of this PR. It is a much larger change which I do not have the time to do right now unfortunately :(

dreiss:
Okay, I guess this is fine.  I see a few issues with the diff.

1. Doing this work in getShellCommandInternal is super confusing.  That method should have no side effects.  This should run in a separate step that happens before AaptStep.
1. Overwriting the same file can be confusing for debugging, especially if it is actually the output of another rule.  This new step should create a new AndroidManifest.xml file that is the result of the substitutions.  It can replace the symlink currently generated at the beginning of AaptPackageResources.getBuildSteps.
1. Make sure to call recordArtifact on the file you generate.

facebook-github-bot-8:

src/com/facebook/buck/android/ReplaceManifestPlaceholdersStep.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @Override
+  public String getShortName() {
+    return "replace_manifest_placeholders";
+  }
+
+  @Override
+  public String getDescription(ExecutionContext context) {
+    return getShortName();
```
dreiss: Can you make this look like a shell command with input and output?  like "shortCommand SOURCE_PATH > OUTPUT_PATH"

src/com/facebook/buck/android/ReplaceManifestPlaceholdersStep.java
```diff
@@ -0,0 +1,105 @@
...
+      ImmutableMap<String, String> manifestEntries) {
+    this.projectFilesystem = projectFilesystem;
+    this.androidManifest = androidManifest;
+    this.replacedManifest = replacedManifest;
+    this.manifestEntries = manifestEntries;
+  }
+
+  @Override
+  public StepExecutionResult execute(ExecutionContext context) throws InterruptedException {
+    Optional<String> content = projectFilesystem.readFileIfItExists(androidManifest);
```
dreiss: This should always be present.  @marcinkosiba , do you know the best way to read this file with ProjectFilesystem?

dreiss:
@facebook-github-bot shipit

dreiss:
@facebook-github-bot shipit

PR #756, opened 2016-05-23
Add a test targets test_host_app as a target dependency in projects

rmaz:
Currently a generated Xcode project with a test target with configured test host will correctly set up the bundle loader for the test target, but will not set a target dependency (strangely this is not implicitly added by Xcode). This PR adds the target dependency to the generated project.


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @yiding, @ryu2 and @andrewjcg to be potential reviewers.

rmaz:
Seems like this is causing some issues on a more complex project of ours, need to take another look at this.

PR #763, opened 2016-05-25
Fix the way environment is handled for worker_tool processes.

mikekap:
Previously the worker process was started with environment variables from whatever first rule happened to start it.

This changes the behavior to:
 - allow specifying environment variables via `env` in worker_tool
 - remove step-specific environment variables from the worker launch environment.
 - moves `$CWD` to project root
 - moves `$TMP` to `buck-out/bin/<worker_tool target>_worker`

This way the worker is started the same way every time. It also avoids including the
(potentially huge) environment from the genrule in an actual call to exec.

In terms of code organization, I followed the existing pattern, but I think perhaps
it might make sense to move `WorkerProcess` startup into `WorkerTool` itself so we
don't have to pass so many arguments around.

### DIFF
--- a/src/com/facebook/buck/shell/WorkerShellStep.java
+++ b/src/com/facebook/buck/shell/WorkerShellStep.java
@@ -29,7 +29,6 @@ import com.facebook.buck.util.environment.Platform;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Optional;
-import com.google.common.base.Predicates;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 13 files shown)


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @sam-swarr, @ryu2 and @andrewjcg to be potential reviewers.

facebook-github-bot-2:

mikekap:
ping

src/com/facebook/buck/shell/WorkerShellStep.java
```diff
@@ -43,22 +42,16 @@
...
-  private Path tmpPath;
-  private Path workingDir;
   private Optional<WorkerJobParams> cmdParams;
   private Optional<WorkerJobParams> bashParams;
   private Optional<WorkerJobParams> cmdExeParams;
 
   public WorkerShellStep(
       ProjectFilesystem filesystem,
-      Path tmpPath,
-      Path workingDir,
```
Coneko: You can't remove this, it's not always the root of the filesystem, for `genrule` it's the sandbox directory.

src/com/facebook/buck/shell/WorkerTool.java
```diff
@@ -50,12 +59,12 @@ public String getArgs() {
...
+  }
+
   @Override
   public ImmutableSortedSet<BuildRule> getRuntimeDeps() {
-    return ImmutableSortedSet.<BuildRule>naturalOrder()
-        .add(exe)
-        .addAll(exe.getExecutableCommand().getDeps(getResolver()))
-        .addAll(depsFromStartupArgs)
-        .build();
+    return getDeps();
```
Coneko: Why this change?

Coneko:
@facebook-github-bot shipit

facebook-github-bot-6:

Coneko:
@facebook-github-bot shipit

facebook-github-bot-2:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/131324783940933/int_phab) to review.

PR #764, opened 2016-05-26
Add a test targets test_host_app as a target dependency in projects

robbertvanginkel:
From [#756](https://github.com/facebook/buck/pull/756)

> Currently a generated Xcode project with a test target with configured test host will correctly set up the bundle loader for the test target, but will not set a target dependency (strangely this is not implicitly added by Xcode). This PR adds the target dependency to the generated project.

#756 crashed projcect generation if the the target dependency is in a different project than the project being generated. This PR adds a check for that to make sure the dependency is only added if it is in the same project, otherwise it is skipped (as xcode projects do not seem to support this).


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @yiding, @ryu2 and @andrewjcg to be potential reviewers.

facebook-github-bot-8:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-0:

PR #765, opened 2016-05-31
targetSdkVersion mismatch should be only warning, don't fail building process

Piasy:


facebook-github-bot-6:
By analyzing the blame information on this pull request, we identified @styurin to be a potential reviewer.

Piasy:
@styurin Could this pr be merged?

Coneko:
cc @marcinkosiba, @michath 

I think we didn't want to modify the manifest merger since it comes from upstream, but I could be wrong.

Piasy:
@Coneko 

> since it comes from upstream

I don't understand what is the upstream means.

Gradle build could handle this problem automatically. Hope BUCK could work well on this too.

styurin:
@Piasy upstream means AOSP: https://android.googlesource.com/platform/sdk/+/a35f8af/manifmerger/src/com/android/manifmerger/ManifestMerger.java

We don't want to introduce fixes that change logic in code that is taken from other libraries. The main concern here is maintenance cost of such changes in cases when we need to update our copy of this library with a new version from upstream.

Piasy:
Got it. But android gradle plugin works well in this case, I'll figure out how does it works.

styurin:
Thanks, let us know what you find, we may only need to update ManifestMerger from upstream.

PR #780, opened 2016-06-18
fixed awk syntax issues when $word has slashes

jas14:


alsutton:
@gsaraf or @beefon are better placed to look this over.

LegNeato:
I think this is no longer needed due to https://github.com/facebook/buck/commit/3972723c49ea0eff00b7e00756a5ec6a47c5860e ?

ryu2:
@LegNeato yes, that seems to be the case.

PR #785, opened 2016-06-22
support for groovy rules in intellij project generation

mread:
Not quite right yet but for your consideration.

This pull request supports `groovy_library` and `groovy_test` rules as intellij source folders.

Outstanding issues:
1. what additional unit or integration tests do you want to see?
2. I'm not sure what `addCompiledShadowIfNeeded()` does but I'm not calling it, should I?
3. there's an unnecessary `<exclude>` and a `packagePrefix` being created in the .iml, any suggestions on the best place to look for a fix?

### DIFF
--- a/src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
+++ b/src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
@@ -21,6 +21,8 @@ import com.facebook.buck.android.AndroidLibraryDescription;
 import com.facebook.buck.android.AndroidResourceDescription;
 import com.facebook.buck.android.RobolectricTestDescription;
 import com.facebook.buck.cxx.CxxLibraryDescription;
+import com.facebook.buck.jvm.groovy.GroovyLibraryDescription;
+import com.facebook.buck.jvm.groovy.GroovyTestDescription;
 import com.facebook.buck.jvm.java.JavaLibraryDescription;
 import com.facebook.buck.jvm.java.JavaTestDescription;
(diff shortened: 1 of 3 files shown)


facebook-github-bot-3:
By analyzing the blame information on this pull request, we identified @alsutton, @yiding and @Coneko to be potential reviewers.

mread:
I've just noticed there's a boolean param for `wantsPackagePrefix`, I'll have look to see if that solves problem 2 above..

mread:
Setting package prefix to false removes the package prefix but there's still an unnecessary `<exclude>`.

marcinkosiba:
Thanks for the PR!

> what additional unit or integration tests do you want to see?
The code looks straightforward, I don't think we need any extra testing.

> I'm not sure what addCompiledShadowIfNeeded() does but I'm not calling it, should I?

It will add a .jar file containing the compiled version of the target to your dependencies list if the target uses annotation processors or has inputs from other buildrules. It's used to make the project support annotation processors (turns out not all APs that run with Javac work with IntelliJ) and generated code (sources coming from genrules/src.zip files that are not available before running the build).

> there's an unnecessary <exclude> and a packagePrefix being created in the .iml, any suggestions on the best place to look for a fix?

Looks like you figured out the `packagePrefix`. Not sure about the exclude. An `<exclude-output>` tag is hardcoded. The excludes are calculated by looking at all the files included into the project and marking any folders on disk that weren't included as excluded.

marcinkosiba:
@facebook-github-bot  shipit

sdwilsh:
@marcinkosiba - it looks like the bot wasn't working when you tried to ship this.

PR #799, opened 2016-07-05
Support XCUITest project generation

robbertvanginkel:
This adds partial support for XCUITests, apple's uitesting framework. The `apple_test` rule will get a new optional parameter `is_ui_test`, which should be used in combination with `test_host_app`. An example:

```
apple_test(
  name = 'UITests',
  test_host_app = ':App',
  is_ui_test = True,
)
```

These test bundles will be skipped when with `buck test`, as xctool doesn't support running them (see [xctool#534](https://github.com/facebook/xctool/issues/534)). They will however be available when the project is generated with `buck project //foo:App`.

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -2695,9 +2695,8 @@ public class ProjectGenerator {
             return ProductType.APPLICATION;
         }
       } else if (binaryNode.getType().equals(AppleTestDescription.TYPE)) {
-        @SuppressWarnings("unchecked")
         TargetNode<AppleTestDescription.Arg> testNode =
-              (TargetNode<AppleTestDescription.Arg>) binaryNode;
+            binaryNode.castArg(AppleTestDescription.Arg.class).get();
         if (testNode.getConstructorArg().isUiTest()) {
           return ProductType.UI_TEST;
         } else {
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -1238,7 +1238,17 @@ public class ProjectGenerator {
     extraSettingsBuilder
         .put("TARGET_NAME", buildTargetName)
         .put("SRCROOT", pathRelativizer.outputPathToBuildTargetPath(buildTarget).toString());
-    if (bundleLoaderNode.isPresent() && isFocusedOnTarget) {
+    if (productType == ProductType.UI_TEST && isFocusedOnTarget) {
+      if (bundleLoaderNode.isPresent()) {
+        BuildTarget testTarget = bundleLoaderNode.get().getBuildTarget();
+        extraSettingsBuilder
+            .put("TEST_TARGET_NAME", getXcodeTargetName(testTarget));
+      } else {
+        throw new HumanReadableException(
+            "The test rule '%s' is configured with 'is_ui_test' but has no test_host_app",
+            buildTargetName);
+      }
+    } else if (bundleLoaderNode.isPresent() && isFocusedOnTarget) {
       TargetNode<AppleBundleDescription.Arg> bundleLoader = bundleLoaderNode.get();
       String bundleLoaderProductName = getProductNameForBuildTarget(bundleLoader.getBuildTarget());
       String bundleLoaderBundleName = bundleLoaderProductName + "." +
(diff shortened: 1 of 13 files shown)


src/com/facebook/buck/cli/TestCommand.java
```diff
@@ -605,6 +609,40 @@ public int compare(TestRule o1, TestRule o2) {
...
+        }
+      }
+
+      if (!isAppleUITest) {
+        builder.add(rule);
+      }
+    }
+
+    return builder.build();
+  }
```
beefon: Should you let AppleTest decide if it can run UI tests or not? There is a `useXctest` boolean flag. I wonder, if it is set to true, can it invoke ui tests using `xctest`?

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -2685,7 +2695,14 @@ ProductType bundleToTargetProductType(
             return ProductType.APPLICATION;
         }
       } else if (binaryNode.getType().equals(AppleTestDescription.TYPE)) {
-        return ProductType.UNIT_TEST;
+        @SuppressWarnings("unchecked")
+        TargetNode<AppleTestDescription.Arg> testNode =
+              (TargetNode<AppleTestDescription.Arg>) binaryNode;
+        if (testNode.getConstructorArg().isUiTest()) {
```
robbertvanginkel: I couldn't find a typesafe way to express that binaryNode has `AppleTestDescription.Arg`, but after the typecheck this should be true. Suggestions welcome.

src/com/facebook/buck/cli/TestCommand.java
robbertvanginkel: As far as I know, there is currently no way to run these test from the commandline except for using xcodebuild with a scheme and project. To run these tests you need to use some special testmanager daeamon and setup some more things, I'm think `xctest` is just for running logic tests. facebook/xctool#534 has some investigation into how involved running uitests is but no one seems to have reverse engineered it yet.

Letting `AppleTest` decide this instead of the `TestRunner` doesn't seem like bad idea. Maybe we should extend the `TestRule` interface with a `isSupported`/`testCanRun` flag that is used for filtering?

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
beefon: There is `TargetNode.castArg()` method that is usually used to cast the arg, and it returns `Optional.absent()` if cast is not possible. Should it help you here?

src/com/facebook/buck/cli/TestCommand.java
beefon: Hm, I don't think we should add such kind of flag to `TestRule`, I think it would be better to just not instantiate `TestRule` at all if we don't want to run it.

facebook-github-bot-7:

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: That works, updated!

src/com/facebook/buck/cli/TestCommand.java
robbertvanginkel: I couldn't find a good way to not instantiate the `TestRule`, instead I propose to return an empty list of test steps for `ImmutableList<Step> runTests(...)`, with a dummy tests response that informs the user these tests have not been run.

The output of this is as follows:
```
RESULTS FOR //mock:TestApp
NOTESTS <100ms  0 Passed   0 Skipped   0 Failed   XCUITest runs not supported with buck test
```

beefon:
Nice! Thanks a lot for implementing this!

beefon:
One last bit: do you think it is possible to add a test into `AppleTestIntegrationTest.java` so you will also cover `buck test ..` invocation as well? I'm interested in target that will have both ui tests and unit tests, making sure UI tests are just skipped, while unit tests still work (and passed/failed). I think `successOnTestPassing` may be easily copy-pasted (including test data folder) and adopted.

facebook-github-bot-7:

test/com/facebook/buck/apple/testdata/apple_test_xcuitest/FooXCUITest.m
```diff
@@ -0,0 +1,8 @@
+#import <XCTest/XCTest.h>
+
+@interface FooXCTest : XCTestCase
```
beefon: I think class name should be `FooXCUITest` here.

test/com/facebook/buck/apple/testdata/apple_test_xcuitest/FooXCUITest.m
```diff
@@ -0,0 +1,8 @@
...
+@end
+
+@implementation FooXCTest
+
```
beefon: Currently there are no tests here. And Buck prints out `NOTESTS`. Which is right-ish. I think what you want is to have some tests here, and then test that Buck actually skips them and still outputs `NOTESTS` for this file.

beefon:
Sorry for continuous ping-pong on this PR, I just want to make sure test suite for your new feature is around. :)

facebook-github-bot-7:

robbertvanginkel:
You're right my test wasn't sound. Don't mind the pingponging with this fast of a turnaround.

beefon:
@facebook-github-bot shipit

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1752637704996600/int_phab) to review.

Issue #803, opened 2016-07-10
Cache size grows with no bounds for large repos


rmaz:
The dir artifact cache is set to clean up either when 2/3 of the cache size has been written to per buck invocation, or as a cleanup operation at the end of each buck invocation. The latter operation is given a 10 second timeout. When the cache size is set sufficiently large, say 10GB, the first condition is unlikely to be met. When the build targets are of a certain size, every build will generate more files than can be indexed and sorted before 10 seconds have elapsed. This results in the cache always growing, as the delete files operation will never have time to complete its cache indexing and start actually deleting files.

The quick solution here is to increase the diskioexecutor timeout to something greater than 10 seconds and to make this parameter customisable via config settings or runtime flags, but open to better suggestions here? Does it ever make sense to kill the cleanup operation at the end of each invocation, should this always be run under certain conditions, say > x files cached?

rmaz:
It looks like buckd is the magic making this work, but we cannot run this on our CI architecture currently.

styurin:
Fixed in 3d611c94df024796d7aa0d7785b663006f9a604d

PR #805, opened 2016-07-11
Disk IO timeout property

rmaz:
Ref #803. Make it possible to specify a custom disk IO executor timeout as a java property. This makes it possible to clean the cache on CI jobs while a better solution is found to fix the issue with the cache growing without bounds.

### DIFF
--- a/src/com/facebook/buck/artifact_cache/DirArtifactCache.java
+++ b/src/com/facebook/buck/artifact_cache/DirArtifactCache.java
@@ -335,6 +335,9 @@ public class DirArtifactCache implements ArtifactCache {
         maxTrimMark = Optional.of(i);
       }
       if (currentSizeBytes > maxSizeBytes) {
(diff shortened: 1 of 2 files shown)


facebook-github-bot-7:
By analyzing the blame information on this pull request, we identified @beefon, @yiding and @ruibm to be potential reviewers.

sdwilsh:
Wouldn't this be better set in the `.buckconfig`?

marcinkosiba:
please make the timeout a setting on .buckconfig

rmaz:
I'm going to close this one for now, we have got buckd working on our CI machines. I am not sure if this setting makes sense in .buckconfig though, it seems like something that shouldn't be customisable. Maybe when buckd is not enabled there should be no timeout? Or at least the task should run until at least some %age of the cache has been cleaned?

PR #810, opened 2016-07-18
Rudimentary kotlin support.

dsyang:
Right now this change doesn't to much more than enable `kotlin_library()` and `kotlin_test()` build rules. You will still need to include the kotlin runtime as a separate prebuilt_jar() build rule or else your kotlin targets will crash and burn.

There isn't a `kotlin_binary()` target yet. Use `java_binary()` instead and remember to add the suffix 'Kt' when specifying the main class (see https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html)

To use this you will need to:

- Have `kotlinc` in your path or specify it in your .buckconfig 
````
[kotlin]
   compiler = ...
````
- Copy the kotlin-runtime.jar into your repo (intellij can do this for you when setting up kotlin) and create a `prebuilt_jar()` build rule for it. Include this target as a dependency under your `kotlin_library()` target.

### CODE REVIEW

**Comment by Coneko** [comment]
@mikekap, @grumpyjames: Could you take a look at this?

File: src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+  private final JavaOptions javaOptions;
+  private final JavacOptions defaultJavacOptions;
+  private final Optional<Long> defaultTestRuleTimeoutMs;
+  private final Optional<Path> testTempDirOverride;
+
+  public KotlinTestDescription(
+      KotlinBuckConfig kotlinBuckConfig,
+      JavaOptions javaOptions,
+      JavacOptions defaultJavacOptions,
+      Optional<Long> defaultTestRuleTimeoutMs, Optional<Path> testTempDirOverride) {
```
grumpyjames: be nice to keep the aligning here

grumpyjames:
Disclaimer: I haven't looked at this code for a fair old while

But: A quick look yields no obvious problems (that one piece of alignment pedantry aside (sorry)): nice work :-)

More general comments about this area:

It would be good to revisit the various LibraryDescription classes to see if there is something common to be extracted there. We've always deferred this in the past to see what other implementations turned up - perhaps now might be a time to reassess. I suppose it depends if we expect any more jvm langs to turn up, and whether _they_ will be similar. Similar motivation might be found if desirable features from JavaLibraryDescription were not being hooked into by groovy, scala, kotlin &c without manual effort (i.e someone copies them across).

There are probably plenty of things in the java package that could probably be pushed into somewhere agnostic, too (DefaultJavaLibrary caught my eye in this commit - that's really DefaultJvmLibrary now, right?); that's something I meant to get to but never did.

bolinfest:
Also cc @bkase for feedback.

bolinfest:
I think it's most important to get feedback on the following [from people who are more familiar with Kotlin]:

* The API for the build rules.
* Whether the implementation of the build rules is efficient.

For style nits and other low-level things, we're pretty good at catching those as part of code review already.

src/com/facebook/buck/rules/KnownBuildRuleTypes.java
```diff
@@ -670,6 +673,16 @@ static Builder createBuilder(
             defaultCxxPlatform,
             testTempDirOverride));
     builder.register(new KeystoreDescription());
+    KotlinBuckConfig kotlinBuckConfig = new KotlinBuckConfig(config);
```
mikekap: I think these are usually declared somewhere above (this "block" is just full of `builder.register`)

mikekap:
@dsyang If you'd like, it would be pretty easy to add the kotlin library as an implicit target. Scala does that already - you would need to mimic https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/scala/ScalaLibraryDescription.java#L82 & https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/scala/ScalaLibraryDescription.java#L144 for kotlin. Otherwise, LGTM.

In terms of (internal) API, the buck JVM rules feel a bit clunky, but that's not the fault of this CR. Particularly, `DefaultJavaLibrary` is a behemoth that has a mix of java specific (e.g. `SuggestBuildRules`) and non-java-specific code & 18 constructor parameters. I think the best refactoring there might be to make it less graph-like - i.e. stop doing transitive closures in the rule itself and defer those to the descriptions. The hope is to make `DefaultJavaLibrary` really dumb - it should really just shell out to javac/scalac/kotlinc/etc. But that change that's outside the scope of this PR.

bkase:
Hi -- Android Kotlin user and Buck fan here. I'll try to invest some time getting a (hopefully) non-trivial kotlin Android example to work with the new rules as soon as I can (which unfortunately is likely to be next week since I'm about to go on vacation)

facebook-github-bot-0:
@dsyang updated the pull request.

facebook-github-bot-0:
@dsyang updated the pull request.

dsyang:
@mikekap Thanks for that tip about how scala implicitly adds the dependency! I think I'll add it as a separate PR though.

dsyang:
n00b questions: How do I...
1) See why AppVeyor couldn't build this (though I think it's probably an autodeps confilct)
2) Update this PR to the current master without adding a ton of commits to the PR

bkase:
Hit a snag when trying to build a Kotlin `MainActivity` in an Android app since there is no `android_library` equivalent kotlin Buck rule.

See https://github.com/bkase/buckitup

Coneko:
The Appveyor failure is a path limit issue, I've got some internal patches to address those but it's all a mess because the windows path limit is really really too short.

You can merge master into this PR, when we import it we'll rebase it anyway.

bolinfest:
@dsyang Can you take this over the finish line?

dsyang:
Yea I should rebase this to make sure the checks pass. 

re: @bkase and the android issues, I think that belongs in a separate PR. I believe this is enough to start using Kotlin along side other plain-old java programs (like buck!) I can create an issue with "next steps" if that's useful.

bkase:
@dsyang makes sense, ship it!

tiestvilee:
@dsyang this is awesome! Let's get it merged.

One thing you might want to look into once it's merged (or I could, but let's wait for the merge) is being a bit more buck like with dependencies. I don't believe you are following them to the degree they should be.

After using the java stuff I would expect the following to work

```python
  kotlin_library(
    name = "app-kotlin",
    srcs = glob(["src/main/**/*.kt"]),
    resources = glob(["src/main/**"], excludes = ["src/main/**/*.java", "src/main/**/*.kt"]),
    # resources_root = "src/main",
    deps = ["//:kotlin-runtime"],
  )

  kotlin_test(
    name = "app-kotlin-test",
    srcs = glob(["src/test/**/*.kt"]),
    resources = glob(["src/test/**"], excludes = ["src/test/**/*.java", "src/test/**/*.kt"]),
    deps = ["//:test-deps", ":app-kotlin"],
  )
```
where `test-deps` is a java_library that includes junit and so forth.

The above compiles in buck, but the tests fail because they can't see junit (a sub-dependency of `test-deps`). If I fix that then it fails because it can't see `kotlin-runtime` - which it should have inherited as a dependency from kotlin_library (would in java_test).

This means i have to define lots of dependencies again and again, which sux.

(also, resources_root doesn't work)

But for now - get this into Buck, it's a godsend!

facebook-github-bot-3:
@dsyang updated the pull request - [view changes](https://github.com/facebook/buck/pull/810/files/43226f40dcda31c4d9be29a680526ce7170b4bb8..850e3370b36bec1b8aafc0a18ffc0678a8513b4a)

dsyang:
@tiestvilee  thanks for trying this out! 

I wonder if that could be fixed by implicitly adding the kotlin-runtime / kotlin-test dependencies like how @mikekap suggested. Definitely something we should look at after this gets in.

dsyang:
I tried adding kotlin to buck as an example of how this would work but that turned out to be too hairy.
Instead I created this extremely simple cli calculator to verify this PR actually works: https://github.com/dsyang/buckotlin

You can see the buck integration is pretty straight forward: https://github.com/dsyang/buckotlin/commit/7326113438366da3e9851ce30dfe2dcd3e89e88b

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
```diff
@@ -0,0 +1,129 @@
...
+                Optional.<SourcePath>absent(),
+                ImmutableList.<String>of(),
+                exportedDeps,
+                resolver.getAllRules(args.providedDeps.get()),
+                new BuildTargetSourcePath(abiJarTarget),
+                /* trackClassUsage */ false,
+                /* additionalClasspathEntries */ ImmutableSet.<Path>of(),
+                new KotlincToJarStepFactory(
+                    kotlinBuckConfig.getKotlinCompiler().get(),
+                    args.extraKotlincArguments.or(ImmutableList.<String>of())),
```
Coneko: No need for the `or` here, the coercer will map an absent to an empty list: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/rules/coercer/ListTypeCoercer.java#L39 .

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
```diff
@@ -0,0 +1,160 @@
...
+    SourcePathResolver pathResolver = new SourcePathResolver(resolver);
+
+    BuildTarget abiJarTarget =
+        BuildTarget.builder(params.getBuildTarget())
+            .addFlavors(CalculateAbi.FLAVOR)
+            .build();
+
+    KotlincToJarStepFactory stepFactory = new KotlincToJarStepFactory(
+        kotlinBuckConfig.getKotlinCompiler().get(),
+        args.extraKotlincArguments.or(ImmutableList.<String>of()));
```
Coneko: Same as above, no need for the `or`.

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
```diff
@@ -0,0 +1,129 @@
...
+      A args) throws NoSuchBuildTargetException {
+    SourcePathResolver pathResolver = new SourcePathResolver(resolver);
+
+    BuildTarget abiJarTarget = params.getBuildTarget().withAppendedFlavors(CalculateAbi.FLAVOR);
+
+    ImmutableSortedSet<BuildRule> exportedDeps = resolver.getAllRules(args.exportedDeps.get());
+    DefaultJavaLibrary defaultJavaLibrary =
+        resolver.addToIndex(
+            new DefaultJavaLibrary(
+                params.appendExtraDeps(
```
Coneko: You're missing `pathResolver.filterBuildRuleInputs(javacOptions.getInputs(pathResolver))` like in https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/JavaLibraryDescription.java#L148 here.

Coneko:
Just a couple of pointers, I'll import it and run our internal tests in the meantime.

Coneko:
@facebook-github-bot shipit

facebook-github-bot-3:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3751191) to review internal test results.

Coneko:
It's failing because of missing dependencies, please make sure you run `buck test` locally.

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
dsyang: I'm not using the coercer here. Are you suggesting that I should use it?

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
dsyang: That looks specific to the java library description (the kotlin library description doesn't ave javacOptions) so I don't think we can add this here.

Unless you're suggesting we add javac options to this.

facebook-github-bot-0:
@dsyang updated the pull request - [view changes](https://github.com/facebook/buck/pull/810/files/850e3370b36bec1b8aafc0a18ffc0678a8513b4a..dd44dd4515f03d63f3e119b25466a630bc731e0d) - [changes since last import](https://github.com/facebook/buck/pull/810/files/850e3370b36bec1b8aafc0a18ffc0678a8513b4a..dd44dd4515f03d63f3e119b25466a630bc731e0d)

dsyang:
@Coneko  I ran `buck test` locally and the only test failures I found were related to the swift integration tests (I think that's because I have an old/incomplete swift installation). Maybe it's something broken with the internal tests?

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
Coneko: No, the coercer is used by the caller of `createBuildRule`, so the `args` object here will have already gone through it.

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
Coneko: Oh I see, `KotlinTestDescription` has that as well so I thought it was needed here.

Coneko:
You're right, the failures seem completely unrelated to your changes, it could be because it got patched on a broken commit. I'll check again on the next import after you address the comments.

facebook-github-bot-1:
@dsyang updated the pull request - [view changes](https://github.com/facebook/buck/pull/810/files/dd44dd4515f03d63f3e119b25466a630bc731e0d..19eee34d422ad43432d8993f96fa6e7e2a485c74) - [changes since last import](https://github.com/facebook/buck/pull/810/files/850e3370b36bec1b8aafc0a18ffc0678a8513b4a..19eee34d422ad43432d8993f96fa6e7e2a485c74)

dsyang:
Can I import my own code? :)

@facebook-github-bot shipit

Coneko:
@facebook-github-bot shipit

facebook-github-bot-1:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3751191) to review internal test results.

Issue #795, opened 2016-06-27
Invalid classes root with IntelliJ


vinz243:
Hi everyone,

I'm trying to add several libraries, such as fresco. Here is the code that adds fresco:
##### `maven/BUCK` 
```py
### Package com.facebook.fresco:fresco
remote_file(
    name =  'com_facebook_fresco__fresco-remote',
    out = 'com_facebook_fresco__fresco-remote.aar',
    sha1 = '86df1ab4b0074e1aeceb419593d2ea6d97cdc3b4',
    url = 'http://repo1.maven.org/maven2/com/facebook/fresco/fresco/0.11.0/fresco-0.11.0.aar'
)

android_prebuilt_aar(
    name = 'com_facebook_fresco__fresco',
    aar = ':com_facebook_fresco__fresco-remote',
    deps = [
        '//maven:com_facebook_fresco__imagepipeline',
        '//maven:com_facebook_fresco__drawee',
        '//maven:com_facebook_fresco__imagepipeline-base',
        '//maven:com_parse_bolts__bolts-tasks',
        '//maven:com_android_support__support-v4',
        '//maven:com_nineoldandroids__library',
        '//maven:com_facebook_fresco__fbcore',
    ],
    visibility = ['PUBLIC']
)
```

Here is for example a file generated by `buck project --ide INTELLIJ`:
##### `.idea/libraries/library_maven_com_facebook_fresco__fresco.xml`

```xml
<component name="libraryTable">
  <library name="library_maven_com_facebook_fresco__fresco">
    <CLASSES>
      <root url="jar://$PROJECT_DIR$/buck-out/gen/maven/com_facebook_fresco__fresco-remote/com_facebook_fresco__fresco-remote.aar!/" />
    </CLASSES>
    <JAVADOC />
    <SOURCES />
  </library>
</component>
```

The aar file exists in said dir.

But I get ` invalid classes root` 

[![alt](http://i.imgur.com/r57yepHm.png)](http://i.imgur.com/r57yepH.png)

Any idea?

Thanks

marcinkosiba:
Thanks! It looks like the new project generator has problems with AARs. For now you can use the old code by running `buck project --deprecated-ij-generation`

vinz243:
Thanks! It works like a charm :)

tonycosentini:
Would it be possible for someone to take a look at this issue?

While the old project generation also works, it seems like there are quite few advantages to the new project generation system, but any Android app won't be able to use it because of this issue.

PR #811, opened 2016-07-20
Unzip AARs and use their classes.jar for Intelilj project generation.

tonycosentini:
(Fixes https://github.com/facebook/buck/issues/795)

This adds back support for unzipping AARs so IntelliJ can reference their classes.jar file. Although this fix works, it's definitely far from ideal.

Open Questions
* Should this have its own pipeline for unzipping AARs? This also happens as part of AAR build rules, but it seems like it would be not so clean and require significant refactoring to make project generation depend on that. In addition, the deprecated intellij.py file also seems to explode the AARs itself.
* Where should I put the unzipped AAR class.jar files? Right now I’m just dumping them in the same directory, but there is definitely a more appropriate place to put them.

Once the general implementation is done, Il’l be sure to add tests as well.


facebook-github-bot-9:
By analyzing the blame information on this pull request, we identified @yiding to be a potential reviewer.

tonycosentini:
Re build falure - it seems like Ant can reference the Unzip class, but not Buck. I'll try to figure it out.

tonycosentini:
@dreiss would you be a good person to review this? (heard you might know about how to approach this)

marcinkosiba:
this approach isn't going to be reliable as the files may not be on disk at the point where your code runs. The UnzipAar rule already does this. What you need to do instead is find the `AndroidPrebuiltAar` BuildRule associated with the AAR, `getPrebuiltJar()` method to figure out where the .jar files are coming from and add the BuildRule to the set of things that need to be built (see getPathIfJavaLibrary in IjProject.java).
Essentially replace the call to `libraryFactoryResolver.getPath` with a new API that does the right thing for AAR files.

facebook-github-bot-0:
@tonycosentini updated the pull request.

src/com/facebook/buck/jvm/java/intellij/DefaultIjLibraryFactory.java
```diff
@@ -165,8 +165,8 @@ public BuildRuleType getType() {
     @Override
     public void apply(
         TargetNode<AndroidPrebuiltAarDescription.Arg> targetNode, IjLibrary.Builder library) {
-      AndroidPrebuiltAarDescription.Arg arg = targetNode.getConstructorArg();
-      library.setBinaryJar(libraryFactoryResolver.getPath(arg.aar));
+      Optional<Path> jarPath = libraryFactoryResolver.getPathIfJavaLibrary(targetNode);
+      library.setBinaryJar(jarPath.get());
```
tonycosentini: This `.get()` will crash if a jar is not present, but that should be guaranteed by the code added in `IjProject`. Let me know if there is a better way to handle this.

facebook-github-bot-0:
@tonycosentini updated the pull request.

facebook-github-bot-4:
@tonycosentini updated the pull request.

facebook-github-bot-0:
@tonycosentini updated the pull request.

tonycosentini:
@marcinkosiba got it working with your approach (I think). Huge thanks for the review and feedback, makes much more sense now.

facebook-github-bot-0:
@tonycosentini updated the pull request.

marcinkosiba:
@facebook-github-bot shipit

facebook-github-bot-6:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1090648877696061/int_phab) to review internal test results.

PR #812, opened 2016-07-22
iOS: setAppIncludesFrameworks to true

robbertvanginkel:
I want to include dynamic frameworks in my build buck app. The work is coming along and I hope to open a PR soon. On iOS these frameworks are available from iOS8+, is there a way to set the `setAppIncludesFrameworks` value conditionally based on specified  [target_sdk_version](https://buckbuild.com/concept/buckconfig.html#apple.target_sdk_version)?


facebook-github-bot-9:
By analyzing the blame information on this pull request, we identified @jernejstrasner, @Coneko and @ryu2 to be potential reviewers.

Coneko:
Sorry I think the property is badly named, it refers to whether applications on that platform will have frameworks or not by default. You can have applications both with or without frameworks regardless of the value of this property by using the right flavours.

robbertvanginkel:
Thanks for the feedback. Don't think the defaults need to be changed then.

Issue #440, opened 2015-09-30
iOS Dynamic Frameworks


crylico:
Does buck support pre-built dynamic frameworks for iOS? After some digging through the docs, it looks like it only supports static libs

sdwilsh:
It does not currently.

crylico:
gotcha! so is prebuilt_cxx_library() currently the only way to get prebuilt code into a project?

sdwilsh:
That's correct.  If there's some specific thing you are trying to do, we might be able to tell you about a better way (or how to change Buck to do what you need), but that's the generic hammer :)

crylico:
thanks for the feedback :) basically just trying to use [Carthage](https://github.com/carthage/carthage) generated frameworks with a buck project.

I understand it's a work in progress and I really like the ideas behind buck - looking forward to seeing it grow!

Coneko:
I'm sure @jspahrsummers would love to add Buck support to Carthage.

petersibley:
Are there any plans to support dynamic frameworks as a target type?

Coneko:
You can already create dynamic frameworks by using `apple_bundle` and setting its `binary` attribute to refer to a `apple_library` with the `#shared` flavour.

We are planning to make the interface nicer though.

petersibley:
O cool. You might consider updating the docs to reflect this. It wasn't clear to me that you could do that. Thanks.

crylico:
Oh neat! What about embedding a dynamic framework that's already been
compiled?

Coneko:
`prebuilt_cxx_library` helps you a bit there, but in practice it's only supported in the cxx workflow, not in the iOS workflow, so project generation and so on won't work.

Most likely not what you're looking for there, sorry.

mindows:
Any updates on this? How would I add something like facebook sdk to my app? generated xcode project that depends on the framework would be pretty broken without the framework + headers included properly, right?

Coneko:
There is a hacky workaround we use, I mentioned it here: https://github.com/facebook/buck/issues/660#issuecomment-202884912

aledalgrande:
Looks like there was a development here, but it was never documented. https://github.com/facebook/buck/pull/1263 adds docs.

PR #818, opened 2016-07-26
[Apple] Add prebuilt dynamic framework support

robbertvanginkel:
This PR proposes a way to embed prebuilt dynamic frameworks into `apple_bundle`s, linking that framework against the binary for the apple bundle. This is a start to solving #440, in buck it (mostly) works, the project generation still has to be done. The rule syntax for this would be:
```
prebuilt_apple_framework(
    name = 'BuckTest',
    framework = 'BuckTest.framework',
)
```

You can add this rule to an apple_library or apple_binary, this should do the following:
- Symlink the `Headers` folder inside the framework as the exposed headers for the rule.
- Export linkerflag `-rpath @loader_path/Frameworks`/ `-rpath @executable_path/Frameworks`
- Export the linkerflag `-framework <frameworkname>`, and `-F path/to/framework/`
- Embed the framework in the bundle's "Frameworks" folder

I'm however encountering the following problem when trying to embed the framework: the `INCLUDE_FRAMEWORKS` flavor is tied to `apple_library`'s `FRAMEWORK` flavor, and I don't want that. This happens in [AppleDescriptions.java](https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/AppleDescriptions.java#L516). I have the following situation: an apple_binary (wrapped in apple_bundle) with two dependencies: a `prebuilt_apple_framework` and an `apple_libary`. I want to statically link the `apple_library`, and embed and dynamically link the `prebuilt_apple_framework`. The probelm is, when I build the `apple_binary` with `#INCLUDE_FRAMEWORKS`, the mentioned code in AppleDescriptions goes through the deps and requires them with the `FRAMEWORK_FLAVOR`, which tries to embed the apple_library as a framework which is broken.

I sort of avoided this by making apple_binary not require the framework flavor, but this is inconsistent and will cause headaches. Suggestions on how to move forward with this?


facebook-github-bot-0:
By analyzing the blame information on this pull request, we identified @Coneko, @beefon and @yiding to be potential reviewers.

ryu2:
@robbertvanginkel We should just propagate the framework flavor if it's a `apple_bundle` (with apple_library) or `prebuilt_apple_framework`.

robbertvanginkel:
Are we in any way concerned that that will break the current undocumented way the framework flavor works?

ryu2:
We currently aren't using the `#framework` flavor; instead we're using `apple_bundle` with a `apple_library` + `#shared` rule (which does the same thing).

src/com/facebook/buck/apple/AbstractApplePlatform.java
```diff
@@ -35,6 +35,7 @@
           .setSwiftName("ios")
           .setArchitectures(ImmutableList.of("armv7", "arm64"))
           .setMinVersionFlagPrefix("-mios-version-min=")
+          .setAppIncludesFrameworks(true)
```
Coneko: As noted in the other PR this shouldn't be changed.

src/com/facebook/buck/apple/AppleBundleDescription.java
```diff
@@ -131,7 +131,7 @@ public boolean hasFlavors(final ImmutableSet<Flavor> flavors) {
     if (!AppleDescriptions.INCLUDE_FRAMEWORKS.getValue(params.getBuildTarget()).isPresent()) {
       return (AppleBundle) resolver.requireRule(
           params.getBuildTarget().withAppendedFlavors(
-              AppleDescriptions.NO_INCLUDE_FRAMEWORKS_FLAVOR));
+              AppleDescriptions.INCLUDE_FRAMEWORKS_FLAVOR));
```
Coneko: This also changes the default behaviour.

Coneko:
Agree with @ryu2.

robbertvanginkel:
Thanks for the response. This does make sense semantically, but it also takes a different direction than some previous commits so to make sure we're aligned please verify the following:

- Never make `apple_library output a framework bundle. Remove `apple_library`'s reaction to `#framework` flavor (as opposed to 8753c2a962492f05c0da17a69a1ba9cacd7eb47c, by @Coneko).
- Do frameworks through `apple_bundles` with apple_library as binary
    - There's a recent commit (ce2322644aecddb1fc54a5ead0f506496c272e54) that explicitly doens't propagate the framework flavor through bundles, but it isn't really clear why.
    - Do we still need a framework flavor? (We can identify a framework through `apple_bundle`'s extension)
    - How would you suggest we expose headers and other build settings with this appoach? From what I could find apple_bundle doesn't pass any cxx flags around. 

Say we have the follwing:
```
apple_binary(
    name = 'BinaryA',
    deps = [':FrameworkB']
)

apple_bundle(
    name = 'FrameworkB',
    binary = ':LibraryB#shared',
)

apple_library(
    name = 'LibraryB',
    headers = ['header.h']
)
```

Now `BinaryA` has `#import <LibraryB/header.h>` somehwere. With the `LibraryB#framework` approach this would be visible as the apple_binary directly depended on the `apple_library`, with the suggested `apple_bundle`+#shared it is not. 

How do you handle this? Conforming to the  `CxxPreprocessorDep` protocol if we're a bundle with framework seems like a semantical stretch. Am I overlooking something obvious here? Xcode does this an option `-ivfsoverlay`, but it seems like buck doens't have any support for that yet.

src/com/facebook/buck/apple/AppleBundleDescription.java
robbertvanginkel: I'll revert this.

src/com/facebook/buck/apple/AbstractApplePlatform.java
robbertvanginkel: I'll revert this.

robbertvanginkel:
Seem like I was not a 100% right in concluding that Xcode uses `-ivfsoverlay`, it seems to widely overspecify headers to include in build rules that depend on frameworks. There are also `-I` and `-F` flags that seem to be used for this.

AppleBundle.java already has a `getCxxPreprocessorInput`, to propagate its headers to a test target that depends on a bundle. Would it be a good approach to propagate the binary's headers there if the bundle is a framework target?

facebook-github-bot-2:

robbertvanginkel:
I've updated the PR to do the following:

- In cxx preprocess deps gathering step, go through the frameworks and see if any is a buildrule. If so add that rule as a preprocessing dep step.
- Simplify the `prebuilt_apple_framework` build rule to copy the framework to a gen folder, and pass the framework search paths option`-F` to clang on linking and as an preprocessor dep.

Lets keep this PR restricted to just the prebuilt framework, and I'll incorporate the feedback for `apple_library`'s with #framework flavor in a future one.

ryu2:
@robbertvanginkel  thanks.  I think `getCxxPreprocessorInput` is the right approach for now. 

We might want to call this `prebuilt_dynamic_apple_framework` to distinguish it from a static framework (e.g. basically a `.a` in a framework bundle)

>  There's a recent commit (ce23226) that explicitly doens't propagate the framework flavor through bundles, but it isn't really clear why.

Before, #no-include-frameworks flavor was propagated down to CxxPreprocessAndCompile rules which could potentially result in cache misses.

For example: Use simple_application_bundle_dwarf_and_dsym to build

buck build //:DemoApp#dwarf

Have a look at gen folder, there you will see

DemoAppBinary#compile-main.m.od32f9e30,iphonesimulator-x86_64,no-include-frameworks
DemoAppBinary#compile-AppDelegate.m.oc53fc391,iphonesimulator-x86_64,no-include-frameworks

no-include-frameworks should not be part of these paths. It affects only the layout of the final bundle, and not the compilation rule outputs.

test/com/facebook/buck/apple/testdata/prebuilt_apple_framework_links/app/Info.plist
```diff
@@ -0,0 +1,43 @@
...
+	<string>1.0</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+	<key>LSRequiresIPhoneOS</key>
+	<true/>
+	<key>UIRequiredDeviceCapabilities</key>
+	<array>
+		<string>armv7</string>
```
ryu2: Do we need this for a simulator `x86_64` build?

test/com/facebook/buck/apple/testdata/prebuilt_apple_framework_links/app/main.m
```diff
@@ -0,0 +1,18 @@
+//
```
ryu2: Please use the standard copyright header.

src/com/facebook/buck/apple/PrebuiltAppleFrameworkDescription.java
```diff
@@ -0,0 +1,93 @@
...
+  public <A extends Arg, U> Optional<U> createMetadata(
+      BuildTarget buildTarget,
+      BuildRuleResolver resolver,
+      A args,
+      Class<U> metadataClass) throws NoSuchBuildTargetException {
+    if (metadataClass.isAssignableFrom(FrameworkDependencies.class)) {
+      ImmutableSet.Builder<SourcePath> sourcePaths = ImmutableSet.builder();
+      resolver.requireRule(buildTarget);
+      sourcePaths.add(new BuildTargetSourcePath(buildTarget));
+      return Optional.of(metadataClass.cast(FrameworkDependencies.of(sourcePaths.build())));
```
ryu2: Easier to just use `ImmutableSet.of()` directly here for a singleton set.

src/com/facebook/buck/apple/PrebuiltAppleFramework.java
```diff
@@ -0,0 +1,279 @@
...
+    this.params = params;
+    this.frameworkPath = frameworkPath;
+    this.ruleResolver = ruleResolver;
+    this.exportedLinkerFlags = exportedLinkerFlags;
+    this.frameworks = frameworks;
+    this.supportedPlatformsRegex = supportedPlatformsRegex;
+
+    BuildTarget target = params.getBuildTarget();
+    this.frameworkName = pathResolver.getAbsolutePath(frameworkPath).getFileName().toString();
+    this.out =
```
ryu2: Easier to say 

`BuildTargets.getGenPath( getProjectFilesystem(), target, "%s")`

test/com/facebook/buck/apple/testdata/prebuilt_apple_framework_links/app/Info.plist
robbertvanginkel: The `apple_bundle` rule needs an info.plist, I just copied this over from another test fixture. I'll remove some of the keys so the file is generic.

facebook-github-bot-6:

robbertvanginkel:
> We might want to call this prebuilt_dynamic_apple_framework to distinguish it from a static framework (e.g. basically a .a in a framework bundle)

You can also link static frameworks with `-framework`, you'd just have to copy the resources into the main apple bundle, not copy the framework to a Frameworks folder and don't pass the rpath flags. I think we should keep it a `prebuilt_apple_framework` and clarify currently only supports dynamic frameworks until someone extends it for bundles with static frameworks.

Coneko:
`apple_library`'s framework flavours should be unrelated since this handles the case of prebuilt frameworks.

The alternative here would be to add framework support to `prebuilt_cxx_library`, or add a `prebuilt_apple_library`.

ryu2:
As we discussed earlier today, the two ways of specifying frameworks (bundle + extension vs. library with #framework flavor) are suited for different things (Xcode project generation vs. build with buck)

To resolve this, maybe we should make new rules (this + `apple_framework`) as the canonical way to specify frameworks, and transform them differently depending on whether we're building or generating a project.

If we want to build libraries for iOS 7 and frameworks for iOS 8+, then best to have a macro in the BUCK file that splits a rule into two (`apple_library` + `apple_framework`).

robbertvanginkel:
> `apple_library`'s framework flavours should be unrelated since this handles the case of prebuilt frameworks.

It should, but it is currently not because of [AppleDescriptions.java#L505](https://github.com/facebook/buck/blob/2910f2ecf78ca93d660114e7af90ab4639b7e067/src/com/facebook/buck/apple/AppleDescriptions.java#L505). That is, if you say your apple_bundle includes frameworks (like we want to to include these prebuilt frameworks), it automatically finds metadata from its dependencies with the `#framework` flavor. So if the apple_bundle depends on an `apple_library` that library is built and included with the broken framework flavor. That's a potential issue that will come up if you depend on both. 

>To resolve this, maybe we should make new rules (this + apple_framework) as the canonical way to specify frameworks, and transform them differently depending on whether we're building or generating a project.

> If we want to build libraries for iOS 7 and frameworks for iOS 8+, then best to have a macro in the BUCK file that splits a rule into two (apple_library + apple_framework).

I agree that a separate apple_framework seems like the way to go.

ryu2:
Thanks!  Looks good to go for now.

There are a few trivial tweaks but I can make them on my end, and I'll merge this manually.

ryu2:
Closed in af4a756e8a982781783f8e643184b25901bf0899

test/com/facebook/buck/apple/testdata/prebuilt_apple_framework_links/app/main.m
```diff
@@ -0,0 +1,10 @@
+#import <BuckTest/BuckTest.h>
```
ryu2: @robbertvanginkel Just out of curiousity, I'm wondering if @import works, since you already have the `.modulemap` file.  That would be a good thing to support if it doesn't already.

robbertvanginkel: Yes, that should work!

fkorotkov:
@robbertvanginkel @ryu2 what do you think about changing type of `framework` from `SourcePath` to `BuildRule`. That way it will be possible to create these prebuilt frameworks from `remote_file`s and `gen_rule`s?

ryu2:
A `SourcePath` can be a build rule, although I haven't tried it in the context of prebuilt frameworks. 

e.g. if you specify `//foo:bar` to any input that takes a SourcePath, it's shorthand for the output of that rule.

fkorotkov:
Nice! Didn't know about that!

src/com/facebook/buck/apple/AppleBinaryDescription.java
```diff
@@ -396,13 +398,36 @@ private boolean shouldUseStubBinary(BuildRuleParams params) {
...
+      AppleDescriptions.populateCxxBinaryDescriptionArg(
+          new SourcePathResolver(resolver),
+          delegateArg,
+          args,
+          buildTarget);
+      return delegate.createMetadata(buildTarget, resolver, delegateArg, metadataClass);
+    }
+
+    Optional<Flavor> cxxPlatformFlavor = delegate.getCxxPlatforms().getFlavor(buildTarget);
+    Preconditions.checkState(
```
yiding: If a bundle depends on another bundle (for instance, an app bundle with an app extension bundle as dependecy), this method is reached via `createAppleBundle() -> AppleBundle.createMetadata() -> AppleBinary.createMetadata()`, the platform flavor may not be set on this build target because a user doesn't usually annotate a bundle's binary parameter with a pre-existing platform.

This also doesn't take into account of multi-architecture binaries.

src/com/facebook/buck/apple/AppleBinaryDescription.java
```diff
@@ -396,13 +398,36 @@ private boolean shouldUseStubBinary(BuildRuleParams params) {
...
+        cxxPlatformFlavor.isPresent(),
+        "Could not find cxx platform in:\n%s",
```
yiding: This error message should just print the full build target. I encountered a case where if the build target has no flavors, it just prints a message "Could not find cxx platform in:" which is entirely unhelpful.

PR #820, opened 2016-07-27
Always use relative paths for Android manifest entries in IML files.

tonycosentini:
These paths always need to be relative, otherwise the IDE will duplicate it like this:
 screen shot 2016-07-27 at 4 32 26 pm


kageiit:
Looks good!

PR #830, opened 2016-08-03
Use package and resource name when computing referenced resource names in dex

### LINKED ISSUE #828: Too many field references in Uber R.java jar to fit in a single dex
In 4cf0eb7a744c9583cc3f45f7952ef3aa6ae8afae , the list of fields to be included in R.java was changed from computing from the resource xml references and just using the R.txt from prebuilt aars instead. This means whatever is in R.txt gets treated as if it was declared in the package that is in the aar's manifest.

This becomes a problem in large apps that consume many prebuilt aars. If these aars depend on libraries like the android support library, gms etc, the field references of used resources coming from such dependencies will be counted against each prebuilt aar.

For example, if gms had a color resource 
```
public static class color {
    public static final int common_google_signin_btn_text_dark=0x7f0700b0;
}
```
referenced in an app and aarX and aarY (both depended on by app) depend on gms, the Uber R.java jar will contain both `com.aarX.R.color.common_google_signin_btn_text_dark` and `com.aarY.R.color.common_google_signin_btn_text_dark` even if the app only refers to `com.aarX.R.color.common_google_signin_btn_text_dark` in its code. 

This leads to a situation where all the R.java files cannot fit in a single jar to be dexed/predex because the number of filed references is > 65535.

```
com.facebook.buck.step.StepFailedException: Failed on step dx_&&_write_file with an exception:
field ID not in [0, 0xffff]: 65536
com.android.dex.DexIndexOverflowException: field ID not in [0, 0xffff]: 65536
```

`trim_resource_ids` does not help much because the field reference count in the custom dx packaged with buck only looks at field names and not the fully qualified name of field references for simplicity.

Had a long conversation regarding this with @dreiss yesterday. I seem to see two possible solutions to this

- Make the referenced resource calculation in dx more fine grained by including the package name + resource name instead of just the resource name
- Provide an option to list a `primary_dex_resource_patterns` and split the Uber R.java jar into multiple jars if `linear_alloc_splt_dex` is enabled so resources needed in the primary dex can still be kept like today.

If there are any other workarounds or other solutions, would love to know so we can try the best alternative

kageiit:
Resolves #828

### DIFF
--- a/src/com/facebook/buck/android/TrimUberRDotJava.java
+++ b/src/com/facebook/buck/android/TrimUberRDotJava.java
@@ -58,6 +58,9 @@ class TrimUberRDotJava extends AbstractBuildRule {
   private static final Pattern R_DOT_JAVA_LINE_PATTERN = Pattern.compile(
       "^ *public static final int(?:\\[\\])? (\\w+)=");
 
+  private static final Pattern R_DOT_JAVA_PACKAGE_NAME_PATTERN = Pattern.compile(
+      "^ *package ([\\w.]+);");
+
   TrimUberRDotJava(
       BuildRuleParams buildRuleParams,
       SourcePathResolver resolver,
@@ -176,12 +179,22 @@ class TrimUberRDotJava extends AbstractBuildRule {
       OutputStream output,
       ImmutableSet<String> allReferencedResources)
       throws IOException {
+    String packageName = null;
+    Matcher m;
     for (String line : rDotJavaLines) {
-      Matcher m = R_DOT_JAVA_LINE_PATTERN.matcher(line);
-      // We ignore the containing nested class and just match on the resource name.
+      if (packageName == null) {
+        m = R_DOT_JAVA_PACKAGE_NAME_PATTERN.matcher(line);
+        if (m.find()) {
+          packageName = m.group(1);
+        } else {
+          continue;
+        }
+      }
+      m = R_DOT_JAVA_LINE_PATTERN.matcher(line);
+      // We match on the package name + resource name.
       // This can cause us to keep (for example) R.layout.foo when only R.string.foo
       // is referenced.  That is a very rare case, though, and not worth the complexity to fix.
-      if (m.find() && !allReferencedResources.contains(m.group(1))) {
+      if (m.find() && !allReferencedResources.contains(packageName + "." + m.group(1))) {
         continue;
       }
       output.write(line.getBytes(Charsets.UTF_8));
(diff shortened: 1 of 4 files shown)


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @dreiss to be a potential reviewer.

src/com/facebook/buck/android/TrimUberRDotJava.java
```diff
@@ -58,6 +58,9 @@
   private static final Pattern R_DOT_JAVA_LINE_PATTERN = Pattern.compile(
       "^ *public static final int(?:\\[\\])? (\\w+)=");
 
+  private static final Pattern R_DOT_JAVA_PACKAGE_NAME_PATTERN = Pattern.compile(
+      "^ *package(?:\\[\\])? ([\\w\\.]+);");
```
dreiss: What is `\\[\\]` for?

`.` doesn't need `\\` inside `[]`

src/com/facebook/buck/android/TrimUberRDotJava.java
```diff
@@ -176,12 +179,22 @@ private static void filterRDotJava(
       OutputStream output,
       ImmutableSet<String> allReferencedResources)
       throws IOException {
+    String packageName = "";
```
dreiss: Make this start as null instead of `""`?

src/com/facebook/buck/android/TrimUberRDotJava.java
kageiit: The `?:\\[\\])?` part I took from the regex above. I dont think it is needed. I can remove it for simplicity

src/com/facebook/buck/android/TrimUberRDotJava.java
kageiit: will do

facebook-github-bot-2:

dreiss:
Looks like AndroidBinaryIntegrationTest and TrimUberRDotJavaTest are failing.  Want to take a look?

dreiss:
Otherwise, the change looks fine.

kageiit:
> Looks like AndroidBinaryIntegrationTest and TrimUberRDotJavaTest are failing. Want to take a look?

Sure I will update the tests. Thanks!

facebook-github-bot-0:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/830/files/73b8d2400ff88df48cc92a444eb67b92fad50577..a6f228a49e38fe06bee539bfb0f46913a10064eb) - [changes since last import](https://github.com/facebook/buck/pull/830/files/73b8d2400ff88df48cc92a444eb67b92fad50577..a6f228a49e38fe06bee539bfb0f46913a10064eb)

kageiit:
@dreiss tests updated and pass

Issue #821, opened 2016-07-28
New project generation no longer sets IDE module SDKs.


tonycosentini:
Previously, the deprecated project generation would set the JDK for the entire project using:

```
[intellij]
    jdk_name = Android API 23 Platform
    jdk_type = AndroidSDK
    language_level = JDK_1_7
```

It seems like the new project generation doesn't respect these values. Is there any rationale behind this - or is it simply not implemented?

marcinkosiba:
It's not implemented. The new project generator doesn't write the misc.xml file at all, so once you set up the SDK once in IJ it should just keep on working.
The main reason this part is non-trivial is because the jdk/sdk settings are bound to your install of IntelliJ, and will potentially vary across developers machines, so the choice is to either have a hardcoded value, which may or may not work (deprecated project), to have none (current project implementation). Another option would be to auto-detect IntellIj SDK/JDK setup and write a misc.xml that matches is, but that then requires knowing which version of IJ the user wants to open the project with (sounds like something the IJ plugin could help with).

tonycosentini:
That makes sense.

What about individual SDKs for specific IntelliJ modules? Does that get overwritten when running `buck project`? We have a mix of Android and Java modules - when setting the global project SDK to Android, you need to manually set all of the Java modules to a non-Android JDK.

kageiit:
> so the choice is to either have a hardcoded value, which may or may not work (deprecated project), to have none (current project implementation)

How about writing only if values are explicitly specified instead? I was also thinking of doing an sdk option per language like

```
android_sdk
java_sdk
```

This can then be set on the module xml of the corresponding targets. This way more jvm based languages can add their own sdk parameters as needed and st them in their module xml.The language level can be inferred from the source and target options set on the java/android targets, to be set on the module xml instead of misc.xml as well.

kageiit:
@marcinkosiba I put up a PR that sets up the jdk and language level in every module.iml file instead. This achieves what @tonycosentini mentioned in the previous comment.

styurin:
That was recently implemented, check .buckconfig docs: https://buckbuild.com/concept/buckconfig.html#intellij.

kageiit:
#832 makes it possible to configure multiple default sdks per project based on the type of module

PR #832, opened 2016-08-05
Allow configuring Android SDK name and type for intellij project modules

kageiit:
Ref #821

This adds support for setting the default android sdk name and type in generated intellij iml files.

```
[intellij]
  android_sdk_name = Android API 23 Platform
  android_sdk_type = AndroidSDK
```

### DIFF
--- a/docs/concept/buckconfig.soy
+++ b/docs/concept/buckconfig.soy
@@ -1585,6 +1585,24 @@ $ buck targets --resolve-alias app#src_jar
 
 {call buckconfig.entry}
   {param section: 'intellij' /}
+  {param name: 'android_module_sdk_type' /}
+  {param example_value: 'Android SDK' /}
+  {param description}
+    Default Android SDK type for android modules.
+  {/param}
+{/call}
+
+{call buckconfig.entry}
+  {param section: 'intellij' /}
+  {param name: 'android_module_sdk_name' /}
+  {param example_value: 'Android API 23 Platform' /}
+  {param description}
+    Default Android SDK name for android modules.
+  {/param}
+{/call}
+
+{call buckconfig.entry}
+  {param section: 'intellij' /}
   {param name: 'language_level' /}
   {param example_value: 'JDK_1_7' /}
   {param description}
--- a/src/com/facebook/buck/jvm/java/intellij/AbstractIjProjectConfig.java
+++ b/src/com/facebook/buck/jvm/java/intellij/AbstractIjProjectConfig.java
@@ -48,6 +48,10 @@ abstract class AbstractIjProjectConfig {
 
   public abstract Optional<String> getProjectJdkType();
 
+  public abstract Optional<String> getAndroidModuleSdkName();
+
+  public abstract Optional<String> getAndroidModuleSdkType();
+
   public abstract Optional<String> getProjectLanguageLevel();
 
   public abstract List<String> getExcludedResourcePaths();
--- a/src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
+++ b/src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
@@ -289,6 +289,7 @@ public class IjModuleFactory {
   }
 
   private static final String SDK_TYPE_JAVA = "JavaSDK";
+  private static final String SDK_TYPE_ANDROID = "Android SDK";
 
   private final Map<BuildRuleType, IjModuleRule<?>> moduleRuleIndex = new HashMap<>();
   private final IjModuleFactoryResolver moduleFactoryResolver;
@@ -357,13 +358,19 @@ public class IjModuleFactory {
     Optional<String> sourceLevel = getSourceLevel(targetNodes);
     // The only JDK type that is supported right now. If we ever add support for Android libraries
     // to have different language levels we need to add logic to detect correct JDK type.
-    String sdkType = SDK_TYPE_JAVA;
-
+    String sdkType;
     Optional<String> sdkName;
-    if (sourceLevel.isPresent()) {
-      sdkName = getSdkName(sourceLevel.get(), sdkType);
+
+    if (context.getAndroidFacet().isPresent()) {
+      sdkType = projectConfig.getAndroidModuleSdkType().orElse(SDK_TYPE_ANDROID);
+      sdkName = projectConfig.getAndroidModuleSdkName();
     } else {
-      sdkName = Optional.empty();
+      sdkType = projectConfig.getProjectJdkType().orElse(SDK_TYPE_JAVA);
+      if (sourceLevel.isPresent()) {
+        sdkName = getSdkName(sourceLevel.get(), sdkType);
+      } else {
+        sdkName = Optional.empty();
+      }
     }
 
     return IjModule.builder()
--- a/src/com/facebook/buck/jvm/java/intellij/IjProjectBuckConfig.java
+++ b/src/com/facebook/buck/jvm/java/intellij/IjProjectBuckConfig.java
@@ -76,6 +76,10 @@ public class IjProjectBuckConfig {
             buckConfig.getValue(INTELLIJ_BUCK_CONFIG_SECTION, "jdk_name"))
         .setProjectJdkType(
             buckConfig.getValue(INTELLIJ_BUCK_CONFIG_SECTION, "jdk_type"))
+        .setAndroidModuleSdkName(
+            buckConfig.getValue(INTELLIJ_BUCK_CONFIG_SECTION, "android_module_sdk_name"))
+        .setAndroidModuleSdkType(
+            buckConfig.getValue(INTELLIJ_BUCK_CONFIG_SECTION, "android_module_sdk_type"))
         .setProjectLanguageLevel(
             buckConfig.getValue(INTELLIJ_BUCK_CONFIG_SECTION, "language_level"))
         .setExcludedResourcePaths(excludedResourcePaths)
--- a/test/com/facebook/buck/jvm/java/intellij/ProjectIntegrationTest.java
+++ b/test/com/facebook/buck/jvm/java/intellij/ProjectIntegrationTest.java
@@ -604,6 +604,11 @@ public class ProjectIntegrationTest {
   }
 
   @Test
+  public void testBuckProjectWithCustomAndroidSdks() throws IOException {
+    runBuckProjectAndVerify("project_with_custom_android_sdks");
+  }
+
+  @Test
   public void testVersion2BuckProjectWithProjectSettings() throws IOException {
     runBuckProjectAndVerify("experimental_project_with_project_settings");
   }
--- a/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/.buckconfig
+++ b/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/.buckconfig
@@ -1,0 +1,11 @@
+[color]
+  ui = false
+[alias]
+  root=//:root_repo
+[project]
+  ignore = buck-out
+  ide = intellij
+  disable_r_java_idea_generator = true
+[intellij]
+  android_module_sdk_name = "Google Inc.:Google APIs:24"
+  android_module_sdk_type = "AndroidSDK"
--- a/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/.idea/modules.xml.expected
+++ b/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/.idea/modules.xml.expected
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectModuleManager">
+    <modules>
+      <module fileurl="file://$PROJECT_DIR$/modules/libA/modules_libA.iml" filepath="$PROJECT_DIR$/modules/libA/modules_libA.iml"  group="modules"  />
+      <module fileurl="file://$PROJECT_DIR$/project_root.iml" filepath="$PROJECT_DIR$/project_root.iml"  />
+
+   </modules>
+  </component>
+</project>
--- a/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/modules/libA/BUCK.fixture
+++ b/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/modules/libA/BUCK.fixture
@@ -1,0 +1,4 @@
+android_library(
+    name='src',
+    visibility=['PUBLIC'],
+)
--- a/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/modules/libA/modules_libA.iml.expected
+++ b/test/com/facebook/buck/jvm/java/intellij/testdata/project_with_custom_android_sdks/modules/libA/modules_libA.iml.expected
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="JAVA_MODULE" version="4">
+  <component name="FacetManager">
+    <facet type="android" name="Android">
+      <configuration>
+        <option name="GEN_FOLDER_RELATIVE_PATH_APT" value="/../../buck-out/android/modules/libA/gen" />
+        <option name="GEN_FOLDER_RELATIVE_PATH_AIDL" value="/../../buck-out/android/modules/libA/gen" />
+        <option name="MANIFEST_FILE_RELATIVE_PATH" value="/../../android_res/AndroidManifest.xml" />
+        <option name="RES_FOLDERS_RELATIVE_PATH" value="" />
+        <option name="ASSETS_FOLDER_RELATIVE_PATH" value="/assets" />
+        <option name="LIBS_FOLDER_RELATIVE_PATH" value="/libs" />
+        <option name="APK_PATH" value="" />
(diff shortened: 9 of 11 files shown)


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @alsutton, @ilya-klyuchnikov and @andrewjcg to be potential reviewers.

facebook-github-bot-6:

src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
```diff
@@ -670,6 +683,7 @@ public void apply(TargetNode<JavaTestDescription.Arg> target, ModuleBuildContext
           true /* wantsPackagePrefix */,
           context);
       addCompiledShadowIfNeeded(target, context);
+      context.setLanguageLevel(getLevel(target.getConstructorArg().source));
```
marcinkosiba: multiple TargetNodes in the same module could have different language levels. The way you've written it means "last one wins", at minimum keep them as a ImmutableSortedSet<String> and get the first/last one.

src/com/facebook/buck/jvm/java/intellij/IjProjectWriter.java
```diff
@@ -109,6 +110,15 @@ private Path writeModule(IjModule module) throws IOException {
         "androidFacet",
         projectDataPreparer.getAndroidProperties(module));
 
+    Optional<String> javaSdk = buckConfig.getValue("intellij", "java_sdk");
```
marcinkosiba: IjProjectWriter should cary as little logic and behavior as possible as it's only covered by integration tests. Move this to IjProjectTemplateDataPreparer and add tests.

Also, passing in the entire BuckConfig is overkill as, again, it makes things harder to test (I know the IJProjectCleaner does it, however, you will notice that that class has no tests :/). We have two patterns:
 1) create a IjProjectBuckConfig and have it wrap BuckConfig (see `ArtifactCacheBuckConfig` for an example), this has the benefit of keeping the config-ish logic in one place, but to test you still need to new up a BuckConfig,
 2) create a data object for the config and have separate code that extracts it from BuckConfig (see `AbstractRageConfig`, and `RageBuckConfig`).
I'd prefer using 2), as it makes testing simpler, but if you go 1) I won't block the review.

marcinkosiba:
Have you considered what happens when aggregation is enabled? I'd expect modules with different levels to not want to merge with each other.

I'd like to see tests for the following scenarios:
 - how are multiple language versions resolved if 2 targets in the same path have different settings
 - a test showing that the buckconfig value is the fallback
 - a test showing how this interacts with the module aggregation code.

src/com/facebook/buck/jvm/java/intellij/ij-module.st
```diff
@@ -49,7 +49,14 @@
         %if(generatedFolder.isAndroidResources)% type="java-resource" %endif%/>
     </content>
 }%
+
+%if(!androidFacet.enabled && java_sdk)%
+    <orderEntry type="jdk" jdkName="%java_sdk%" jdkType="JavaSDK" />
```
marcinkosiba: I had a bunch of code that would attempt to read IntelliJ's config and determine whether these SDKs are actually defined (with the intent to figure out if the .jar files they include would be the ones buck would use during the build). IIRC that was like 80% complete, would you be interested in taking that over?

Assuming that the developers will have the correct SDK set up has not worked well for us internally. Any changes to the default are painful and it takes a long time for everyone to update their IJ setup.

kageiit: This is purely scoped to java_sdk. If someone wants to use groosdk, it would need to be a separate config key `groovy_sdk` in the intellij block of buckconfig and could be set here.

I would like to get this in to atleast mirror the legacy support and make it module level granularity. The auto detection change should not be part of this change imo

kageiit: > I had a bunch of code that would attempt to read IntelliJ's config

What if there are multiple instalaltions of inteeli/android studio or its variants? There is a lot of complexity involved in that calculation and it should be kept separate from this change

kageiit:
> Have you considered what happens when aggregation is enabled?

I dont know what aggregation means. Could you please clarify?

> how are multiple language versions resolved if 2 targets in the same path have different settings

I am not sure if intellij supports this? Is this a common enough scenario to support?
I want to keep this change as small as possible to support the most common case. More features etc. can be added in later

src/com/facebook/buck/jvm/java/intellij/ij-module.st
marcinkosiba: > This is purely scoped to java_sdk. If someone wants to use groovysdk, they would need to be a separate config key groovy_sdk in the intellij block of buckconfig and it could be used here.

I think one of my comments wasn't clear. I didn't mean you should add support for more kind of language SDKs, I meant that you have no guarantee that these SDKs are defined in IntelliJ's config nor that they are correctly set up. To put it another way - you're effectively hardcoding strings and expecting those to be correctly set up on every machine you run `buck project` on.

> The auto detection change should not be part of this change imo
Agreed, that was not what I was asking for. My question was "do you want the code?", not "merge it into this PR". The real question behind it is: are you planning more changes to this feature (project generation)?

src/com/facebook/buck/jvm/java/intellij/ij-module.st
```diff
@@ -49,7 +49,14 @@
         %if(generatedFolder.isAndroidResources)% type="java-resource" %endif%/>
     </content>
 }%
+
+%if(!androidFacet.enabled && java_sdk)%
```
marcinkosiba: just like with IjProjectWriter, the templates are only tested through integration tests, so it's more convenient to have any logic live at higher abstraction levels. Here I suggest adding a Jdk type on the module, if it's null, print 'inheritedJdk', otherwise print it's .name and .type:

```
%if(jdk)%
  <orderEntry type="jdk" jdkName="%jdk.name%" jdkType="%jdk.type%" />
%else%
 <orderEntry type="inheritedJdk" />
%endif%
```

marcinkosiba:
>> Have you considered what happens when aggregation is enabled?

> I dont know what aggregation means. Could you please clarify?

aggregates multiple modules into one, see https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/intellij/IjModuleGraph.java

>> how are multiple language versions resolved if 2 targets in the same path have different settings

> I am not sure if intellij supports this? Is this a common enough scenario to support?
> I want to keep this change as small as possible to support the most common case. More features etc. can be added in later

Correct, intellij doesn't support this, but buck does. We need a deterministic way to translate the more expressive buck model into something IntelliJ can understand. Yes, this is common enough to support.

kageiit:
Seems like part of this was addressed in 60058a4f32681808ce5776d9f8bbb348282b5633

I will rebase this pr and include only the android sdk changes

facebook-github-bot-7:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/832/files/2d27703f0910b2823ca312bcb9753be11c3f9ed8..b6aa77cebeb690e1185b1ad030abbffca6a5e28d)

kageiit:
@marcinkosiba I have rewritten this with your suggestions and isolated all logic to `IjModuleFactory`. Let me know if this approach looks ok and I will add tests to check all the scenarios.

facebook-github-bot-9:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/832/files/b6aa77cebeb690e1185b1ad030abbffca6a5e28d..353a6e532e506053e6e6310ffce5a2516ea42d0b)

facebook-github-bot-8:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/832/files/353a6e532e506053e6e6310ffce5a2516ea42d0b..5bea36ce9d7cf5cd2c6393bc2b5753bb2633e5a6)

facebook-github-bot-8:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/832/files/5bea36ce9d7cf5cd2c6393bc2b5753bb2633e5a6..6d6adedb847210e9ddb0624c43c35ab2b23e3fb8)

src/com/facebook/buck/jvm/java/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -400,6 +400,31 @@ public ModuleIndexEntry apply(IjModule module) {
     return androidProperties;
   }
 
+  public Map<String, String> getJdkProperties(IjModule module)  throws IOException {
```
marcinkosiba: Why do you to convert this to a map? ST handles Java objects just fine.

marcinkosiba: double space after `throws`

src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
```diff
@@ -359,29 +361,65 @@ public IjModule createModule(
...
+      chosenJdkType = intellijConfig.getAndroidSdkType().or(DEFAULT_ANDROID_SDK_TYPE);
+    } else {
+      if (chosenJdkName.isPresent()) {
+        String jdkName = chosenJdkName.get();
+        if (!jdkName.contains(".")) {
+          chosenJdkName = Optional.of(String.format("1.%s", jdkName));
+        }
       }
+
+      chosenJdkName = chosenJdkName.or(intellijConfig.getJdkName());
```
marcinkosiba: this is confusing, put it in an `else` clause of the `if` statement above

marcinkosiba: We already have a global language and target level configured in buck: https://buckbuild.com/rule/java_library.html (`source`/`target`) and https://buckbuild.com/concept/buckconfig.html#java.source_level

src/com/facebook/buck/jvm/java/intellij/IjProjectTemplateDataPreparer.java
kageiit: ah I was not aware. I will clean it up :)

src/com/facebook/buck/jvm/java/intellij/IjModuleFactory.java
kageiit: Sure I will make the changes. Thanks for the feedback

kageiit:
Seems like the approach looks fine. So i will address comments from @marcinkosiba and add tests

kageiit:
Just did a quick rebase. will be adding tests tomorrow

test/com/facebook/buck/jvm/java/intellij/IjModuleFactoryTest.java
```diff
@@ -525,50 +525,10 @@ public void testOverrideSdkFromBuckConfig() throws Exception {
         moduleBasePath,
         ImmutableSet.of(java8Node));
 
-    assertThat(moduleWithDefault.getSdkName(), equalTo(Optional.<String>absent()));
-    assertThat(moduleWithJava8.getSdkName(), equalTo(Optional.of("TestSDK")));
-    assertThat(moduleWithJava8.getLanguageLevel(), equalTo(Optional.of("1.8")));
-  }
-
-  @Test
-  public void testAndroidPrebuiltAar() {
```
marcinkosiba: why did you remove this?

kageiit:
I have not added tests yet. See my previous comment. I just did a quick rebase to resolve some merge conflicts. I will fix the line limit and test scenarios

styurin:
This is already implemented, can you check existing functionality and see if we are missing something?

kageiit:
@styurin please take a look

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
   {param name: 'jdk_name' /}
   {param example_value: 'Java SDK 1.6' /}
   {param description}
-    IntelliJ project SDK name.
+    IntelliJ project Java SDK name.
```
styurin: `jdk_name` is a bad name here, it's an SDK name, not JDK name. It's better leave it as it was before (i.e. just SDK name) because it can point to an Android SDK or Intellij Plugin SDK.

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
...
 {/call}
 
 {call buckconfig.entry}
   {param section: 'intellij' /}
   {param name: 'jdk_type' /}
-  {param example_value: 'Android SDK or JavaSDK' /}
+  {param example_value: 'JavaSDK' /}
   {param description}
-    IntelliJ project SDK type.
+    IntelliJ project Java SDK type.
```
styurin: the same, JDK is one type, there might be other types.

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
...
+  {/param}
+{/call}
+
+{call buckconfig.entry}
+  {param section: 'intellij' /}
+  {param name: 'android_sdk_type' /}
```
styurin: We already have `java_library_sdk_names`, can you follow the same logic and rename this option to `android_library_sdk_type`?

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
...
+  {param example_value: 'Android SDK' /}
+  {param description}
+    IntelliJ project Android SDK type.
```
styurin: "Default SDK type for android_library modules" or something like that. In other words, please, indicate that this is applied to android_library only.

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
...
+  {/param}
+{/call}
+
+{call buckconfig.entry}
+  {param section: 'intellij' /}
+  {param name: 'android_sdk_name' /}
```
styurin: The same, `android_library_sdk_name`.

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
...
+  {param example_value: 'Google Inc.:Google APIs:24' /}
+  {param description}
+    IntelliJ project Android SDK name.
```
styurin: The same comment as in SDK type

src/com/facebook/buck/jvm/java/intellij/AbstractIjProjectConfig.java
```diff
@@ -48,6 +48,10 @@ public boolean isAutogenerateAndroidFacetSourcesEnabled() {
 
   public abstract Optional<String> getProjectJdkType();
 
+  public abstract Optional<String> getProjectAndroidSdkName();
```
styurin: `Project` means it's a project level setting. In this case methods should be `getAndroidLibrarySdkName` and `getAndroidLibrarySdkType`.

docs/concept/buckconfig.soy
```diff
@@ -1570,16 +1570,34 @@ $ buck targets --resolve-alias app#src_jar
...
+  {param example_value: 'Android SDK' /}
+  {param description}
+    IntelliJ project Android SDK type.
+  {/param}
+{/call}
+
+{call buckconfig.entry}
+  {param section: 'intellij' /}
+  {param name: 'android_sdk_name' /}
+  {param example_value: 'Google Inc.:Google APIs:24' /}
```
styurin: Can you replace the example with the default SDK name, for example, "Android API 23 Platform"

test/com/facebook/buck/jvm/java/intellij/testdata/experimental_project_with_custom_sdks/.buckconfig
```diff
@@ -9,3 +9,5 @@
 [intellij]
   java_library_sdk_names = 1.6 => Java SDK 1.6, \
                            1.8 => Java SDK 1.8
+  android_sdk_name = "Google Inc.:Google APIs:24"
```
styurin: The same about sdk name, please, replace with the default name "Android API 23 Platform".

test/com/facebook/buck/jvm/java/intellij/testdata/experimental_project_with_custom_sdks/modules/libA/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
styurin: Do we really need real code here?

docs/concept/buckconfig.soy
kageiit: This applies to all android rules. I think the original argument should instead be renamed to `java_sdk_names` instead because it would apply to all java rules as well

docs/concept/buckconfig.soy
kageiit: Same comment as above. I think this should be named `android_sdk_name` as it applies to all android rules

test/com/facebook/buck/jvm/java/intellij/testdata/experimental_project_with_custom_sdks/modules/libA/src/main/AndroidManifest.xml
kageiit: We dont. I will remove

PR #839, opened 2016-08-10
Change OSX framework bundle paths

robbertvanginkel:
I don't know where the `Contents` and `MacOs` folders originally came from, that commit (d44f10e6258859badb0d5ce9ee3ff99d65e1aaf6) doesn't contain a lot of background. I think those folders are wrong for the following reasons:

- They do not conform to the [apple docs](https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html)
- Checking a random library on my mac also doesn't show them.
```
$ tree -L 3 /Library/Frameworks/iTunesLibrary.framework
/Library/Frameworks/iTunesLibrary.framework
├── Headers -> Versions/Current/Headers
├── Modules -> Versions/Current/Modules
├── Resources -> Versions/Current/Resources
├── Versions
│   ├── A
│   │   ├── Headers
│   │   └── iTunesLibrary
│   └── Current -> A
└── iTunesLibrary -> Versions/Current/iTunesLibrary
```
- When using a framework generated from an `apple_library` (I know this doesn't fully work but I'm working on a PR to fix that), clang complains during linking/preprocessing (with `-F` and `-framework`) that it cannot find the framework. When I changed the binaries and headers to not be in the contents and MacOS paths everything works.

### DIFF
--- a/src/com/facebook/buck/apple/AppleDescriptions.java
+++ b/src/com/facebook/buck/apple/AppleDescriptions.java
@@ -488,9 +488,15 @@ public class AppleDescriptions {
         params.getBuildTarget(),
         MultiarchFileInfos.create(appleCxxPlatforms, params.getBuildTarget()));
 
-    AppleBundleDestinations destinations =
-        AppleBundleDestinations.platformDestinations(
-            appleCxxPlatform.getAppleSdk().getApplePlatform());
+    AppleBundleDestinations destinations = null;
+
+    if (extension.isLeft() && extension.getLeft().equals(AppleBundleExtension.FRAMEWORK)) {
+      destinations = AppleBundleDestinations.platformFrameworkDestinations(
+          appleCxxPlatform.getAppleSdk().getApplePlatform());
+    } else {
+      destinations = AppleBundleDestinations.platformDestinations(
+                        appleCxxPlatform.getAppleSdk().getApplePlatform());
+    }
 
     AppleBundleResources collectedResources = AppleResources.collectResourceDirsAndFiles(
         targetGraph,
--- a/src/com/facebook/buck/apple/AbstractAppleBundleDestinations.java
+++ b/src/com/facebook/buck/apple/AbstractAppleBundleDestinations.java
@@ -46,6 +46,12 @@ abstract class AbstractAppleBundleDestinations implements RuleKeyAppendable {
   @Value.Parameter
   public abstract Path getWatchAppPath();
 
+  @Value.Parameter
+  public abstract Path getHeadersPath();
+
+  @Value.Parameter
+  public abstract Path getModulesPath();
+
   @Override
   public void appendToRuleKey(RuleKeyObjectSink sink) {
     sink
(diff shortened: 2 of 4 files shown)


facebook-github-bot-7:
By analyzing the blame information on this pull request, we identified @beefon, @Coneko and @andrewjcg to be potential reviewers.

Coneko:
`Contents/MacOS` is for macOS applications, not frameworks.

Does this not break `AppleBinaryIntegrationTest.testAppleBinaryWithLibraryDependencyBuildsApp`?

robbertvanginkel:
Ah, I only checked frameworks not apps. It does break that test and I didn't notice as I only ran the `AppleBundle`/`AppleLibrary` IntegrationTests.

I think the best way forward would be to add a `AppleBundleDestinations.FRAMEWORK_DESTINATIONS`? Maybe also rename the `OSX`/`IOS` destinations to `OSX_APP_DESTINATIONS` and `IOS_APP_DESTINATIONS`.

Coneko:
Sounds good.

Are the framework destinations the same between macOS and iOS or do those need to be split as well? Can you double check?

robbertvanginkel:
Looking at the iOS Simulator frameworks it seems that they are slightly different and don't have a `Resources` folder. I can't find a good doc about it thought, except that xcode uses a setting `UNLOCALIZED_RESOURCES_FOLDER_PATH`, which for OSX is [defined in the docs](https://developer.apple.com/library/mac/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html)  (by default `CONTENTS_FOLDER_PATH/Contents/Resources`). 

There's no iOS counterpart though, but when creating a fresh framework project the value
`export UNLOCALIZED_RESOURCES_FOLDER_PATH=TestFramework.framework` is used. This seems like there is a difference and I'll update the PR accordingly.

```
16:56 $ tree -L 2/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/System/Library/Frameworks/GLKit.framework
.
├── GLKit
├── Headers
│   ├── ...
│   └── GLKitBase.h
├── Info.plist
├── en.lproj
│   └── InfoPlist.strings
├── fsh.xml
├── Modules
│   └── module.modulemap
├── module.map
└── vsh.xml
```

facebook-github-bot-6:

src/com/facebook/buck/apple/AppleDescriptions.java
```diff
@@ -488,9 +488,15 @@ static AppleBundle createAppleBundle(
         params.getBuildTarget(),
         MultiarchFileInfos.create(appleCxxPlatforms, params.getBuildTarget()));
 
-    AppleBundleDestinations destinations =
-        AppleBundleDestinations.platformDestinations(
-            appleCxxPlatform.getAppleSdk().getApplePlatform());
+    AppleBundleDestinations destinations = null;
```
robbertvanginkel: I'm not sure if this is the best way to do this in java. Style comments appreciated.

ryu2: Usually we use `Optional<>`.

ryu2:
@facebook-github-bot shipit

facebook-github-bot-6:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3698142) to review internal test results.

PR #842, opened 2016-08-11
[WIP] apple_library fix the framework flavor

robbertvanginkel:
This is is a work in progress, and based on #839. The goal is to make it easy to switch between statically linked apple_libraries or packaging them as frameworks.

This mostly works, but there are some parts that need to be updated. I'll put some comments in line and we can discuss the approach.


src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -495,6 +524,27 @@ public CodeSignIdentity get() {
...
+    Preconditions.checkArgument(isFrameworkBundle());
+    Preconditions.checkArgument(headers.isPresent());
+
+    Path headersFolder = bundleRoot.resolve(this.destinations.getHeadersPath());
+    stepsBuilder.add(
+        new MakeCleanDirectoryStep(
+            getProjectFilesystem(),
+            headersFolder));
+
+    for (SourcePath header : headers.get().getLinks().values()) {
```
robbertvanginkel: This copies each header individually, and doens't work correctly if the symlinktree has subfolder. Is there an easy way to copy the contents of a Symlinktree?

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -866,4 +942,46 @@ public BuildRule getBinaryBuildRule() {
...
+  @Override
+  public Iterable<? extends NativeLinkable> getNativeLinkableExportedDeps(CxxPlatform cxxPlatform) {
+    return ImmutableSet.of();
+  }
+
+  @Override
+  public NativeLinkableInput getNativeLinkableInput(
+      CxxPlatform cxxPlatform, Linker.LinkableDepType type) throws NoSuchBuildTargetException {
+    ImmutableList.Builder<Arg> linkerArgsBuilder = ImmutableList.builder();
+    linkerArgsBuilder.addAll(StringArg.from(
```
robbertvanginkel: These rpaths should be setup differently depending on what type of bundle the binary ends up in. For iOS and for framework bundles it needs to be `@loader_path/Frameworks`, but for OSX bundles `@loader_path/../Frameworks`. I think this might not be the best place to add this, and that it could be more appropriate as a flavor on the binary. Including frameworks a bundle that contains a binary influences that binary's linker flags because these rpaths should be set or not (although I'm not sure if its bad if they're always set). The flavor on the binary would add the appropriate linker flags for the platform/bundle combination.

Thoughts?

src/com/facebook/buck/apple/AppleDescriptions.java
```diff
@@ -508,7 +509,8 @@ static AppleBundle createAppleBundle(
         Optional<FrameworkDependencies> frameworkDependencies =
             resolver.requireMetadata(
                 BuildTarget.builder(dep)
-                    .addFlavors(FRAMEWORK_FLAVOR)
+                    //this seems wrong
```
robbertvanginkel: When including frameworks, I do not want to convert all `apple_library` rules into frameworks by default. I might have a huge library like opencv that would be better linked statically so you can still code strip and don't end up with huge frameworks.

We need to figure out an easy way to switch between libs and frameworks that allows fine grained control and doesn't force your whole tree one way or another.

test/com/facebook/buck/apple/AppleBinaryIntegrationTest.java
```diff
@@ -855,6 +855,22 @@ public void testAppleBundleDebugFormatRespectsDefaultConfigSettingNoDebug() thro
...
+    assumeTrue(Platform.detect() == Platform.MACOS);
+    ProjectWorkspace workspace = TestDataHelper.createProjectWorkspaceForScenario(
+        this, "apple_binary_with_library_dependencies_as_frameworks", tmp);
+    workspace.setUp();
+
+    BuildTarget target = BuildTargetFactory.newInstance("//TestApp:TestAppBundle#include-frameworks");
+    ProjectWorkspace.ProcessResult result = workspace.runBuckCommand(
+        "build", target.getFullyQualifiedName());
+    result.assertSuccess();
+
```
robbertvanginkel: I need to write some more asserts for this. It currently builds but doesn't run because of the aforementioned rpath problem. After I fix that I'll update the test case. The test case is elaborate on mixing frameworks/non frameworks and should give a good impression of the requirements for what I'm trying to achieve.

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -838,10 +898,26 @@ public CxxPreprocessorInput getCxxPreprocessorInput(
             cxxPlatform,
             headerVisibility);
       }
+      if (isFrameworkBundle() && headerVisibility == HeaderVisibility.PUBLIC) {
```
robbertvanginkel: Some of this stuff is very similar to the prebuilt frameworks linking support from #818. It'd be good if I can find a way to consolidate the logic for linking with a framework.

src/com/facebook/buck/apple/AppleBundle.java
ryu2: Not currently I believe; feel free to implement it. ;)

src/com/facebook/buck/apple/AppleBundle.java
ryu2: Perhaps a static method in a `AppleFrameworks` utility class?

src/com/facebook/buck/apple/AppleDescriptions.java
ryu2: Perhaps we can make a non-framework flavor and explicitly specify it in a rule's `deps` to include libraries that we want to force to be normal libs.  (Or analogously specify the framework flavor to force to be frameworks)

src/com/facebook/buck/apple/AppleBundle.java
ryu2: Since these flags will only be added when a binary is included in a framework bundle, I'm thinking we just do it here, and use the bundle's `platformName` to choose.

robbertvanginkel: The problem is that when specifying it here, you do not know what kind of bundle you end up in. Right not this is configured like an exported linker flag: the dependency framework tells the binary that when it links with it that the parent should also set an rpath value so it can find the framework with dyld. 

Problem is that the dependency itself doesn't know if the binary that calls the `getNativeLinkableInput` is a binary that will go into an osx app bundle, framework bundle or ios app bundle. Based on that info these paths should be different. That's why I don't think this setting can work correctly from here. I'll try to see if I can modify the binary's linker flags from `AppleBundleDescription`.

src/com/facebook/buck/apple/AppleDescriptions.java
robbertvanginkel: Yes, I think we'll need some way of saying "by default everything frameworks or libraries", but it should be overridable.

What would you say is a better approach, having a non framework flavor or an extra field on `apple_library` like `allows_framework_packaging`.

src/com/facebook/buck/apple/AppleBundle.java
robbertvanginkel: I was thinking more like Java's interfaces with default methods, but it seems like they're only available from 8 and onwards.

The methods for implementing `CxxPreprocessorDep, NativeLinkable` as a famework bundle are very short, so I'll see if it makes sense to put them in an apple frameworks utility class.

src/com/facebook/buck/apple/AppleBundle.java
robbertvanginkel: Fixed this with no linker flags and an `install_name_tool` step in the apple bundle build rule.

facebook-github-bot-8:

facebook-github-bot-8:

facebook-github-bot-8:

ryu2:
Looks mostly good.  One thing: you can simplify the code somewhat and get rid of the need for a `InstallToolStep` by using a anonymous subclass of `ShellStep`.

See eaddCoreDataCompilationSteps()e in `AppleBundle` for an example.

facebook-github-bot-8:

facebook-github-bot-4:
@robbertvanginkel updated the pull request - [view changes](https://github.com/facebook/buck/pull/842/files/d9d749aa0a7e9bcb00728a55603a66baca7d9276..103578880e62d38b4642e1202fa2d9ca8e85357b) - [changes since last import](https://github.com/facebook/buck/pull/842/files/d9d749aa0a7e9bcb00728a55603a66baca7d9276..103578880e62d38b4642e1202fa2d9ca8e85357b)

ryu2:
@facebook-github-bot shipit

facebook-github-bot-0:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3722323) to review internal test results.

src/com/facebook/buck/apple/AppleDescriptions.java
```diff
@@ -545,7 +546,8 @@ static AppleBundle createAppleBundle(
         Optional<FrameworkDependencies> frameworkDependencies =
             resolver.requireMetadata(
                 BuildTarget.builder(dep)
-                    .addFlavors(FRAMEWORK_FLAVOR)
```
ryu2: We still need this line; this applies to the dependencies and not the bundle itself.

test/com/facebook/buck/apple/AppleCxxPlatformsTest.java
```diff
@@ -130,6 +130,7 @@ public void iphoneOSSdkPathsBuiltFromDirectory() throws Exception {
         .add(Paths.get("Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib"))
         .add(Paths.get("Toolchains/XcodeDefault.xctoolchain/usr/bin/strip"))
         .add(Paths.get("Toolchains/XcodeDefault.xctoolchain/usr/bin/nm"))
+        .add(Paths.get("Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool"))
```
ryu2: Please also add `install_name_tool` to `FakeAppleRuleDescriptions` so the tests can work.

ryu2:
OK, coming together.  Some tests are failing (`ant test`).  I commented in the areas where I had a quick fix.

Also please fix the lint errors (mostly just line length); you should be able run `ant lint` to find them.

facebook-github-bot-0:
@robbertvanginkel updated the pull request - [view changes](https://github.com/facebook/buck/pull/842/files/103578880e62d38b4642e1202fa2d9ca8e85357b..4d23c50b762275e3ee76b309d75bb4d4dc580c4f) - [changes since last import](https://github.com/facebook/buck/pull/842/files/103578880e62d38b4642e1202fa2d9ca8e85357b..4d23c50b762275e3ee76b309d75bb4d4dc580c4f)

facebook-github-bot-5:
@robbertvanginkel updated the pull request - [view changes](https://github.com/facebook/buck/pull/842/files/4d23c50b762275e3ee76b309d75bb4d4dc580c4f..305c14b9c321c04a3cf2c139ecb8c255cf5002e0) - [changes since last import](https://github.com/facebook/buck/pull/842/files/4d23c50b762275e3ee76b309d75bb4d4dc580c4f..305c14b9c321c04a3cf2c139ecb8c255cf5002e0)

ryu2:
Thanks!  

Still getting a failure here: 

`ant test -Dtest.class=AppleLibraryIntegrationTest`
Testcase: frameworkDependenciesDoNotContainTransitiveDependencies

facebook-github-bot-4:
@robbertvanginkel updated the pull request - [view changes](https://github.com/facebook/buck/pull/842/files/305c14b9c321c04a3cf2c139ecb8c255cf5002e0..5e47bdd46c316a65c3347b9e6c31086ae23c0372) - [changes since last import](https://github.com/facebook/buck/pull/842/files/305c14b9c321c04a3cf2c139ecb8c255cf5002e0..5e47bdd46c316a65c3347b9e6c31086ae23c0372)

Coneko:
Travis build is red:

```
/home/travis/build/facebook/buck/src/com/facebook/buck/cxx/CxxLinkableEnhancer.java:19: error: package com.facebook.buck.apple does not exist
import com.facebook.buck.apple.AppleBundle;
                              ^
/home/travis/build/facebook/buck/src/com/facebook/buck/cxx/CxxLinkableEnhancer.java:20: error: package com.facebook.buck.apple does not exist
import com.facebook.buck.apple.FrameworkDependencies;
                              ^
Errors: 2. Warnings: 0.
BUILD FAILED: //src/com/facebook/buck/cxx:rules failed with exit code 1:
javac
[-] BUILDING...FINISHED 104.3s [100%]
src/com/facebook/buck/cxx/BUCK (:rules) is missing deps:
    '//src/com/facebook/buck/apple:rules',

```

Coneko:
I don't think you can depend on apple stuff from cxx stuff, it would introduce a cyclical dependency.

You can run `buck autodeps` to update the dependencies.

facebook-github-bot-3:
@robbertvanginkel updated the pull request - [view changes](https://github.com/facebook/buck/pull/842/files/5e47bdd46c316a65c3347b9e6c31086ae23c0372..c9915e3fb02e9296139e9902534446866f09378e) - [changes since last import](https://github.com/facebook/buck/pull/842/files/5e47bdd46c316a65c3347b9e6c31086ae23c0372..c9915e3fb02e9296139e9902534446866f09378e)

robbertvanginkel:
I'm working on removing the cyclic dependency.

talgi:
qq, why is packageLibraryAsFramework determined by global config? 
Wouldn't that block users from building some targets as frameworks and some as static libraries in the same build?

jernejstrasner:
@robbertvanginkel any updates on this?

src/com/facebook/buck/apple/umbrella_header.st
```diff
@@ -0,0 +1,6 @@
+#import \<Foundation/Foundation.h>
+
+FOUNDATION_EXPORT double <framework_name>VersionNumber;
+FOUNDATION_EXPORT const unsigned char <framework_name>VersionString[];
+
+<exported_headers:{x| #import \<<x>><\n>}>
```
piotrwach: I would like to be able to define module map myself, as we have a lot of Objective-C, Objective-C++, and C++ headers, and to define module correctly for library like that, we have to specify correct `requires objc/cplusplus` option for each header. This will not be possible with auto generated header.

robbertvanginkel:
Update: after dropping this work for some other prio's I'll be working on this again this week.

robbertvanginkel:
Replacement PR's will start from #1032

PR #843, opened 2016-08-15
[WIP] Add support for swift dependencies

nguyentruongtho:
Maybe I have over simplified the problem but I'm putting up a very thin PR to implement following features.
- Fix `swift_library` depends on `swift_library`
- Fix `apple_library`, `apple_binary` depend on `swift_library`

This will need to be tested thoughtfully (unit test & integration), but I've added a manual test case here: https://github.com/nguyentruongtho/buck_swift_test/tree/master/swift_on_swift

### DIFF
--- a/src/com/facebook/buck/swift/SwiftLibrary.java
+++ b/src/com/facebook/buck/swift/SwiftLibrary.java
@@ -16,6 +16,7 @@
 
 package com.facebook.buck.swift;
 
+import static com.facebook.buck.cxx.CxxPreprocessables.IncludeType.LOCAL;
 import static com.facebook.buck.swift.SwiftUtil.Constants.SWIFT_MAIN_FILENAME;
 import static com.facebook.buck.swift.SwiftUtil.normalizeSwiftModuleName;
 import static com.facebook.buck.swift.SwiftUtil.toSwiftHeaderName;
@@ -46,12 +47,15 @@ import com.facebook.buck.rules.BuildableContext;
 import com.facebook.buck.rules.HasRuntimeDeps;
 import com.facebook.buck.rules.SourcePath;
 import com.facebook.buck.rules.SourcePathResolver;
+import com.facebook.buck.rules.SourcePaths;
 import com.facebook.buck.rules.Tool;
 import com.facebook.buck.rules.args.SourcePathArg;
 import com.facebook.buck.rules.args.StringArg;
 import com.facebook.buck.rules.coercer.FrameworkPath;
 import com.facebook.buck.step.Step;
 import com.facebook.buck.step.fs.MkdirStep;
+import com.facebook.buck.util.MoreIterables;
+import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicate;
 import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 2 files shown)


facebook-github-bot-0:
By analyzing the blame information on this pull request, we identified @ryu2, @yiding and @bhamiltoncx to be potential reviewers.

nguyentruongtho:
This only works for macos platform at this moment as it supports static linking. I'll update this PR for ios soon.

nguyentruongtho:
Interestingly, static linking also works for iOS module, please have a look at the third example from here: https://github.com/nguyentruongtho/buck_swift_test/tree/master/swift_on_swift. I haven't check if cyclic dependences is a problem here. For example: 
- A depends on B, B depends on X
- A depends on C, C depends on X
Will X be statically linked to A twice (compilation error)?

However it would be interesting to link swift libraries statically (I believe it is not documented by Apple).

The problem I had earlier was that, flavors of parent target weren't populated to swift deps before preprocess stage (swift header generation happened before this). Because of that, swift deps used the default cxx platform, `macosx`, specified in `.buckconfig`, as the result parent target couldn't get linked with its swift dependency. When I [specify swift platform explicitly](https://github.com/nguyentruongtho/buck_swift_test/blob/master/swift_on_swift/BUCK#L36), it works.

ryu2:
No, this shouldn't be a problem as we deduplicate linker inputs.  You're still producing a final dylib as output, right?

nguyentruongtho:
getting final dylib would not be a problem, xcode always does that at the end for a swift library compilation. We can just add an additional step for `swift_library` for it.

This is what xcode does:
```
clang
-arch x86_64
-dynamiclib
-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk
-L$BUILD/Products/Debug
-F$BUILD/Products/Debug
-filelist $BUILD/Intermediates/swift_on_swift.build/Debug/dep1.build/Objects-normal/x86_64/dep1.LinkFileList
-install_name /usr/local/lib/libdep1.dylib
-Xlinker
-rpath
-Xlinker @executable_path/../Frameworks
-Xlinker
-rpath
-Xlinker @loader_path/../Frameworks
-mmacosx-version-min=10.11
-Xlinker
-no_deduplicate
-L/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift/macosx
-Xlinker
-add_ast_path
-Xlinker $BUILD/Intermediates/swift_on_swift.build/Debug/dep1.build/Objects-normal/x86_64/dep1.swiftmodule
-single_module
-compatibility_version 1
-current_version 1
-Xlinker
-dependency_info
-Xlinker $BUILD/Intermediates/swift_on_swift.build/Debug/dep1.build/Objects-normal/x86_64/dep1_dependency_info.dat
-o $BUILD/Products/Debug/libdep1.dylib
```

We should run some tests to measure the time to run this task. If it can take long, I think it make sense to make the dylib generation step optional.

The problem I mentioned with the platform flavor is more complicated, I'm working on it right now.

ryu2:
Thanks -- added the `swift_on_swift` test manually; merging now.

ryu2:
@facebook-github-bot shipit

PR #853, opened 2016-08-22
Use the system default XSLT engine when generating junit reports

shs96c:
Because we use the user-provided classpath when running junit tests
it's possible that they will specify an XSLT engine that can not be
loaded. Avoid this problem by using the platform default one.


Coneko:
@facebook-github-bot shipit

facebook-github-bot-0:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3750882) to review internal test results.

PR #859, opened 2016-08-24
Replace Agent receive-file action with ddmlib file push.

zbsz:
Changed `ExopackageInstaller.installFile` to use ddmlib `IDevice.pushFile` instead
of `receive-file` action of android agent app.

`receive-file` was used to upload a file and put it in specified path, using `IDevice.pushFile` is much easier and seems to be working better (at least on my devices).
This was the only place where 'receive-file' was used, so I removed it completely, also from the native app version.

I have no idea why this was done this way, only tested it on couple devices, so this is just a suggestion, someone with more experience (or test devices) should verify if that is fine.


Coneko:
cc @dreiss

facebook-github-bot-9:

dreiss:
We originally added receive-file so we could use run-as to put the files into the app's private data directory.  Now that we put them into /data/local/tmp, this might work fine.  I need to think about it a bit.

dreiss:
Oh, I think I remember the other problem.  This creates files with mode 777 (or 666), so there's a window where a malicious process could tamper with them.  The agent's umask forces the file to be created with the correct permissions, if I recall correctly.

zbsz:
True, this files are in 666, and they stay this way.
Didn't realize that anybody cares about that, isn't exopackage supposed to only be used in development? 
Thanks for info.

dreiss:
Yeah, but I still consider it unsafe to allow any app on the phone to inject code into your development app.

PR #862, opened 2016-08-25
Dynamic library support for swift

nguyentruongtho:
Enable `swift_library` target to create dynamic library.

### DIFF
--- a/src/com/facebook/buck/rules/KnownBuildRuleTypes.java
+++ b/src/com/facebook/buck/rules/KnownBuildRuleTypes.java
@@ -515,7 +515,8 @@ public class KnownBuildRuleTypes {
         new SwiftLibraryDescription(
             cxxBuckConfig,
             cxxPlatforms,
-            platformFlavorsToAppleCxxPlatforms, defaultCxxPlatform);
+            platformFlavorsToAppleCxxPlatforms,
+            defaultCxxPlatform);
     builder.register(swiftLibraryDescription);
 
     CodeSignIdentityStore codeSignIdentityStore =
--- a/src/com/facebook/buck/rules/KnownBuildRuleTypes.java
+++ b/src/com/facebook/buck/rules/KnownBuildRuleTypes.java
@@ -513,9 +513,9 @@ public class KnownBuildRuleTypes {
 
     SwiftLibraryDescription swiftLibraryDescription =
         new SwiftLibraryDescription(
+            cxxBuckConfig,
             cxxPlatforms,
-            platformFlavorsToAppleCxxPlatforms,
-            defaultCxxPlatform);
+            platformFlavorsToAppleCxxPlatforms, defaultCxxPlatform);
     builder.register(swiftLibraryDescription);
 
     CodeSignIdentityStore codeSignIdentityStore =
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -182,21 +182,13 @@ class SwiftCompile
       BuildableContext buildableContext) {
     buildableContext.recordArtifact(outputPath);
     return ImmutableList.of(
-            new MkdirStep(getProjectFilesystem(), outputPath),
-            makeCompileStep());
+        new MkdirStep(getProjectFilesystem(), outputPath),
+        makeCompileStep());
   }
 
   @Override
   public Path getPathToOutput() {
     return outputPath;
-  }
-
-  Path getModulePath() {
-    return modulePath;
-  }
-
-  Path getObjectPath() {
-    return objectPath;
   }
 
   /**
@@ -249,10 +241,10 @@ class SwiftCompile
         .addArgs(
             new SourcePathArg(
                 getResolver(),
-                new BuildTargetSourcePath(getBuildTarget(), getModulePath())))
+                new BuildTargetSourcePath(getBuildTarget(), modulePath)))
         .addArgs(
             new SourcePathArg(
                 getResolver(),
-                new BuildTargetSourcePath(getBuildTarget(), getObjectPath())));
+                new BuildTargetSourcePath(getBuildTarget(), objectPath)));
   }
 }
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -22,6 +22,7 @@ import static com.facebook.buck.swift.SwiftUtil.normalizeSwiftModuleName;
 import static com.facebook.buck.swift.SwiftUtil.toSwiftHeaderName;
 
 import com.facebook.buck.cxx.CxxHeaders;
+import com.facebook.buck.cxx.CxxLink;
 import com.facebook.buck.cxx.CxxPlatform;
 import com.facebook.buck.cxx.CxxPreprocessables;
 import com.facebook.buck.cxx.CxxPreprocessorInput;
@@ -78,6 +79,7 @@ class SwiftCompile
   private final boolean hasMainEntry;
   private final boolean enableObjcInterop;
   private final Optional<SourcePath> bridgingHeader;
+  private final CxxLink delegateLinkRule;
 
   private final Iterable<CxxPreprocessorInput> cxxPreprocessorInputs;
 
@@ -91,6 +93,7 @@ class SwiftCompile
       };
 
   SwiftCompile(
+      CxxLink delegateLinkRule,
       final CxxPlatform cxxPlatform,
       BuildRuleParams params,
       SourcePathResolver resolver,
@@ -101,6 +104,7 @@ class SwiftCompile
       Optional<Boolean> enableObjcInterop,
       Optional<SourcePath> bridgingHeader) throws NoSuchBuildTargetException {
     super(params, resolver);
+    this.delegateLinkRule = delegateLinkRule;
     this.cxxPreprocessorInputs =
         CxxPreprocessables.getTransitiveCxxPreprocessorInput(cxxPlatform, params.getDeps());
     this.swiftCompiler = swiftCompiler;
@@ -177,9 +181,13 @@ class SwiftCompile
       BuildContext context,
       BuildableContext buildableContext) {
     buildableContext.recordArtifact(outputPath);
-    return ImmutableList.of(
-        new MkdirStep(getProjectFilesystem(), outputPath),
-        makeCompileStep());
+    delegateLinkRule.getBuildSteps(context, buildableContext);
+    return ImmutableList.<Step>builder()
+        .add(
+            new MkdirStep(getProjectFilesystem(), outputPath),
+            makeCompileStep())
+        .addAll(delegateLinkRule.getBuildSteps(context, buildableContext))
+        .build();
   }
 
   @Override
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -22,21 +22,24 @@ import static com.facebook.buck.swift.SwiftUtil.normalizeSwiftModuleName;
 import static com.facebook.buck.swift.SwiftUtil.toSwiftHeaderName;
 
 import com.facebook.buck.cxx.CxxHeaders;
-import com.facebook.buck.cxx.CxxLink;
 import com.facebook.buck.cxx.CxxPlatform;
 import com.facebook.buck.cxx.CxxPreprocessables;
 import com.facebook.buck.cxx.CxxPreprocessorInput;
+import com.facebook.buck.cxx.NativeLinkableInput;
 import com.facebook.buck.io.ProjectFilesystem;
 import com.facebook.buck.parser.NoSuchBuildTargetException;
 import com.facebook.buck.rules.AbstractBuildRule;
 import com.facebook.buck.rules.AddToRuleKey;
 import com.facebook.buck.rules.BuildContext;
 import com.facebook.buck.rules.BuildRuleParams;
+import com.facebook.buck.rules.BuildTargetSourcePath;
 import com.facebook.buck.rules.BuildableContext;
 import com.facebook.buck.rules.SourcePath;
 import com.facebook.buck.rules.SourcePathResolver;
 import com.facebook.buck.rules.SourcePaths;
 import com.facebook.buck.rules.Tool;
+import com.facebook.buck.rules.args.SourcePathArg;
+import com.facebook.buck.rules.args.StringArg;
 import com.facebook.buck.step.Step;
 import com.facebook.buck.step.fs.MkdirStep;
 import com.facebook.buck.util.MoreIterables;
@@ -79,7 +82,6 @@ class SwiftCompile
   private final boolean hasMainEntry;
   private final boolean enableObjcInterop;
   private final Optional<SourcePath> bridgingHeader;
-  private final CxxLink delegateLinkRule;
 
   private final Iterable<CxxPreprocessorInput> cxxPreprocessorInputs;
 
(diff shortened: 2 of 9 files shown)


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @ryu2, @andrewjcg and @Coneko to be potential reviewers.

facebook-github-bot-7:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/22aca68b27efcee7a41c4ccf1d104534fa2ee23d..b0a79d69f3fba855c056f9d2655bd3921b54f244) - [changes since last import](https://github.com/facebook/buck/pull/862/files/22aca68b27efcee7a41c4ccf1d104534fa2ee23d..b0a79d69f3fba855c056f9d2655bd3921b54f244)

ryu2:
Thanks!  Please rebase past 8f92c86ddc77bfac25629b3772332b0728892e31.

> Should we make this dylib generation optional
> Currently, I create the dylib inside SwiftCompile, right after the process of creating .swiftmodule and object files since I think it is not necessary to use another flavor (shared_flavor) for it. Any objection?

Eventually, I feel it'd be a good idea to support static libraries on platforms such as OS X which can generate them.  We can do this in a follow-on diff.

>  The name of dylib is currently based on the swift_library target name, maybe we should use moduleName instead.

Is there a case where these would or should ever differ?

> Currently ignore Mach-O-Bundle, maybe we are going to need it for swift_test?

Can do this as a follow-on diff.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -168,9 +172,13 @@ public String apply(SourcePath input) {
       BuildContext context,
       BuildableContext buildableContext) {
     buildableContext.recordArtifact(outputPath);
-    return ImmutableList.of(
-        new MkdirStep(getProjectFilesystem(), outputPath),
-        makeCompileStep());
+    delegateLinkRule.getBuildSteps(context, buildableContext);
```
ryu2: This doesn't seem to be needed.

src/com/facebook/buck/rules/KnownBuildRuleTypes.java
```diff
@@ -513,9 +513,9 @@ static Builder createBuilder(
 
     SwiftLibraryDescription swiftLibraryDescription =
         new SwiftLibraryDescription(
+            cxxBuckConfig,
             cxxPlatforms,
-            platformFlavorsToAppleCxxPlatforms,
-            defaultCxxPlatform);
+            platformFlavorsToAppleCxxPlatforms, defaultCxxPlatform);
```
Coneko: Revert the whitespace change, one argument per line is the correct formatting.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -90,6 +92,7 @@ public String apply(String input) {
       };
 
   SwiftCompile(
+      CxxLink delegateLinkRule,
```
Coneko: We take the output of the compilation and pass it as an input to the linking in cxx, is there a particular reason for this pattern of delegating the build steps here?

src/com/facebook/buck/swift/SwiftCompile.java
nguyentruongtho: image

src/com/facebook/buck/rules/KnownBuildRuleTypes.java
nguyentruongtho: Didn't notice that, IntelliJ should be smarter with style auto-formatting after refactoring.

src/com/facebook/buck/swift/SwiftCompile.java
nguyentruongtho: This is definitely something I'll change. I originally wanted to copy XCode behavior of always creating dylib during `swiftmodule` compilation, this is just 1 to 1 behavior -> implementation mapping.

ryu2:
This diff doesn't work properly with swift libraries that depend on ObjC libraries (see the test in test/com/facebook/buck/swift/testdata/swift_calls_objc)

ryu2:
`ant test -Dtest.class=SwiftOSXBinaryIntegrationTest`

nguyentruongtho:
@ryu2 No worries, I'll update it in a bit. Just tried with my new version, it is working :)

facebook-github-bot-2:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/b0a79d69f3fba855c056f9d2655bd3921b54f244..3f27571a310e883553e3eee714f4472488e84395) - [changes since last import](https://github.com/facebook/buck/pull/862/files/b0a79d69f3fba855c056f9d2655bd3921b54f244..3f27571a310e883553e3eee714f4472488e84395)

nguyentruongtho:
image

@ryu2 @Coneko no more `WIP`, plz help review the PR. Thanks.

facebook-github-bot-9:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/3f27571a310e883553e3eee714f4472488e84395..df35172257e37364db2df7a663fe6724e75419df) - [changes since last import](https://github.com/facebook/buck/pull/862/files/b0a79d69f3fba855c056f9d2655bd3921b54f244..df35172257e37364db2df7a663fe6724e75419df)

ryu2:
@facebook-github-bot shipit

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -238,4 +234,17 @@ Path getObjectPath() {
 
     return args.build();
   }
+
+  void populateLinkableInput(NativeLinkableInput.Builder inputBuilder) {
```
Coneko: Change this to return the args instead of passing in a builder to add them to.

src/com/facebook/buck/swift/SwiftDescriptions.java
```diff
@@ -70,4 +71,15 @@ private SwiftDescriptions() {
     output.bridgingHeader = args.bridgingHeader;
   }
 
+  static boolean isSharedRequested(Linker.LinkableDepType type) {
```
Coneko: Add this method to `Linker.LinkableDepType` instead of here.

src/com/facebook/buck/swift/SwiftRuntimeNativeLinkable.java
```diff
@@ -0,0 +1,109 @@
...
+import com.facebook.buck.rules.args.StringArg;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import java.nio.file.Path;
+
+/**
+ * Pseudo linkable for representing Swift runtime library's linker arguments.
+ */
+final class SwiftRuntimeNativeLinkable implements NativeLinkable {
```
Coneko: The runtime should be flags added to the link, not a `NativeLinkable`.

See `CxxPlatform.getRuntimeLdflags()`.

src/com/facebook/buck/swift/SwiftLibrary.java
```diff
@@ -210,9 +173,28 @@ SwiftCompile requireSwiftCompileRule(Flavor... flavors)
...
+              requiredBuildTarget));
+    }
+    return rule;
+  }
+
   @Override
   public NativeLinkable.Linkage getPreferredLinkage(CxxPlatform cxxPlatform) {
-    return Linkage.ANY;
+    if (getBuildTarget().getFlavors().contains(SWIFT_LIBRARY_FLAVOR)) {
+      return Linkage.STATIC;
```
Coneko: Why this?

Coneko:
Very nice! Just a couple of small comments.

facebook-github-bot-9:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/df35172257e37364db2df7a663fe6724e75419df..926f8c0e0f980e93bd57b820826f28c51824b8d6) - [changes since last import](https://github.com/facebook/buck/pull/862/files/df35172257e37364db2df7a663fe6724e75419df..926f8c0e0f980e93bd57b820826f28c51824b8d6)

src/com/facebook/buck/swift/SwiftLibrary.java
nguyentruongtho: only companion swift library does contain `SWIFT_LIBRARY_FLAVOR`, we don't want to generate dylib for it but link statically swift source and objc in a mix target.

Maybe I should rename `SWIFT_LIBRARY_FLAVOR` to `SWIFT_COMPANION_FLAVOR`.

src/com/facebook/buck/swift/SwiftDescriptions.java
nguyentruongtho: I wonder if we can just do the comparison check: `type == SHARED` instead of this, but I wanted to keep the old code here. I would prefer to use the comparison instead of having a method/attribute in the enum.

nguyentruongtho: hmm, i think i wont use the comparison, Apple could be crazy enough to have another type of dynamic library in the future.

src/com/facebook/buck/swift/SwiftRuntimeNativeLinkable.java
nguyentruongtho: hmm, I thought about it too, but `CxxPlatform` is also used for non-swift targets, make no sense to add swift flags for those targets. Additionally, flag won't work for `getNativeLinkableDeps`, see [SwiftLibrary](https://github.com/nguyentruongtho/buck/blob/df35172257e37364db2df7a663fe6724e75419df/src/com/facebook/buck/swift/SwiftLibrary.java#L109-L115)

src/com/facebook/buck/swift/SwiftDescriptions.java
nguyentruongtho: image

Indeed, we use the comparison everywhere, I guess it is safe to use it again :trollface:

facebook-github-bot-8:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/926f8c0e0f980e93bd57b820826f28c51824b8d6..34fedbea306a4b02bd47d1e5d8847a6bac87f780) - [changes since last import](https://github.com/facebook/buck/pull/862/files/df35172257e37364db2df7a663fe6724e75419df..34fedbea306a4b02bd47d1e5d8847a6bac87f780)

facebook-github-bot-0:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/34fedbea306a4b02bd47d1e5d8847a6bac87f780..aa13fd8f9ee6b8826b88fe36eda54bc44f427439) - [changes since last import](https://github.com/facebook/buck/pull/862/files/df35172257e37364db2df7a663fe6724e75419df..aa13fd8f9ee6b8826b88fe36eda54bc44f427439)

nguyentruongtho:
Sorry for some spam commits, this is when I start to like `arc lint` 😐

facebook-github-bot-3:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/aa13fd8f9ee6b8826b88fe36eda54bc44f427439..3e88e3000e28c8a02f244cdb66a90f6ab55f9603) - [changes since last import](https://github.com/facebook/buck/pull/862/files/df35172257e37364db2df7a663fe6724e75419df..3e88e3000e28c8a02f244cdb66a90f6ab55f9603)

src/com/facebook/buck/swift/SwiftRuntimeNativeLinkable.java
nguyentruongtho: I've updated `SwiftRuntimeNativeLinkable` so that the pseudo build target remains the same for all instances, it will benefit the performance since we don't have to do graph transformation for this class instances more than one.

Coneko: You can add the method for getting the flags to `AppleCxxPlatform` and pass them in from `SwiftLibraryDescription`.

nguyentruongtho: The problem I [mentioned above](https://github.com/facebook/buck/pull/862#discussion_r77158301) doesn't seem to work with flags. By using flags, I will have to put those flags inside `SwiftLibrary#getNativeLinkableInput`, and this will cause duplication.

Let take an example, we have an `apple_binary` A depends on `apple_library` B and another `swift_library` C which B depends on. A has another `swift_library` D. We want to create a binary for A with dynamic linking to all its dependencies (`link_style=shared`). We would ended up with duplication of swift runtime flags like this:

nguyentruongtho: @ryu2 can provide you more context since I took action on his comment about this duplication of runtime flags.

Coneko: The pseudo build target is a problem.

How was using the native linkable solving the duplicate flags before they were all using the same pseudo build target?

Would adding the flags at the binary level work? Although that's going to introduce a massive dependency cycle cxx -> swift -> apple -> cxx.

Alternatively can we create a type of NativeLinkable that doesn't have a build target? But then how do you identify it to deduplicate it?

Still thinking about this.

nguyentruongtho: I think the problem with NativeLinkableInput is that it is concatenated in order (that is why we are using `ImmutableList` in [here](https://github.com/facebook/buck/blob/1e1d14f23616505309e4349bdc83867c75379672/src/com/facebook/buck/cxx/AbstractNativeLinkableInput.java#L62-L62)).

Using pseudo `NativeLinkable` works, since all `NativeLinkable` created in `SwiftLibrary#getNativeLinkableDeps` has the same build target, therefore you will have only one `SwiftRuntimeNativeLinkable` instance in the list of accumulated linkables after this: https://github.com/facebook/buck/blob/b83ae5538a7c553ccd10f06b2ee034dccfa412d6/src/com/facebook/buck/cxx/NativeLinkables.java#L231-L238. 

I've already tested. It works pretty well. There could be still a duplication because of this https://github.com/facebook/buck/blob/34fedbea306a4b02bd47d1e5d8847a6bac87f780/src/com/facebook/buck/swift/SwiftLibraryDescription.java#L287-L287. But there can only be a duplication of 2 at most.

Coneko: I think adding the flags at the binary level would work, and be the best way to do this.

We could discover them using `requireMetadata`, no need to make it a generic mechanism, just find out if any of the dependencies need a swift runtime.

Do you think it would work?

Given that it's a complex undertaking I don't mind it being done in a different PR and merging this one as is, assuming it is a good idea.

nguyentruongtho: I haven't looked at how `metadata` works in buck, which I'm going to dig in more to support `prebuilt_framework` for swift after this. The solution is very clear that we just need a way to communicate with the top level target that it has a swift dependency and it needs swift runtime flags for linking. We just need to find the most suitable implementation.
I would happy to merge this in for now. I'll see what I can do after working with the metadata.

nguyentruongtho: agree that adding the flags at binary level is the best solution, I already did that anyway for `SwiftLibraryDescription` (since we know for sure with this description, swift stuff is included): https://github.com/facebook/buck/blob/34fedbea306a4b02bd47d1e5d8847a6bac87f780/src/com/facebook/buck/swift/SwiftLibraryDescription.java#L287-L287

facebook-github-bot-1:
@nguyentruongtho updated the pull request - [view changes](https://github.com/facebook/buck/pull/862/files/3e88e3000e28c8a02f244cdb66a90f6ab55f9603..dd40cfa7d356adf50e4d47d8a09a82d5e0be1b59) - [changes since last import](https://github.com/facebook/buck/pull/862/files/df35172257e37364db2df7a663fe6724e75419df..dd40cfa7d356adf50e4d47d8a09a82d5e0be1b59)

ryu2:
@facebook-github-bot shipit

PR #866, opened 2016-08-31
Add scala compler_plugins setting.

zbsz:
This change adds setting to enable compiler plugins for Scala libraries. 
Most of the time plugins are needed to use advanced macros, and this case is used in added test.


facebook-github-bot-7:
By analyzing the blame information on this pull request, we identified @mikekap, @aiked and @yiding to be potential reviewers.

Coneko:
Thank you!

PR #869, opened 2016-09-06
Allow exported Maven artifacts to have a template pom

shs96c:
Without this template, there is not enough information available
within the build files to correctly generate a valid and useful,
and more importantly, valid `pom.xml`.

Selenium uses this functionality extensively within its own build
in order to successfully export artifacts to Maven Central.

### DIFF
--- a/src/com/facebook/buck/maven/Pom.java
+++ b/src/com/facebook/buck/maven/Pom.java
@@ -162,8 +162,7 @@ public class Pom {
     }
 
     if (file.isFile()) {
-      Model fromFile = constructModel(file, model);
-      model = merge(model, fromFile);
+      model = constructModel(file, model);
     }
 
     return model;
--- a/src/com/facebook/buck/maven/Pom.java
+++ b/src/com/facebook/buck/maven/Pom.java
@@ -58,9 +58,8 @@ import org.eclipse.aether.artifact.DefaultArtifact;
 
 import java.io.File;
 import java.io.IOException;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.nio.charset.Charset;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashSet;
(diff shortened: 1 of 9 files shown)


facebook-github-bot-9:
By analyzing the blame information on this pull request, we identified @sdwilsh, @aiked and @yiding to be potential reviewers.

shs96c:
For an example of this in action:

https://github.com/SeleniumHQ/selenium/blob/selenium-3.0.0-beta-3/java/client/src/org/openqa/selenium/BUCK#L23

src/com/facebook/buck/maven/Pom.java
```diff
@@ -114,27 +137,212 @@ public Artifact apply(HasMavenCoordinates input) {
...
-      try {
-        ModelBuildingResult modelBuildingResult = modelBuilder.build(modelBuildingRequest);
+      Model fromFile = constructModel(file, model);
+      model = merge(model, fromFile);
+    }
+
+    return model;
+  }
+
+  private Model merge(Model first, @Nullable Model second) {
```
Coneko: This method is massive, is there really no method in the library to do this?

shs96c: I can't find a public method that does this. Most irksome. It's not regular enough to use the JavaBeans APIs to introspect over it either.

sdwilsh:
Can you add a test to handle the case where you pass this in as well?  It looks like all tests you touched just handle the `Optional.absent()` case.

facebook-github-bot-2:
@shs96c updated the pull request - [view changes](https://github.com/facebook/buck/pull/869/files/db2bd46f555b2386a8072862ae0f25c11073bf77..81e3e04813e593ab7ccc2008f7280c553cf2d122)

facebook-github-bot-7:
@shs96c updated the pull request - [view changes](https://github.com/facebook/buck/pull/869/files/81e3e04813e593ab7ccc2008f7280c553cf2d122..87b6a6902b993131c2358711182a527f8b3e1e04)

shs96c:
Added a test

PR #871, opened 2016-09-07
Compiling swift files for appleTV

talgi:
When trying to compile swift files for appletvsimulator platform, the target parameter passed to the swift compiler is wrong.
It is x86_64-apple-appletvsimulator9.2 but should be x86_64-apple-tvos9.2


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @k21, @jernejstrasner and @kangzhang to be potential reviewers.

PR #882, opened 2016-09-15
Place MANIFEST.MF as the second entry of built jars

shs96c:
`JarInputStream.getManifest()` only looks at the first two
entries of a zip to find the manifest. The way buck builds
jar files meant the manifest was the last entry in the zip
and that's clearly Not Right.

This diff also means we correctly merge manifest sections,
rather than using a "last in wins" approach.

Finally, we now read manifests from directories & not just
zip files.


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @aiked, @jkeljo and @Coneko to be potential reviewers.

PR #894, opened 2016-09-22
Provide an option to keep certain resource names during trimming

### LINKED ISSUE #877: Android res/raw resources from libraries are not included in the final APK
We're seeing this issue while trying to integrate Buck into our project. We first noticed this due to a crash on the [jodatime-android](https://github.com/dlew/joda-time-android) library, which uses reflection to load resources from `R.raw`.
Turns out that the raw class for library dependencies (aar) is completely empty when building with buck. 
The same class is correctly populated when building with gradle. Is this a known issue/limitation?

kageiit:
This lets users configure a regex matcher to keep resources that confirm to a particular format when they are accessed via reflection

Fixes #877


facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @dreiss, @andrewjcg and @russellporter to be potential reviewers.

hzsweers:
Related to this, it would probably reduce friction if parity support could be added for Android's keep.xml pattern.

kageiit:
`keep.xml`is a larger change and could be done separately.

facebook-github-bot-1:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/894/files/71bf49a9cf4241d81908b5b5a9c5dfa426dc484c..45108e099237725cff27f518d49a0ccaba3d9acb)

marcinkosiba:
@facebook-github-bot shipit

marcinkosiba:
@facebook-github-bot shipit

marcinkosiba:
@kageiit, actually - there seem to be legit test failures on travis, could you take a look please?

kageiit:
@marcinkosiba thanks for the heads up. The integration test did indeed catch a real error. I updated the code to make the test pass

facebook-github-bot-0:
@kageiit updated the pull request - [view changes](https://github.com/facebook/buck/pull/894/files/4e8bb87ba131b3304d18273f7f51f2a3be12d992..8323011f01534c26e5081e0640138e2dce19f233)

src/com/facebook/buck/android/TrimUberRDotJava.java
```diff
@@ -212,9 +212,12 @@ public Pattern apply(@NonNull String input) {
       // is referenced.  That is a very rare case, though, and not worth the complexity to fix.
       if (m.find()) {
        final String resource = m.group(1);
-       if (!allReferencedResources.contains(packageName + "." + resource) &&
-           (pattern.isPresent() && !pattern.get().matcher(resource).find())) {
-             continue;
+       if (!allReferencedResources.contains(packageName + "." + resource)) {
+         if (pattern.isPresent() && pattern.get().matcher(resource).find()) {
```
marcinkosiba: why can't we integrate this into the condition of the previous `if` statement?

test/com/facebook/buck/android/AndroidBinaryIntegrationTest.java
```diff
@@ -779,10 +779,10 @@ public void testResourcesTrimming() throws IOException {
     workspace.replaceFileContents(
         "apps/multidex/BUCK",
         "# ARGS_FOR_APP",
-        "trim_resource_ids = True,  # ARGS_FOR_APP");
+        "keep_resource_pattern = '^app_.*', trim_resource_ids = True,  # ARGS_FOR_APP");
```
marcinkosiba: newline after comma

src/com/facebook/buck/android/TrimUberRDotJava.java
kageiit: It was before, but that caused the integration test to fail. This is because the continue needs to happen for both conditionals, but the write only needs to happen if the inner conditional is satisfied

facebook-github-bot-1:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3930820).

src/com/facebook/buck/android/TrimUberRDotJava.java
marcinkosiba:
```
      if (m.find()) {
        final String resource = m.group(1);
        if (!allReferencedResources.contains(packageName + "." + resource) &&
            (!pattern.isPresent() || !pattern.get().matcher(resource).find())) {
          continue;
        }
      }
```
though I personally find inverting it is clearer:
```
        boolean shouldWriteLine = allReferencedResources.contains(packageName + "." + resource)
            || (keepPattern.isPresent() && keepPattern.get().matcher(resource).find());
        if (!shouldWriteLine) {
          continue;
        }
```
boolean algebra

kageiit:
Hmm I don't remember touching the unit tests in the last commit. Didn't look because they were green on travis

PR #899, opened 2016-09-24
Add option to specify swift version

nguyentruongtho:
After upgrading to Xcode8, the default swift compiler will be 3.0. This PR will allow buck to specify swift version to be used by adding following setting to the buck config file:

```
[swift]
  version = 2.3
```

Extracted from https://github.com/facebook/buck/pull/893

PR #901, opened 2016-09-26
Fix swift linking when swift version is specified

nguyentruongtho:
Given swift version is specified in .buckconfig, which is different to the default swift, https://github.com/facebook/buck/pull/899 works for standalone swift target, but doesn't cover the case when there are multiple swift targets depend on each other, which will fail at linking time.

This PR will fix that problem.


facebook-github-bot-2:
By analyzing the blame information on this pull request, we identified @Coneko, @jernejstrasner and @yiding to be potential reviewers.

dinhviethoa:
@ryu2 Could you take a look at this one?

nguyentruongtho:
image

facebook-github-bot-9:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3927225).

PR #903, opened 2016-09-27
Add SwiftSupport directory and copy Swift stdlibs when creating .ipa packages

AquaGeek:
To submit an `.ipa` containing Swift code to the App Store for review, the Swift standard libraries need to be copied to a "SwiftSupport" directory at the root of the archive.

### DIFF
--- a/src/com/facebook/buck/apple/AppleBundle.java
+++ b/src/com/facebook/buck/apple/AppleBundle.java
@@ -489,7 +489,8 @@ public class AppleBundle
       addSwiftStdlibStepIfNeeded(
         bundleRoot.resolve(Paths.get("Frameworks")),
         Optional.of(codeSignIdentitySupplier),
-        stepsBuilder
+        stepsBuilder,
+        false /* is for packaging? */
       );
 
       stepsBuilder.add(
@@ -504,7 +505,8 @@ public class AppleBundle
       addSwiftStdlibStepIfNeeded(
         bundleRoot.resolve(Paths.get("Frameworks")),
         Optional.<Supplier<CodeSignIdentity>>absent(),
-        stepsBuilder
+        stepsBuilder,
+        false /* is for packaging? */
       );
     }
 
@@ -674,7 +676,8 @@ public class AppleBundle
   public void addSwiftStdlibStepIfNeeded(
       Path destinationPath,
       Optional<Supplier<CodeSignIdentity>> codeSignIdentitySupplier,
-      ImmutableList.Builder<Step> stepsBuilder) {
+      ImmutableList.Builder<Step> stepsBuilder,
+      boolean isForPackaging) {
     // It's apparently safe to run this even on a non-swift bundle (in that case, no libs
     // are copied over).
     if (swiftStdlibTool.isPresent()) {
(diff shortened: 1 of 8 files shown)


facebook-github-bot-8:
By analyzing the blame information on this pull request, we identified @ryu2, @k21 and @beefon to be potential reviewers.

facebook-github-bot-8:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

ryu2:
lgtm -- as soon as you sign the CLA, we can merge this.  Thanks!

AquaGeek:
I have signed the CLA. I reached out to the email address listed for assistance.

Also, it looks like we need to make further changes to ensure that the standard libraries are copied without being code signed (the current implementation signs them by default). However, there seems to be a bug in `swift-stdlib-tool` where it doesn't properly support the `--unsigned-destination` argument (it crashes due to the missing `--destination` flag). I've opened a radar.

AFAIK, the stdlibs need to be copied without being code signed, so we're going to need to figure out a workaround. Hopefully it doesn't come down to "iterate Frameworks + PlugIns, copy any matching libs manually from the Toolchain dir." 😞

AquaGeek:
OK, it looks like `swift-stdlib-tool` doesn't codesign unless the `--sign` option is provided, so we _should_ be able to use `--destination` instead of `--unsigned-destination` (though once the bug is fixed we probably want to migrate to the latter).

Also, it turns out the libraries were being codesigned in the build phase — the packaging phase was essentially a no-op because the libraries had already been copied (due to the reused temp dir). I added a flag to use a separate temp dir for the packaging phase, and now the libraries are properly copied as-is.

Happy to hear any thoughts on how to make this better (not super happy with the flag approach), but it appears to at least be functioning properly now.

ryu2:
@AquaGeek I think this is probably good for now.  Going forward, if we can see how to sign a directory with existing libraries, we can avoid multiple invocations of `swift-stdlib-tool` which would help build times a bit.

AquaGeek:
Cool. I'll update when I hear back from the CLA folks.

Any idea why `//test/com/facebook/buck/android:integration` is failing?

ryu2:
@AquaGeek Not sure about the failing test, but it was happening before this revision so I wouldn't worry too much about it (some of them are flakey).

nguyentruongtho:
@ryu2 plz don't import it yet as we'll need to add a constraint to not adding SwiftSupport folder when there is no swift file included.

AquaGeek:
@nguyentruongtho The copy swift stdlibs step that the packaging class leverages already takes this into account. I've added another integration test to check that scenario, though.

sdwilsh:
@AquaGeek - it looks like you signed the personal CLA without adding your GitHub handle.  With that said, you need to reach out to tpham in your org to get yourself added to the corporate CLA.

AquaGeek:
@sdwilsh Our point of contact here for the CLA (@myhrvold) has been trying to get ahold of anybody at FB to add me to the CLA. I think he emailed you late last week with the relevant info. Anything else we need to do on our end?

sdwilsh:
I'm oncall this week, so I'm the person new requests go to.  \o/

I see the email from last week, so I'll figure out this process and get 
it moving today (and I'll figure out why it didn't happen last week). 
Sorry for the delay here.  We're actively looking at ways to improve the 
corporate CLA process because it's not great right now.

sdwilsh:
Okay, you've finally been added.  The bot should come along and comment any time now.

sdwilsh:
o_O

Don't mind our bot.  It appears to be drunk.

AquaGeek:
Rebased to fix the merge conflict and did some minor cleanup.

PR #907, opened 2016-09-28
Incremental build for swift - First iteration.

nguyentruongtho:
Chapter one:
- Use swift driver (swiftc) for incremental build (give swiftc input and get information from its output).
- Fix all problems related to linking when switching to incremental build.
- Implement log parsing for output from swift compiler.

TODO (in the second iteration):
- Instead of using swift driver for run all the compilation jobs and caching, we pass them to buck. This is also for parallelizing the jobs.
- Better output parsing.

### DIFF
--- a/src/com/facebook/buck/apple/AppleCxxPlatforms.java
+++ b/src/com/facebook/buck/apple/AppleCxxPlatforms.java
@@ -446,7 +446,6 @@ public class AppleCxxPlatforms {
       ImmutableList<Path> toolSearchPaths,
       ExecutableFinder executableFinder) {
     ImmutableList<String> swiftParams = ImmutableList.of(
-        "-frontend",
         "-sdk",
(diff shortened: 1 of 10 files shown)


facebook-github-bot-5:
By analyzing the blame information on this pull request, we identified @ryu2, @Coneko and @yiding to be potential reviewers.

nguyentruongtho:
I've just discussed with @Coneko on this, we will need to cover an edge case about the building order and caching problem.

This PR is planed for changes. You can have a look to understand what is going on, and what we are trying to do, but this is subject to change soon.

Coneko:
Is this still relevant now that #917 has been merged?

nguyentruongtho:
no, let's close this as there are issues described in https://github.com/facebook/buck/pull/907#issuecomment-250240761, and also swift3 probably changed the way this works.

PR #911, opened 2016-09-30
Include headers in search path during swift compilation(resolve #909)

### LINKED ISSUE #909: SwiftCompile should include headers if bridging header is presented
It seems when a bridging header is presented header symlink trees should be provided to swift compiler via `-I` flag.

I've created an example here: https://github.com/fkorotkov/buck-sandbox/commit/46d25d39a07f52f3e2f981d080c1eef0845ca26d

Basically my bridging header **next to a BUCK** file has `#import "objc/Foo.h"` and _objc/Foo.h_ contains `#import "Version.h"` where _Version.h_ suppose to be next to the BUCK file as well. With  this special setup I'm getting:

```
complex_objc_swift/Foo-Bridging-Header.h:1:
#import "objc/Foo.h"
        ^
/Users/fedor_korotkov/workspace/buck-sandbox/complex_objc_swift/objc/Foo.h:3:9: error: 'Version.h' file not found
#import "Version.h"
        ^
<unknown>:0: error: failed to import bridging header 'complex_objc_swift/Foo-Bridging-Header.h'
```

I was able to fix the command line by adding `-I complex_objc_swift`.

fkorotkov:
Summary:

A bridging header needs headers of the current target to resolve imports. Included private headers if a bridging header is presented.

Test plan:

Added an integration test that was failing before.


fkorotkov:
Note: This does not work for an `apple_test` target since there is an extra `apple-test-library` flavor. Now I'm not sure if this approach is right :-( /cc @ryu2

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -147,6 +149,15 @@ private SwiftCompileStep makeCompileStep() {
       compilerCommand.add(
           "-import-objc-header",
           getResolver().getRelativePath(bridgingHeader.get()).toString());
+
+      // bridging header needs exported headers for imports
+      Path headerPath = CxxDescriptionEnhancer.getHeaderSymlinkTreePath(
+          getProjectFilesystem(),
+          BuildTargets.unflavored(getBuildTarget()),
```
ryu2: @fkorotkov Does it work for tests with with the `apple-test-library` flavor propagated?

fkorotkov: @ryu2 yeah. Doesn't work. I also saw it while testing but didn't quickly find a way to fix it. Ideally I should somehow just get rid of extra flavors(https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/AppleTestDescription.java#L174) instead of hardcoding a check on AppleTestLibrary. 

Do you have any suggestion how can I do it in a clear fashion?

fkorotkov: Yeah. Doing it at the moment. But need to refactor dependencies a bit because I need to filter flavors from `AppleDebugFormat` as well.

This is my additional refactoring https://github.com/facebook/buck/pull/916

src/com/facebook/buck/model/BuildTargets.java
```diff
@@ -116,6 +116,14 @@ public static BuildTarget createFlavoredBuildTarget(
         .build();
   }
 
+  /**
+   * Takes the {@link BuildTarget} and derives a new {@link BuildTarget}
+   * from it with no flavors at all.
+   */
+  public static BuildTarget unflavored(BuildTarget buildTarget) {
```
nguyentruongtho: I really don't like this syntactic sugar.

test/com/facebook/buck/swift/testdata/swift_calls_complex_objc/objc/Foo.h
```diff
@@ -0,0 +1,10 @@
+#import <Foundation/Foundation.h>
+
+// make sure it's available from Foo-Bridging-Header.h
+#import "SwiftCallsComplexObjC/Empty.h"
+
+@interface Foo: NSObject
+
+@property (nonatomic, readonly) NSString *name;
+
+@end
```
nguyentruongtho: newline

test/com/facebook/buck/swift/testdata/swift_calls_complex_objc/.buckconfig
```diff
@@ -0,0 +1,2 @@
+[cxx]
```
nguyentruongtho: do you need this file?

src/com/facebook/buck/swift/SwiftCompile.java
nguyentruongtho: I don't think it's gonna work for swift companion target. In general, I dont think it is a good idea to strip off all flavors like this.

fkorotkov: @nguyentruongtho me too. I'm testing stripping `SwiftLibraryDescription.SUPPORTED_FLAVORS`, `AppleDebugFormat` and `SwiftLibraryDescription.Type` now.

I was looking for a method that returns header tree for a given target but haven't found one.

ryu2: Maybe a better approach is `CxxDescriptionEnhancer.requireHeaderSymlinkTree`?

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -207,6 +210,24 @@ public BuildRule apply(SwiftLibrary input) {
...
               .toSortedSet(Ordering.natural()));
+
+      if (args.exportedHeaders.isPresent()) {
+        UnflavoredBuildTarget unflavoredBuildTarget =
+            params.getBuildTarget().getUnflavoredBuildTarget();
+
+        for (HeaderVisibility headerVisibility : HeaderVisibility.values()) {
+          // unflavoredBuildTarget because #headers can collide with any other flavor
+          // from the same domain.
+          CxxDescriptionEnhancer.requireHeaderSymlinkTree(
```
ryu2: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/AppleTestDescription.java#L177

ryu2: It looks like the flavor isn't actually used for headers; you should be able to safely remove it from `AppleTestDescription` and `AppleTestIntegrationTest`.

cc: @k21 who introduced it just in case there's something I overlooked.

k21: I think that flavor is used to distinguish between targets belonging to the test library from the ones from the test bundle. The headers usually belong to apple_library, so they have that flavor here. It should be fine to remove that flavor but it might require changing `createBuildRule` for the test description and probably some other code that looks up this rule by the build target.

fkorotkov:
^^ just a rebase

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -332,6 +353,7 @@ public static boolean isSwiftTarget(BuildTarget buildTarget) {
     public Optional<SourcePath> bridgingHeader;
     public Optional<ImmutableSortedSet<BuildTarget>> deps;
     public Optional<NativeLinkable.Linkage> preferredLinkage;
+    public Optional<ImmutableMap<Path, SourcePath>> exportedHeaders;
```
ryu2: Need to update `SwiftLibraryIntegrationTest` as well to set `args.exportedHeaders` to `Optional.absent()`

test/com/facebook/buck/swift/SwiftOSXBinaryIntegrationTest.java
```diff
@@ -121,4 +121,18 @@ public void swiftCallingObjCRunsAndPrintsMessageOnOSX() throws IOException {
         runResult.getStdout(),
         containsString("Hello ObjC\n"));
   }
+
+  @Test
```
ryu2: You should use `fbxctest` here.  See https://gist.github.com/ryu2/350a5d773e86fb70d5383cfd1e2c0ee0 for an example.

Unfortunately, after I made that change, it fails with this

```
    [junit] Testcase: swiftCallingComplexObjCRunsAndPrintsMessageOnOSX took 1.007 sec
    [junit] 	Caused an ERROR
    [junit] The same include path maps to multiple files:
    [junit]   Include path: SwiftCallsComplexObjC/FBASLParser.h
    [junit]   Conflicting files:
    [junit]     /private/var/folders/79/r8x23p3x5z962kr42g6qfnt0x713m8/T/junit-temp-path5100542076165353880/fbxctest/bin/FBControlCore.framework/Versions/A/Headers/FBASLParser.h
    [junit]     /private/var/folders/79/r8x23p3x5z962kr42g6qfnt0x713m8/T/junit-temp-path5100542076165353880/fbxctest/bin/FBControlCore.framework/Headers/FBASLParser.h
    [junit] com.facebook.buck.util.HumanReadableException: The same include path maps to multiple files:
    [junit]   Include path: SwiftCallsComplexObjC/FBASLParser.h
    [junit]   Conflicting files:
    [junit]     /private/var/folders/79/r8x23p3x5z962kr42g6qfnt0x713m8/T/junit-temp-path5100542076165353880/fbxctest/bin/FBControlCore.framework/Versions/A/Headers/FBASLParser.h
    [junit]     /private/var/folders/79/r8x23p3x5z962kr42g6qfnt0x713m8/T/junit-temp-path5100542076165353880/fbxctest/bin/FBControlCore.framework/Headers/FBASLParser.h
    [junit] 	at com.facebook.buck.apple.AppleDescriptions.convertToFlatCxxHeaders(AppleDescriptions.java:197)
    [junit] 	at com.facebook.buck.apple.AppleDescriptions.parseAppleHeadersForUseFromOtherTargets(AppleDescriptions.java:151)
    [junit] 	at com.facebook.buck.apple.AppleDescriptions.convertAppleHeadersToPublicCxxHeaders(AppleDescriptions.java:114)
    [junit] 	at com.facebook.buck.apple.AppleDescriptions.populateCxxConstructorArg(AppleDescriptions.java:224)
    [junit] 	at com.facebook.buck.apple.AppleDescriptions.populateCxxLibraryDescriptionArg(AppleDescriptions.java:288)
    [junit] 	at com.facebook.buck.apple.AppleLibraryDescription.requireSingleArchUnstrippedBuildRule(AppleLibraryDescription.java:352)
    [junit] 	at com.facebook.buck.apple.AppleLibraryDescription.requireUnstrippedBuildRule(AppleLibraryDescription.java:330)
    [junit] 	at com.facebook.buck.apple.AppleLibraryDescription.createLibraryBuildRule(AppleLibraryDescription.java:256)
    [junit] 	at com.facebook.buck.apple.AppleTestDescription.createTestLibraryRule(AppleTestDescription.java:366)
    [junit] 	at com.facebook.buck.apple.AppleTestDescription.createBuildRule(AppleTestDescription.java:225)
    [junit] 	at com.facebook.buck.apple.AppleTestDescription.createBuildRule(AppleTestDescription.java:1)
    [junit] 	at com.facebook.buck.rules.DefaultTargetNodeToBuildRuleTransformer.transform(DefaultTargetNodeToBuildRuleTransformer.java:48)
    [junit] 	at com.facebook.buck.rules.BuildRuleResolver.requireRule(BuildRuleResolver.java:120)
    [junit] 	at com.facebook.buck.rules.ActionGraphCache$1.visit(ActionGraphCache.java:152)
    [junit] 	at com.facebook.buck.rules.ActionGraphCache$1.visit(ActionGraphCache.java:1)
    [junit] 	at com.facebook.buck.graph.AbstractBottomUpTraversal.traverse(AbstractBottomUpTraversal.java:62)
    [junit] 	at com.facebook.buck.rules.ActionGraphCache.createActionGraph(ActionGraphCache.java:161)
    [junit] 	at com.facebook.buck.rules.ActionGraphCache.getActionGraph(ActionGraphCache.java:90)
    [junit] 	at com.facebook.buck.cli.TestCommand.runWithoutHelp(TestCommand.java:482)
    [junit] 	at com.facebook.buck.cli.AbstractCommand.run(AbstractCommand.java:198)
    [junit] 	at com.facebook.buck.cli.BuckCommand.run(BuckCommand.java:92)
    [junit] 	at com.facebook.buck.cli.Main.runMainWithExitCode(Main.java:1234)
    [junit] 	at com.facebook.buck.testutil.integration.ProjectWorkspace.runBuckCommandWithEnvironmentOverridesAndContext(ProjectWorkspace.java:478)
    [junit] 	at com.facebook.buck.testutil.integration.ProjectWorkspace.runBuckCommandWithEnvironmentOverridesAndContext(ProjectWorkspace.java:420)
    [junit] 	at com.facebook.buck.testutil.integration.TestDataHelper$CacheClearingProjectWorkspace.runBuckCommandWithEnvironmentOverridesAndContext(TestDataHelper.java:111)
    [junit] 	at com.facebook.buck.testutil.integration.ProjectWorkspace.runBuckCommand(ProjectWorkspace.java:354)
    [junit] 	at com.facebook.buck.swift.SwiftOSXBinaryIntegrationTest.swiftCallingComplexObjCRunsAndPrintsMessageOnOSX(SwiftOSXBinaryIntegrationTest.java:146)
```

fkorotkov:
@ryu2 I've addressed the comments but CI is being failing for days on master so I can't rely on it :-(

ryu2:
@fkorotkov Does this test fail for you locally with "The same include path maps to multiple files"?  It is for me.

```
ant test -Dtest.class=SwiftOSXBinaryIntegrationTest
```

fkorotkov:
@ryu2 indeed! I don't know what's going on there so I moved the test back to just `build` since we are testing a proper compilation.

I've verified though that this test is failing without my changes and it can compile and run tests for my sandbox: https://github.com/fkorotkov/buck-sandbox

src/com/facebook/buck/swift/SwiftDescriptions.java
```diff
@@ -54,6 +59,18 @@ private SwiftDescriptions() {
...
         input -> filterSwiftSources(sourcePathResolver, args.srcs.get()));
+    output.exportedHeaders = args.exportedHeaders.transform(
+         input -> FluentIterable.from(input.getPaths())
+            .uniqueIndex(new Function<SourcePath, Path>() {
+              @Override
+              public Path apply(SourcePath input) {
+                if (input instanceof PathSourcePath) {
+                  return ((PathSourcePath)input).getRelativePath();
+                }
+                return null;
```
ryu2: In theory, this will never be a non `PathSourcePath`.

Instead of `if` you should use `Preconditions.checkArgument(input instanceof PathSourcePath)` to assert this.  Don't return `null` in any case.

src/com/facebook/buck/swift/SwiftDescriptions.java
```diff
@@ -54,6 +59,18 @@ private SwiftDescriptions() {
       BuildTarget buildTarget) {
     output.srcs = args.srcs.transform(
         input -> filterSwiftSources(sourcePathResolver, args.srcs.get()));
+    output.exportedHeaders = args.exportedHeaders.transform(
+         input -> FluentIterable.from(input.getPaths())
+            .uniqueIndex(new Function<SourcePath, Path>() {
+              @Override
+              public Path apply(SourcePath input) {
+                if (input instanceof PathSourcePath) {
+                  return ((PathSourcePath)input).getRelativePath();
```
ryu2: Space after `(PathSourcePath)`

fkorotkov:
@ryu2 nice! Just fixed it! PTAL

src/com/facebook/buck/swift/SwiftDescriptions.java
```diff
@@ -54,6 +60,16 @@ private SwiftDescriptions() {
       BuildTarget buildTarget) {
     output.srcs = args.srcs.transform(
         input -> filterSwiftSources(sourcePathResolver, args.srcs.get()));
+    output.exportedHeaders = args.exportedHeaders.transform(
```
nguyentruongtho: why does swift has exportedHeaders? isn't it something Apple-specific? Does swift on *nix have it?

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -212,7 +233,7 @@ public boolean hasFlavors(ImmutableSet<Flavor> flavors) {
               params.getProjectFilesystem(),
               buildTarget, "%s"),
           args.srcs.get(),
-          Optional.absent(),
+          args.enableObjcInterop,
```
nguyentruongtho: what is the reason for this change?

src/com/facebook/buck/swift/SwiftLibraryDescription.java
nguyentruongtho: Again, can you confirm that swift does generate c-headers on Linux too?

src/com/facebook/buck/swift/SwiftDescriptions.java
piotrwach: We pass a flag to generate it, so yes: https://github.com/facebook/buck/pull/911/files#diff-6d7c1429a28d3416822c55287d317a40R190

nguyentruongtho: ... and also it shouldn't affect compilation on linux machine.

fkorotkov:
@nguyentruongtho sorry I missed the ` c-headers on Linux` but could you please clarify what exactly does it mean? 

Maybe `exportedHeaders` is a miss-leading naming. I should probably rename it to `additionalHeaders` since it's only used if there is a bridging header which needs them.

PR #912, opened 2016-10-01
Add an option to run tests only if they are cached failed

mikekap:
Useful for retrying tests.

### DIFF
--- a/docs/command/test.soy
+++ b/docs/command/test.soy
@@ -32,6 +32,32 @@ contains a <code>tests = ['...']</code> field to specify its tests.</p>
 {/call}
 
 {call buck.param}
+  {param name: 'no-results-cache' /}
+  {param desc}
+  Disables caching results for tests between runs. If specified, all tests will
+  be run regardless of previous results.
+
+  <p>
+  Buck normally caches test results. This means unless you change the test code
+  (or any transitive dependency) buck will not re-run the test. This flag
+  disables the caching entirely and always runs the specified tests.
+  {/param}
+{/call}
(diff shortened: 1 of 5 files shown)


kageiit:
Is the reason for this to retry flaky tests?

mikekap:
Exactly that.

kageiit:
What if a flaky test passed on the first run? There would never be a way to detect a real failure in that case right?

mikekap:
Yes. This does not attempt to fix the flaky test problem generally - that problem already exists with Buck test result caching. What this attempts to do is give the user the chance to re run failing tests without rerunning the whole test suite. It's particularly useful in speeding up flaky CI when there are a lot of tests and only one fails.

kageiit:
So it only reruns the failing tests in a whole set of tests defined by a test rule?

That seems like a good option to have in general as well

mikekap:
It doesn't do that unfortunately. This re-runs the entire test targets that failed. Buck only has test suite unrolling for JVM based tests and that's only for UI. We could in theory generate the proper filter to re-run only the failed test cases here. That might be for another time though.

kageiit:
Got it. so it only preserves results cache if all tests in the target passed. But would definitely be good to have a filter to run the failed tests from previous run without running every other test as well

mikekap:
@yiding @Coneko ping

marcinkosiba:
@mikekap - could you please update the docs to mention this new option?

mikekap:
@marcinkosiba ping

Issue #857, opened 2016-08-23
Add Kotlin Android support


dsyang:
Vanilla kotlin library support was added in https://github.com/facebook/buck/pull/810. the next big step would be to add android support. 

From @bkase's comment on https://github.com/facebook/buck/pull/810#issuecomment-238069818 it looks like the problem is we need the equivalent of `android_library()` for kotlin. I think that could be pretty straight-forward.

 We would need create a new `AndroidKotlinLibraryDescription` that behaves almost exactly like `AndroidLibraryDescription` excepts it compiles with kotlinc instead of javac.  

A quick look through the source code suggests we'd need to do the following:
 - [ ] Refactor `AndroidLibrary` to take in a `BaseCompileToJarStepFactory` (it currently defaults to javac)
 - [ ] Copy `AndroidLibraryDescription` into a `AndroidKotlinLibraryDescription`. Basically doing the same thing except passing in `KotlincToJarStepFactory` to the `AndroidLibrary`
 - [ ] Refactor the common stuff out of the two android library descriptions.

I'm not sure when I'll have time to tackle this yet. If someone else wants to take a shot go for it!

bolinfest:
@bkase Want to sign up for this one?

bkase:
Very busy through September, but afterwards I may have some spare cycles

kageiit:
Does this account for `kapt` support as well?

zbsz:
Got something similar in #913, but instead of creating specific `Android{LANG}LibraryDescription` I opted to use `language` config parameter. So far have it implemented for Scala, adding Kotlin and other jvm languages should be quite straightforward then, just wanted to get it accepted for Scala first.

dsyang:
#937  enables this.  @kageiit I don't think so, but I think it'd be best to open up a new issue for that

PR #913, opened 2016-10-02
Add Scala support to AndroidLibraryDescription.

zbsz:
Added AndroidLibraryCompiler abstraction to allow for different jvm languages be used in android library.
Added language parameter to AndroidLibrary arguments.

It should be quite easy to later add Kotlin and Groovy support using the same abstraction.


kageiit:
@dsyang has some plans for android kotlin support in #857

Adding them to ensure this is compatible

dsyang:
I haven't implemented (or really thought closely) about doing this for
kotlin yet. This sounds like a much better approach. If it works for scala
that's awesome and we should merge it. I'm sure we can do something similar
for kotlin.

Thanks for doing this @zbsz!!

src/com/facebook/buck/android/AndroidLibraryCompiler.java
```diff
@@ -0,0 +1,158 @@
...
+      AndroidLibraryDescription.Arg arg,
+      JavacOptions javacOptions,
+      BuildRuleResolver resolver) {
+
+    return new JavacToJarStepFactory(javacOptions, new BootClasspathAppender());
+  }
+}
+
+
+class ScalaAndroidLibraryCompiler extends AndroidLibraryCompiler {
```
dsyang: Why leave these here instead of in their own file (or package)?

src/com/facebook/buck/android/AndroidLibraryCompiler.java
```diff
@@ -0,0 +1,158 @@
...
+            .addAll(scalaBuckConfig.getCompilerFlags())
+            .addAll(arg.extraArguments.get())
+            .addAll(Iterables.transform(
+                resolver.getAllRules(scalaBuckConfig.getCompilerPlugins()),
+                new Function<BuildRule, String>() {
+                  @Override
+                  public String apply(BuildRule input) {
+                    return "-Xplugin:" + input.getPathToOutput();
+                  }
+                }))
```
dsyang: Maybe we should extract this to a utility/static method instead of duplicating between here and `ScalaLibraryDescription`

src/com/facebook/buck/android/AndroidLibraryCompiler.java
```diff
@@ -0,0 +1,158 @@
...
+                    new Function<String, Path>() {
+                      @Override
+                      public Path apply(String input) {
+                        return new File(input).toPath();
+                      }
+                    }
+                );
+          }
+        }
+    );
```
dsyang: nit: feels like this can (and should) be consolidated

src/com/facebook/buck/android/AndroidLibraryDescription.java
```diff
@@ -164,17 +187,42 @@ public Arg createUnpopulatedConstructorArg() {
...
     public Optional<SourcePath> manifest;
     public Optional<String> resourceUnionPackage;
     public Optional<String> finalRName;
+
+    @Hint(isInput = false)
+    public Optional<JvmLanguage> language;
+  }
+
+  public enum JvmLanguage {
+    JAVA, SCALA
```
dsyang: nit: split into 2 lines

src/com/facebook/buck/android/AndroidLibraryDescription.java
```diff
@@ -47,17 +52,20 @@
...
   public static final BuildRuleType TYPE = BuildRuleType.of("android_library");
 
   private static final Flavor DUMMY_R_DOT_JAVA_FLAVOR =
       AndroidLibraryGraphEnhancer.DUMMY_R_DOT_JAVA_FLAVOR;
 
   private final JavacOptions defaultOptions;
+  private final ScalaBuckConfig scalaBuckConfig;
 
-  public AndroidLibraryDescription(JavacOptions defaultOptions) {
+  public AndroidLibraryDescription(JavacOptions defaultOptions, ScalaBuckConfig scalaConfig) {
```
dsyang: I wonder if it's possible to do this without hardcoding `ScalaBuckConfig` into the default `AndroidLibraryDescription`. It's not going to be scalable when we start implementing support for kotlin (and other jvm languages).

zbsz:
@dsyang thanks for comments. Do you have suggestions how to solve AndroidPlatformTarget and ScalaBuckConfig issues? I'm not very familiar with the code and didn't see any obvious way to do it better, so any tips would be appreciated.

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: Shoulf I add 'AndroidPlatformTarget' to 'BuildContext'? Or is there some better way?

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: Moved to separate files.

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: Fixed

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: Replaced android boot classpath supplier in `BuildContext` with `AndroidPlatformTarget` supplier. Seems better this way.

src/com/facebook/buck/android/AndroidLibraryDescription.java
zbsz: Not sure how to improve that, we could introduce some additional class for this, something like `JvmLanguagesConfig`, but it's just moving the same things from arguments list to fields. Any better suggestions?

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: Actually, I could extract this whole method, will do that later.

src/com/facebook/buck/android/AndroidLibraryDescription.java
dsyang: What if we had a `JvmLanguagesConfig` interface or base class that the`Scala/Kotlin/BuckConfig` classes implement.  Then we use `JvmLanguagesConfig` here. 

2) Pass in an android compiler instead of a buck config. Your `getCompiler()` implementation doesn't look like it uses anything inside this class and can probably be pulled out into some selector class that choose

zbsz: 1) `JvmLanguagesConfig` would need to include almost everything that is in language specific configs, I mean there doesn't seem to be a common subset of settings that would be enough to instantiate any `AndroidLibraryCompiler`
2) `getCompiler` uses `language` param, so it can not be created before `AndroidLibraryDescription` 

3)
Other option would be to have class providing suppliers for all configs, something like:
```
abstract class AndroidLibraryConfig {
    Supplier<ScalaBuckConfig> getScalaConfigSuplier();
    Supplier<KotlinBuckConfig> getKotlinConfigSupplier();
    ...
}
```
All methods would have default implementations throwing exceptions, so there is no need to specify all configs every time.
What do you think about that?

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: To achieve better reuse it could be good to replace `AndroidLibraryCompiler` with common `JvmLibraryCompiler`, this would be more general compiler provider without references to android (placed in `jvm.common`), then `ScalaLIbraryDescription` (and others) would use it the same way `AndroidLibraryDescription` does.
I'm just afraid that this will end up being a bigger refactoring, and I'm not sure if I want to do it as a part of this PR.

src/com/facebook/buck/android/AndroidLibraryDescription.java
dsyang: @Coneko @mikekap @grumpyjames what do you guys think?

src/com/facebook/buck/android/AndroidLibraryCompiler.java
dsyang: Yea I agree. definitely sounds like something we should to in a separate PR

src/com/facebook/buck/android/AndroidLibraryDescription.java
zbsz: Just tried adding `AndroidLibraryCompilerFactory`, in last commit. Not sure if that's worth it, it ads some indirection, library description doesn't need to know all the configs, but factory will still need them.

src/com/facebook/buck/android/AndroidLibraryCompiler.java
```diff
@@ -0,0 +1,68 @@
...
+import com.facebook.buck.rules.BuildRuleResolver;
+import com.facebook.buck.rules.CellPathResolver;
+import com.facebook.buck.rules.ImplicitDepsInferringDescription;
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+
+import java.nio.file.Path;
+
+
+@SuppressWarnings("unused")
```
Coneko: Why is this necessary?

zbsz: Ant complains about unused parameters in default implementation, for example: 'The value of the parameter resolver is not used'.

src/com/facebook/buck/rules/KnownBuildRuleTypes.java
```diff
@@ -525,7 +526,7 @@ static Builder createBuilder(
 
     KotlinBuckConfig kotlinBuckConfig = new KotlinBuckConfig(config);
 
-    ScalaBuckConfig scalaConfig = new ScalaBuckConfig(config);
+    final ScalaBuckConfig scalaConfig = new ScalaBuckConfig(config);
```
Coneko: This doesn't need to be final, I assume it's a leftover from some earlier version of the code.

test/com/facebook/buck/android/AndroidLibraryIntegrationTest.java
```diff
@@ -51,4 +51,20 @@ public void testAndroidLibraryDoesNotUseTransitiveResources() throws IOException
     result.assertFailure();
     assertTrue(result.getStderr().contains("package R does not exist"));
   }
+
+  @Test
```
Coneko: All the other scala integration tests have increased timeouts because of how long they take on CI. Adjust these accordingly as well.

src/com/facebook/buck/rules/BuildContext.java
```diff
@@ -78,8 +50,8 @@ public String get() {
   public abstract BuckEventBus getEventBus();
 
   @Value.Default
-  public Supplier<String> getAndroidBootclasspathSupplier() {
```
Coneko: Please extract this change to a separate PR.

zbsz: Created #915

src/com/facebook/buck/rules/KnownBuildRuleTypes.java
zbsz: fixed, yup forgot to revert that

test/com/facebook/buck/android/testdata/android_project/scala/com/sample/lib/BUCK.fixture
```diff
@@ -0,0 +1,35 @@
+android_library(
+  name = 'lib',
+  srcs = glob(['*.scala']),
+  language = 'SCALA',
```
dsyang: can we also add a test that says compilation will fail if the language parameter isn't something recognized? ("java", "scala",etc...)

test/com/facebook/buck/rules/BuildContextTest.java
```diff
@@ -17,85 +17,29 @@
...
 
-import java.io.File;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.List;
-
 public class BuildContextTest {
 
-  @Test
-  public void testGetAndroidBootclasspathSupplierWithAndroidPlatformTarget() {
```
dsyang: I realize this was extracted into #915 but do we have an equivalent test that covers this case? Seems weird to just remove test cases instead of moving them to a different unit test. I could be wrong but that PR didn't seem to have one.

src/com/facebook/buck/android/AndroidLibraryDescription.java
```diff
@@ -172,10 +198,28 @@ public boolean hasFlavors(ImmutableSet<Flavor> flavors) {
...
+        .findDepsForTargetFromConstructorArgs(buildTarget, cellRoots, constructorArg);
+  }
+
   @SuppressFieldNotInitialized
   public static class Arg extends JavaLibraryDescription.Arg {
     public Optional<SourcePath> manifest;
     public Optional<String> resourceUnionPackage;
     public Optional<String> finalRName;
+
+    @Hint(isInput = false)
```
dsyang: (Just trying to understand the codebase better) why did we need to add this?

src/com/facebook/buck/android/AndroidLibraryCompilerFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.android;
+
+public interface AndroidLibraryCompilerFactory {
```
dsyang: Javadoc

src/com/facebook/buck/android/AndroidLibraryCompiler.java
```diff
@@ -0,0 +1,76 @@
...
+public abstract class AndroidLibraryCompiler
```
dsyang: Javadoc

test/com/facebook/buck/android/testdata/android_project/scala/com/sample/lib/BUCK.fixture
zbsz: Added in `AndroidLibraryIntegrationTest`, although I'm not sure if that's exactly right, this fails on Arg parsing level, so it's not reported as build failure, but `runBuckBuild` throws an exception.

test/com/facebook/buck/rules/BuildContextTest.java
zbsz: You're right, I was just lazy, added test in `AndroidLibraryDescriptionTest`

src/com/facebook/buck/android/AndroidLibraryCompilerFactory.java
zbsz: Added small description, any improvements will be welcome.

test/com/facebook/buck/jvm/java/DefaultJavaLibraryTest.java
```diff
@@ -1382,7 +1384,9 @@ private BuildRule createJavaLibraryRule(ProjectFilesystem projectFilesystem)
           /* providedDeps */ ImmutableSortedSet.<BuildRule>of(),
           /* abiJar */ new FakeSourcePath("abi.jar"),
           /* additionalClasspathEntries */ ImmutableSet.<Path>of(),
-          options.build(),
+          options,
+          options.trackClassUsage(),
+          new JavacToJarStepFactory(options, new BootClasspathAppender()),
```
dsyang: @shs96c Correct me if I'm wrong but your TODO from line 1352 should be relatively simple to fix right? We're making c/f/b/android/BootClasspathAppender public because of this test and I don't think we need to.

@zbsz if you think it's easy to update go for it. Otherwise lets create an issue.

src/com/facebook/buck/android/AndroidLibraryCompilerFactory.java
```diff
@@ -0,0 +1,39 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.android;
+
+import com.facebook.buck.android.AndroidLibraryDescription.JvmLanguage;
+
+/**
+ * Factory providing implementations of {@link AndroidLibraryCompiler}
```
dsyang: How about this?
```
/**
  * Factory providing implementations of {@link AndroidLibraryCompiler}
  * for the specified {@code language}.
  * {@link AndroidLibraryDescription} uses this factory to handle multiple JVM languages.
  * Implementations should provide a compiler implementation for every {@link JvmLanguage}. 
  * See {@link DefaultAndroidLibraryCompiler}
  */
```

src/com/facebook/buck/android/AndroidLibraryCompiler.java
```diff
@@ -0,0 +1,81 @@
...
+import com.facebook.buck.rules.BuildRuleResolver;
+import com.facebook.buck.rules.CellPathResolver;
+import com.facebook.buck.rules.ImplicitDepsInferringDescription;
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+
+import java.nio.file.Path;
+
+/**
+ * Jvm compiler abstraction for android library.
```
dsyang:
```
/**
  * Jvm compiler abstraction for android.
  * Implementations of this class are used in {@link AndroidLibraryDescription} to provide
 * the actual compilation step.
  * This allows us to use different compilers for different {@link JvmLanguage}.
  */

File: src/com/facebook/buck/android/AndroidLibraryDescription.java [hunk 18]
```diff
@@ -172,10 +197,26 @@ public boolean hasFlavors(ImmutableSet<Flavor> flavors) {
...
+
   @SuppressFieldNotInitialized
   public static class Arg extends JavaLibraryDescription.Arg {
     public Optional<SourcePath> manifest;
     public Optional<String> resourceUnionPackage;
     public Optional<String> finalRName;
+    public Optional<JvmLanguage> language;
+  }
+
+  public enum JvmLanguage {
```
zbsz: fixed

src/com/facebook/buck/android/AndroidLibraryCompilerFactory.java
zbsz: replaced

src/com/facebook/buck/android/AndroidLibraryCompiler.java
zbsz: replaced

zbsz:
Rebased with master.
Android integration test keeps failing on Travis, seems to be cause by memory limits, not sure how to fix that.

PR #916, opened 2016-10-04
Optimized dependencies

fkorotkov:
Ran `buck autodeps` to fix dependencies. Mostly removed unused ones.


ryu2:
Our internal version of Buck actually uses those dependencies for stripped-out things in the open source version (e.g. logging to our own internal endpoints).  In theory, we should refactor so that the autodeps are clean for the open-source version, but this PR as is breaks things.

fkorotkov:
My bad :-)

PR #917, opened 2016-10-04
Parallelize swift compilation by compiling file individually

nguyentruongtho:
This PR is one of the effort to rework the incremental build support https://github.com/facebook/buck/pull/907.

Instead of compiling all swift files in a target in one shot, we create build rule for each of the file so that we can compile them in parallel, therefore speed up the compilation process, especially useful for a target with a lot of swift files.

### DIFF
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -404,7 +404,9 @@ class SwiftCompile
           swiftCompiler,
           frameworkPaths,
           moduleName,
-          Optional.of(srcs.iterator().next()),
+          srcs.isEmpty() ?
+              Optional.empty() :
+              Optional.of(srcs.iterator().next()),
           enableObjcInterop,
           bridgingHeader,
           Optional.empty());
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -83,9 +83,6 @@ class SwiftCompile
           .stream()
           .anyMatch(flavor -> flavor.getName().startsWith(COMPILE_FLAVOR_PREFIX));
 
-  // Prepend "-I" before the input with no space (this is required by swift).
-  private static final Function<String, String> PREPEND_INCLUDE_FLAG = INCLUDE_FLAG::concat;
-
   @AddToRuleKey
   private final Tool swiftCompiler;
 
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -21,18 +21,25 @@ import static com.facebook.buck.swift.SwiftUtil.normalizeSwiftModuleName;
 import static com.facebook.buck.swift.SwiftUtil.toSwiftHeaderName;
 
 import com.facebook.buck.cxx.CxxDescriptionEnhancer;
+import com.facebook.buck.cxx.CxxFlavorSanitizer;
 import com.facebook.buck.cxx.CxxHeaders;
 import com.facebook.buck.cxx.CxxPlatform;
 import com.facebook.buck.cxx.CxxPreprocessables;
 import com.facebook.buck.cxx.CxxPreprocessorInput;
 import com.facebook.buck.cxx.HeaderVisibility;
+import com.facebook.buck.io.MorePaths;
 import com.facebook.buck.io.ProjectFilesystem;
 import com.facebook.buck.model.BuildTarget;
+import com.facebook.buck.model.BuildTargets;
+import com.facebook.buck.model.Flavor;
+import com.facebook.buck.model.ImmutableFlavor;
 import com.facebook.buck.parser.NoSuchBuildTargetException;
 import com.facebook.buck.rules.AbstractBuildRule;
 import com.facebook.buck.rules.AddToRuleKey;
 import com.facebook.buck.rules.BuildContext;
+import com.facebook.buck.rules.BuildRule;
 import com.facebook.buck.rules.BuildRuleParams;
+import com.facebook.buck.rules.BuildRuleResolver;
 import com.facebook.buck.rules.BuildTargetSourcePath;
 import com.facebook.buck.rules.BuildableContext;
 import com.facebook.buck.rules.SourcePath;
@@ -40,14 +47,17 @@ import com.facebook.buck.rules.SourcePathResolver;
 import com.facebook.buck.rules.SourcePaths;
 import com.facebook.buck.rules.Tool;
 import com.facebook.buck.rules.args.Arg;
+import com.facebook.buck.rules.args.FileListableLinkerInputArg;
 import com.facebook.buck.rules.args.SourcePathArg;
 import com.facebook.buck.rules.args.StringArg;
 import com.facebook.buck.rules.coercer.FrameworkPath;
 import com.facebook.buck.step.Step;
 import com.facebook.buck.step.fs.MkdirStep;
+import com.facebook.buck.util.MoreCollectors;
 import com.facebook.buck.util.MoreIterables;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -66,6 +76,15 @@ class SwiftCompile
     extends AbstractBuildRule {
 
   private static final String INCLUDE_FLAG = "-I";
+  private static final String COMPILE_FLAVOR_PREFIX = "compile-swift-";
+  private static final Flavor PREPARE_FLAVOR = ImmutableFlavor.of("prepare-compile");
+  private static final java.util.function.Predicate<SwiftCompile> IS_COMPILE_INDIVIDUAL =
+      input -> input.getBuildTarget().getFlavors()
+          .stream()
+          .anyMatch(flavor -> flavor.getName().startsWith(COMPILE_FLAVOR_PREFIX));
+
+  // Prepend "-I" before the input with no space (this is required by swift).
+  private static final Function<String, String> PREPEND_INCLUDE_FLAG = INCLUDE_FLAG::concat;
 
   @AddToRuleKey
   private final Tool swiftCompiler;
@@ -80,7 +99,10 @@ class SwiftCompile
   private final Path objectPath;
 
   @AddToRuleKey
-  private final ImmutableSortedSet<SourcePath> srcs;
+  private final Optional<SourcePath> src;
+
+  @AddToRuleKey
+  private final Optional<SourcePath> bridgingHeader;
 
   private final Path headerPath;
   private final CxxPlatform cxxPlatform;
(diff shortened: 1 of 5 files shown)


src/com/facebook/buck/swift/SwiftPrepareForCompile.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private Step createFileListStep() {
+    ImmutableList<Arg> sourceListArgs = FileListableLinkerInputArg.from(
+        FluentIterable.from(SourcePathArg.from(getResolver(), srcs))
+            .filter(SourcePathArg.class)
+            .toList());
+
+    ProjectFilesystem fileSystem = getProjectFilesystem();
+    Path argFilePath = fileSystem.getRootPath().resolve(
```
nguyentruongtho: we actually don't need argfile for swift, but I don't want to duplicate stuff by creating similar class like `CxxPrepareForLinkStep`.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -175,28 +228,32 @@ private SwiftCompileStep makeCompileStep() {
...
+          "-primary-file",
+          getResolver().getAbsolutePath(src.get()).toString(),
+          "-emit-module-path",
+          modulePath.toString(),
+          "-o",
+          objectPath.toString());
+    } else {
+      compilerCommand.add(
+          "-o",
+          modulePath.toString());
```
nguyentruongtho: When merging all single file outputs, the output is another swift module, not an object file.

nguyentruongtho:
CI became crazy~

nguyentruongtho:
interestingly, travis rebase your diff autoz and build with the rebased version, that's why it failed!

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -265,15 +324,97 @@ public Path getPathToOutput() {
...
+              cxxPlatform,
+              swiftBuckConfig,
+              params.withFlavor(
+                  ImmutableFlavor.of(
+                      COMPILE_FLAVOR_PREFIX + outputName)),
+              sourcePathResolver,
+              swiftCompiler,
+              frameworkPaths,
+              moduleName,
+              outputPath,
```
nguyentruongtho: should we put all individual compiled file to the same folder like this, or use separate folders like cxx precompile?
I thought it could be a race condition when you compile `file1.swift` and `file2.swift` in parallel given `file2.swift` depending on `file1.swift`, but apparently compiling `file2.swift` doesn't result to compiling `file1.swift` (maybe it will do parsing & semantic analysis but won't actually compile `file1.swift`?). This is a really interesting topic and we need to understand more about swift compiler.

Coneko: Use separate output directories. If you need them all in the same directory you will need to create a symlink view of them.

nguyentruongtho: done. Also rebased to resolve conflict

src/com/facebook/buck/swift/SwiftPrepareForCompile.java
```diff
@@ -0,0 +1,89 @@
...
+import com.facebook.buck.step.Step;
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+import java.nio.file.Path;
+
+import javax.annotation.Nullable;
+
+class SwiftPrepareForCompile extends AbstractBuildRule {
```
Coneko: This doesn't need to be a separate build rule, it can be a step in the compilation rule, at least that's how we do it in all other cases.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -176,28 +230,32 @@ private SwiftCompileStep makeCompileStep() {
...
-    for (SourcePath sourcePath : srcs) {
-      compilerCommand.add(getResolver().getRelativePath(sourcePath).toString());
+
+    if (bridgingHeader.isPresent()) {
+      compilerCommand.add(
+          "-import-objc-header",
+          getResolver().getRelativePath(bridgingHeader.get()).toString());
+    }
+
+    if (src.isPresent()) {
```
Coneko: It's really strange to have the rule behave so differently based on whether it has a source or not, can you split this into two rules?

src/com/facebook/buck/swift/SwiftPrepareForCompile.java
nguyentruongtho: no, coz the file list is used by all separate file build rules (dependencies of `SwiftPrepareForCompile` build target), better to create it upfront.

src/com/facebook/buck/swift/SwiftCompile.java
nguyentruongtho: make sense, will do.

nguyentruongtho:
Just rebased for huge java8 changes ^.

src/com/facebook/buck/swift/SwiftPrepareForCompile.java
```diff
@@ -0,0 +1,97 @@
...
+    super(buildRuleParams, resolver);
+    this.srcs = srcs;
+    this.linker = linker;
+    this.fileListPath = BuildTargets.getScratchPath(
+        getProjectFilesystem(), getBuildTarget(), "%s__filelist.txt");
+  }
+
+  private Step createFileListStep() {
+    ImmutableList<Arg> sourceListArgs = FileListableLinkerInputArg.from(
+        SourcePathArg.from(getResolver(), srcs).stream()
```
Coneko: Is this going to work for build target source paths?

If so, then absolute paths would be in the file list, so it shouldn't be cached (see https://github.com/facebook/buck/blob/master/src/com/facebook/buck/rules/SymlinkTree.java#L198-L205 for example).

If not, it should.

src/com/facebook/buck/swift/SwiftPrepareForCompile.java
Coneko: I see, makes sense.

Coneko:
Back to you for the rule split.

src/com/facebook/buck/cxx/CxxCreateArgFileStep.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright 2014-present Facebook, Inc.
```
Coneko: 2016

src/com/facebook/buck/cxx/CxxCreateFileListStep.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright 2014-present Facebook, Inc.
```
Coneko: 2016

src/com/facebook/buck/cxx/CxxCreateArgFileStep.java
```diff
@@ -0,0 +1,101 @@
...
+
+  private final Path argFilePath;
+  private final Path output;
+  private final ImmutableList<Arg> args;
+  private final Linker linker;
+
+  public CxxCreateArgFileStep(
+      Path argFilePath,
+      Path output,
+      ImmutableList<Arg> args,
```
Coneko: It's really strange that this filters out the file list args.

src/com/facebook/buck/cxx/CxxCreateArgFileStep.java
```diff
@@ -0,0 +1,101 @@
...
+    this.args = args;
+    this.linker = linker;
+  }
+
+  private void createArgFile() throws IOException {
+    if (Files.notExists(argFilePath.getParent())) {
+      Files.createDirectories(argFilePath.getParent());
+    }
+    LOG.verbose("Cleaning up previous argfile at " + argFilePath.toString());
+    Files.deleteIfExists(argFilePath);
```
Coneko: No other step deletes its output before writing it, remove this.

src/com/facebook/buck/cxx/CxxCreateFileListStep.java
```diff
@@ -0,0 +1,86 @@
...
+    this.fileListPath = fileListPath;
+    this.args = args;
+  }
+
+  private void createFileList() throws IOException {
+    if (Files.notExists(fileListPath.getParent())) {
+      Files.createDirectories(fileListPath.getParent());
+    }
+    LOG.verbose("Cleaning up previous filelist at " + fileListPath.toString());
+    Files.deleteIfExists(fileListPath);
```
Coneko: Same as above.

Coneko:
Rather than having `CxxCreateArgFileStep` and `CxxCreateFileListStep` encode so much information about the specifics of their use case, making them very unintuitive to use, merge them into a single class that is more reusable and behaves more like a normal step: don't delete the output before writing out the new one, and pass in an optional escaper to use.

Also you still have to address the issue of splitting `SwiftCompile`.

nguyentruongtho:
> Rather than having CxxCreateArgFileStep and CxxCreateFileListStep encode so much information about the specifics of their use case, making them very unintuitive to use, merge them into a single class that is more reusable and behaves more like a normal step: don't delete the output before writing out the new one, and pass in an optional escaper to use.

> Also you still have to address the issue of splitting SwiftCompile.

I preserved `SwiftPrepareForCompile` as a composite step of the other two, one has to follow the other in the right order. Not quite follow the deletion suggestion, I thought we did the same thing in `CxxPrepareForLinkStep`? 

I think creating filelist step is worth to be taken out since it is going to be used in both linking (`CxxLink`) and compiling (`SwiftCompile`), maybe we can remove `CxxCreateArgFileStep`, but it's not clearly better though.

Any suggestion for splitting the step?

Coneko:
I mean add a step that takes a list of args, a path to a file and an optional escaper, and writes out the args to the file optionally using the escaper, and change the code to use this generic step instead of having the two new steps.

This should be done in a separate PR, then rebase this PR on that one.

nguyentruongtho:
hmm, but for `SwiftCompile`, I don't need the args file, `CxxCreateFileListStep` is what I actually need. We don't want to duplicate the logic for creating file list in both `CxxCreateFileListStep` and `CxxPrepareForLinkStep`, that's why I want to split it into 2 steps.

So, if we want to keep `CxxPrepareForLinkStep` class, we need to rename it, since it is going to be used for `SwiftCompile` also, agree?

Coneko:
It's fine to split it into two steps, I'm saying make it two instances of the same step class, instead of making two different step classes.

nguyentruongtho:
Brilliant! Didn't notice that we could make one common step out of the two.

Coneko:
This looks good but is blocked on the test failure in #944.

Coneko:
Back to you to rebase after #944 lands.

nguyentruongtho:
@Coneko Rebased. It should pass the test now but plz don't import it yet. I've just found an edge case when this fails. I'll fix it and add more test soon.

nguyentruongtho:
Pushed a wrong branch, just force pushed again with the right one. But the failing edge case is still not handled.

nguyentruongtho:
ok, let's come back to this mess 😿

Coneko:
Seems like it's not merging cleanly.

steeve:
Btw, I'm actually using a custom buck build from this PR, and it works great !
Hopefully this will get merged soon

nguyentruongtho:
Not going to work more further on this, as we had some benchmarking @Uber and come to a conclusion that building one target in one shot is much faster than building file by file.

steeve:
We had the exact opposite experience. Compiling file by file is much faster (and also incremental).
We can open a new PR if you guys want (we made some modifications).

ultramiraculous:
With a large mixed ObjC/Swift project we had a huge (Around 60%) speedup from building the target all at once and turning on whole module optimization.

nguyentruongtho:
that's very interesting result, can you put up a branch with your modifications?

daedric:
here you go :)
https://github.com/daedric/buck/tree/swift-zenly-rebased

steeve:
We also have some CocoaPods helpers (made in python) to parse and use pods from inside a BUCK file.

nguyentruongtho:
Nice, thanks @daedric and @steeve I'll check if we can get a significant improvement on our side, if yes, we should reopen this and work to get it landed together :).

daedric:
That would be great, do not hesitate asking me question. My branch is unfortunately not clean as it was a bit rushed, but I'm willing to provide any assistance you need to clean it

cheers,

gbrhaz:
@steeve I'd be interested in seeing those CocoaPod scripts - looking at doing this myself. Any chance of putting them up?

nguyentruongtho:
I had a look through https://github.com/daedric/buck/tree/swift-zenly-rebased, there is not a lot of changes comparing to this original pr, I'm glad that it works for your project. 

I think this approach would work for a large target (many files) with a lot of separate parts in it, since in this case, there shouldn't be lot of dependencies between the files/classes, and buck will be able to parallelize the jobs better when compiling per-file targets. I would ague that if this situation happens, we should refactor the big target and split it into smaller targets. However, this approach won't benefit much in general, and doesn't guarantee that it will be always faster, at least in our case, it is slower than compiling per-target with [whole module optimization](https://swift.org/blog/whole-module-optimizations/) enabled.

We found a better way of doing this and will not continue to support this pr anymore, but we will revisit supporting better parallelization compilation for swift later.

PR #922, opened 2016-10-06
Better vim swap file ignore rule

Qix-:
Swap files start from `p` and traverse upwards towards `a`. They do go other places (e.g. `.svp` after `.sw*` gets exhausted) but it's rare and might conflict with other file formats.


Coneko:
Thank you! 🐶

PR #896, opened 2016-09-23
Decouple swift from apple platform

nguyentruongtho:
Code base for supporting swift is getting larger, we should think about how to make buck to build swift platform-independently. This PR is a starting point for that, it is extracted from https://github.com/facebook/buck/pull/893, and should make it easier to specify swift version to be used with buck.

PR #924, opened 2016-10-07
Validate correctness of the tests

nguyentruongtho:
To avoid something like this in the future
https://github.com/facebook/buck/pull/896#discussion_r82299410.

Please feel free to contribute better idea for the validation.

### DIFF
--- a/scripts/travisci_run.sh
+++ b/scripts/travisci_run.sh
@@ -17,6 +17,9 @@ export BUCK_NUM_THREADS=3
 # Make sure that everything builds in case a library is not covered by a test.
 ./bin/buck build --num-threads=$BUCK_NUM_THREADS src/... test/...
 
+# Validate existence of all test rules.
+./bin/buck test --dry-run --all
+
 # Until we get https://github.com/facebook/buck/pull/479, we cannot run the
 # Go tests in Travis
 ./bin/buck test --num-threads=$BUCK_NUM_THREADS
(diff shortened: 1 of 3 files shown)


scripts/travisci_run.sh
```diff
@@ -17,6 +17,9 @@ export BUCK_NUM_THREADS=3
 # Make sure that everything builds in case a library is not covered by a test.
 ./bin/buck build --num-threads=$BUCK_NUM_THREADS src/... test/...
 
+# Validate existence of all test rules.
+./bin/buck test --dry-run --all
+
 # Until we get https://github.com/facebook/buck/pull/479, we cannot run the
```
Coneko: This comment seems stale, the PR was merged.

scripts/travisci_run.sh
```diff
@@ -17,6 +17,9 @@ export BUCK_NUM_THREADS=3
 # Make sure that everything builds in case a library is not covered by a test.
 ./bin/buck build --num-threads=$BUCK_NUM_THREADS src/... test/...
 
+# Validate existence of all test rules.
+./bin/buck test --dry-run --all
```
Coneko: `--dry-run` doesn't work.

scripts/travisci_run.sh
```diff
@@ -17,6 +17,9 @@ export BUCK_NUM_THREADS=3
...
+
 # Until we get https://github.com/facebook/buck/pull/479, we cannot run the
 # Go tests in Travis
 ./bin/buck test --num-threads=$BUCK_NUM_THREADS
```
Coneko: I think you should be able to change this to `./bin/buck test --num-threads=$BUCK_NUM_THREADS //...` to have it resolve all the `tests` attributes.

scripts/travisci_run.sh
nguyentruongtho: shouldn't it be removed from the docs & code?

PR #925, opened 2016-10-07
Use java8 sdk as default in IntelliJ

nguyentruongtho:
`oshi` lib is compiled with java8, but the current intellij project is using java sdk 7, which causes the failure when running test inside Intellij:

image

We should set default java sdk to 8.


nguyentruongtho:
its already in ~.~ https://github.com/facebook/buck/commit/8457c4d9095eaab80a3fa1d0646816b2d09eeb7b

PR #930, opened 2016-10-10
Use realpath instead of readlink in binary wrappers

aarongable:
Readlink directly returns the target of a symlink. If
the symlink is relative, readlink only returns the
relative path. This path is relative to the buck executable
on $PATH, not to the current working directory, so the
Python invocation fails.

Using realpath instead directly resolves the full absolute
path referred to by the symlink, allowing users to use
relative symlinks in their bin/ directory on $PATH.

### DIFF
--- a/bin/buck
+++ b/bin/buck
@@ -5,7 +5,7 @@
 # systems don't have this flag. *sigh*
 BUCK_PATH="$0"
 while [ -h "$BUCK_PATH" ]; do
-  BUCK_PATH=$(readlink "$BUCK_PATH")
+  BUCK_PATH=$(realpath "$BUCK_PATH")
 done
 BUCK_DIR=$(dirname "$BUCK_PATH")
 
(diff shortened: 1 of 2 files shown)


aarongable:
(I'm not sure if facebook-github-bot will auto-update the status here or not, so:)

I am an employee of Google. I just signed the Corporate CLA with three addresses: my corporate (@google.com), open-source (@chromium.org), and personal (@gmail.com) addresses.

sdwilsh:
The corporate CLA process tends to take a bit longer (it's a manual process for lawyers on both ends), but the bot will comment once we have the data that it's been signed :)

davido:
Can we revert this? After pulling new Buck version, build is broken on vanilla Max Os X (El Capitan):

```
/Users/davido/bin/buck: line 8: realpath: command not found
```

And please, don't tell me how i could install `realpath`.

PR #933, opened 2016-10-11
Add a space in time format

kageiit:
Makes it a bit more readable at first glance

### DIFF
--- a/src/com/facebook/buck/event/listener/AbstractConsoleEventBusListener.java
+++ b/src/com/facebook/buck/event/listener/AbstractConsoleEventBusListener.java
@@ -197,7 +197,7 @@ public abstract class AbstractConsoleEventBusListener implements BuckEventListen
   protected String formatElapsedTime(long elapsedTimeMs) {
     long minutes = elapsedTimeMs / 60_000L;
     String seconds = TIME_FORMATTER.format(locale, elapsedTimeMs / 1000.0 - (minutes * 60));
-    return String.format(minutes == 0 ? "%s" : "%2$dm%1$s", seconds, minutes);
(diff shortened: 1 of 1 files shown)


kageiit:
cc @JonShemitz @dreiss

src/com/facebook/buck/event/listener/AbstractConsoleEventBusListener.java
```diff
@@ -197,7 +197,7 @@ public void setPublicAnnouncements(ImmutableList<String> announcements) {
   protected String formatElapsedTime(long elapsedTimeMs) {
     long minutes = elapsedTimeMs / 60_000L;
     String seconds = TIME_FORMATTER.format(locale, elapsedTimeMs / 1000.0 - (minutes * 60));
-    return String.format(minutes == 0 ? "%s" : "%2$dm%1$s", seconds, minutes);
+    return String.format(minutes == 0 ? "%s" : "%2$dm %1$s", seconds, minutes);
```
JonShemitz: Looks reasonable - fwiw, though, I went with this format because it's pretty much what the `time` command uses. I'm not against this change, but I do think there is some value to using what people are used to seeing.

kageiit: What people are used to seeing is totally based on who are seeing it :)

intellij, gradle and other build systems typically report "x minutes and y seconds"

This change only makes it slightly more readable instead of tally changing it :)

kageiit:
@Coneko why does this need revision? Can you please elaborate. The travis failure is not related to this change it is being addressed in #934

Coneko:
I put the PR back on your queue to respond to @JonShemitz's comment, since you have I'll restart the import.

PR #936, opened 2016-10-11
Fixed `No rule found` on some platforms.

fkorotkov:
`//test/com/facebook/buck/apple/BUCK` uses unsafe imports that's why `apple` and `apple_test_integration` targets are only available on `darwin` os which makes CI fail on any other platform.

to: @Coneko @dreiss

### DIFF
--- a/src/com/facebook/buck/apple/BUCK
+++ b/src/com/facebook/buck/apple/BUCK
@@ -1,3 +1,6 @@
+with allow_unsafe_import():
+    from sys import platform as os_platform
+
 PLATFORM_SRCS = [
   'AbstractAppleCxxPlatform.java',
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/apple/BUCK
```diff
@@ -34,10 +34,6 @@ java_immutables_library(
     'GroupedSource',
     'ProvisioningProfileMetadata',
   ],
-  tests = [
```
Coneko: Instead of removing them outright, make them conditional on os being darwin.

fkorotkov:
@Coneko PTAL. I've fixed failing tests here: https://github.com/facebook/buck/pull/941

PR #937, opened 2016-10-11
Enable kotlin support for android_library() targets

dsyang:
As titled. This completes #857 and enables kotlin for android!


src/com/facebook/buck/android/DefaultAndroidLibraryCompilerFactory.java
```diff
@@ -16,22 +16,27 @@
...
 import com.facebook.buck.util.HumanReadableException;
 
 public class DefaultAndroidLibraryCompilerFactory implements AndroidLibraryCompilerFactory {
 
   private final ScalaBuckConfig scalaConfig;
+  private final KotlinBuckConfig kotlinBuckConfig;
 
-  public DefaultAndroidLibraryCompilerFactory(ScalaBuckConfig scalaConfig) {
+  public DefaultAndroidLibraryCompilerFactory(
+      ScalaBuckConfig scalaConfig, KotlinBuckConfig kotlinBuckConfig) {
```
dsyang: Will break this out into multiple lines

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
```diff
@@ -40,6 +41,14 @@
  */
 public abstract class BaseCompileToJarStepFactory implements CompileToJarStepFactory {
 
+  public static final Function<BuildContext, Iterable<Path>> EMPTY_EXTRA_CLASSPATH =
```
dsyang: will update this to java 8 syntax

src/com/facebook/buck/jvm/scala/ScalacToJarStepFactory.java
```diff
@@ -1,4 +1,4 @@
-/*
+<<<</*
```
dsyang: oops. will remove

test/com/facebook/buck/android/AndroidLibraryIntegrationTest.java
```diff
@@ -53,6 +54,26 @@ public void testAndroidLibraryDoesNotUseTransitiveResources() throws IOException
     assertTrue(result.getStderr().contains("package R does not exist"));
   }
 
+  @Test()
```
dsyang: will remove extra parens

test/com/facebook/buck/android/testdata/android_project/kotlin/BUCK.fixture
```diff
@@ -0,0 +1,5 @@
+prebuilt_jar(
+  name='runtime',
+  binary_jar='kotlin-runtime.jar',
+  visibility= [ 'PUBLIC' ],
+)
```
dsyang: will add newline

test/com/facebook/buck/android/testdata/android_project/kotlin/com/sample/lib/Activity.kt
```diff
@@ -0,0 +1,5 @@
+package com.sample.lib
+
+object Activity {
+    var myString = "__FIND_ME_HERE__content=1__GOODBYE__"
+}
```
dsyang: same here

test/com/facebook/buck/android/testdata/android_project/kotlin/com/sample/lib/Sample2.kt
```diff
@@ -0,0 +1,15 @@
...
+import android.os.Bundle
+
+class Sample2 : Activity() {
+
+    override fun onCreate(savedInstanceState: Bundle) {
+        super.onCreate(savedInstanceState)
+
+        setTitle(R.string.title)
+    }
+}
```
dsyang: add newline

test/com/facebook/buck/android/testdata/android_project/kotlin/com/sample/lib/BUCK.fixture
```diff
@@ -0,0 +1,35 @@
+android_library(
+  name = 'lib',
+  srcs = glob(['*.kt']),
+  language = 'KOTLIN',
```
dsyang: We should test this but my feeling is java_library (and by association android_library before #913) expected all of it's sources to be java files. I've made the languages param optional in the latest update to this PR so we can see what that would be like. After testing the old behavior of java/scala we should decide what to do.

bolinfest: The only thing `java_library()` was allowed to receive in its `srcs` in the past were files that ended in `.java` or `.src.zip`. Ignoring `.src.zip` for the moment, `language` seems like something that the `Description` could figure out, so I don't see why this is a big deal. I would rather ban the `language` param quickly before people start relying on it.

Also, I don't see `language` documented on https://buckbuild.com/rule/android_library.html =\

bolinfest: @zbsz So specifically, no, there should not be any backwards-compatibility issues. I believe we would pass those other types of files to `javac` and the compiler should reject them and fail.

sdwilsh:
Are we going to make sure we have Kotlin installed in the travis/appveyor jobs so these tests can run?

dsyang:
Re auto detecting the language. Sure I can remove `JvmLanguage`

Re documentation. Yea I need to write some stuff up for Kotlin. Maybe do a
blog post too when this lands and gets tested.

Re testing & installing: I think we can do something like what @mikekap did
for scala: add a .buckconfig flag that points to a build target for the
compiler. Then we can check in a version of the compiler and use that to
run tests. Since its just a .jar we should be able to run this without a
problem on existing infrastructure.

src/com/facebook/buck/android/DefaultAndroidLibraryCompilerFactory.java
Coneko: Still have to break this one per line.

src/com/facebook/buck/android/KotlinAndroidLibraryCompiler.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Override
+  public boolean trackClassUsage(JavacOptions javacOptions) {
+    return false;
+  }
+
+  @Override
+  public CompileToJarStepFactory compileToJar(
+      AndroidLibraryDescription.Arg args, JavacOptions javacOptions, BuildRuleResolver resolver) {
```
Coneko: One per line.

Coneko:
Seems good, but split out the changes to remove the language and use the file extensions into a separate PR.

src/com/facebook/buck/android/DefaultAndroidLibraryCompilerFactory.java
```diff
@@ -16,23 +16,39 @@
...
+    this.kotlinBuckConfig = kotlinBuckConfig;
   }
 
   @Override
-  public AndroidLibraryCompiler getCompiler(AndroidLibraryDescription.JvmLanguage language) {
-    switch (language) {
-      case JAVA: return new JavaAndroidLibraryCompiler();
-      case SCALA: return new ScalaAndroidLibraryCompiler(scalaConfig);
+  public AndroidLibraryCompiler getCompiler(AndroidLibraryDescription.Arg args) {
+    // We currently expect all non-java languages to only contain files of that extension.
```
Coneko: Do we check this assumption anywhere? If not, we should, and throw a `HumanReadableException` that explains what is wrong.

dsyang:
Removed changes for removing the language param. Will put out the new PR after this lands.

Something's broken with the LogFormatter which causes buck builds to fail. Not sure what though since this code doesn't touch that at all

PR #947, opened 2016-10-14
[IntelliJ] include generated sources for resolution

fkorotkov:
Without these folders many of classes are not resolved in IntelliJ.


Coneko:
What do you mean? Did you run the build in IntelliJ? Some classes are generated as part of it.

Is there a specific instance of a class not being indexed even though you ran a successful build?

fkorotkov:
@Coneko maybe I'm using `buck.iml` incorrectly but I just opened `buck` folder in IJ ultimate 2016.3 EAP and classes like `com.facebook.buck.model.BuildTarget` and some other are not resolved. I found them in `build/gen` and `build/src-gen`. After including them everything is resolved correctly and I don't see any "red code" in IJ.

Coneko:
They're in `build/` because that's the ant build output directory.

You should run the `Make Project` command in IntelliJ, in the `Build` menu.

fkorotkov:
Indeed! Now it's in `build-ij` folder. I was always running `buck build buck` and `buck test <target>` before and haven't tried building from IJ even once. Thanks!

PR #948, opened 2016-10-14
Remove language param and auto-detect jvm compiler for android_library()

dsyang:
Address comments from https://github.com/facebook/buck/pull/937#discussion_r83084051


src/com/facebook/buck/android/DefaultAndroidLibraryCompilerFactory.java
```diff
@@ -32,12 +37,68 @@ public DefaultAndroidLibraryCompilerFactory(
...
+      case KOTLIN:
+        previousExtension = KotlinBuckConfig.EXTENSION;
+        break;
+      case SCALA:
+        previousExtension = ScalaBuckConfig.EXTENSION;
+        break;
     }
-    throw new HumanReadableException("Unsupported `language` parameter value: %s", language);
+    throw new HumanReadableException(
+        "Non-java jvm languages can only have sources of one filetype. Found .%s and .%s files",
```
kageiit: This seems limiting. Why can the sourcepaths not be compiled separately and combined together later? Gradle supports this already for example

kageiit: nevermind, this is fine since there can be multiple android library targets with different jvm languages that can still be combined later

Coneko:
I just realised this isn't going to work with generated sources, or source zips.

dsyang:
@Coneko what specifically about source zips and generated sources do you think this breaks? I think this doesn't break any backwards compatibility (since it tries to default to javac)

Coneko:
Yeah but what if you want to generate kotlin sources or kotlin source zips? If you specify the language explicitly it will work, but the autodetection will always choose Java.

dsyang:
hmm fair enough. @bolinfest  do you have any thoughts on how we can improve this?

marcinkosiba:
@dsyang, you could check whether all SourcePaths are instances of PathSourcePath and require and explicit language setting.
While I support auto-detect in obvious cases I don't agree with removing the possibility to specify the language, should the user want to.

dsyang:
@kageiit 

kotlinc does support mixed source compilation but it's not as straightforward as invoking bin/kotlinc with some flags.  You have to call kotlinc on the .kt files first then call javac to compile the rest (https://discuss.kotlinlang.org/t/compiling-mixed-java-and-kotlin-files-on-the-command-line/1553) 

We should make a new issue for this.

dsyang:
This approach won't work so I'm closing the PR

PR #949, opened 2016-10-14
fix is_git detection when .git dir is located elsewhere 

andrewzhaosc:
The .git dir may be in a different location, and have .git be a text file that points to that dir.  This fixes is_git detection when the .git dir is somewhere else.

https://git-scm.com/docs/gitrepository-layout

### DIFF
--- a/programs/buck_version.py
+++ b/programs/buck_version.py
@@ -34,17 +34,17 @@ class EmptyTempFile(object):
 
 def is_git(dirpath):
     dot_git = os.path.join(dirpath, '.git')
-    if(all([
-        os.path.exists(dot_git),
-        os.path.isdir(dot_git),
-        which('git'),
-        sys.platform != 'cygwin',
-    ])):
-        return True
-    output = check_output(
-        ['git', 'rev-parse', '--is-inside-work-tree'],
-        cwd=dirpath)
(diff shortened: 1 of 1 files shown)


programs/buck_version.py
```diff
@@ -34,12 +34,17 @@ def fileno(self):
...
-        os.path.exists(dot_git),
-        os.path.isdir(dot_git),
-        which('git'),
-        sys.platform != 'cygwin',
-    ])
+    if(which('git') and sys.platform != 'cygwin'):
+        if(os.path.exists(dot_git) and os.path.isdir(dot_git)):
+            return True
+        output = check_output(
+            ['git', 'rev-parse', '--is-inside-work-tree'],
```
kageiit: this assumes git is on the path. it may not be a valid assumption to make

kageiit: nevermind. I see you are checking it above :)

k21:
This breaks when Buck does not run from a Git repo:
```
~$ git rev-parse --is-inside-work-tree; echo $?
fatal: Not a git repository (or any of the parent directories): .git
128
```
Since this is blocking update of the Buck version in Brew, I am going to revert this PR.

Issue #712, opened 2016-04-12
Keep worker tools around between buck invocations


mikekap:
Since compile errors and the like are common, it would be nice if the tool didn't start out cold each time.

mikekap:
I have a proof-of-concept for this at https://github.com/mikekap/buck/tree/persistent-workers if anybody is interested.

Coneko:
This is a good idea and something we wanted to do, but the proof of concept adds a worker hash when since the worker is a binary rule it could use the rule key.

Coneko:
cc @sam-swarr

mikekap:
Totally agree about the rule key, but I couldn't find a way to get it. If nobody else is working on this, I wouldn't mind picking this up, but I think I'd need a tip on how to pull out the rule key - it seems to currently reside deep in `CachingBuildEngine`. I could add something to `ExecutionContext` that exposes a `getRuleKey` - would that be the way to go?

PR #953, opened 2016-10-17
Add the ability to have worker stick around between buck commands

mikekap:
This adds the idea of a `persistent` worker tool as mentioned in #712 . These workers have the same lifetime as `buckd`, unless the worker inputs themselves change, which causes a restart. We use the `RuleKey` via `InitializableFromDisk` to decide when that needs to happen.

### DIFF
--- a/docs/rule/worker_tool.soy
+++ b/docs/rule/worker_tool.soy
@@ -75,6 +75,17 @@
   {/param}
 {/call}
 
+{call buck.arg}
+  {param name: 'persistent' /}
+  {param default: 'False' /}
+  {param desc}
+    If set to true, buck will not restart the tool unless the tool itself changes. This means the
+    tool will persist across multiple buck commands without being shut down and may see the same
+    rule being built more than once. Be careful when using this with tools that don't expect
+    to process the same input (with different contents) twice!
+  {/param}
+{/call}
+
 {/param}
 
 {param examples}
--- a/src/com/facebook/buck/rules/args/WorkerMacroArg.java
+++ b/src/com/facebook/buck/rules/args/WorkerMacroArg.java
@@ -29,8 +29,10 @@ import com.facebook.buck.rules.macros.WorkerMacroExpander;
 import com.facebook.buck.shell.WorkerTool;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.hash.HashCode;
 
 import java.nio.file.Path;
+import java.util.Optional;
 
 public class WorkerMacroArg extends MacroArg {
 
@@ -38,6 +40,7 @@ public class WorkerMacroArg extends MacroArg {
   private final ImmutableList<String> startupCommand;
   private final ImmutableMap<String, String> startupEnvironment;
   private final String jobArgs;
+  private final BuildTarget buildTarget;
 
   public WorkerMacroArg(
       MacroHandler macroHandler,
@@ -67,11 +70,12 @@ public class WorkerMacroArg extends MacroArg {
           unexpanded,
           target));
     }
-    BuildRule workerTool = resolver.getRule(targets.get(0));
+    this.buildTarget = targets.get(0);
+    BuildRule workerTool = resolver.getRule(buildTarget);
     if (!(workerTool instanceof WorkerTool)) {
       throw new MacroException(String.format("%s used in worker macro, \"%s\", of target %s does " +
           "not correspond to a worker_tool",
-          targets.get(0),
+          buildTarget,
           unexpanded,
           target));
     }
@@ -97,6 +101,18 @@ public class WorkerMacroArg extends MacroArg {
 
   public String getStartupArgs() {
     return workerTool.getArgs();
+  }
+
+  public Optional<String> getPersistentWorkerKey() {
+    if (workerTool.isPersistent()) {
+      return Optional.of(buildTarget.getCellPath().toString() + buildTarget.toString());
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  public HashCode getWorkerHash() {
+    return workerTool.getInstanceKey();
   }
 
   public int getMaxWorkers() {
--- a/src/com/facebook/buck/shell/DefaultWorkerTool.java
+++ b/src/com/facebook/buck/shell/DefaultWorkerTool.java
@@ -17,28 +17,43 @@
 package com.facebook.buck.shell;
 
 import com.facebook.buck.model.BuildTargets;
+import com.facebook.buck.rules.AddToRuleKey;
 import com.facebook.buck.rules.BinaryBuildRule;
+import com.facebook.buck.rules.BuildInfo;
+import com.facebook.buck.rules.BuildOutputInitializer;
 import com.facebook.buck.rules.BuildRule;
 import com.facebook.buck.rules.BuildRuleParams;
 import com.facebook.buck.rules.CommandTool;
 import com.facebook.buck.rules.HasRuntimeDeps;
+import com.facebook.buck.rules.InitializableFromDisk;
 import com.facebook.buck.rules.NoopBuildRule;
+import com.facebook.buck.rules.OnDiskBuildInfo;
+import com.facebook.buck.rules.RuleKey;
 import com.facebook.buck.rules.SourcePathResolver;
 import com.facebook.buck.rules.SourcePaths;
 import com.facebook.buck.rules.Tool;
 import com.facebook.buck.util.MoreCollectors;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.hash.HashCode;
 
+import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Map;
+import java.util.Optional;
 
-public class DefaultWorkerTool extends NoopBuildRule implements HasRuntimeDeps, WorkerTool {
+public class DefaultWorkerTool extends NoopBuildRule implements
+    HasRuntimeDeps, WorkerTool, InitializableFromDisk<DefaultWorkerTool.Data> {
+
+  @AddToRuleKey
+  private final String args;
+  @AddToRuleKey
+  private final ImmutableMap<String, String> env;
 
   private final BinaryBuildRule exe;
-  private final String args;
-  private final ImmutableMap<String, String> env;
   private final int maxWorkers;
+  private final boolean isPersistent;
+  private final BuildOutputInitializer<Data> buildOutputInitializer;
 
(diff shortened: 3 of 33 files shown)


docs/rule/worker_tool.soy
```diff
@@ -75,6 +75,17 @@
...
+  {param name: 'persistent' /}
+  {param default: 'False' /}
+  {param desc}
+    If set to true, buck will not restart the tool unless the tool itself changes. This means the
+    tool will persist across multiple buck commands without being shut down and may see the same
+    rule being built more than once. Be careful when using this with tools that don't expect
+    to process the same input (with different contents) twice!
+  {/param}
+{/call}
+
```
Coneko: I don't think this needs to be configurable at the tool level: a global configuration option in `.buckconfig` for migration purposes is ok, but generally worker tools should be safe to persist across multiple commands anyway.

From the tool's perspective there is no difference between being persisted or not.

src/com/facebook/buck/shell/ShBinary.java
```diff
@@ -136,7 +136,8 @@ public Tool getExecutableCommand() {
   // for this rule to be usable.
   @Override
   public ImmutableSortedSet<BuildRule> getRuntimeDeps() {
-    return ImmutableSortedSet.copyOf(getResolver().filterBuildRuleInputs(main));
+    return ImmutableSortedSet.copyOf(getResolver().filterBuildRuleInputs(
```
Coneko: Nice catch!

Can you split this out in a separate PR?

src/com/facebook/buck/shell/WorkerProcess.java
```diff
@@ -147,6 +149,26 @@ public void close() {
...
+          LOG.debug(e2, "Failed to read stderr file");
+        } finally {
+          try {
+            fileReader.get().close();
+          } catch (IOException e3) {
+            LOG.debug(e3, "Failed to close stderr reader");
+          }
+        }
+      }
+
```
Coneko: Please split out these changes in a separate PR.

src/com/facebook/buck/shell/WorkerProcessPool.java
```diff
@@ -16,40 +16,36 @@
...
-import com.google.common.base.Optional;
+import com.google.common.hash.HashCode;
 
 import java.io.IOException;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.Semaphore;
 
 public abstract class WorkerProcessPool {
 
-  public static final int UNLIMITED_CAPACITY = 0;
```
Coneko: Seems like a good change, but please pull it out into a separate PR.

src/com/facebook/buck/step/AbstractTestStep.java
```diff
@@ -83,7 +83,7 @@ public StepExecutionResult execute(ExecutionContext context) throws InterruptedE
     }
     ProcessExecutorParams params = ProcessExecutorParams.builder()
         .setCommand(command)
-        .setDirectory(workingDirectory)
+        .setDirectory(workingDirectory.transform(filesystem::resolve))
```
Coneko: What is this change for?

Coneko:
Seems generally ok, there's a lot of changes that should be split out into separate PRs.

Only practical problem I can see is that if the worker tool changes target we'll keep around persistent copies of all the different targets it had until the daemon restarts.

src/com/facebook/buck/shell/WorkerProcessProtocolZero.java
```diff
@@ -239,6 +239,8 @@ public void close() throws IOException {
       processStdinWriter.close();
       processStdoutReader.endArray();
       processStdoutReader.close();
+    } catch (IllegalStateException ex) {
+      throw new IOException(ex);
```
mikekap: the `processStdoutReader.endArray()` can throw an `IllegalStateException`.

src/com/facebook/buck/step/AbstractTestStep.java
Coneko: Make a separate PR just for this then.

src/com/facebook/buck/shell/WorkerProcessProtocolZero.java
Coneko: But why do you need to wrap it in an `IOException`?

docs/rule/worker_tool.soy
Coneko: I'm just not sure we guarantee we will run the tools only once with certain inputs during the build to begin with, for example consider a multi platform build.

src/com/facebook/buck/shell/WorkerProcessProtocolZero.java
mikekap: Opened #960 and reverted this. Originally the callers only caught `IOException` and this seemed IO-ish, but it's probably more error-proof to just catch `Exception` in callers.

docs/rule/worker_tool.soy
mikekap: Ah I see, I didn't realize that. I think from the perspective of the code, it's still one build though - each time a `BuildCommand` gets created a new `ExecutionContext` gets created with it (which is what holds the ephemeral workers right now).

The issue I'm facing is with the Clojure compiler (though I think Scala may have a similar issue). The basic problem is that loading files - compiled or not - is really slow because it executes a lot of `static`-level code. A good way to "cache" that load step is simply to not unload the output. The current `worker_tool` is perfect for that. Let's say I have 2 rules with dependencies: `//:a -> //:b`.
 - On a clean build of `//:a`, a new `worker_tool` simply compiles `//:b` and keeps it in-memory. Then when `//:a` is compiled, we add the output of `//:b` to the classpath, but the classloader ignores it since the compiled class is already loaded.
 - If we change `//:a`'s input and build `//:a` again, the new `worker_tool` will add `//:b`'s output to the classpath to load the compiled code rather than recompiling `//:b` again.
 - Unfortunately this doesn't work if a new "version" of `//:b` is compiled with the same worker - there's really no way to force unload a class in java cleanly.

With this approach the `worker_tool` never ends up reloading code unless it really needs to. The downside is that targets need precise deps or the builds won't be deterministic. The way that's accomplished in the project I linked above is via an autodeps-like mechanism that reads the code to figure out the exact dependency graph.

Is there a chance we can keep the two kinds of persistent processes coexisting (though perhaps not as a default)?

Coneko: We can leave it a non default for now, but the exact semantics of what executions are going to be done in a single build seem very implementation specific, and not something you can rely on.

It really sounds like something that just happens to work right now and could break at any moment.

Coneko:
Needs a rebase.

Coneko:
Does the persistent worker die correctly if you `kill -9` `buckd`?

src/com/facebook/buck/shell/DefaultWorkerTool.java
```diff
@@ -85,7 +103,47 @@ public int getMaxWorkers() {
...
+
+  @Override
+  public HashCode getInstanceKey() {
+    return buildOutputInitializer.getBuildOutput().getRuleKey().getHashCode();
+  }
+
+  @Override
+  public Data initializeFromDisk(OnDiskBuildInfo onDiskBuildInfo) throws IOException {
+    Optional<RuleKey> ruleKey =
+        onDiskBuildInfo.getRuleKey(BuildInfo.METADATA_KEY_FOR_RULE_KEY);
```
k21: Why does the worker tool need to read its rule key from the disk? Cannot it use a `RuleKeyBuilderFactory`? I would prefer that over relying on mutable external state.

src/com/facebook/buck/rules/args/WorkerMacroArg.java
```diff
@@ -99,6 +103,18 @@ public String getStartupArgs() {
     return workerTool.getArgs();
   }
 
+  public Optional<String> getPersistentWorkerKey() {
+    if (workerTool.isPersistent()) {
+      return Optional.of(buildTarget.toString());
```
k21: The key should probably include the target's cell in some form to avoid false sharing of workers between rules in different cells.

src/com/facebook/buck/shell/DefaultWorkerTool.java
mikekap: This seems to be a unified way to do something regardless whether a rule has been built or restored from cache (i.e. when steps aren't run). At the point `initializeFromDisk` is called, `OnDiskBuildInfo` actually has `getRuleKey` called on it quite a few times, so I think this is actually faster than rehashing the key via the factory. If not - seems like `OnDiskBuildInfo` could use a cache.

mikekap: Sorry - I misread the question there. We could definitely recompute the rule key to avoid relying on the build info, but I don't know if that's much better. `RuleKeyBuilderFactory` will have to hash the source files anyway, which are likely build outputs and as mutable as the written rule key file. If anything, I think it would help if the data in `OnDiskBuildInfo` wasn't coming directly from disk but rather from the data with which the build was started.

mikekap:
@Coneko not sure about the `kill -9` bit. I don't think it's possible to force terminate the worker processes, but they will definitely get their stdin closed. With some luck they will terminate too, but there's no way to be sure of it.

In case it's a more graceful exit, we do tear down the worker pools on daemon shutdown.

src/com/facebook/buck/rules/args/WorkerMacroArg.java
k21: Unfortunately I think outside of the parser the cell prefix is not present in the build targets because it would have to change based on which cell references the target (two cells can be using a different names for another cell or the same name for different cells). Explicitly adding the `cellPath` from the build target should work well though.

src/com/facebook/buck/shell/DefaultWorkerTool.java
k21: The `RuleKeyBuilderFactory` classes have a cache of all computed rule keys on them, so it should be cheap. The part I worry about when using `OnDiskBuildInfo` is that it can only be safely called after the rule has been built and the call to it is buried in the call stack of a method that doesn't communicate this requirement in any way (`WorkerMacroArg#getWorkerHash`). But since violating that would result in an exception which should be caught by the tests, it is probably not such a big deal.

src/com/facebook/buck/rules/args/WorkerMacroArg.java
mikekap: Ah I see. Makes sense - thanks for the explanation!

Coneko:
Strange, github says it can be merged cleanly, but appveyor and our internal CI fail to patch it, can you rebase it please?

PR #957, opened 2016-10-18
Fix sh_binary runtime deps to include the resources

mikekap:
Part of #953


Coneko:
It doesn't matter for this PR but just so you know for next time, since we have Java 8 now we prefer using the stream API.

PR #959, opened 2016-10-18
Fix (at least) Go tests

mikekap:
The working directory was set to be in the middle of nowhere. Part of #953


Coneko:
Importing to run some tests internally, will get back to you.

PR #960, opened 2016-10-18
Catch JSON errors in processes better.

mikekap:
They're not IOExceptions. Part of #953


Coneko:
What other exceptions can happen? Why are they unchecked exceptions if we're meant to catch them?

mikekap:
`JsonReader` throws `IllegalStateException` for some classes of parse errors unfortunately. In particular `beginObject`/`endObject`/`beginArray`/`endArray` all throw `IllegalStateException`s. Otherwise it can throw `MalformedJsonException` (a subclass of `IOException`) and raw `IOException`s.

Coneko:
I see, sounds good.

PR #963, opened 2016-10-18
Exclude all testdata folders

nguyentruongtho:
Just found out `testdata` folder for swift is not excluded, so I put it and all other similar folders to the ignore list.


nguyentruongtho:
Unrelated failure because of special characters:
https://github.com/facebook/buck/blob/779d7cd99fe6c2b8bedc08706e52b9e52b03359f/test/com/facebook/buck/cxx/CxxSourceRuleFactoryTest.java#L502-L502

PR #977, opened 2016-10-28
Publishing to Maven sometimes needs a username and password.

shs96c:
It sometimes needs far more, but this is what we need to get
the Selenium project's jars published to the staging server.

### DIFF
--- a/src/com/facebook/buck/cli/PublishCommand.java
+++ b/src/com/facebook/buck/cli/PublishCommand.java
@@ -75,6 +75,19 @@ public class PublishCommand extends BuildCommand {
       usage = "Just print the artifacts to be published")
   private boolean dryRun = false;
 
+  @Option(
+      name = "--username",
+      aliases = "-u",
+      usage = "User name to use to authenticate with the server")
+  private String username;
+
+  @Option(
+      name = "--password",
+      aliases = "-p",
+      usage = "Password to use to authenticate with the server")
+  private String password;
+
+
   @Override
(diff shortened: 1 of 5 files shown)


src/com/facebook/buck/cli/PublishCommand.java
```diff
@@ -75,6 +75,19 @@
...
+  @Option(
+      name = "--username",
+      aliases = "-u",
+      usage = "User name to use to authenticate with the server")
+  private String username;
+
+  @Option(
+      name = "--password",
+      aliases = "-p",
+      usage = "Password to use to authenticate with the server")
```
sdwilsh: Some people don't like to pass in passwords on the command line (it shows up in history).  Is there a way to provide an option to prompt for it too (I'm not sure if buckd is going to rain heavily on this idea).

LegNeato: Perhaps we copy what the AWS stuff does? It has a multi-layered approach, starting with a file in the homedir, moving to ENV variables, and then finally using values specified on the command line. We could even do he whole `AWS_PROFILE` jazz for the homedir stuff.

shs96c: We could, but I'd prefer to do that in a follow-up diff.

sdwilsh:
To be clear, my comment was a drive-by.  I'm too far removed from Buck's day-to-day development to consider myself qualified to review it, so someone else from the team should look at it.

PR #978, opened 2016-10-28
Signing maven pushes with PGP

shs96c:


shs96c:
Please land PR #977 first!

src/com/facebook/buck/maven/Publisher.java
```diff
@@ -246,15 +273,65 @@ public DeployResult publish(List<Artifact> toPublish) throws DeploymentException
...
+          .addCommand(
+              "gpg",
+              "-ab",
+              "--batch",
+              "--passphrase", signingPassphrase.get(),
+              file.getAbsolutePath())
+          .build();
+      ProcessExecutor.Result result = processExecutor.launchAndExecute(args);
+      if (result.getExitCode() != 0) {
+        throw new HumanReadableException("Unable to sign %s", file);
```
illicitonion: Actually include the stdout/stderr here.

illicitonion:
Throw in a check in PublisherIntegrationTest#testPublishFiles that we get the .asc and I'm happy.

test/com/facebook/buck/maven/PublisherIntegrationTest.java
```diff
@@ -87,4 +108,67 @@ public void testPublishFiles() throws Exception {
...
+          }
+        } catch (IOException e) {
+          throw new RuntimeException(e);
+        }
+        return executor;
+      }
+    };
+
+    publisher.publish(ImmutableSet.of(publishable));
+  }
```
illicitonion: This is missing the assert that the .asc file was actually PUT?

shs96c: If the `.asc` is missing and signing has been enabled, the check at `Publisher.java:322` will cause an exception to be thrown. The test will fail if the executor isn't called, and I've added an assert that every expected command is executed.

test/com/facebook/buck/maven/PublisherIntegrationTest.java
```diff
@@ -87,4 +108,75 @@ public void testPublishFiles() throws Exception {
...
+          throw new RuntimeException(e);
+        }
+        return executor;
+      }
+    };
+
+    publisher.publish(ImmutableSet.of(publishable));
+
+    for (ProcessExecutorParams params : commands.keySet()) {
+      executor.isProcessLaunched(params);
```
illicitonion: Actually assert on this random boolean value that you're just computing? :)

shs96c: Hahaha! I thought the method made the assert already. *facepalm*

v-jizhang:
Sorry we never got back to this! Thank you for your contributions. At this point, we are closing this because it is quite out of date and it is not clear if there is still interest in this feature. If you are still interested in adding PGP signing support, please submit a new PR and we promise to look at it in less than 3 years. Thank you (and sorry) again.

PR #364, opened 2015-06-24
Bump JaCoCo to 0.7.5

somechris:
Summary:
By bumping JaCoCo to 0.7.5, JaCoCo's exec file format version gets
bumped from 1006 to 1007, and we can use current tools to analyze
the jacoco.exec again.

PR #984, opened 2016-11-01
Bump jacoco to 0.7.7, enable code coverage with robolectric tests.

justinmuller:
jacoco [272](https://github.com/jacoco/jacoco/pull/272) broke Robolectric coverage at jacoco 0.7.3
buck #364 bumped jacoco 0.7.2 to 0.7.5

Allow `[test] incl_no_location_classes = true` to be set in `.buckconfig` to pass `inclnolocationclasses=true` to the jacoco javaagent when code coverage is enabled. This will ensure that code coverage is generated when running Android tests.


docs/concept/buckconfig.soy
```diff
@@ -2480,12 +2480,23 @@ your <code>.buckjavaargs</code> file</a>:
 
 {call buckconfig.entry}
   {param section: 'test' /}
+  {param name: 'incl_no_location_classes' /}
+  {param example_value: 'true' /}
+  {param description}
+    This specifies whether jacoco code coverage is enabled on classes without source location.
+    The default is false.  Set to true to enable code coverage with robolectric tests, although
+    setting to true may cause issues with generated source such as jmockit.
```
styurin: what kind of issues with generated code?

src/com/facebook/buck/cli/BuildCommand.java
```diff
@@ -302,6 +302,7 @@ Build createBuild(
         console,
         buckConfig.getDefaultTestTimeoutMillis(),
         isCodeCoverageEnabled(),
+        buckConfig.getBooleanValue("test", "incl_no_location_classes", false),
```
styurin: why it's not true by default? because of issues with generated code?

justinmuller: Correct.  For example, if this were true by default the classes generated by jmockit would also be included in code coverage.

PR #985, opened 2016-11-01
Implement --wait-for-debugger (-w) for install android apk command

korniltsev:
Pass -D flag to "am start" command if -w set. Should be used in together with -r.


marcinkosiba:
Thanks! It would be awesome if we could also warn the user if the app might not be debuggable. Because of the fantastic stuff we do to manifests it might be a bit hard to check, but for AndroidBinary we know the packageType, so there's at least that..

Issue #991, opened 2016-11-03
ObjC imports not found when trying to importing inter-op apple_library()


stowy:
Check out the Readme at https://github.com/stowy/buck_swift_test/tree/dependencyTests

Or just do:

```
git clone git@github.com:stowy/buck_swift_test.git
git checkout origin/dependencyTests
buck build :MainBinary
```

### Bug description
- The problem is, building `:MainBinary`, when imported any headers from a MixedDependency, fails with header not found. 
- When not importing a header, even with the dependency still in place, it builds correctly
- To test, just clone this branch and run `buck build :MainBinary` and you should get the following error:

```
$ fbuck build :MainBinary
Not using buckd because watchman isn't installed.
MainApp/AppDelegate.m:5:9: fatal error: 'MixedDependency1/MixedDep1.h' file not found
#import "MixedDependency1/MixedDep1.h"
        ^
1 error generated.

BUILD FAILED: //:MainBinary#compile-AppDelegate.m.o4513ea3e,iphonesimulator-x86_64 failed with exit code 1:
objective-c compile_munge_debuginfo
[-] PROCESSING BUCK FILES...FINISHED 0.1s [100%] 🐳  New buck daemon
[-] DOWNLOADING... (0.00 B/S AVG, TOTAL: 0.00 B, 0 Artifacts)
[-] BUILDING...FINISHED 0.6s [100%] (11/13 JOBS, 0 UPDATED, 0 [0.0%] CACHE MISS)
```
- Note that if you remove the line `#import "MixedDependency1/MixedDep1.h"` from the AppDelegate in `MainApp/AppDelegate.m`, it will build successfully.
- Note also that the other apple_libraries with the same dependency (eg `:MixedDependency2`, `:MixedDependency3`) also build and run tests successfully, with dependencies

@robbertvanginkel

stowy:
Looking into this further, it seems that the file organisation in the 'buck-out/gen' folder for the ObjC headers changes when the 'swift-compile' step is included in the build process (because of the presence of swift in the library).

See this file output below, notice that the headers are included in a single flat directory for the ObjC output, but in a hierarchical directory structure for the swift one. The libraries that rely on this MixedDependency1 then fail to compile because headers are not found. 

@nguyentruongtho

<img width="771" alt="screenshot 2016-11-04 09 16 31" src="https://cloud.githubusercontent.com/assets/5924628/20013204/8fd985e4-a26f-11e6-8926-0a2e2e5d8e41.png">

[...]

PR #996, opened 2016-11-05
Added interop tests and fixed header map for Swift step

stowy:
Fixes issue: https://github.com/facebook/buck/issues/991

Probably needs a lot of tidy up and more tests but it's a start.

@nguyentruongtho


nguyentruongtho:
Can you briefly explain about it, there is too many files to look into.

stowy:
Sure sorry, the main change is at https://github.com/stowy/buck/blob/21d4cf21eacb3191c774973772b354db37043b51/src/com/facebook/buck/swift/SwiftDescriptions.java#L67

It changes the header map to be the shortened version of `{ 'HeaderPrefixPath/Filename.h' : 'absolute/path/to/file/filename.h'}` instead of what it is currently, which is `{ 'relative/path/to/Filename.h' : 'absolute/path/to/file/filename.h'}`.

stowy:
The remainder of the change is to add an interop integration test (https://github.com/stowy/buck/blob/21d4cf21eacb3191c774973772b354db37043b51/test/com/facebook/buck/swift/InteropLibraryIntegrationTest.java#L41) which replicates issue https://github.com/facebook/buck/issues/991 - this passes on this branch but not in the main project.

nguyentruongtho:
> Sure sorry, the main change is at https://github.com/stowy/buck/blob/21d4cf21eacb3191c774973772b354db37043b51/src/com/facebook/buck/swift/SwiftDescriptions.java#L67
It changes the header map to be the shortened version of { 'header_prefix_path/filename.h' : 'absolute/path/to/file/filename.h'} instead of what it is currently, which is { 'relative/path/to/filename.h' : 'absolute/path/to/file/filename.h'}. The former allows normal imports in ObjC, the latter breaks ObjC imports.

I think you are right, swift companion target should know about its parent `header_path_prefix` and should be able to search for its parent header during compilation. The approach looks good, I suggest you to create a simpler sample for the issue.

Since this is issue regarding headers, @ryu2 would know more. And also, there was some work on this part was done by @fkorotkov https://github.com/facebook/buck/pull/911. Are you guys reviewing this PR?

nguyentruongtho:
You might not need this PR. Indeed, I'm finishing up something that could cover this case, and also the previous case https://github.com/facebook/buck/pull/911.

fkorotkov:
Part about headers seems reasonable to me. Seems my logic had bugs for more complicated project structure.

PR #1003, opened 2016-11-10
fix is_git detection when .git dir is located elsewhere

andrewzhaosc:
Fixed the problem I had in #949

### DIFF
--- a/programs/buck_version.py
+++ b/programs/buck_version.py
@@ -34,17 +34,17 @@ class EmptyTempFile(object):
 
 def is_git(dirpath):
     dot_git = os.path.join(dirpath, '.git')
-    if(all([
-        os.path.exists(dot_git),
-        os.path.isdir(dot_git),
-        which('git'),
-        sys.platform != 'cygwin',
-    ])):
-        return True
-    output = check_output(
-        ['git', 'rev-parse', '--is-inside-work-tree'],
-        cwd=dirpath)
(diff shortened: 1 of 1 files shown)


programs/buck_version.py
```diff
@@ -34,12 +34,20 @@ def fileno(self):
...
 def is_git(dirpath):
     dot_git = os.path.join(dirpath, '.git')
-    return all([
-        os.path.exists(dot_git),
-        os.path.isdir(dot_git),
-        which('git'),
-        sys.platform != 'cygwin',
-    ])
+    if(which('git') and sys.platform != 'cygwin'):
+        if(os.path.exists(dot_git) and os.path.isdir(dot_git)):
```
k21: In Python code we generally don't use parentheses around the condition in if statements, please remove those here and on the line above.

programs/buck_version.py
```diff
@@ -34,12 +34,20 @@ def fileno(self):
...
+            return True
+        try:
+            output = check_output(
+                ['git', 'rev-parse', '--is-inside-work-tree'],
+                cwd=dirpath)
+            return output.strip() == 'true'
+        except:
```
k21: If you want to use exceptions here, change this block to only handle the exact error type that gets raised when the process exits with a nonzero code. Although I would prefer using `subprocess.run` instead and checking the `returncode` manually.

PR #1005, opened 2016-11-11
Add rt.jar to java bootclasspath

seanabraham:
This is in the same vein as https://github.com/facebook/buck/pull/732. 

This adds `rt.jar` to the `javac` classpath so that buck can build java 8 sources correctly. This is needed, for example, with the use of retrolambda in an Android project.

Previously we worked around this by adding something like the following to our `android_library` rules (a la https://github.com/uber/okbuck/pull/81):

```
 extra_arguments = [
  ...
  '-bootclasspath',
  '/Users/some_user/path/to/android-24/android.jar:/Library/Java/JavaVirtualMachines/jdk1.8.0_74.jdk/Contents/Home/jre/lib/rt.jar',
 ],
```

The problem with this is that the absolute paths in the BUCK files break caching as the jvm arguments are part of definition of the rule key as far as I can tell.

cc @kageiit @felipecsl

### DIFF
--- a/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java
+++ b/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java
@@ -440,6 +440,7 @@ public class DefaultJavaLibrary extends AbstractBuildRule
     ImmutableSortedSet<Path> declared = ImmutableSortedSet.<Path>naturalOrder()
         .addAll(declaredClasspaths)
         .addAll(additionalClasspathEntries)
+        .add(Paths.get(System.getProperty("java.home")).resolve("lib").resolve("rt.jar"))
         .addAll(provided)
         .build();
 


kageiit:
Can we add tools.jar to the bootclasspath as well @dreiss

dreiss:
Hrm.  Is this for android_library or java_library?  For android_library, you should be building against android.jar instead of rt.jar.  Does android.jar not have the classes you need for lambdas?

kageiit:
android.jar is already in the boot class path, but it does not have the required classes.

The retrolambda gradle plugin also adds rt.jar to compile class path for it to work correctly.
https://github.com/evant/gradle-retrolambda/blob/37b4418274553022b540084f139d1983454ac556/gradle-retrolambda/src/main/groovy/me/tatarka/RetrolambdaPluginAndroid.groovy#L78

dreiss:
Discussed this in chat.  Okay to close?

seanabraham:
Yup, will close now. For future readers (especially those also using okbuck), this was resolved via: https://github.com/uber/okbuck/pull/239 and https://github.com/uber/okbuck/pull/241

PR #983, opened 2016-11-01
Allow apple_library to be imported as a module

nguyentruongtho:
This PR will enable Swift target to be able to import apple_library. Quite a few things are imported from https://github.com/facebook/buck/pull/842

PR #1007, opened 2016-11-11
Revert "Include headers in search path during swift compilation"

nguyentruongtho:
This reverts https://github.com/facebook/buck/pull/911, with some minor fixes after merging.

The reason is here: https://github.com/facebook/buck/pull/983#issuecomment-259938575

/cc @ryu2


ryu2:
@facebook-github-bot shipit

PR #1012, opened 2016-11-13
WIP: Add cgo support.

daedric:
This is a work in progress to add CGo support in Buck based on @mikekap work.

There are still a couple of issue to fix:

* On Mac OS X this does not totally works:
 * ~-fno-pie does not exist and I did not yet find a standard way to check C/CXX compiler flags availability~ -> I've a switch-case on the platform;
 * The scrub on Mac OS X does not like, at all, the generated binary
* Unit tests

But before going further I'd like a feedback on how I've been doing.
First, about cgo
* cgo sources *need* to be handled specifically by the cgo tool, the cgo tool acts a bit like a preprocessor
* Thus upon cgo tool execution a couple of file are generated:
 * some C files from which at first we generate 1 binary and from it, we generate a go file
 * Then again, we regenerate a binary that will be linked by go during the last stage
* All the generated Go sources and "regular" Go sources are to be compiled together and then linked.

First I've implemented a simple Cgo support with only Go sources. It allows one to have C source in Go source.
The second step was to allow Cgo extension to be linked against a native library and let all the transitivity rules kicking in (esp. w.r.t the include paths).
The main caevat of using buck for Cgo sources is that we _can't_ depends on the {c/l}flags in the sources because we don't use `go build`

The approach taken minimally changed the existing code.

### DIFF
--- a/src/com/facebook/buck/cxx/CxxBinaryDescription.java
+++ b/src/com/facebook/buck/cxx/CxxBinaryDescription.java
@@ -341,6 +341,10 @@ public class CxxBinaryDescription implements
     return defaultCxxPlatform;
   }
 
+  public CxxBuckConfig getCxxBuckConfig() {
+    return cxxBuckConfig;
+  }
+
   @Override
   public <A extends Arg, U> Optional<U> createMetadata(
       BuildTarget buildTarget,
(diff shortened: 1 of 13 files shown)


mkaczanowski:
hey @daedric I have just landed cgo support for buck. Since your diff was a year old I reworked quite a bit.
Thanks a lot for contributing and enjoy the cgo feature in buck!

Commit: https://github.com/facebook/buck/commit/28b54757c5eb81137eae088dc7327b5c0f696701

PR #1013, opened 2016-11-13
ideabuck: tests debugging

korniltsev:
In this patch ideabuck decides to attach debugger every time it see `Debugging. Suspending JVM. Connect a JDWP debugger to port (\d+) to proceed.` line from `buck test`  command
Also removed `NopProcessHandler`


kageiit:
Looks good. But please be aware that a lot of users of buck now use the okbuck gradle plugin for their builds. When they hook into `buckw` which wraps running the gradle okbuck task before running buck (only if needed), the output from the buck binary set in the ide preferences will have output different from vanilla buck.

As long as the plugin looks for key text on all output lines, it should be okay.

korniltsev:
To be honest I don't really like my idea of relying on warning message in the console. We should probably introduce new bus event, something like `JVMIsWaitingForDebuggerEvent` and handle it in ws client. If we agree on this I can fix it.

kageiit:
New bus event would be the best solution and won't break the ide if the message changes in future buck updates.

There are still a few points in the plugin that seem to rely on text out put like progress printing etc. If we make bus events, we will have the best customizability and can make new ide plugin features do the same to talk to buck.

korniltsev:
Replaced Runnable with lambdas

Issue #299, opened 2015-05-01
Update to a newer version of the Android Manifest Merger


joshzana:
This is either a bug or something so hard to figure out that it should be documented.

I have a debug directory with it's own manifest like this:

<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  package="com.joshzana"
  xmlns:tools="http://schemas.android.com/tools">

    <application
        tools:replace="android:name"
        android:name=".DebugApplication"/>
</manifest>

Which under grade gets picked up in debug build types, overriding the non-debug Application subclass.
    
With buck I have rules that look something like this:

android_manifest(
  name = 'manifest',
  skeleton = 'AndroidManifest.xml',
  deps = [
    ':debug_manifest'
  ],
)

android_resource(
  name = 'debug_manifest',
  manifest = 'build-types/debug/AndroidManifest.xml',
  deps = [ ],
)

And I get this warning:

[AndroidManifest.xml:48, AndroidManifest.xml:3] Main manifest has <application android:name='com.joshzana.MyApplication'> but library uses name='com.joshzana.DebugApplication'. 

The merged manifest in buck-out shows the wrong Application name, and my debug-only code doesn't execute.

Is there a setting I'm missing, or is just incorrect conflict handling?

marcinkosiba:
Hey! The manifest merger wasn't intended for rewriting parts of the manifest like that and I don't think there is a way to make it do what you want. I'm guessing the problematic bit is the fact that you're merging a bunch of manifests and so you need to reference the same application name everywhere? In that case would using a genrule to replace the name be an acceptable solution?

joshzana:
I would much prefer for that logic to live in the manifest with the "tools:*" directives.  In gradle these are called "markers" and they instruct the merger how to deal with conflicts:  http://tools.android.com/tech-docs/new-build-system/user-guide/manifest-merger#TOC-Markers    These are useful for more than just name (eg. icon, label, etc on your debug variant) so duplicating all of that in a general seems duplicative.

Do you think this is something buck could support?

marcinkosiba:
From that link it sounds like gradle upgraded to a newer version of the Manifest merger which has the nice features you're looking for. It's something we might look at in the future but is not a high priority at the moment. If you'd like to speed that along pull requests are always welcome - the instructions under third-party/java/aosp would be a good starting point.

facebook-github-bot-9:
[...]

PR #1018, opened 2016-11-16
Update manifest merger

marcinkwiatkowski:
This switches to `ManifestMerger2` implementation and fixes https://github.com/facebook/buck/issues/299

Majority of this PR is updating aosp to `25.2.0` version of the tools.

The only interesting change is in `GenerateManifestStep.java`.

For the sake of not maintaining two versions of aosp, this PR replaces old `ManifestMerger` with `ManifestMerger2`. Having both is possible (we could take `ManifestMerger` from tools `24.x.x`, the last version that shipped with it) but if the new merger works for everyone, probably best to just go with that.


kageiit:
ManifestMerger2 also has a lot of improvements we contributed to aosp, most notably the fix to remove duplication of library manifest loading https://code.google.com/p/android/issues/detail?id=222034#c7

Excited to see this change in buck finally

sdwilsh:
@dreiss or @marcinkosiba should probably be the ones to look at this.

dreiss:
Can you rebase this?  I fixed some of the annotations that should not have been referencing the Android versions.

Can you also update the readme with enough detail for me to ensure that the steps can be reproduced?

I'll also need to check that this doesn't break any of our existing manifest merging at FB.

marcinkwiatkowski:
@dreiss updated.

Coneko:
`GenerateManifestStepTest` is failing both on Travis and on Appveyor, seems related to the change.

build.xml
```diff
@@ -317,7 +317,7 @@
     <ecj srcdir="${aosp.src.dir}" destdir="${aosp.classes.dir}" gensrcdir="${aosp.gensrc.dir}">
       <ecj-params>
         <!-- ManifestMerger.java has redundant null checks -->
-        <compilerarg value="-err:-null" />
```
dreiss: Why did this change?

third-party/java/aosp/README.txt
```diff
@@ -1,13 +1,16 @@
...
+com.android.tools:sdklib:25.2.0
+com.android.tools.layoutlib:layoutlib-api:25.2.0
+2) Copied com.android.manifmerger.ManifestMerger2 and all required dependencies to the aosp/ module.
 3) In the Android SDK's source code manifmerger, sdklib and utils are separate modules. These are
 copied into a single module which alters certain package paths. For example,
-com.android.SDKConstants becomes com.android.common.SDKConstants.
-4) Added support for re-ordering <activity-alias> so they always go after <activity> elements.
-5) Added support for merging <meta-data> elements of <application>.
\ No newline at end of file
+com.android.SDKConstants becomes com.android.common.SDKConstants. This is done to avoid conflicts
```
dreiss: Renaming SDKConstants is gross.  What libraries does this conflict with?  Should we update them?  This hack is fine for now, but you should file a task to clean it up.

third-party/java/aosp/aosp.iml
```diff
@@ -17,6 +17,6 @@
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="library" name="buck-lib" level="project" />
```
dreiss: Why is buck-lib in here?

src/com/facebook/buck/android/GenerateManifestStep.java
```diff
@@ -86,35 +84,48 @@ public StepExecutionResult execute(ExecutionContext context) {
...
-        String xmlText = Files.toString(outManifestFile, Charsets.UTF_8);
-        xmlText = xmlText.replace("\r\n", "\n");
-        Files.write(xmlText.getBytes(Charsets.UTF_8), outManifestFile);
-      } catch (IOException e) {
-        throw new HumanReadableException("Error converting line endings of manifest file");
-      }
+      xmlText = xmlText.replace("\r\n", "\n");
+    }
+    try {
+      File outManifestFile = outManifestPath.toFile();
```
dreiss: Can you use ProjectFileSystem.writeContentsToPath here?

build.xml
marcinkwiatkowski: There is a ton of new warnings in `ManifestMerger2` (73!) that ant turns into errors

third-party/java/aosp/README.txt
marcinkwiatkowski: Buck uses `common-24.2.3`, `sdk-common-24.2.3`, etc. `25.2.0` comes with different api in many places. If we can upgrade to use `25.2.0` everywhere, we can get rid of the whole `aosp` hack and just use prebuild jars

kageiit: @dreiss any luck checking this PR out on FB's codebase?

sdwilsh:
A bunch, but not all, builds failed with this: `Error generating manifest file: Manifest merger failed with multiple errors, see logs`.  Unfortunately, I don't see anything in the logs that gives me more details, so it's a bit difficult to tell what's wrong here.

src/com/facebook/buck/android/GenerateManifestStep.java
```diff
@@ -86,35 +83,47 @@ public StepExecutionResult execute(ExecutionContext context) {
...
+          .newMerger(mainManifestFile, logger, ManifestMerger2.MergeType.APPLICATION)
+          .withFeatures(
+              ManifestMerger2.Invoker.Feature.NO_PLACEHOLDER_REPLACEMENT,
+              ManifestMerger2.Invoker.Feature.REMOVE_TOOLS_DECLARATIONS,
+              ManifestMerger2.Invoker.Feature.SKIP_BLAME)
+          .addLibraryManifests(Iterables.toArray(libraryManifestFiles, File.class))
+          .merge();
+      if (mergingReport.getResult().isError()) {
+        mergingReport.log(logger);
+        throw new HumanReadableException("Error generating manifest file");
```
dreiss: This is still showing up with no indication of what the actual error is.  I think we need to display all available diagnostic info in the console to make it easier for engineers (and me) to debug their build failures.

dreiss:
Okay, I'm going to start sorting through the errors now.  The first one is really annoying: when there's a duplicate element conflict, it shows you the full path to the second occurrence, but only the basename of the first.  But all of our basenames are the same!

kageiit:
Any update on this?

dreiss:
Yeah, it breaks some stuff in our builds.  So far, it's all been valid errors in our code, but it's taking my a while to slog through them all.

marcinkwiatkowski:
@dreiss let me know, we could probably support both versions of the merger for some time.

dreiss:
Okay, I found one thing that I think is a legit bug.  If you have a WRITE_EXTERNAL_STORAGE, even if you have a READ_EXTERNAL_STORAGE elsewhere, it adds a READ_EXTERNAL_STORAGE permission, but with an "android:" prefix on the element name, which seems incorrect.  Want to check it out?

marcinkwiatkowski:
So looks like they use "android:" for all implicit permissions. Here's the logic that adds READ_EXTERNAL_STORAGE https://github.com/marcinkwiatkowski/buck/blob/e984badb83ee93f95f52bd2e80629b6e45a9a595/third-party/java/aosp/src/com/android/manifmerger/XmlDocument.java#L489-L489

What's the failure?

kageiit:
> What's the failure?

I think the issue here is that even if `READ_EXTERNAL_STORAGE` is already present elsewhere, it is being re-added as `android: READ_EXTERNAL_STORAGE` from @dreiss 's comment

dreiss:
There's no failure that I know of.  I'm just worried that something is going to go wrong at runtime.

kageiit:
All the apps in the wild using gradle have been on manifest merger 2 for many months now. If there was indeed something to worry about at runtime, it would have been quite obvious :)

You could also sanity test the apk produced by the new merger if needed

dreiss:
There's an internal diff now to upgrade our sdklib to android/sdk-common-25.2.0.jar.  Does that make this easier or harder?

marcinkwiatkowski:
@dreiss doesn't affect this PR directly but we could potentially remove the whole `aosp` and make the whole setup simpler. Are you upgrading all jars in `third-party/java/android`?

dreiss:
I think so.  Not sure about sdklib, but definitely the others.  I'll let you know when it lands.

gastoncesarf:
@marcinkwiatkowski sorry for my question (in the case that is possible) I need to know an approximate date for this. Can you estimate this? Thanks.

dreiss:
Okay, I've narrowed down the problem to addImplicitElements.  Because it thinks that all of our library manifests are targeting sdk 1, it's adding a bunch of (unwanted) stuff to our manifest.  I commented out the call to that method and it seems to be working.  Can you comment it out in the PR, then we can see if we can find a cleaner way to disable these?

marcinkwiatkowski:
@kageiit @dreiss I agree, this is only for a test. 
I think only
```
        if (libraryTargetSdk < 4) {
            addIfAbsent(mergingReport.getActionRecorder(),
                    USES_PERMISSION,
                    permission("WRITE_EXTERNAL_STORAGE"),
                    lowerPriorityDocument.getPackageName() + " has a targetSdkVersion < 4");
            hasWriteToExternalStoragePermission = true;

            addIfAbsent(mergingReport.getActionRecorder(),
                    USES_PERMISSION,
                    permission("READ_PHONE_STATE"),
                    lowerPriorityDocument.getPackageName() + " has a targetSdkVersion < 4");
        }
```
block is problematic for @dreiss

dreiss:
I think the next block (`if (hasWriteToExternalStoragePermission) {`) is also potentially problematic, though we could work around that.  I'm not sure about the last one.

I think what I'd really like (upstream) is that if the library doesn't specify a target, instead of assuming that it is "1" (because, really, what are the odds of that), assume that it doesn't care and set it to the target of the main manifest.

marcinkwiatkowski:
It's a bit of patching but we could do that @dreiss (see latest commit), I think it makes sense. Do most of your libraries don't specify sdk? I updated test data to reflect that.

dreiss:
It looks like this removed both "*Test.java" files.  Was that intentional?  It was messing up our CI with a very confusing error message (which is one reason this has taken so long).

marcinkwiatkowski:
Those were aosp tests, I don't think they need to be run as part of buck pipeline. `GenerateManifestStepTest` is an integration test that covers our use of the merger.

dreiss:
Looks like another failure has been introduced on our side: https://gist.github.com/dreiss/f22391e27e0c65c53c8684781d6591f4

Code looks pretty reasonable to me.  Suggestions?  Maybe create a subclass to use as the second receiver?

marcinkwiatkowski:
Yep, subclass should be good.

kageiit:
Subclass should work fine

dreiss:
This had to be reverted.  I'm looking into it.

dreiss:
Looks like the old version was (somehow) ensuring that all activity-alias elements came after the activities they referenced, but the new one does not, resulting in some `<activity-alias> target activity [...] not found in manifest`.  I guess I'm going to have to track down cases where this happens in our codebase and reorder them manually.  I'm not sure what will happen if the activity and the alias appear in separate files.

I'm open to other suggestions if people have them.

marcinkwiatkowski:
@dreiss I found the diff changed it in old version https://github.com/facebook/buck/commit/2d63b10fe12522283ddb4a5d62cd22dbf8ba0e8a
I think instead of patching the merger, the reordering (if needed) should be done as a separate step on a merged document.

dreiss:
idk, this seems like a bug in the merger.  Why would you require the user to run a separate postprocessing step?

dreiss:
Yep, I found a case in our codebase where the activity and the alias are in separate files, so there's no way to fix this short of postprocessing.  I can do it for Facebook, but this doesn't seem like a reasonable thing for us to ask every Buck user to do.  I'd much rather the merger just use a safe ordering.

PR #1020, opened 2016-11-17
Add support for custom compiler class names

### LINKED ISSUE #1019: Add support for custom compiler class names
Currently, the jar backed javac hardcodes `com.sun.tools.javac.api.JavacTool` as the implementation of `JavaCompiler`

https://github.com/facebook/buck/blob/8e3c38fcf5403808aa0529afe16b8e9978c9b427/src/com/facebook/buck/jvm/java/AbstractJavacOptions.java#L149

Proposal is to open that up with a `compiler_class_name` option that can work with `javac_jar` option.

This lets us integrate alternative compilers like Google's Error prone (https://github.com/google/error-prone) into the build process

kageiit:
Fixes #1019

### DIFF
--- a/docs/__jvm_common.soy
+++ b/docs/__jvm_common.soy
@@ -10,3 +10,16 @@
   {/param}
 {/call}
 {/template}
+
+/***/
+{template .compiler_class_name_arg}
+{call buck.arg}
+  {param name: 'compiler_class_name' /}
+  {param default : '<global value>' /}
+  {param desc}
+  Specifies the Java compiler class name to use in tandem with javac_jar.
+  Overrides the value in <a
+  href="{ROOT}concept/buckconfig.html#tools.compiler_class_name"><code>.buckconfig</code></a>.
+  {/param}
+{/call}
+{/template}
--- a/docs/concept/buckconfig.soy
+++ b/docs/concept/buckconfig.soy
@@ -2671,6 +2671,15 @@ your <code>.buckjavaargs</code> file</a>:
 
 {call buckconfig.entry}
   {param section: 'tools' /}
+  {param name: 'compiler_class_name' /}
+  {param description}
+    When javac_jar is set, Buck loads the referenced compiler class name from the jar.
+    When it is not set but javac_jar is set, Buck uses the default compiler class.
+  {/param}
+{/call}
+
+{call buckconfig.entry}
+  {param section: 'tools' /}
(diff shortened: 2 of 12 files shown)


kageiit:
Happy to add tests, if I can be pointed to where to add them

Coneko:
`JavacOptionsTest`, there's already some that test what is returned by `getJavac` there.

Coneko:
In general if a certain functionality is completely enclosed in a class the test for it should go in the corresponding test suite, you can always create a new one if one doesn't exist.

kageiit:
Added tests and docs

Coneko:
Changes look good, but extract the copy pasta in the soy docs into a common template and call it.

kageiit:
@Coneko fixed the templates. Also tested locally by running `./docs/soy2html.sh` and checking the output docs

Please take a look. Thanks

zhihuawensc:
Is there a concrete example about how to use it with Google's Error prone compiler?

I am also interested in about how to use it with aspectj compiler?
The format used by Hugo at https://github.com/JakeWharton/hugo/blob/master/hugo-runtime/build.gradle is like below in gradle

android.libraryVariants.all { variant ->
  JavaCompile javaCompile = variant.javaCompile
  javaCompile.doLast {
    String[] args = [
        "-showWeaveInfo",
        "-1.5",
        "-inpath", javaCompile.destinationDir.toString(),
        "-aspectpath", javaCompile.classpath.asPath,
        "-d", javaCompile.destinationDir.toString(),
        "-classpath", javaCompile.classpath.asPath,
        "-bootclasspath", android.bootClasspath.join(File.pathSeparator)
    ]

    MessageHandler handler = new MessageHandler(true);
    new Main().run(args, handler)

    def log = project.logger
    for (IMessage message : handler.getMessages(null, true)) {
      switch (message.getKind()) {
        case IMessage.ABORT:
        case IMessage.ERROR:
        case IMessage.FAIL:
          log.error message.message, message.thrown
          break;
        case IMessage.WARNING:
        case IMessage.INFO:
          log.info message.message, message.thrown
          break;
        case IMessage.DEBUG:
          log.debug message.message, message.thrown
          break;
      }
    }
  }
}

kageiit:
@zhihuawensc I think for your purpose, it should be done as a `postprocess_classes_command` instead. Similar to how retrolambda works right now in okbuck

PR #1025, opened 2016-11-21
Use a ProjectFileSystem if one is set

shs96c:
Oops.


shs96c:
Took me ages to track down why the test Daniel asked me to write wasn't working.

PR #1027, opened 2016-11-22
Duplicate TargetNodeSpecs don't seem ideal. Use a Set.

shs96c:
We don't want duplicates, and ordering may matter (though it
should not). Use an `ImmutableSet` not an `ImmutableList`.


Coneko:
I don't think this will work: target node specs aren't generally equatable, and in particular the ones returned by `CommandLineTargetNodeSpecParser` are `TargetNodePredicateSpec` with lambdas as predicates so they'll never be equal.

styurin:
Not sure this would work correctly with the current logic in `DefaultParser`, it depends on the order of `TargetNodeSpecs`.

PR #1031, opened 2016-11-30
Enhance post process classes commands with additional env vars

kageiit:
This makes the `classpath` and `bootclasspath` (if available) of a jvm target available to `postProcessClassesCommands`. This is very useful to be able to do bytecode manipulation of the output classes with tools like retrolamdba etc.

### DIFF
--- a/src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
+++ b/src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
@@ -23,12 +23,16 @@ import com.facebook.buck.rules.BuildContext;
 import com.facebook.buck.rules.BuildableContext;
 import com.facebook.buck.rules.SourcePathResolver;
 import com.facebook.buck.shell.BashStep;
+import com.facebook.buck.step.ExecutionContext;
 import com.facebook.buck.step.Step;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 
 import java.nio.file.Path;
@@ -83,9 +87,11 @@ public abstract class BaseCompileToJarStepFactory implements CompileToJarStepFac
         buildableContext);
 
     steps.addAll(Lists.newCopyOnWriteArrayList(addPostprocessClassesCommands(
-        filesystem.getRootPath(),
+        filesystem,
         postprocessClassesCommands,
-        outputDirectory)));
+        outputDirectory,
+        declaredClasspathEntries,
+        getBootClasspath())));
 
     steps.add(
         new JarDirectoryStep(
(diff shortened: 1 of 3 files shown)


kageiit:
cc @dreiss

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
```diff
@@ -16,19 +16,24 @@
 
 package com.facebook.buck.jvm.java;
 
+import com.facebook.buck.android.AndroidPlatformTarget;
```
dreiss: I'd like to avoid having this Java code relying on anything Android-specific.  Can the required information be passed from above in some generic way?  ("extraEnvironmentForPostprocessing" or something like that?)

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
```diff
@@ -109,19 +116,51 @@ public void createCompileToJarStep(
...
       List<String> postprocessClassesCommands,
-      Path outputDirectory) {
+      Path outputDirectory,
+      ImmutableSortedSet<Path> declaredClasspathEntries) {
     ImmutableList.Builder<Step> commands = new ImmutableList.Builder<Step>();
+
+    ImmutableMap.Builder<String, String> envVarBuilder = ImmutableMap.builder();
+    if (!postprocessClassesCommands.isEmpty()) {
+      envVarBuilder.put(
+          "CLASSPATH",
```
dreiss: Doesn't the "CLASSPATH" variable get special treatment by `java`?  Can this be renamed to something more specific like `COMPILATION_CLASSPATH`?

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
```diff
@@ -109,19 +116,51 @@ public void createCompileToJarStep(
...
+      BuildContext context,
+      ProjectFilesystem filesystem,
       List<String> postprocessClassesCommands,
-      Path outputDirectory) {
+      Path outputDirectory,
+      ImmutableSortedSet<Path> declaredClasspathEntries) {
     ImmutableList.Builder<Step> commands = new ImmutableList.Builder<Step>();
+
+    ImmutableMap.Builder<String, String> envVarBuilder = ImmutableMap.builder();
+    if (!postprocessClassesCommands.isEmpty()) {
```
dreiss: Maybe just check this at the top and return an empty list?

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
kageiit: Done

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
```diff
@@ -98,6 +104,10 @@ public void createCompileToJarStep(
             classesToRemoveFromJar));
   }
 
+  Optional<String> getBootClasspath() {
```
dreiss: This feels like it should be abstract and return a non-optional String.  Do kotlin/groovy/scala not have this concept?  Can you document when and how a subclass should be overriding this?

src/com/facebook/buck/jvm/java/BaseCompileToJarStepFactory.java
```diff
@@ -109,19 +119,46 @@ public void createCompileToJarStep(
...
+        "COMPILATION_CLASSPATH",
+        Joiner.on(':').join(
+            Iterables.transform(
+                declaredClasspathEntries,
+                filesystem::resolve)));
+
+    if (bootClasspath.isPresent()) {
+      envVarBuilder.put("COMPILATION_BOOTCLASSPATH", bootClasspath.get());
+    }
+    final ImmutableMap<String, String> envVars = envVarBuilder.build();
```
dreiss: I don't think this needs to be final in Java 8.

PR #1032, opened 2016-11-30
Make frameworks a linking style at cxx level

robbertvanginkel:
Start replacement for #842. Some notes:

- The old approach included framewrorks at the apple package level, which introduced complications at the linking level. As the linker/compiler at a clang/ld level accept framework related arguments, implement these at the level of the cxx package.
- Instead of specifying a `#framework` flavor for dependencies, use the `preferred_linkage` and `link_style` arguments to controll if apple library will be a framework, dylib or static library.
- Split into a `framework-bundle` and `framework-binary` flavor for clariness.

Todo's:
- [ ] Modulemaps and umbrella headers
- [ ] Framework resources are not collected properly
- [ ] Embedding frameworks

### DIFF
--- a/src/com/facebook/buck/apple/AppleBundleDescription.java
+++ b/src/com/facebook/buck/apple/AppleBundleDescription.java
@@ -55,7 +55,8 @@ public class AppleBundleDescription implements Description<AppleBundleDescriptio
 
   public static final ImmutableSet<Flavor> SUPPORTED_LIBRARY_FLAVORS = ImmutableSet.of(
       CxxDescriptionEnhancer.STATIC_FLAVOR,
-      CxxDescriptionEnhancer.SHARED_FLAVOR);
+      CxxDescriptionEnhancer.SHARED_FLAVOR,
+      CxxDescriptionEnhancer.FRAMEWORK_BINARY_FLAVOR);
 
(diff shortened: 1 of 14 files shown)


robbertvanginkel:
A problem I am running into is correctly collecting the resources for a framework bundle. Currently this happens in `AppleBundleResources.collectResourceDirsAndFiles`, which uses`AppleBuildRules.getRecursiveTargetNodeDependenciesOfTypes`, which traverses the `TargetGraph`. This is in issue, because depending on the `link_style`/`preferred_linkage` argument of the `apple_library` (and its dependencies), transitive resources should be included or not. But at this point, the target graph that is being traversed is the general graph without flavors, which is unaware of the way it will be linked.

A possible solution would be to modify the `collectResourceDirsAndFiles` call, to collect the resources through `requireMetaData` in combination with a new `ResourceMetadata` class. Any suggestions on tackling this problem?

ryu2:
@robbertvanginkel I think that solution works for now.

ryu2:
@robbertvanginkel  I've fixed the duplicate resources issue for now in  87eb248acecc0947d4a32244a7a73689900cf50487eb248acecc0947d4a32244a7a73689900cf504 by not propagating the resources dependencies of any shared `apple_library` or `cxx_library`

PR #1043, opened 2016-12-11
Error prone fixes

davido:
Activate error prone on Buck by adding these lines to .buckconfig:

```
[tools]
  javac_jar = third-party/error_prone_ant-2.0.15.jar
  compiler_class_name = com.google.errorprone.ErrorProneJavaCompiler
[java]
  extra_arguments = "-Xep:OptionalEquality:OFF"
```

Put `error_prone_ant-2.0.15.jar` to `third-party` directory, run `buck build buck`
and fix the problem flagged by error prone.

### DIFF
--- a/src/com/facebook/buck/jvm/java/AccumulateClassNamesStep.java
+++ b/src/com/facebook/buck/jvm/java/AccumulateClassNamesStep.java
@@ -151,7 +151,7 @@ public class AccumulateClassNamesStep implements Step {
             HashCode existing = classNames.putIfAbsent(key, value);
             if (existing != null && !existing.equals(value)) {
               throw new IllegalArgumentException(String.format(
-                  "Multiple entries with same key but differing values: %1$s=%s and %1$s=%s",
+                  "Multiple entries with same key but differing values: %1$s=%2$s and %1$s=%3$s",
                   key, value, existing));
             }
           }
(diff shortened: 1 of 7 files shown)


src/com/facebook/buck/jvm/java/AccumulateClassNamesStep.java
```diff
@@ -152,7 +152,7 @@ public InputStream openStream() throws IOException {
             if (existing != null && !existing.equals(value)) {
               throw new IllegalArgumentException(String.format(
                   "Multiple entries with same key but differing values: %1$s=%s and %1$s=%s",
-                  key, value, existing));
+                  key, value));
```
k21: This and the next change don't look right, I think the format should be changed to something like `%1$s=%2$s and %1$s=%3$s` (or the positional arguments can be removed and we can just pass `key, value, key, existing`).

src/com/facebook/buck/util/Escaper.java
```diff
@@ -410,7 +410,7 @@ public static int decodeNumericEscape(
       boolean valid = false;
       for (int i = 0; i < maxCodeLength && pos < escaped.length(); i++) {
         final char digit = (char) table.indexOf(Character.toLowerCase(escaped.charAt(pos)));
-        if (digit == -1 || digit >= base) {
+        if (digit >= base) {
```
k21: This looks like it will work correctly, but I would prefer to keep `digit` as an `int` instead.

Coneko:
Oh this is very nice, thanks!

src/com/facebook/buck/jvm/java/AccumulateClassNamesStep.java
davido: @k21 Done. PTAL.

davido:
I addressed all comments.

PR #1049, opened 2016-12-14
Create src jars using JarShape

shs96c:
It's fast, free, and funky.


src/com/facebook/buck/jvm/java/JavaSourceJar.java
```diff
@@ -81,12 +99,21 @@ public JavaSourceJar(
     steps.add(new MakeCleanDirectoryStep(getProjectFilesystem(), temp));
 
     Set<Path> seenPackages = Sets.newHashSet();
+    Map<ProjectFilesystem, JavaPackageFinder> finders = new HashMap<>();
+
+    for (SourcePath sourcePath : sources) {
+      Path source = getResolver().getAbsolutePath(sourcePath);
+      if (!"java".equals(MorePaths.getFileExtension(source))) {
```
kageiit: should make the list of extensions configurable as other jvm based libraries can have scala/groovy/kotlin files etc and can easily override this

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+
+  private final ProjectFilesystem filesystem;
+
+  public SourceReadingPackageFinder(ProjectFilesystem filesystem) {
+    this.filesystem = filesystem;
+  }
+
+  @Override
+  public Path findJavaPackageFolder(Path pathRelativeToProjectRoot) {
+    if ("java".equals(MorePaths.getFileExtension(pathRelativeToProjectRoot))) {
```
kageiit: Same with file extension here

shs96c: Same logic applies.

src/com/facebook/buck/jvm/java/JavaSourceJar.java
kageiit: Why do you need a parser to create source jars?

kageiit: If you just need to extract the package, a full blown java source code parser might be overkill. Java/Scala/Groovy/Kotlin do not differ in how they declare their package names except for maybe the omission of a semi colon at the end. The package can be extracted via a simpler means (regex?) and the logic can then be applied to all jvm based language source files. Thoughts?

shs96c: This isn't a full blown java parser, it just handles enough to get to the package declaration. Parsing files using regex is generally considered a pretty [bad idea](https://blog.codinghorror.com/parsing-beyond-regex/), and possibly error-prone since there might be a package declaration in a comment. Put another way, we either need a very complex regex or a simple parser in order to be consistently correct. I opted for a parser.

Given that this parser is only used from the `JavaSourceJar` class, I'm not sure why you mention kotlin, Scala, and groovy. It seems beside the point of this diff.

shs96c: cf: parsing things with regex: http://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+      value.append((char) last);
+      for (;;) {
+        int ch = reader.read();
+
+        if (ch == -1) {
+          return new Token(Token.Kind.UNKNOWN, "");
+        }
+
+        if (last == '*' && ch == '/') {
+          return new Token(Token.Kind.COMMENT, value.toString());
```
k21: I am not sure why the value is needed here, but since it is passed, it looks like this will contain the last `*`, but not the `/`.

src/com/facebook/buck/jvm/java/JavaLibraryDescription.java
```diff
@@ -91,46 +92,69 @@ public Arg createUnpopulatedConstructorArg() {
...
+            sources);
+      } else if (flavors.contains(JavaLibrary.SRC_JAR)) {
+        ImmutableSortedSet<SourcePath> sources = summary.getPackagedRules().stream()
+            .filter(HasSources.class::isInstance)
+            .flatMap(dep -> ((HasSources) dep).getSources().stream())
+            .collect(MoreCollectors.toImmutableSortedSet(Ordering.natural()));
+
+        BuildRuleParams emptyParams = params.copyWithDeps(
+            Suppliers.ofInstance(
+                ImmutableSortedSet.copyOf(pathResolver.filterBuildRuleInputs(sources))),
```
shs96c: Good catch. Done.

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  @Override
+  public Path findJavaPackageFolder(Path pathRelativeToProjectRoot) {
+    if ("java".equals(MorePaths.getFileExtension(pathRelativeToProjectRoot))) {
+      String packageName = findJavaPackage(pathRelativeToProjectRoot);
+
+      Iterator<String> segments = Splitter.on('.').split(packageName).iterator();
+      if (!segments.hasNext()) {
+        throw new IllegalStateException("Unable to create path from " + packageName);
```
shs96c: Changed to a `HumanReadableException` and added the file name.

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+    for (;;) {
+      Token token = lexer.getNextToken();
+
+      switch (token.getKind()) {
+        case PACKAGE:
+          return token.getValue();
+
+        case UNKNOWN:
+          // We have no idea what to do. Assume the default package
+          return "";
```
shs96c: If we never see a package declaration, we're in the default package, and that might happen. We also use the `UNKNOWN` token to handle the case where we hit anything that might precede a package statement, so `interface`, `class`, and `enum` (and modifiers) all get treated that way. I'd chosen to play it safe for now, and the approach seems to work on both Buck's code and Selenium's

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    private Token readWhiteSpace(Reader reader) throws IOException {
+      // Just keep going until we see something that's not whitespace.
+      for (;;) {
+        reader.mark(1);
+        int ch = reader.read();
+
+        if (ch == -1) {
+          return new Token(Token.Kind.UNKNOWN, "");
```
shs96c: Because we've run out of parseable data, and `UNKNOWN` is a terminal condition.

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+    private Token readSingleLineComment(Reader reader) throws IOException {
+      // Read up until we see a line ending. Assume Mac, UNIX or DOS line endings.
+      StringBuilder value = new StringBuilder("//");
+
+      for (;;) {
+        int ch = reader.read();
+
+        switch (ch) {
+          case -1:
+            return new Token(Token.Kind.UNKNOWN, "");
```
shs96c: Same logic. I could add an `EOF` token, I guess.

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
```diff
@@ -0,0 +1,278 @@
...
+          // New-line always means the end of the line
+          case '\n':
+            return new Token(Token.Kind.COMMENT, value.toString());
+
+          // Carriage return might mean either a Mac EOL or a DOS one. Peek ahead
+          case '\r':
+            reader.mark(1);
+            ch = reader.read();
+            if (ch == -1) {
+              return new Token(Token.Kind.UNKNOWN, "");
```
shs96c: As above.

src/com/facebook/buck/jvm/java/SourceReadingPackageFinder.java
shs96c: I thought it might be useful when debugging. Stripped that last character.

shs96c:
I think this is ready to go now. Anything else to do?

k21:
AFAICT you didn't address my comment above about [unicode escapes](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.3). Otherwise it looks good.

shs96c:
This is old, tired, and woefully out of date. Abandoning.

PR #1050, opened 2016-12-14
Add swift compilerflags from buckconfig to rulekey

robbertvanginkel:
Fix rebuilding when swift compiler flags are updated in the buckconfig. Also marks the bridging header for rulekey.

Any suggestions for test are welcome.

### DIFF
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -126,8 +126,8 @@ class SwiftCompile
 
     this.srcs = ImmutableSortedSet.copyOf(srcs);
     ImmutableList.Builder<String> compilerFlagsBuilder = ImmutableList.builder();
-    compilerFlagsBuilder.addAll(compilerFlags);
     compilerFlagsBuilder.addAll(swiftBuckConfig.getFlags().orElse(ImmutableSet.of()));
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -125,7 +125,10 @@
     this.objectPath = outputPath.resolve(escapedModuleName + ".o");
 
     this.srcs = ImmutableSortedSet.copyOf(srcs);
-    this.compilerFlags = compilerFlags;
+    ImmutableList.Builder<String> compilerFlagsBuilder = ImmutableList.builder();
+    compilerFlagsBuilder.addAll(swiftBuckConfig.getFlags().orElse(ImmutableSet.of()));
```
illicitonion: Why not just make SwiftBuckConfig#getFlags always just return a (maybe empty) ImmutableList?

Then I'd slightly prefer this to look like:

```java
this.compilerFlags = new ImmutableList.Builder<String>()
    .addAll(swiftBuckConfig.getFlags())
    .addAll(compilerFlags)
    .build();
```

or something along those lines.

Coneko:
Assuming this will be merged from the fork together with the rest of the changes, closing for inactivity.

PR #1053, opened 2016-12-14
Extend zip external attributes to indicate files are files

shs96c:
Without this, node's zip-decompress library has no idea what
kind of file it's dealing with and gets grumpy with us. That
can't be a Good Thing, so just indicate that files are files
as we build the zip.


Coneko:
`//test/com/facebook/buck/rules:rules - inputBasedRuleKeyCacheHitAvoidsBuildingLocally` is failing.

PR #1054, opened 2016-12-15
Improved Kotlin support.

brettwooldridge:
kotlin_test execution now implicitly includes Kotlin runtime.

Kotlin-specific classes introduced which extend their Java peers rather than referencing
the Java implementations directly, for semantic separation and preparation for first-class
native (in VM) compilation, abi, and autodeps support.

### DIFF
--- a/src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibrary.java
+++ b/src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibrary.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2016-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.jvm.kotlin;
+
+import com.facebook.buck.jvm.java.CompileToJarStepFactory;
+import com.facebook.buck.jvm.java.DefaultJavaLibrary;
+import com.facebook.buck.model.BuildTarget;
+import com.facebook.buck.rules.BuildRule;
+import com.facebook.buck.rules.BuildRuleParams;
+import com.facebook.buck.rules.SourcePath;
+import com.facebook.buck.rules.SourcePathResolver;
+import com.facebook.buck.rules.SourcePathRuleFinder;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSortedSet;
+
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class DefaultKotlinLibrary extends DefaultJavaLibrary {
+
+  public DefaultKotlinLibrary(
+      BuildRuleParams params,
+      final SourcePathResolver resolver,
+      SourcePathRuleFinder ruleFinder,
+      Set<? extends SourcePath> srcs,
+      Set<? extends SourcePath> resources,
+      Optional<Path> generatedSourceFolder,
+      Optional<SourcePath> proguardConfig,
+      ImmutableList<String> postprocessClassesCommands,
+      ImmutableSortedSet<BuildRule> exportedDeps,
+      ImmutableSortedSet<BuildRule> providedDeps,
+      BuildTarget abiJar,
+      ImmutableSortedSet<SourcePath> abiInputs,
+      final boolean trackClassUsage,
+      ImmutableSet<Path> additionalClasspathEntries,
+      CompileToJarStepFactory compileStepFactory,
+      Optional<Path> resourcesRoot,
+      Optional<SourcePath> manifestFile,
+      Optional<String> mavenCoords,
+      ImmutableSortedSet<BuildTarget> tests,
+      ImmutableSet<Pattern> classesToRemoveFromJar) {
+
+    super(
+        params,
+        resolver,
+        ruleFinder,
+        srcs,
+        resources,
+        generatedSourceFolder,
+        proguardConfig,
+        postprocessClassesCommands,
+        exportedDeps,
+        providedDeps,
+        abiJar,
+        abiInputs,
+        trackClassUsage,
+        additionalClasspathEntries,
+        compileStepFactory,
+        resourcesRoot,
+        manifestFile,
+        mavenCoords,
+        tests,
+        classesToRemoveFromJar);
+  }
+}
--- a/src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
+++ b/src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
@@ -20,26 +20,119 @@ import com.facebook.buck.cli.BuckConfig;
 import com.facebook.buck.io.ExecutableFinder;
 import com.facebook.buck.rules.HashedFileTool;
 import com.facebook.buck.rules.Tool;
+import com.facebook.buck.util.HumanReadableException;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 
+import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Optional;
+
+import javax.annotation.Nullable;
 
 public class KotlinBuckConfig {
+  private static final String SECTION = "kotlin";
+
   private static final Path DEFAULT_KOTLIN_COMPILER = Paths.get("kotlinc");
 
   private final BuckConfig delegate;
+  private @Nullable Path kotlinHome;
 
   public KotlinBuckConfig(BuckConfig delegate) {
     this.delegate = delegate;
   }
 
+  /**
+   * Get the Tool instance for the Kotlin compiler.
+   * @return the Kotlin compiler Tool
+   */
   public Supplier<Tool> getKotlinCompiler() {
-    Path compilerPath = delegate.getPath("kotlin", "compiler").orElse(DEFAULT_KOTLIN_COMPILER);
+    Path compilerPath = getKotlinHome().resolve("kotlinc");
+    if (!Files.isExecutable(compilerPath)) {
+      compilerPath = getKotlinHome().resolve(Paths.get("bin", "kotlinc"));
+      if (!Files.isExecutable(compilerPath)) {
+        throw new HumanReadableException("Could not resolve kotlinc location.");
+      }
+    }
 
     Path compiler = new ExecutableFinder().getExecutable(compilerPath, delegate.getEnvironment());
-
     return Suppliers.ofInstance(new HashedFileTool(compiler));
+  }
+
+  /**
+   * Get the path to the Kotlin runtime jar.
+   * @return the Kotlin runtime jar path
+   */
+  public Path getPathToRuntimeJar() {
+    Optional<String> value = delegate.getValue(SECTION, "runtime_jar");
+    boolean isAbsolute = (value.isPresent() && Paths.get(value.get()).isAbsolute());
+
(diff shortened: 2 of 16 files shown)


dsyang:
On mobile so I can't properly review right now but do you need to change a test as well? If we're implicitly including the runtime there should be a kotlin_test integration test that doesn't explicitly use the runtime

brettwooldridge:
I did consider that.  But as far as I can tell, none of the tests are run anyway because of [KotlinTestAssumptions](https://github.com/facebook/buck/blob/b6edaa49524c13269c57002030d6ec002e44969d/test/com/facebook/buck/jvm/kotlin/KotlinTestAssumptions.java).
```
ASSUME  <100ms  0 Passed   1 Skipped   0 Failed   com.facebook.buck.jvm.kotlin.KotlinBuckConfigTest
ASSUME   171ms  0 Passed   3 Skipped   0 Failed   com.facebook.buck.jvm.kotlin.KotlinLibraryIntegrationTest
ASSUME  <100ms  0 Passed   4 Skipped   0 Failed   com.facebook.buck.jvm.kotlin.KotlinTestIntegrationTest
```
My intention is to actually remove that assumption when I implement in-VM compilation (and include ``kotlin-compiler.jar`` in third-party).

brettwooldridge:
BTW, the build failure is unrelated to these changes.  Should I rebase against a commit that passes the build?

kageiit:
We should let users specify a kotlinc_jar similar to javac_jar because kotlin is constantly evolving and comes with breaking changes. This will let users use the latest toolchain.

kageiit:
I would recommend getting rid of kotlin home and just let users specify the path to the kotlinc_jar when in memory compilation is available

brettwooldridge:
@kageiit I disagree.  While it is certainly valid to allow specifying a kotlin compiler and runtime jar in buckconfig, it should not be required.  Doing so makes a project non-portable unless the project also includes both the jars in its own repository.  It is the *least* desirable option.

I have three versions of Java installed, and two versions of Scala, and I switch accordingly, without the need to edit anyone's project build files, by setting the appropriate environment variable and environment path.  The same is possible with Kotlin.

kageiit:
> It is the least desirable option.

Probably we should make it one available option then. Because as it stands using KOTLIN_HOME means anyone using a remote build cache cannot reuse the artifacts if the compilers are different

kageiit:
> I have three versions of Java installed, and two versions of Scala, and I switch accordingly, without the need to edit anyone's project build files, by setting the appropriate environment variable and environment path. The same is possible with Kotlin.

I dont think you are using a remote cache, but if you did this would mean that artifacts on one machine can never be reused on another machine

sdwilsh:
We should probably fall back to the environment variable, but otherwise 
use what's specified in the config.  IIRC, that's how we do it with Java.

brettwooldridge:
@sdwilsh Agreed, if the compiler/runtime is defined in the buckconfig that should override everything.

brettwooldridge:
@kageiit I don't see Java or its runtime components in the repositories of projects I've encountered, maybe that is done "at scale", but the majority of ant, maven, gradle, and buck users probably don't expect that.

As far as I know, Kotlin has committed to binary compatibility, so I don't know under what conditions a shared cache wouldn't work in any case.

kageiit:
> As far as I know, Kotlin has committed to binary compatibility, so I don't know under what conditions a shared cache wouldn't work in any case.

I dont think its related to binary compatibility, but more about whats added as a rulekey. for java libraries, the javac version is added to the rulekey for example. And even if kotlin has commited to binary compatibility, the intermediate representation of tools like kapt1/2/3 are still evolving and will need caching. So it is important to consider them as part of the kotlin build toolchain as well

kageiit:
> but the majority of ant, maven, gradle, and buck users probably don't expect that.

The main idea is to support flexibility with config options here because we really do not have data to claim how everyone is using buck in their own unique setups. As long as we have the config options, it should be fine to use all the available options as a consumer sees fit

brettwooldridge:
That I definitely agree with 100%.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -16,17 +16,24 @@
 
 package com.facebook.buck.jvm.kotlin;
 
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+
```
dsyang: nit: move these after com.facebook.....

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -35,11 +42,74 @@ public KotlinBuckConfig(BuckConfig delegate) {
...
 
+  /**
+   * Get the Tool instance for the Kotlin compiler.
+   * @return the Kotlin compiler Tool
+   */
   public Supplier<Tool> getKotlinCompiler() {
-    Path compilerPath = delegate.getPath("kotlin", "compiler").orElse(DEFAULT_KOTLIN_COMPILER);
+    Path compilerPath = getKotlinHome().resolve("kotlinc");
+    if (!Files.isRegularFile(compilerPath)) {
+      compilerPath = getKotlinHome().resolve(Paths.get("bin", "kotlinc"));
```
dsyang: can we reuse the previous getKotlinHome() result instead of recomputing?

src/com/facebook/buck/jvm/kotlin/KotlinLibrary.java
```diff
@@ -0,0 +1,23 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.jvm.kotlin;
+
+import com.facebook.buck.jvm.java.JavaLibrary;
+
+public interface KotlinLibrary extends JavaLibrary {
```
dsyang: I'm assuming you're going to add more stuff in here in later PRs? Can we either 1) hold off until we actually populate this with new stuff or 2) add a comment explaining why this is hear (and different from JavaLibrary)

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
```diff
@@ -18,37 +18,57 @@
 
 import static com.facebook.buck.jvm.common.ResourceValidator.validateResources;
 
+import java.util.Optional;
```
dsyang: nit: alphabetize imports

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
```diff
@@ -89,65 +89,76 @@ public Arg createUnpopulatedConstructorArg() {
...
                 args.srcs,
                 ResourceValidator.validateResources(
                     pathResolver,
                     params.getProjectFilesystem(),
                     args.resources),
-                defaultJavacOptions.getGeneratedSourceFolderName(),
-                /* proguardConfig */ Optional.empty(),
-                /* postprocessClassesCommands */ ImmutableList.of(),
-                /* exportDeps */ ImmutableSortedSet.of(),
-                /* providedDeps */ ImmutableSortedSet.of(),
```
dsyang: why remove the /* */ hints?

src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibrary.java
```diff
@@ -0,0 +1,79 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.jvm.kotlin;
+
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
```
dsyang: nit: move behind com....

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
brettwooldridge: They weren't consistently present for all parameters, and were not present in the JavaLibraryDescription.  We can add them back in; though if we do, they should probably appear after the parameters rather than before to make the code "prettier".

brettwooldridge:
Rebased w/changes.  Tests coming soon'ish.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -20,26 +20,102 @@
 import com.facebook.buck.io.ExecutableFinder;
 import com.facebook.buck.rules.HashedFileTool;
 import com.facebook.buck.rules.Tool;
+import com.facebook.buck.util.HumanReadableException;
+
```
dsyang: nit: no space. doesn't look like other files have a newline between com.facebook and com.google

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
dsyang: Got it. lets just leave them off then. Better to conform to one of the existing styles than introduce a third variant.

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
```diff
@@ -122,36 +134,36 @@ public Arg createUnpopulatedConstructorArg() {
...
-                /* additionalClasspathEntries */ ImmutableSet.of(),
+                false,
+                ImmutableSet.of(),
                 stepFactory,
-                /* resourcesRoot */ Optional.empty(),
-                /* manifest file */ Optional.empty(),
-                /* mavenCoords */ Optional.empty(),
-                /* tests */ ImmutableSortedSet.of(),
-                /* classesToRemoveFromJar */ ImmutableSet.of()
+                args.resourcesRoot,      /* resourcesRoot */
```
dsyang: +1 to just using args.X, lets leave off the comments

brettwooldridge:
@dsyang Just pushed a commit, didn't see your most recent comments, so it does not include those changes.  But it does include a set of new unit tests.

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
brettwooldridge: @dsyang I did reintroduce these.  I actually to find them useful, where there might be a series of:
```java
Optional.empty(),
ImmutableList.of(),
ImmutableSortedSet.of(),
ImmutableSortedSet.of(),
```
It actually is useful to read:
```
Optional.empty(),        /* proguardConfig */
ImmutableList.of(),      /* postprocessClassesCommands */
ImmutableSortedSet.of(), /* exportedDeps */
ImmutableSortedSet.of(), /* providedDeps */
```

brettwooldridge:
@dsyang I think this is ready for final review.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -20,26 +20,117 @@
...
+    runtime = getKotlinHome().resolve(Paths.get("lib", "kotlin-runtime.jar"));
+    if (Files.isRegularFile(runtime)) {
+      return runtime.normalize();
+    }
+
+    throw new HumanReadableException("Could not resolve kotlin runtime JAR location.");
+  }
+
+  private synchronized Path getKotlinHome() {
```
dsyang: question. Why is this synchronized?

brettwooldridge: Probably coding too late at night ... because of the check below my brain switched into "what if two threads are accessing this instance" mode.  But actually that does not seem possible.  If that is true, and you know better than I, then I can take the synchronized out.

src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibrary.java
brettwooldridge: done.

brettwooldridge:
Removed synchronized.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -20,26 +20,117 @@
...
 import java.nio.file.Paths;
+import java.util.Optional;
 
 public class KotlinBuckConfig {
+  private static final String SECTION = "kotlin";
+
   private static final Path DEFAULT_KOTLIN_COMPILER = Paths.get("kotlinc");
 
   private final BuckConfig delegate;
+  private Path kotlinHome;
```
dsyang: mark as @nullable?

brettwooldridge: Added @Nullable.

brettwooldridge:
@dsyang Thanks!

PR #1056, opened 2016-12-16
Bump to gcc5

illicitonion:


brettwooldridge:
@illicitonion I don't think this can work because of how buck invokes g++.  This is a long standing issue with travis-ci.  The only fix I've found that works is [this](https://github.com/facebook/buck/pull/1060/files#diff-354f30a63fb0907d4ad57269548329e3R25).

brettwooldridge:
@illicitonion [This build](https://travis-ci.org/facebook/buck/builds/185076230) demonstrates the "fix".  I think the actual fix is for buck to honor the CXX environment variable [over the hard-coded default path](https://github.com/facebook/buck/blob/22840aa89e27ee0a01bd458843cc02845e86e670/src/com/facebook/buck/cxx/DefaultCxxPlatforms.java#L46).  If it does, the travis build can be made to run with a simple ``export CXX=...`` instead of ``sudo update-alternatives ...``.

illicitonion:
Obsoleted by b503b5a6f3de1a7735d7b8d81410668abdffa01a

PR #1063, opened 2016-12-20
Make cxx complier location portable...

brettwooldridge:
... and return travis-ci to a container-based build.

### DIFF
--- a/src/com/facebook/buck/cxx/DefaultCxxPlatforms.java
+++ b/src/com/facebook/buck/cxx/DefaultCxxPlatforms.java
@@ -16,6 +16,7 @@
 
 package com.facebook.buck.cxx;
 
+import com.facebook.buck.io.ExecutableFinder;
 import com.facebook.buck.io.ProjectFilesystem;
 import com.facebook.buck.model.Flavor;
 import com.facebook.buck.model.ImmutableFlavor;
@@ -66,16 +67,17 @@ public class DefaultCxxPlatforms {
     Archiver archiver;
     DebugPathSanitizer compilerSanitizer;
     Optional<String> binaryExtension;
+    ImmutableMap<String, String> env = config.getEnvironment();
     switch (platform) {
       case LINUX:
         sharedLibraryExtension = "so";
         sharedLibraryVersionedExtensionFormat = "so.%s";
         staticLibraryExtension = "a";
         objectFileExtension = "o";
-        defaultCFrontend = DEFAULT_C_FRONTEND;
-        defaultCxxFrontend = DEFAULT_CXX_FRONTEND;
+        defaultCFrontend = getExecutablePath("gcc", DEFAULT_C_FRONTEND, env);
+        defaultCxxFrontend = getExecutablePath("g++", DEFAULT_CXX_FRONTEND, env);
         linkerType = LinkerProvider.Type.GNU;
-        archiver = new GnuArchiver(new HashedFileTool(DEFAULT_AR));
+        archiver = new GnuArchiver(new HashedFileTool(getExecutablePath("ar", DEFAULT_AR, env)));
         compilerSanitizer = new PrefixMapDebugPathSanitizer(
             config.getDebugPathSanitizerLimit(),
             File.separatorChar,
@@ -90,10 +92,10 @@ public class DefaultCxxPlatforms {
         sharedLibraryVersionedExtensionFormat = ".%s.dylib";
         staticLibraryExtension = "a";
         objectFileExtension = "o";
-        defaultCFrontend = DEFAULT_OSX_C_FRONTEND;
-        defaultCxxFrontend = DEFAULT_OSX_CXX_FRONTEND;
+        defaultCFrontend = getExecutablePath("clang", DEFAULT_OSX_C_FRONTEND, env);
+        defaultCxxFrontend = getExecutablePath("clang++", DEFAULT_OSX_CXX_FRONTEND, env);
         linkerType = LinkerProvider.Type.DARWIN;
-        archiver = new BsdArchiver(new HashedFileTool(DEFAULT_AR));
+        archiver = new BsdArchiver(new HashedFileTool(getExecutablePath("ar", DEFAULT_AR, env)));
         compilerSanitizer = new PrefixMapDebugPathSanitizer(
             config.getDebugPathSanitizerLimit(),
             File.separatorChar,
@@ -108,10 +110,11 @@ public class DefaultCxxPlatforms {
         sharedLibraryVersionedExtensionFormat = "dll";
         staticLibraryExtension = "lib";
         objectFileExtension = "obj";
-        defaultCFrontend = DEFAULT_C_FRONTEND;
-        defaultCxxFrontend = DEFAULT_CXX_FRONTEND;
+        defaultCFrontend = getExecutablePath("gcc", DEFAULT_C_FRONTEND, env);
+        defaultCxxFrontend = getExecutablePath("g++", DEFAULT_CXX_FRONTEND, env);
         linkerType = LinkerProvider.Type.WINDOWS;
-        archiver = new WindowsArchiver(new HashedFileTool(DEFAULT_AR));
+        archiver = new WindowsArchiver(new HashedFileTool(
+            getExecutablePath("ar", DEFAULT_AR, env)));
         compilerSanitizer = new PrefixMapDebugPathSanitizer(
             config.getDebugPathSanitizerLimit(),
             File.separatorChar,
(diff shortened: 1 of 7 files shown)


brettwooldridge:
@illicitonion @Coneko Can you review this?  It removes the hard-coded paths for the C/C++ compilers (well, it does fall-back to them if they are not in the path), which allows the travis-ci build to return to running in container mode; cutting build times from ~45min to ~25min.

src/com/facebook/buck/cxx/DefaultCxxPlatforms.java
```diff
@@ -42,15 +43,44 @@ private DefaultCxxPlatforms() { }
...
+  private static final Path DEFAULT_RANLIB;
+  private static final Path DEFAULT_NM;
 
-  private static final Path DEFAULT_OSX_C_FRONTEND = Paths.get("/usr/bin/clang");
-  private static final Path DEFAULT_OSX_CXX_FRONTEND = Paths.get("/usr/bin/clang++");
+  private static final Path DEFAULT_OSX_C_FRONTEND;
+  private static final Path DEFAULT_OSX_CXX_FRONTEND;
+
+  static {
+    ImmutableMap<String, String> env = ImmutableMap.copyOf(System.getenv());
```
illicitonion: Instead of using System.getenv() here, please pass it down through the chain of functions all the way from Main.runMainWithExitCode's clientEnvironment (yeah, it's a long way :()

I would possibly also be tempted to just pass down {PATH: "..."} rather than the whole environment map to be explicit about what's being used.

src/com/facebook/buck/cxx/DefaultCxxPlatforms.java
```diff
@@ -42,15 +43,44 @@ private DefaultCxxPlatforms() { }
...
+  private static final Path DEFAULT_AR;
+  private static final Path DEFAULT_STRIP;
+  private static final Path DEFAULT_RANLIB;
+  private static final Path DEFAULT_NM;
 
-  private static final Path DEFAULT_OSX_C_FRONTEND = Paths.get("/usr/bin/clang");
-  private static final Path DEFAULT_OSX_CXX_FRONTEND = Paths.get("/usr/bin/clang++");
+  private static final Path DEFAULT_OSX_C_FRONTEND;
+  private static final Path DEFAULT_OSX_CXX_FRONTEND;
+
```
illicitonion: Rather than a static initialiser, let's look up these paths inside build. Specifically, these are platform-specific, and not all of them need to exist on all platforms, so let's not try to look then up if we're not going to use them.

test/com/facebook/buck/cxx/CxxCompilationDatabaseIntegrationTest.java
```diff
@@ -599,12 +600,13 @@ private void assertHasEntry(
     String key = tmp.getRoot().toRealPath().resolve(fileName).toString();
     CxxCompilationDatabaseEntry entry = fileToEntry.get(key);
     assertNotNull("There should be an entry for " + key + ".", entry);
-    assertEquals(
-        Joiner.on(' ').join(
-            Iterables.transform(
-                command,
-                Escaper.SHELL_ESCAPER)),
-        entry.getCommand());
+    assertThat(
```
illicitonion: Can you switch this back to equality, but do the ExecutableFinder lookup again?

brettwooldridge:
@illicitonion I appreciate the review.  Unfortuntely, I don't have any more time to invest in this side-trip.  My goal was to minimize the amount of change while making the code "no worse" than it already was.  Given the previous hardcoded values, I feel it is appreciably better.   My intention was re-enabling the containerized travis-ci build, which we in the community depend on quite heavily, to complete in a reasonable time.  I am more than happy if someone else wants to take up the challenge of the above refactors.

sdwilsh:
Wouldn't the easier fix on Travis be just adding a `.buckconfig.local` file with the right locations for these tools?

illicitonion:
Unfortunately we'd need to do that for every single testdata directory, because each one creates its own .buckconfig.

The other way of doing this would be to inherit whitelisted keys from the top-level .buckconfig in each of the test invocations.

brettwooldridge:
I feel like I've entered into a parallel universe where it is "correct" to hardcode tool paths.

I'm on MacOS, why not change [buck_tool.py](https://github.com/facebook/buck/blob/master/programs/buck_tool.py#L203) from using that ``which("java")`` function and just drop in the string ``"/Library/Java/JavaVirtualMachines/jdk1.8.0_111.jdk/Contents/Home/bin/java"``?  I know, I know, maybe there are *some* users that doesn't work for, but we could allow specifying the location of Java in ``.buckconfig``.  Like Lua, we could add a setting in the ``[java]`` section:
```ini
[java]
  java = /usr/bin/java
```
Or -- *now just let me finish because I know how crazy it sounds* -- maybe *all* external tool executables should be located by the equivalent of ``which()`` by default.

brettwooldridge:
This refactor represents my final effort on this issue.

illicitonion:
@brettwooldridge I understand your frustration. I would definitely like to have good support for toolchains in Buck, but right now sadly our situation is that we don't have a good way of expressing dependencies on tools, and in the absence of a good way, we are very inconsistent about how we do so.

For finding tools, reaching into the environment is one answer, and right now we don't really have a better alternative. But it's important that we track when that environment changes. If we just call System.getenv() anywhere, it makes it very hard to correctly cache things. So we have built some data flows which e.g. redact some environment variables. Some of us have been talking recently about being even more specific about tracking which environment variables affect which bits of the graph, so that we can do selective invalidation. I appreciate your help in making this work easier in the future by using one of the existing flows, rather than just calling System.getenv().

I think that this pull request is the right thing to do, but because it *does* introduce a dependency on the environment where there wasn't one before, I'd really like to see a test which makes sure that we actually re-build if that environment changes (or if a matching binary is added to a directory earlier to the PATH). Right now, I'm pretty sure that if the environment variable actually changes, we'll re-build, but if a new gcc is added to an existing PATH entry, we probably *don't* re-build, and will give incorrect results. That kind of correctness issue causes real problems, so a test to show that it's not an issue, and that we're not introducing new correctness bugs, would really be nice.

brettwooldridge:
@illicitonion If you look at the latest commit, ``System.getenv()`` is no longer called.  The code uses the environment as it is passed up through ``CxxBuckConfig``, which delegates to the ``BuckConfig`` that is passed up the chain.

The code *always* had an issue in that if gcc is updated buck may not rebuild.  Just because the compiler path was hardcoded to ``/usr/bin/gcc`` doesn't imply that gcc can't and hasn't changed.  Installation paths such as that are as much ``the environment`` as anything.

illicitonion:
Yes, the code always had an issue that if the binary is updated buck may not rebuild. And that's gradually being improved with things like HashedFileTool which track when those files change.

The code *didn't* have an issue that creating a file which isn't referenced anywhere in Buck can change the output. Consider:

```
$ echo $PATH
/foo:/usr/bin
$ which gcc
/usr/bin/gcc
$ buck build blah
...
$ touch /foo/gcc && chmod a+x /foo/gcc
$ which gcc
/foo/gcc
$ buck build blah
...
```

This is the case I'm worried about. It *could* be that the code already handles this (if we eagerly drop caches when the DefaultCxxPlatform changes); I'm not sure. Which is why I'd like a simple testcase which shows that this is handled ok. That doesn't feel like a huge burden to ask?

The other testcase I was asking for, which I'm pretty sure *will* pass because we eagerly drop caches for all environment variable changes is:

```
$ echo $PATH
/usr/bin
$ which gcc
/usr/bin/gcc
$ buck build blah
...
$ export PATH="/foo:$PATH"
$ which gcc
/foo/gcc
$ buck build blah
...
```

brettwooldridge:
@illicitonion While it is not a "huge burden", it has the feeling of busy work.  The existing tests never covered the case where ``/usr/bin/gcc`` might change, and it feels like I'm being asked to add a test to cover a quite frankly more obscure scenario.  I consider it *highly* more likely that a user might do a ``yum upgrade gcc`` than the likelihood that a file named ``gcc`` that was previously in the path but not executable suddenly becomes so.

I don't mean to be a dick, but honestly I've spent a full day on this, and I'm not collecting a paycheck from Facebook.  My company doesn't even *have* any C/C++ code in a buck project; this honestly has nothing to do with our use of buck.  I've merely made this change because the process of submitting *other contributions* (namely improved kotlin support) became increasing painful when the continuous build jumped from 24 minutes to 45.  I understand that is not a pain that Facebook feels, as your continuous build servers are internal.

I was trying to get this in before the holidays, but unfortunately in 16 hours from now I will be on an airplane (and it is 11pm here in Tokyo).

I am sorry if you are not comfortable with the request as it stands, or if you yourself are not willing to pitch-in to add the unit test you'd like to see, but as I said above, the last commit was my final one on this pull request.  Thanks for you time.

LegNeato:
Thanks for the change @brettwooldridge! I am willing to write the unit test, I'll try to get to it today.

LegNeato:
I added a test in https://github.com/LegNeato/buck/commit/bb317679b29631d7530917c92a8d01ec1d14951b but can't get it to pass. Not sure if it is failing because the behavior is indeed incorrect or that my test is wrong. I'll take another look at it later tonight, but feel free to point out any obvious silliness 😄

brettwooldridge:
@LegNeato Thanks for the test.  I fixed it.  Essentially an issue of needing to make the generated shell scripts executable.  I've merged into my  branch, and running a last local test before pushing it.

brettwooldridge:
@illicitonion @sdwilsh Thanks to @LegNeato a test was added verifying that of the contents (hash) of the clang++/g++ binary change, that a build is re-triggered appropriately.  This pull request is ready for review.

sdwilsh:
I'm going to leave this for @illicitonion since I don't actively hack on 
Buck much these days anymore.

LegNeato:
@brettwooldridge ah, I suspected that was the issue and was going to try to fix it today! Thanks for getting it working :-).

brettwooldridge:
@illicitonion Is there anything else needed for this pull request?

illicitonion:
This looks great, thanks a lot @LegNeato and @brettwooldridge for putting together that test :)

Sorry for the delay in getting this merged, I've been away from a computer for a couple of weeks. I'll get this merged right away :)

PR #1064, opened 2016-12-20
Static prebuilt framework

robbertvanginkel:
The undocumented rule prebuilt_apple_framework worked for dynamic frameworks, but not for static frameworks. Some 3rd party integrations (googlemaps, crashlytics, frabric) offer static frameworks for easy integration of a prebuilt static lib + resources.

Follow up prs will come for making the prebuilt frameworks work for project gen and then write some docs on the rule.


ryu2:
@facebook-github-bot shipit

PR #1065, opened 2016-12-20
Add instructions on sharing macros to docs

ultramiraculous:
I was recently trying to get some macros setup for a project and I couldn't remember how includes were supposed to work, and it would have been helpful if the page on macros was more instructive. This is just adding a section to introduce the concept of project includes so it's easier to come across/is in more than one place. 

Also add some section headers to break up the article.


ttsugriy:
Thank you for your contribution @ultramiraculous!

PR #879, opened 2016-09-14
Populate Styleable entries correctly in R.java generated by MiniAapt

raviagarwal7:
- Styleable entries array now contain right int values equal to its correspondding attr values. Android attributes are still given random values since R.java is not available at MergeAndroidResourcesStep for libraries.
- Set R idValues of a parent library equal to it’s dependent libraries idValue.

PR #1068, opened 2016-12-25
Sort RDotTxtEntry such that styleable entries are kept together in R.txt

raviagarwal7:
Correct Sorting:
```
int[] styleable ActionBar {  }
int styleable ActionBar_background 10
int styleable ActionBar_backgroundSplit 12
int styleable ActionBar_backgroundStacked 11
int styleable ActionBar_contentInsetEnd 21
...
int[] styleable ActionBarLayout { 0x7f060008 }
int styleable ActionBarLayout_android_layout_gravity 0
```

Incorrect Sorting
```
int[] styleable ActionBar {  }
int[] styleable ActionBarLayout { 0x7f060008 }
int styleable ActionBarLayout_android_layout_gravity 0
int styleable ActionBar_background 10
int styleable ActionBar_backgroundSplit 12
int styleable ActionBar_backgroundStacked 11
int styleable ActionBar_contentInsetEnd 21
...
```

### DIFF
--- a/src/com/facebook/buck/android/aapt/RDotTxtEntry.java
+++ b/src/com/facebook/buck/android/aapt/RDotTxtEntry.java
@@ -216,10 +216,24 @@ public class RDotTxtEntry implements Comparable<RDotTxtEntry> {
       return 0;
     }
 
-    return ComparisonChain.start()
-        .compare(this.type, that.type)
-        .compare(this.name, that.name)
-        .result();
+    ComparisonChain comparisonChain = ComparisonChain.start().compare(this.type, that.type);
+
+    String [] thisNameParts = this.name.split("_");
+    String [] thatNameParts = that.name.split("_");
+
+    int index = 0;
+    while (index < thisNameParts.length && index < thatNameParts.length) {
+      comparisonChain = comparisonChain.compare(thisNameParts[index], thatNameParts[index]);
+      index++;
+    }
+
+    if (index < thisNameParts.length) {
+      comparisonChain = comparisonChain.compare(thisNameParts[index], "");
+    } else if (index < thatNameParts.length) {
+      comparisonChain = comparisonChain.compare("", thatNameParts[index]);
+    }
+
+    return comparisonChain.result();
   }
 
   @Override
--- a/src/com/facebook/buck/android/aapt/RDotTxtEntry.java
+++ b/src/com/facebook/buck/android/aapt/RDotTxtEntry.java
@@ -227,10 +227,8 @@ public class RDotTxtEntry implements Comparable<RDotTxtEntry> {
       index++;
     }
 
-    if (index < thisNameParts.length) {
-      comparisonChain = comparisonChain.compare(thisNameParts[index], "");
-    } else if (index < thatNameParts.length) {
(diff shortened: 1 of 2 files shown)


raviagarwal7:
This is a part of a bigger pull request - https://github.com/facebook/buck/pull/879#issuecomment-266749552. Breaking it down so that we can figure out what exactly breaks FB's robolectric tests.

src/com/facebook/buck/android/aapt/RDotTxtEntry.java
```diff
@@ -216,10 +216,24 @@ public int compareTo(RDotTxtEntry that) {
...
+    while (index < thisNameParts.length && index < thatNameParts.length) {
+      comparisonChain = comparisonChain.compare(thisNameParts[index], thatNameParts[index]);
+      index++;
+    }
+
+    if (index < thisNameParts.length) {
+      comparisonChain = comparisonChain.compare(thisNameParts[index], "");
+    } else if (index < thatNameParts.length) {
+      comparisonChain = comparisonChain.compare("", thatNameParts[index]);
+    }
```
dreiss: It looks like this is not guaranteed to consume the entire string.  What if the longer one has a double-underscore in it?  Couldn't that give a false "equal" comparison with a different string?  What about comparing the two array lengths instead of this block.  Would that work?

test/com/facebook/buck/android/aapt/RDotTxtEntryTest.java
```diff
@@ -0,0 +1,57 @@
...
+        new RDotTxtEntry(IdType.INT, RType.STYLEABLE, "ActionBarLayout_android", "0"),
+        new RDotTxtEntry(IdType.INT, RType.STYLEABLE, "ActionBar_backgroundStack", "1")
+    );
+
+    ImmutableList<RDotTxtEntry> sortedEntries = ImmutableList.copyOf(
+        Ordering.natural().sortedCopy(entries)
+    );
+
+    assertEquals(
+        sortedEntries,
```
dreiss: Expected value should be the first argument, and actual value the second.

src/com/facebook/buck/android/aapt/RDotTxtEntry.java
raviagarwal7: nice catch! fixed the issue, using array length comparison as you suggested.

test/com/facebook/buck/android/aapt/RDotTxtEntryTest.java
raviagarwal7: fixed.

sdwilsh: Notably, this is why I think `assertThat` is better (plus it gives better error messages).  I thought the team was encouraging the use of that instead?

raviagarwal7:
@dreiss -- are the tests passing in FB's codebases with this change? would love to get this merged soon so I can break https://github.com/facebook/buck/pull/879 further.

dreiss:
Sorry, we had some internal bugs that were making it tough for me to test this.  Giving it another shot.

raviagarwal7:
nw, thanks for looking into this.

PR #1079, opened 2017-01-06
Add NDK cxx_runtime_type to link static C++ runtime

### LINKED ISSUE #741: Missing NDK support for static runtimes/no runtime
There's no way to specify static or shared runtime (always shared) in the `[ndk]` section of `.buckconfig`.

We currently rely on `libc++`; however, there's a [bug](http://developer.android.com/ndk/guides/cpp-support.html#rc) in `c++_shared` where exceptions do not always work on arm devices. The solution is to use `c++_static`, but this is not possible using `cxx_library` rules as they are now. A fix for this would be much appreciated.

EDIT: A more detailed explanation of the exceptions issue is [here](http://gosuwachu.io/2016/04/09/android-ndk-bug/)

splhack:
Fix #741

Example:

    [ndk]
      ndk_version = r10e
      app_platform = android-21
      compiler = clang
      cxx_runtime = libcxx
      cxx_runtime_type = static

### DIFF
--- a/src/com/facebook/buck/android/AndroidBuckConfig.java
+++ b/src/com/facebook/buck/android/AndroidBuckConfig.java
@@ -17,6 +17,7 @@
 package com.facebook.buck.android;
 
 import com.facebook.buck.cli.BuckConfig;
+import com.facebook.buck.cxx.Linker;
 import com.facebook.buck.util.environment.Platform;
 import com.google.common.collect.ImmutableSet;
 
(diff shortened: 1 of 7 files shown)


dreiss:
Andrew, you wrote the ndk linking stuff, right?  Thoughts on this?

andrewjcg:
Bleh, sorry for taking forever on this.  Changes look good to me.

styurin:
This was implemented in 3dddf893259dae0e4a492d29f4c5ecb575d96810

PR #1082, opened 2017-01-09
Add snapshot_reference_images_path to apple_test

robbertvanginkel:
For usage with the https://github.com/facebook/ios-snapshot-test-case repo.

Snapshot test reference image paths are controlled by an environment variable that is set on the test run process. Currently though it is not possible to add an environment variable to a specific test target. This commit allows indivudual apple_test rules to reference different paths for snapshot test reference images.

This is a specific field instead of a more general thing like `test_environment_variables`, to allow the value to be a SourcePath so it can be combined with the `export_file` rule.


ryu2:
@facebook-github-bot shipit

PR #1106, opened 2017-01-11
snapshot_reference_images_path can be either a path or a string

### LINKED ISSUE #1091: snapshot_reference_images_path can't be verified
I'm trying to use `napshot_reference_images_path` introduced in #1082. Here is the relevant line in my BUCK file:
```
snapshot_reference_images_path='AsyncDisplayKitTests/ReferenceImages',
```

I got this error when I tried to run buck test:
```
BUILD FAILED: //:Tests: parameter 'snapshotReferenceImagesPath': no such file or directory 'AsyncDisplayKitTests/ReferenceImages'
```

If I change the path to a valid path, say `AsyncDisplayKitTests/ReferenceImages_iOS_10`, the build will proceed. But then FBSnapshotTestCase will try to append a suffix to this path, thus turn it into `ReferenceImages_iOS_10_iOS_10` and tests will fail because it's no longer a valid path.

I think `snapshot_reference_images_path` should be either a string or an unverified path. Or maybe I'm using it wrong?

nguyenhuy:
Summary:
- Allow an unverified string to be used as snapshot_reference_images_path in apple_test.
- An invalid string will cause snapshot test cases to fail.

Test Plan: unit test

Resolves #1091


src/com/facebook/buck/apple/XctoolRunTestsStep.java
```diff
@@ -96,7 +97,7 @@
   private final Optional<String> logLevelEnvironmentVariable;
   private final Optional<String> logLevel;
   private final Optional<Long> timeoutInMs;
-  private final Optional<Path> snapshotReferenceImagePath;
+  private final Optional<Either<Path, String>> snapshotReferenceImagesPath;
```
robbertvanginkel: I think you can simplify this by just making it a String, and pass in the String version of the path you get from `getResolver().getAbsolutePath(this.snapshotReferenceImagesPath.get().getLeft())` when calling the constructor for this.

Its important that this is a sourcepath at AppleTest.java line 113, so that the rulekey reference for the cache uses a relative path. After you've called absolute path on it though you can just treat this as a string. No use in two code paths if the only thing you do with the `Path` is call `toString` at the end.

nguyenhuy:
@robbertvanginkel Ready for another round. Thanks!

robbertvanginkel:
Looks good to me, but I don't have merging powers.

ryu2:
@facebook-github-bot shipit

src/com/facebook/buck/apple/AppleTest.java
```diff
@@ -303,6 +304,20 @@ public boolean hasTestResultFiles() {
...
+
+      Optional<String> snapshotReferenceImagesPath = Optional.empty();
+      if (this.snapshotReferenceImagesPath.isPresent()) {
+        if (this.snapshotReferenceImagesPath.get().isLeft()) {
+          snapshotReferenceImagesPath = Optional.of(
+              getResolver().getAbsolutePath(this.snapshotReferenceImagesPath.get().getLeft())
+                  .toString());
+        } else if (this.snapshotReferenceImagesPath.get().isRight()) {
+          snapshotReferenceImagesPath = Optional.of(
+              getProjectFilesystem().getRootPath().getFileSystem().toString() +
```
ryu2: Calling `toString()` here is incorrect.  You want `getPathForRelativePath()` in `ProjectFilesystem` instead.

test/com/facebook/buck/apple/XctoolRunTestsStepTest.java
```diff
@@ -655,4 +656,61 @@ public void testDirectoryAndLevelPassedInEnvironment() throws Exception {
...
+        Optional.empty(),
+        Suppliers.ofInstance(Optional.of(Paths.get("/path/to/developer/dir"))),
+        TestSelectorList.EMPTY,
+        Optional.empty(),
+        Optional.empty(),
+        Optional.empty(),
+        Optional.empty(),
+        Optional.empty(),
+        Optional.of(Either.ofRight("path/to/snapshotimages")));
+    String snapshotPath = projectFilesystem.getRootPath().getFileSystem().toString() +
```
ryu2: Same issue as above.

test/com/facebook/buck/apple/XctoolRunTestsStepTest.java
```diff
@@ -655,4 +656,61 @@ public void testDirectoryAndLevelPassedInEnvironment() throws Exception {
...
+        Paths.get("/path/to/output.json"),
+        Optional.empty(),
+        Suppliers.ofInstance(Optional.of(Paths.get("/path/to/developer/dir"))),
+        TestSelectorList.EMPTY,
+        Optional.empty(),
+        Optional.empty(),
+        Optional.empty(),
+        Optional.empty(),
+        Optional.empty(),
+        Optional.of(Either.ofRight("path/to/snapshotimages")));
```
ryu2: Same as above.

test/com/facebook/buck/apple/XctoolRunTestsStepTest.java
```diff
@@ -620,7 +621,7 @@ public void testDirectoryAndLevelPassedInEnvironment() throws Exception {
         Optional.of("TEST_LOG_LEVEL"),
         Optional.of("verbose"),
         Optional.empty(),
-        Optional.of(Paths.get("/path/to/snapshotimages")));
+        Optional.of(Either.ofLeft(Paths.get("/path/to/snapshotimages"))));
```
ryu2: Need to change this, since you made `XctoolRunTestsStep` take an `Optional<String>`.

ryu2:
@facebook-github-bot shipit

PR #1127, opened 2017-01-13
Signed binary if needed in Apple bundle

fkorotkov:
I was getting `Foo.app/Foo: code object is not signed at all` when I was running `buck install --emulator FooApp` where `FooApp` is an `apple_bundle`.

Testing:

After this change I'm able to launch the app but will appreciate direction on where and how to add tests :-)

### DIFF
--- a/src/com/facebook/buck/apple/AppleBundle.java
+++ b/src/com/facebook/buck/apple/AppleBundle.java
@@ -547,6 +547,11 @@ public class AppleBundle
         };
       }
 
+      addBinarySigningStepIfNeeded(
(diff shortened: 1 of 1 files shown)


ryu2:
Thanks!  You can look at `checkCodeSigning()` in `AppleBundleIntegrationTest` for test examples.

ryu2:
Once tests are added, this should be good to merge.

fkorotkov:
I've also noticed that we had a Buck version from December and after updating it I can't actually reproduce the issue. Going to close for now.

PR #1147, opened 2017-01-20
Correct talk title

kageiit:


kageiit:
Should be a quick one @sdwilsh

PR #1155, opened 2017-01-30
Fix shouldRefuseToRunAppleTestIfXctestNotPresent

robbertvanginkel:
This currently fails if the xctool is available in the PATH, because
it auto resolves making this test result inconsistent across environments. 
Making the path fake makes this test consistently fail regardless of environment.


Coneko:
The test has a `.buckconfig`, so you should set the config there: https://github.com/facebook/buck/blob/master/test/com/facebook/buck/apple/testdata/apple_test_xctest/.buckconfig .

Coneko:
Oh sorry I noticed now the same testdata directory is used across multiple tests, it's likely some of them won't pass with this change in `.buckconfig` since it will apply to all.

Travis runs on Linux and Appveyor runs on Windows so these tests aren't shown in CI, we should probably set up a mac Travis configuration.

Sorry about the back and forth, make sure you test your change locally.

PR #1157, opened 2017-01-31
Add ability to specify multiple project res and assets directories

### LINKED ISSUE #1156: Add support for multiple dirs in project_res and project_assets args in android_resource rule
The newly added `project_res` and `project_assets` arguments on the `android_resouce` rule only support one directory but there are cases where you would want to specify multiple, such as in a top level application module that has multiple buildTypes or flavors.

cc @kageiit

kageiit:
Fixes #1156


kageiit:
cc @k21 @styurin @dreiss

kageiit:
This is not really required as the real issues are with buck project generation not picking up the right folders and excluding actual resource folders as discussed in #1156

PR #1158, opened 2017-02-02
[swift] respect prebuilt_apple_framework in deps

fkorotkov:
It appeared swift compilation wasn't respecting prebuilt frameworks

Testing:

`ant test -Dtest.class=SwiftLibraryIntegrationTest` passed


milend:
As per @ryu2's comments, there are a few changes that need to happen. Since the initial PR, `swift_library` is no longer supported and instead `apple_library` should be used (with `apple.use_swift_delegate = false` in the config). `apple_library` correctly handles prebuilt frameworks which can be used via Swift.

Issue #816, opened 2016-07-23
Styleable entries in R.txt and R.java is not generated correctly for libraries.


raviagarwal7:
**R.java generated when running with buck:**

```
public static class attr {
    public static final int popupTheme=0x07f010b0;

    public static final int floatingLabelLabelText=0x07f015ea;
    public static final int floatingLabelLabelTextAppearance=0x07f015eb;
    public static final int floatingLabelStartingText=0x07f015ec;
    ...
}

public static class styleable {
    public static final int[] Spinner={  };
    public static final int Spinner_android_dropDownWidth=3;
    public static final int Spinner_android_entries=0;
    public static final int Spinner_android_popupBackground=1;
    public static final int Spinner_android_prompt=2;
    public static final int Spinner_popupTheme=4;

    public static final int[] FloatingLabelEditText={ 0x7f060003,0x7f060004,0x7f060005,0x7f060006,0x7f060007,0x7f060008,0x7f060009,0x7f06000a,0x7f06000b,0x7f06000c,0x7f06000d,0x7f06000e,0x7f06000f,0x7f060010,0x7f060011,0x7f060012,0x7f060013,0x7f060014,0x7f060015,0x7f060016,0x7f060017,0x7f060018,0x7f060019,0x7f06001a,0x7f06001b,0x7f06001c,0x7f06001d };
    public static final int FloatingLabelEditText_android_enabled=23;
    public static final int FloatingLabelEditText_android_imeActionId=25;
    ...
    public static final int FloatingLabelEditText_floatingLabelLabelText=0;
    public static final int FloatingLabelEditText_floatingLabelLabelTextAppearance=3;
    public static final int FloatingLabelEditText_floatingLabelStartingText=1;
    ...
}
```

**R.java generated when running with gradle:**

```
public static final class attr {
    public static final int popupTheme=0x7f010034;

    public static int floatingLabelLabelText=0x7f0100f7;
    public static int floatingLabelLabelTextAppearance=0x7f0100fa;
    public static int floatingLabelStartingText=0x7f0100f8;
    ...
}

public static final class styleable {

    public static final int[] Spinner = {
        0x010100b2, 0x01010176, 0x0101017b, 0x01010262, 0x7f010034
    };

    public static int Spinner_android_dropDownWidth = 3;
    public static int Spinner_android_entries = 0;
    public static int Spinner_android_popupBackground = 1;
    public static int Spinner_android_prompt = 2;
    public static int Spinner_popupTheme = 4;   


[...]

PR #1159, opened 2017-02-03
Get resId value of an attribute from module’s styleable array if there is no explicit attribute entry.

raviagarwal7:
This applies for attributes which come from android sdk.

Correct Output
```
    public static int[] AlertDialog={ 0x010100f2,0x07f01074,0x07f010b3,0x07f010b4,0x07f010c0,0x07f010df };
    public static int AlertDialog_android_layout=0;
    public static int AlertDialog_buttonPanelSideLayout=1;
    public static int AlertDialog_listItemLayout=2;
    public static int AlertDialog_listLayout=3;
    public static int AlertDialog_multiChoiceItemLayout=4;
    public static int AlertDialog_singleChoiceItemLayout=5;
```

Previous Output
```
    public static int[] AlertDialog={ 0,0x07f01074,0x07f010b3,0x07f010b4,0x07f010c0,0x07f010df };
    public static int AlertDialog_android_layout=0;
    public static int AlertDialog_buttonPanelSideLayout=1;
    public static int AlertDialog_listItemLayout=2;
    public static int AlertDialog_listLayout=3;
    public static int AlertDialog_multiChoiceItemLayout=4;
    public static int AlertDialog_singleChoiceItemLayout=5;
```

### DIFF
--- a/src/com/facebook/buck/android/MergeAndroidResourcesStep.java
+++ b/src/com/facebook/buck/android/MergeAndroidResourcesStep.java
@@ -45,6 +45,7 @@ import java.io.IOException;
 import java.io.PrintWriter;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
@@ -54,6 +55,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class MergeAndroidResourcesStep implements Step {
   private static final Logger LOG = Logger.get(MergeAndroidResourcesStep.class);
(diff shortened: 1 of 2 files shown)


raviagarwal7:
@marcinkosiba @dreiss -- as discussed in https://github.com/facebook/buck/issues/816#issuecomment-236587468, there is an eaiser way to get resource id's for `android_*` resources. This change does that. 

Basically we can get the resource values of missing android attributes from `int[] styleable` of external aar's which generate their R.txt using AAPT. So for a module since we iterate in dependent order this works perfectly.

kageiit:
@marcinkosiba @dreiss or @Coneko can any of you please update on the status of this?

raviagarwal7:
bump @dreiss -- this change is solely inside the new styleable method `getStyleableResources`, so should not affect any old behaviour.

PR #1164, opened 2017-02-06
[swift] Allow bridging header import headers under current module

zayhero:
Right now when buck building swift with bridging header, it is not able to import header files when it reference from current library.

For example, in build rule with this setup

```
apple_library(
    name = 'Greeter',
    bridging_header = 'bridging-header.h',
    headers = glob([
        '*.h',
    ]),
    srcs = glob([
        '*.m',
        '*.swift',
    ]),
    frameworks = [
        '$SDKROOT/System/Library/Frameworks/Foundation.framework'
    ],
)
```

We can not do `#import <Greeter/Greeter.h>` in bridging header.

After investigation, I figure its due to buck didn't include the header map generated for current library in SwiftCompile.

This PR is to improve buck so it can support above use case.

**Testing**
And a new test case and verify it with `buck test --all`

### DIFF
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -25,8 +25,10 @@ import com.facebook.buck.cxx.CxxHeaders;
 import com.facebook.buck.cxx.CxxPlatform;
 import com.facebook.buck.cxx.CxxPreprocessables;
 import com.facebook.buck.cxx.CxxPreprocessorInput;
+import com.facebook.buck.cxx.HeaderVisibility;
 import com.facebook.buck.cxx.LinkerMapMode;
 import com.facebook.buck.io.ProjectFilesystem;
+import com.facebook.buck.model.BuildTarget;
 import com.facebook.buck.parser.NoSuchBuildTargetException;
 import com.facebook.buck.rules.AbstractBuildRuleWithResolver;
 import com.facebook.buck.rules.AddToRuleKey;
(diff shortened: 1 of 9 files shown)


aiked:
Hello @zayhero, I think @ryu2 will be able to help you with this.

zayhero:
@ryu2 could you help to take a look?

Coneko:
Can you import the header as `#import "Greeter.h"` in the bridging header?

zayhero:
@Coneko yes, using quote format works with bridging header.
However, in a complicate file system, this cause too much mess.

For example, we have two header files under `folderA/A.h` and `folderB/B.h`. In `A.h`, we do `#import <Greeter/B.h>`, this won't work with buck right now.
If we want to change to this import to quote format, we can either 1) move all these headers to same folder, or 2) have a very complicate import logic in each header. Both ways are quite mess to me.

Thus, I figure out the best way is adding the header map generated for current library into SwiftCompile, so we can import the libraries easier.

ryu2:
Looks like `AppleLibraryIntegrationTest` is failing:

```
    [junit] Testcase: testBuildAppleLibraryUsingBridingHeaderAndSwiftDotH took 10.48 sec
    [junit] 	FAILED
    [junit] Expected exit code 0 but was 1.
    [junit] junit.framework.AssertionFailedError: Expected exit code 0 but was 1.
    [junit] 	at com.facebook.buck.testutil.integration.ProjectWorkspace$ProcessResult.assertExitCode(ProjectWorkspace.java:785)
    [junit] 	at com.facebook.buck.testutil.integration.ProjectWorkspace$ProcessResult.assertSuccess(ProjectWorkspace.java:746)
    [junit] 	at com.facebook.buck.apple.AppleLibraryIntegrationTest.testBuildAppleLibraryUsingBridingHeaderAndSwiftDotH(AppleLibraryIntegrationTest.java:702)
```

zayhero:
@ryu2 thanks for reviewing! I accidentally remove the private header map from the list, so the new test case no long work. Already fixed, please take a look.

ryu2:
@facebook-github-bot shipit

fkorotkov:
@ryu2 we are seeing a race condition when `CxxDescriptionEnhancer.getHeaderSymlinkTreePath` is empty when we are compiling swift companion. It make sense since apple library depends on the companion and not visa versa. 

Can you suggest how to fix it? I tried to use `requireHeaderSymlinkTree` but it goes in infinite loop because of apple library/swift library dependencies.

PR #1168, opened 2017-02-08
[swift] Add -iquote rootPath into clang args

zayhero:
When we use bridging header to generate swift modules, swift sometimes include the bridging header into `-Swift.h`. For example, sth like this

```
#import "bridging-header/bridging-header.h"
```

Theoretically, swift should only import `bridging-header.h` instead of the full path, but that's not what they are doing right now.
And BUCK is not able to build the objC code if its importing the `-Swift.h`. `clang` throws this exception

```
In file included from CallHello.m:1:
In file included from CallHello.h:1:
buck-out/gen/Greeter#macosx-x86_64,static,swift-compile/Greeter-Swift.h:121:9: fatal error: 'bridging-header/bridging-header.h' file not found
#import "bridging-header/bridging-header.h"
```

To fix this problem, this PR is adding `-iquote rootPath` into clang flags, so clang can build the objC files.

For testing, update the tests to verify the change.

PS. I am not sure if making change on `AppleCxxPlatforms` is the best approach, please let me know if there is a better way. I am also thinking adding `compiler_flags = ['-iquote', os.getcwd()]` in the build rule, but this seems too hacky to me.


zayhero:
@ryu2 could you take a look at this PR as well?

ryu2:
@facebook-github-bot shipit

yiding:
I'm going to revert this diff since adding the repo root to quoted include paths is a very poor way of solving the problem you have here.  Not only does it break invariants and allow rules to access arbitrary headers relative to the root via quoted includes, it doesn't even fix your problem for any case other than when the target is defined at the root of the repo. It will also breaks caching if not for DebugPathSanitizer doing a last-ditch effort to clean up repo roots.

If you can give a detailed description of what the problem is, complete with a repro case, I can probably replace this solution with something more sensible.

zayhero:
@yiding Actually this fix our problems and we are using this change to build our entire iOS app with Buck. In our use case, even the targets are not defined at the root, Buck still use the relative path to root path to generate the `*-Swift.h`, so it does solve the problem.

So our problem is, when we pass bridging header file (e.g. `bridging-header.h`) to `apple_library`, Buck will call `swiftc` to generate `*-Swift.h`. In the generated `*-Swift.h` file, it contains `#import buckPath/bridging-header.h`.
When another objC file is importing the `*-Swift.h`, it will try to import `buckPath/bridging-header.h`, but this doesn't work since `clang` doesn't able to locate `buckPath/bridging-header.h`. By adding `-iquote` into the compile args, it ensure `clang` can locate the bridging header file correctly.
Please let me know if you have any idea about how to fix this problem. Thanks!

Coneko:
@yiding I believe Danil from the AR Studio team has hit this issue last week, which means this fix doesn't even fix this problem in general. He has a patch to fix it for his case and I told him to send it to @ryu2 to review, I have not seen the patch so I can't tell if it makes sense or not.

yiding:
I think the correct action here is to ensure the include directives in the swiftc generated file follows the pattern of include directives for buck-based apple targets. This might involve a rewrite step in SwiftCompile to rewrite the include directives in the `-Swift.h` bridging header if swiftc doesn't allow this to be specified manually. Additionally, the input objc bridging header should be added to the exported headers, so that the output `-Swift.h` bridging header can be accessed from other libraries.

zayhero:
@yiding We are doing something similar internally. We introduce a script to replace the import statement, so we can ensure it always doing `#import <Lib/Lib.h>`.
Your idea here seems better, that just do the replace logic on the `*-Swift.h` in `SwiftCompile`.
Am I understand correctly?

yiding:
Correct, essentially, SwiftCompile.getBuildSteps will return another step after the compile step which fixes the generated header file.

However do note that this bridging header should either be explicitly or implicitly added to the exported headers of the library in question, so that the inclusion directive actually references a valid file, especially when used from another library.

zayhero:
@yiding Thanks for the information! Will you implement this logic into Buck directly? 
If you are not planning to do that, I can try to implement this logic myself.

yiding:
Feel free to have a go at it. If you leave it to me who knows when it'll be done.

PR #1189, opened 2017-02-17
Fix default value in docs for `allow_symlinks`

jaimeagudo:
As per https://github.com/facebook/buck/blob/5faa80d0a10e252b79274aa59edead8e0e4e0d20/src/com/facebook/buck/parser/AbstractParserConfig.java#L136


davido:
Thanks for the fix!

PR #1194, opened 2017-02-21
Fix Guava 20.0 migration

davido:
Some places were erroneously migrated to calling throwIfUnchecked(e)
even though the thrown exception was known to be a checked exception.


marcinkosiba:
getting a bunch of `arc lint` failures about unused imports.

davido:
My bad, will address it and update the PR.

davido:
@marcinkosiba Done. PTAL.

PR #1197, opened 2017-02-22
Add option to specify max heap size for out of process dx

kageiit:
- Also move `dx_threads` into `DxConfig`.
- Documentation update will follow in the next commit if code looks good.

### DIFF
--- a/src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java
+++ b/src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java
@@ -119,6 +119,7 @@ public class AndroidBinaryGraphEnhancer {
   private final AndroidNativeLibsPackageableGraphEnhancer nativeLibsEnhancer;
   private final APKModuleGraph apkModuleGraph;
   private final ListeningExecutorService dxExecutorService;
+  private final DxConfig dxConfig;
 
   AndroidBinaryGraphEnhancer(
       BuildRuleParams originalParams,
@@ -154,7 +155,8 @@ public class AndroidBinaryGraphEnhancer {
       ListeningExecutorService dxExecutorService,
       ManifestEntries manifestEntries,
       CxxBuckConfig cxxBuckConfig,
-      APKModuleGraph apkModuleGraph) {
+      APKModuleGraph apkModuleGraph,
+      DxConfig dxConfig) {
     this.buildRuleParams = originalParams;
     this.manifestEntries = manifestEntries;
     this.originalBuildTarget = originalParams.getBuildTarget();
@@ -198,6 +200,7 @@ public class AndroidBinaryGraphEnhancer {
             relinkerMode,
             apkModuleGraph);
     this.apkModuleGraph = apkModuleGraph;
+    this.dxConfig = dxConfig;
   }
 
   AndroidGraphEnhancementResult createAdditionalBuildables() throws NoSuchBuildTargetException {
@@ -597,7 +600,8 @@ public class AndroidBinaryGraphEnhancer {
         allPreDexDeps,
         dexForUberRDotJava,
         dxExecutorService,
-        xzCompressionLevel);
+        xzCompressionLevel,
+        dxConfig.getDxMaxHeapSize());
     ruleResolver.addToIndex(preDexMerge);
 
     return preDexMerge;
--- a/src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java
+++ b/src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java
@@ -454,7 +454,8 @@ public class AndroidBinaryGraphEnhancer {
         Suppliers.ofInstance(ImmutableSortedSet.of(compileUberRDotJava)),
         /* extraDeps */ Suppliers.ofInstance(ImmutableSortedSet.of()));
     DexProducedFromJavaLibrary dexUberRDotJava =
-        new DexProducedFromJavaLibrary(paramsForDexUberRDotJava, compileUberRDotJava, dxConfig.getDxMaxHeapSize());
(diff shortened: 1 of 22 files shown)


kageiit:
cc @dreiss @malbano

src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java
```diff
@@ -597,7 +600,8 @@ PreDexMerge createPreDexMergeRule(
         allPreDexDeps,
         dexForUberRDotJava,
         dxExecutorService,
-        xzCompressionLevel);
+        xzCompressionLevel,
+        dxConfig.getDxMaxHeapSize());
```
dreiss: Unclear whether this is necessary, because merging is much less memory intensive, and we usually run it in-process, but okay.

src/com/facebook/buck/android/DxConfig.java
```diff
@@ -0,0 +1,44 @@
...
+   */
+  public Optional<String> getDxMaxHeapSize() {
+    return delegate.getValue("dx", "dx-max-heap-size");
+  }
+
+  /**
+   * @return The dx thread count.
+   */
+  public Optional<Integer> getDxThreadCount() {
+    return delegate.getInteger("dx", "dx-threads");
```
dreiss: Why does this start with "dx-"?  I think it had the "dx" prefix before because it was in the java section, but now it probably makes more sense to just call it "threads" or "max_threads".

I guess anyone using the old directive will need to update their .buckconfig?  Should it fall back to checking the old value, with a warning?

src/com/facebook/buck/android/DxConfig.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public DxConfig(BuckConfig delegate) {
+    this.delegate = delegate;
+  }
+
+  /**
+   * @return The upper heap size limit for out of process dx if specified.
+   */
+  public Optional<String> getDxMaxHeapSize() {
+    return delegate.getValue("dx", "dx-max-heap-size");
```
dreiss: I think the convention is underscore separators, rather than dashes, right?

kageiit: proguard config was using dashes. I can use `_` if thats the convention, but was unsure

src/com/facebook/buck/android/DxConfig.java
kageiit: I can remove the `dx` prefix. I can also add the fallback. Whats the best way to show a warning?

src/com/facebook/buck/android/DxConfig.java
dreiss: Oh, interesting.  Yeah, almost everything uses underscores, though.

src/com/facebook/buck/android/DxConfig.java
dreiss: That I'm not sure.  It looks like KnownBuildRuleTypes has instances of LOG.warn and LOG.error.  Maybe try those?

dreiss:
This looks good to me.  I need to get the proper config values added to our .buckconfig before I can import it.

src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java
kageiit: @dreiss Do you want want me to change this or is there any other change you requested?

dreiss: Nope.  I just had a few speedbumps on the import.  No other changes required IMO.

dreiss:
I'm trying to land this.  I did notice two non-critical issues that you might want to address.

1/ This isn't being added to rule keys, which means that if you set the limit differently on CI and laptops, you might have engineers getting OOMs on rules that are building cleanly in CI, or vice-versa.  I kind of think that is okay because lots of things can already affect whether you get an OOM.

2/ You're not passing this down to DexProducedFromJavaLibrary, which is the rule for dexing a single java library in a debug build.

kageiit:
I will look into fixing the second part. The first as you explained can already happen in many other cases

kageiit:
@dreiss I addressed your comments and rebased onto latest

dreiss:
```
[Thu, 02 Mar 2017 08:40:01] ant lint
checkstyle:
[checkstyle] Running Checkstyle 6.16.1 on 3281 files
[checkstyle] [ERROR] src/com/facebook/buck/android/AndroidBinaryGraphEnhancer.java:457: Line is longer than 100 characters (found 115). [LineLength]
```

:(

kageiit:
ah i forgot to run lint. lemme fix that :(

kageiit:
ok fixed and pushed. sry abt that

kageiit:
@dreiss Any issues with import I can help resolve?

kageiit:
@dreiss this file seems to get more conflicts the longer its not merged. Any update on whats holding back the merge?

test/com/facebook/buck/android/DxStepTest.java
```diff
@@ -200,6 +201,33 @@ public void testVerbose10AddsVerboseFlagToDx() throws IOException {
...
+      DxStep dx = new DxStep(
+          filesystem,
+          SAMPLE_OUTPUT_PATH,
+          SAMPLE_FILES_TO_DEX,
+          EnumSet.noneOf(DxStep.Option.class),
+          Optional.of("2g"));
+
+      String expected = String.format(
+          "%s --output %s %s",
+          "/usr/bin/dx -JXmx2g --dex",
```
dreiss: FYI, this broke the windows test because it's using hard-coded "/" instead of EXPECTED_DX_PREFIX.  I'm fixing it locally.

PR #1201, opened 2017-02-22
Close test result output file properly

kageiit:
Currently, the `BaseRunner` does not correctly close the underlying `FileOutputStream` when writing test results out. This is because the current `close()` call is only performed on `BufferedOutputStream` which simply class `flush()` on the underlying stream but never closes it. This leads to too many files open exception and the test run exiting prematurely when there are several test rules running on a machine at the same time. This change fixes that problem

### DIFF
--- a/src/com/facebook/buck/testrunner/BaseRunner.java
+++ b/src/com/facebook/buck/testrunner/BaseRunner.java
@@ -157,9 +157,11 @@ public abstract class BaseRunner {
       testSelectorSuffix += ".dry_run";
     }
     OutputStream output;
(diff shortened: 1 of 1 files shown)


kageiit:
cc @dreiss

dreiss:
`BufferedOutputStream` inherits close from `FilterOutputStream`, which says " The close method of FilterOutputStream calls its flush method, and then calls the close method of its underlying output stream."  It seems like this should work fine (unless BufferedOutputStream's constructor throws).

kageiit:
@dreiss thanks for the info. Going to close this. Seems like the problem is something else

PR #1205, opened 2017-02-27
Make Groovy descriptions extend java

kageiit:
- Make the groovy description classes extend the java versions similar to what the kotlin rules do.
- Also make sure to set the right resource root etc. so that tests can work correctly


kageiit:
cc @dreiss @dsyang

PR #1208, opened 2017-03-01
Fix race condition on Swift build

zayhero:
In PR https://github.com/facebook/buck/pull/1164, we introduce the logic to allow `SwiftCompile` use the header maps to built bridging headers.
However, we encounter race condition issue when we building with this logic.

The headers maps, is generated by `CxxLibraryDescription`, and `SwiftCompile` step is generated by `SwiftLibraryDescription`. These two build rules are running in parallel, so its not guarantee that header maps are ready when we running `SwiftCpmile`.

To fix this issue, I am making changes to:
1. call `CxxDescriptionEnhancer.requireHeaderSymlinkTree ` to create the header symlink tree BuildRule
2. let `SwiftLibraryDescription` depends on the symlink tree BuildRule, also make changes to not filter out BuildRules if its building for headers

@ryu2 could you help to take a look on this? Thanks!

### DIFF
--- a/src/com/facebook/buck/apple/AppleLibraryDescription.java
+++ b/src/com/facebook/buck/apple/AppleLibraryDescription.java
@@ -25,6 +25,8 @@ import com.facebook.buck.cxx.CxxLibraryDescription;
 import com.facebook.buck.cxx.CxxPlatform;
 import com.facebook.buck.cxx.CxxStrip;
 import com.facebook.buck.cxx.FrameworkDependencies;
+import com.facebook.buck.cxx.HeaderSymlinkTree;
+import com.facebook.buck.cxx.HeaderVisibility;
 import com.facebook.buck.cxx.Linker;
 import com.facebook.buck.cxx.LinkerMapMode;
 import com.facebook.buck.cxx.ProvidesLinkedBinaryDeps;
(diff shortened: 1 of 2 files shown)


nguyentruongtho:
I think there is a disconnection between two parts, create the header rule (this PR) and using the output header (https://github.com/facebook/buck/pull/1164). Why don't we require the header rule somewhere before creating swift compile?

zayhero:
@nguyentruongtho Just to make sure I understand your idea, are you suggesting that we pass the header information in `SwiftLibraryDescription`, and then call `requireHeaderSymlinkTree ` before we generate the `SwfitCompile` build rule?
I think about this before, and I think it require more changes to pass the arguments over, but with my current approach, we save the effort on changing the interface between `AppleLibraryDescription` and `SwiftLibraryDescription`. 
How do you think?

src/com/facebook/buck/apple/AppleLibraryDescription.java
```diff
@@ -369,6 +375,55 @@ public boolean hasFlavors(ImmutableSet<Flavor> flavors) {
...
+        cxxDelegateArgs);
+    ImmutableMap<Path, SourcePath> privateHeaders =  CxxDescriptionEnhancer.parseHeaders(
+        params.getBuildTarget(),
+        pathResolver,
+        Optional.of(defaultCxxPlatform),
+        cxxDelegateArgs);
+
+    BuildRuleParams untypedParams = params
+        .withoutFlavor(SwiftLibraryDescription.SWIFT_COMPILE_FLAVOR)
+        .withoutFlavor(SwiftLibraryDescription.SWIFT_COMPILE_FLAVOR);
```
ryu2: `SWIFT_COMPANION_FLAVOR`?

Also consider using a FlavorDomain, then you can call `withoutFlavors(flavorDomain.getFlavors())`

nguyentruongtho:
@zayhero Yes, I think that is the idea. The approach is a bit hacky to me, since you created the rules (for private and public headers) way before they are actually used in `SwiftCompile`. The disconnection will make it hard to understand about why we need to add those rules as dependencies before we know that they will be used for swift compile.

Furthermore, if the target is not a mixed target, then there will be no swift target and rule, and we are adding header targets for nothing.

zayhero:
@ryu2 @nguyentruongtho I am closing this PR since I have #1232 submitted to address the comments (by moving the header logic into `SwiftLibraryDescription`). Please take a look.

PR #1213, opened 2017-03-02
Do not return empy annotation processor params if annotation processor deps are present

kageiit:
Tools like [error-prone](https://github.com/google/error-prone) pickup custom error prone checks from the `-processorpath`. This change makes it so that annotation processor deps are always added to processor path if they exist regardless of the existence of annotation processors

### DIFF
--- a/src/com/facebook/buck/jvm/java/JvmLibraryArg.java
+++ b/src/com/facebook/buck/jvm/java/JvmLibraryArg.java
@@ -56,7 +56,8 @@ public class JvmLibraryArg extends AbstractDescriptionArg {
       ProjectFilesystem filesystem,
       BuildRuleResolver resolver,
       Set<String> safeAnnotationProcessors) {
-    if (annotationProcessors.isEmpty() && plugins.isEmpty()) {
+    if (annotationProcessors.isEmpty() && plugins.isEmpty()
+        && annotationProcessorDeps.isEmpty()) {
       return AnnotationProcessingParams.EMPTY;
     }
 


kageiit:
cc @dreiss @Coneko @davido

jkeljo:
Hey @kageiit, have you seen the new `java_annotation_processor` rule? My intention is to remove the old `annotation_processors`/`annotation_processor_deps` stuff and then document `java_annotation_processor` as the way to work with APs in Buck.

As the implementation of that rule stands, the way to do what you want would be to define a `java_annotation_processor` rule whose `deps` includes both error-prone and any custom checks. Does that work for you?

kageiit:
@jkeljo 

It works if everything is documented and visible. Currently I had no idea anything of that sorts existed. Can you point me at more documentation around this so we can explore the approach you suggested?

kageiit:
@jkeljo taking a cusrosry look at the code, the existing `java_annotation_processor` rule will not work for this use case as it requires the `processorClass` argument as non optional. https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/JavaAnnotationProcessorDescription.java#L81

Error prone checkers do not have an annotation processor class but still need to show up on the processor path

jkeljo:
@kageiit on documentation, I'm letting the design bake a little bit before putting docs up on the site, but I will do so soon.

The `processor_class` would be error prone's processor itself. Or is it a custom compiler or something?

kageiit:
there is no processor class in error prone. Its a java compiler

we just need some entries on the processrpath without adding a processor because the error prone java compiler looks for custom checkers on the processrpath only via the service loader

jkeljo:
Making `processor_class` optional should make that work as long as error-prone is run out of process from Buck. Is that the scenario? I don't think this PR or making `processor_class` optional would be sufficient to make it run in-process, because of the way Buck creates annotation processors for use with an inproc compiler.

kageiit:
error prone is an in process compiler as it is loaded via the javac_jar api

kageiit:
It feels like we need a processorpath plugin/rule where we can add deps that need to end up on the processor path. This way they can be decoupled from annotation processors

kageiit:
also, the current PR works and we have verified it internally in our buck fork where we apply a custom error prone checker on all our java/android modules.

jkeljo:
Hm, and it finds checkers even on modules that don't use another AP?

kageiit:
> Hm, and it finds checkers even on modules that don't use another AP?

Yes. since it uses the service loader api it finds checkers through its own discovery process. annotation processors are just one type of class found by the [service loader](http://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)

jkeljo:
Yeah, I'm just surprised there's a class loader getting created for that at all.

jkeljo:
Ah, I see. It's because the `processorpath` option is still handled by the `FileManager`. OK, that makes sense. So making the `processor_class` optional should also work.

PR #1218, opened 2017-03-04
Coloured dot outputs for queries

rspencer01:
Adds colors to the dot nodes when visualizing graphs. They are colored according to their types, on a random (but consistent basis). We reserve certain colors for the more common targets.  This list could be expanded.

### DIFF
--- a/src/com/facebook/buck/graph/Dot.java
+++ b/src/com/facebook/buck/graph/Dot.java
@@ -70,7 +70,7 @@ public class Dot<T> {
 
         try {
           output.append(String.format(
-              "  %s [style=filled,color=%s]\n",
+              "  %s [style=filled,color=%s];\n",
               source,
               Dot.colorFromType(sourceType)));
         } catch (IOException e) {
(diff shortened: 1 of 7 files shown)


src/com/facebook/buck/graph/Dot.java
```diff
@@ -22,22 +22,40 @@
...
 
   private final DirectedAcyclicGraph<T> graph;
   private final String graphName;
   private final Function<T, String> nodeToName;
+  private final Function<T, String> nodeToTypeName;
   private final Appendable output;
+  private static final Map<String, String> typeColors;
+
+  static {
+    HashMap<String, String> _typeColors = new HashMap<>();
```
illicitonion: Could you make this an ImmutableMap?

```
typeColors = new ImmutableMap.Builder() {{
  put("android_aar", "springgreen2");
  put("android_library", "springgreen3");
  ...
}}.build()

illicitonion:
This also breaks some tests in QueryCommandIntegrationTest - can you fix those up?

PR #1229, opened 2017-03-09
Use filelist for swift output dynamic library and object files

nguyentruongtho:
Pull out from https://github.com/facebook/buck/pull/917

Before this PR:
* Swift object files stay in the link command of any parent target.
* Swift dylib files are in the link command of any parent target, along with object files from the same swift target. This is wrong because object files should only be used for static linking.

After:
* Swift object files will stay in an filelist, along with other object files from cxx rules. If the target is a shared target, only its dylib output file will be used for linking (not its object files)
* Swift dylib files stay in an output filelist of the link command.

This also helps to remove duplication of swift object files in a link command.

**Sample**:
Input: A depends on B (swift)

A link command:
```
...
-Xlinker -add_ast_path
/path/to/B.swiftmodule
/path/to/B.o
...
```

will change to 
```
-Xlinker -add_ast_path
/path/to/B.swiftmodule
...
-filelist
/path/to/B_filelist.txt <--- /path/to/B.o is in here
```

### DIFF
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -275,8 +275,8 @@ class SwiftCompile extends AbstractBuildRule {
     return args.build();
   }
 
-  ImmutableSet<Arg> getAstLinkArgs() {
-    return ImmutableSet.<Arg>builder()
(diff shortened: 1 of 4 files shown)


ryu2:
@facebook-github-bot shipit

PR #1232, opened 2017-03-10
Fix race condition on Swift build

zayhero:
In PR #1164, we introduce the logic to allow SwiftCompile use the header maps to built bridging headers.
However, we encounter race condition issue when we building with this logic.

The headers maps, is generated by CxxLibraryDescription, and SwiftCompile step is generated by SwiftLibraryDescription. These two build rules are running in parallel, so its not guarantee that header maps are ready when we running SwiftCpmile.

To fix this issue, I am making changes to:
1. passing the headers from `AppleLibraryDescription` and `AppleBinaryDescription` to `SwiftLibraryDescription`
2. in `SwiftLibraryDescription`, use the headers to generate symlink tree build rules for headers
3. force `SwiftCompile` build rule depends on the symlink tree build rules

@ryu2 @nguyentruongtho Could you take a look? Thanks!

### DIFF
--- a/src/com/facebook/buck/apple/AppleBinaryDescription.java
+++ b/src/com/facebook/buck/apple/AppleBinaryDescription.java
@@ -20,6 +20,7 @@ import static com.facebook.buck.swift.SwiftLibraryDescription.isSwiftTarget;
 
 import com.facebook.buck.cxx.CxxBinaryDescription;
 import com.facebook.buck.cxx.CxxCompilationDatabase;
+import com.facebook.buck.cxx.CxxLibraryDescription;
 import com.facebook.buck.cxx.CxxPlatform;
 import com.facebook.buck.cxx.CxxStrip;
 import com.facebook.buck.cxx.FrameworkDependencies;
@@ -198,8 +199,16 @@ public class AppleBinaryDescription
       BuildRuleParams params,
       BuildRuleResolver resolver,
       A args) throws NoSuchBuildTargetException {
+    SourcePathResolver pathResolver = new SourcePathResolver(new SourcePathRuleFinder(resolver));
+    CxxLibraryDescription.Arg cxxDelegateArgs = CxxLibraryDescription.createEmptyConstructorArg();
+    AppleDescriptions.populateCxxLibraryDescriptionArg(
+        pathResolver,
+        cxxDelegateArgs,
+        args,
+        params.getBuildTarget());
+
     Optional<BuildRule> swiftCompanionBuildRule = swiftDelegate.createCompanionBuildRule(
-        targetGraph, params, resolver, args);
+        targetGraph, params, resolver, delegate.getDefaultCxxPlatform(), cxxDelegateArgs, args);
     if (swiftCompanionBuildRule.isPresent()) {
       // when creating a swift target, there is no need to proceed with apple binary rules,
       // otherwise, add this swift rule as a dependency.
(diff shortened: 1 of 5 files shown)


src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -221,6 +224,10 @@ public boolean hasFlavors(ImmutableSet<Flavor> flavors) {
               "Could not find SwiftCompile with target %s", buildTarget);
         }
       };
+
+      params = params.appendExtraDeps(
```
nguyentruongtho: Will it be required for mixed `apple_binary` target?

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -353,6 +371,56 @@ private BuildRule createSharedLibraryBuildRule(
     }
   }
 
+  private <A extends CxxLibraryDescription.Arg> void populdateHeadersArg(
```
nguyentruongtho: `populateHeadersArg`

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -371,6 +439,9 @@ public static boolean isSwiftTarget(BuildTarget buildTarget) {
     public Optional<SourcePath> bridgingHeader;
     public ImmutableSortedSet<BuildTarget> deps = ImmutableSortedSet.of();
     public Optional<NativeLinkable.Linkage> preferredLinkage;
+
+    public ImmutableMap<Path, SourcePath> headers = ImmutableMap.of();
```
nguyentruongtho: not sure about the reason we need this, can we just passing the headers around without relying on the args. The args is for different purpose.

src/com/facebook/buck/apple/AppleBinaryDescription.java
```diff
@@ -198,8 +199,16 @@ private boolean shouldWrapIntoAppleDebuggableBinary(
...
+    CxxLibraryDescription.Arg cxxDelegateArgs = CxxLibraryDescription.createEmptyConstructorArg();
+    AppleDescriptions.populateCxxLibraryDescriptionArg(
+        pathResolver,
+        cxxDelegateArgs,
+        args,
+        params.getBuildTarget());
+
     Optional<BuildRule> swiftCompanionBuildRule = swiftDelegate.createCompanionBuildRule(
-        targetGraph, params, resolver, args);
+        targetGraph, params, resolver, delegate.getDefaultCxxPlatform(), cxxDelegateArgs, args);
```
nguyentruongtho: We should not use `defaultCxxPlatform` here. I think you should use an optional of specified `cxxPlatform`, otherwise, it might work for default platform specified in the `.buckconfig` but it won't work for target with platform flavors, for example if you want to build `A#iphoneos-x86_64`, but you have `default_platform = iphonesimulator-x86_64`, you will get problem with this.

Above of this, I would argue that we don't need extra params here, `createCompanionBuildRule` simply for creating `SwiftLibrary` rule (a place holder rule without any compile steps), `SwiftLibrary` doesn't care about `cxxPlatform`. Also, `createCompanionBuildRule` forwards all inputs from `apple_binary` and `apple_library`'s `createBuildRule` method, so as long as apple description can get those cxx information, why can't Swift library description?

I haven't looked much at bridging header support, I need to look through the issue again and get back with suggestion.

styurin:
Is that still valid?

zayhero:
closing this. It's fixed after @milend rewrite the swift logic

PR #1238, opened 2017-03-13
Generate modulemap automatically for symlink tree with headermap

nguyentruongtho:
Split out from https://github.com/facebook/buck/pull/983 in effort of supporting clang module.

We create `buck.modulemap` instead of `module.modulemap` for all public header symlink trees with header map. Those module maps will enable us to import a target as module later on.

The reason for this special module map naming is that, if module is enabled for a project (`-fmodules`), all `module.modulemap` files in that project's search paths will be pre-compiled and imported implicitly (see https://clang.llvm.org/docs/Modules.html, search for `-fimplicit-module-maps`), which leads to nondeterministic compilation results if we don't handle module cache carefully. 
Instead of looking into how to handle module cache, we will name module map that created by buck `buck.modulemap`, so that it will not be used implicitly, and will explicitly specified them via [`-fmodule-map-file`](https://clang.llvm.org/docs/Modules.html) in compile commands when needed.


src/com/facebook/buck/cxx/HeaderSymlinkTreeWithHeaderMap.java
```diff
@@ -72,10 +120,44 @@ public SourcePath getSourcePathToOutput() {
...
     }
-    return ImmutableList.<Step>builder()
+    ImmutableList.Builder<Step> builder = ImmutableList.<Step>builder()
         .addAll(super.getBuildSteps(context, buildableContext))
-        .add(new HeaderMapStep(getProjectFilesystem(), headerMapPath, headerMapEntries.build()))
-        .build();
+        .add(new HeaderMapStep(getProjectFilesystem(), headerMapPath, headerMapEntries.build()));
+
+    if (shouldCreateModule) {
+      Optional<String> umbrellaHeader = getUmbrellaHeader(getBuildTarget().getShortName());
```
nguyentruongtho: Implicitly use `targetName.h` as the umbrella header file of a module if exists (which is the way xcode structures a framework), otherwise we use umbrella directory (http://clang.llvm.org/docs/Modules.html#umbrella-directory-declaration). 

Another way for doing this more explicitly is that we can add an extra argument for `cxx_*`, `apple_*` rules to point to the umbrella header.

src/com/facebook/buck/cxx/HeaderSymlinkTreeWithHeaderMap.java
```diff
@@ -72,10 +120,44 @@ public SourcePath getSourcePathToOutput() {
...
       headerMapEntries.put(key, buckOut.relativize(getRoot().resolve(key)));
     }
-    return ImmutableList.<Step>builder()
+    ImmutableList.Builder<Step> builder = ImmutableList.<Step>builder()
         .addAll(super.getBuildSteps(context, buildableContext))
-        .add(new HeaderMapStep(getProjectFilesystem(), headerMapPath, headerMapEntries.build()))
-        .build();
+        .add(new HeaderMapStep(getProjectFilesystem(), headerMapPath, headerMapEntries.build()));
+
+    if (shouldCreateModule) {
```
nguyentruongtho: Always create module map doesn't hurt, except that if the target name is not in a good format for module name, then we have to normalize it (see below)

ryu2:
@facebook-github-bot shipit

PR #1241, opened 2017-03-14
Fix travis-ci build.

brettwooldridge:
Ignore this pull request, just poking at travis to see if I can make it pass.


Coneko:
Travis is still red after the latest update, I see some timeouts still happening.

brettwooldridge:
@Coneko I think the real solution is to break up some of the Android unit tests into multiple files.

Coneko:
Latest run's failures don't seem to be any of the usual ones, and might in fact be a legit failure that was never noticed because Travis was always red anyway.

Importing.

brettwooldridge:
@Coneko Woohooo! Only one failure ``TESTS FAILED: 1 FAILURE``, in the ``go`` tests.

PR #1245, opened 2017-03-14
Don't set enable-testing by default

rmaz:
This should be configurable in the `buckconfig`, otherwise there is no way to disable it for release builds.


ryu2:
```//test/com/facebook/buck/swift:integration - testAppleTestToWorkWithSwift (com.facebook.buck.swift.SwiftTestIOSIntegrationTest)```

seems to be failing; you'll probably need to re-add `enable-testing` to the test fixture's `.buckconfig`.

rmaz:
Whoops, thought I had run that one. Updated.

ryu2:
@facebook-github-bot shipit

PR #1239, opened 2017-03-13
Update showcase.soy

darkforestzero:
trying to add Brainium Studios

PR #1248, opened 2017-03-15
Fix broken showcase page

sdwilsh:
https://github.com/facebook/buck/pull/1239 was a little broken:
1) Image links weren't right, so they didn't render on the page
2) The code used the old `moreInfo` instead of `links` in the map,
   causing the links to not render

This change fixes that.  It also removes "Word Search Star" because
there is no HTTPS image available from the iTunes store, and it is
not published in the Play store.

Test Plan:
`.\docs\soyweb-local.ps1` and then loaded
http://localhost:9811/about/showcase.html

### DIFF
--- a/docs/about/showcase.soy
+++ b/docs/about/showcase.soy
@@ -108,8 +108,8 @@
   {param items}
     {call .item}
       {param name: 'Sudoku' /}
-        {param imgSrc: 'https://http://brainiumstudios.com/icons/sudoku.png' /}
-        {param moreInfo: quoteKeysIfJs([
+        {param imgSrc: 'https://lh6.ggpht.com/m92klrf80HKcoOW10p565ooRzYsP7ktAgDLGT6QnK2KrfMxgcxsfgqJ1LwE08J8z7_br=w100' /}
+        {param links: quoteKeysIfJs([
             'iOS/Android':
               'http://www.brainiumstudios.com/site/sudoku.html',
               ]) /}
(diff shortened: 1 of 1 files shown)


sdwilsh:
cc @darkforestzero as an FYI

darkforestzero:
I have a fixed version with correct links to our icons, there was just an extraneous https// before each link.  it's really hard to iterate on this without being able to see the end result.  i committed my fix to patch 5, but I guess that's closed now.  I'd prefer if we use the links on our server which are higher resolution and more up to date.  Also, would be nice to include Wordsearch since it's built with buck.  

I've attached a "fixed' version of the showcase.soy .  Sorry this is dragging on so long - again it's hard to iterate through a black box :)

[showcase.soy.txt](https://github.com/facebook/buck/files/845418/showcase.soy.txt)

sdwilsh:
@darkforestzero, the problem with linking on your server is that you have no SSL configured, and Buck's documentation pages all use SLL, so end users will get a mixed-content warning if we use images elsewhere.

As for resolution; we can use a higher resolution image pretty trivially by changing the URL parameters (I believe; I haven't actually checked the DPI), but we're trying to keep image size down as well so this page doesn't take an eternity to load.


As for local testing, see the test plan in this change (there's a unix script named very similarly in the same folder).

darkforestzero:
@sdwilsh thanks for the soyweb-local info.  dang, it looks so good!  too bad about ssl
screen shot 2017-03-15 at 3 15 54 pm

sdwilsh:
I'm going to land this as-is, @darkforestzero, so we can have a non-broken page.  Why don't you put up a PR with your suggested changes once this lands and we'll figure out the best path forward :)

PR #1254, opened 2017-03-17
Add delegate for getSystemService in ApplicationLike.

runningcode:
We have an application that delegates the `getSystemService` to a another service lookup similar to the pattern found here: https://github.com/square/mortar/blob/master/mortar-hellodagger2/src/main/java/com/example/hellodagger2/HelloDagger2Application.java

In order to use Exopackage, we need to be able to delegate our Application class to our service lookup.

### DIFF
--- a/android/com/facebook/buck/android/support/exopackage/DefaultApplicationLike.java
+++ b/android/com/facebook/buck/android/support/exopackage/DefaultApplicationLike.java
@@ -42,4 +42,9 @@ public class DefaultApplicationLike implements ApplicationLike {
 
   @Override
   public void onConfigurationChanged(Configuration newConfig) {}
+
+  @Override
+  public Object getSystemService(String name) {
+    return null;
+  }
 }
--- a/android/com/facebook/buck/android/support/exopackage/ExopackageApplication.java
+++ b/android/com/facebook/buck/android/support/exopackage/ExopackageApplication.java
@@ -162,4 +162,15 @@ public abstract class ExopackageApplication<T extends ApplicationLike> extends A
       delegate.onConfigurationChanged(newConfig);
     }
   }
+
+  @Override
+  public Object getSystemService(String name) {
+    if (delegate != null) {
+      Object service = delegate.getSystemService(name);
+      if (service != null) {
+        return service;
+      }
+    }
+    return super.getSystemService(name);
+  }
 }
--- a/android/com/facebook/buck/android/support/exopackage/ApplicationLike.java
+++ b/android/com/facebook/buck/android/support/exopackage/ApplicationLike.java
@@ -53,4 +53,9 @@ public interface ApplicationLike {
    */
   void onConfigurationChanged(Configuration newConfig);
 
+  /**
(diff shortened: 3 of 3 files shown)


Coneko:
Can you explain what this change is for?

runningcode:
Description updated!

Coneko:
cc @dreiss, @cjhopman

android/com/facebook/buck/android/support/exopackage/DefaultApplicationLike.java
```diff
@@ -42,4 +42,9 @@ public void onTerminate() {}
 
   @Override
   public void onConfigurationChanged(Configuration newConfig) {}
+
+  @Override
+  public Object getSystemService(String name) {
```
rjrjr: Should mirror the `@ServiceName @NonNull` annotations found in `Context.java`.

rjrjr: Actually, the most important place to do that is on the interface.

runningcode: This module has no dependencies on those annotations. I'm not sure we want to add them, I'll leave that up to the buck maintainers to decide.

dreiss:
How common is this pattern?  If it's something particular to your app, I'd rather not push this all the way into ExopackageApplication.

I think the alternative would be for you to define a `HasSystemService` interface that goes into the primary dex and is implemented by your delegate, then you can move the `getSystemService` definition that you currently have in `ExopackageApplication` into your `AppShell`

kageiit:
> I think the alternative would be for you to define a HasSystemService interface that goes into the primary dex and is implemented by your delegate

We actually do a similar thing to provide the top level dagger component/delegate via an interface thats available in the main dex.

dreiss:
Would this change let you simplify your code?  If so, I'm okay with importing it.

android/com/facebook/buck/android/support/exopackage/ExopackageApplication.java
```diff
@@ -162,4 +162,15 @@ public void onConfigurationChanged(Configuration newConfig) {
       delegate.onConfigurationChanged(newConfig);
     }
   }
+
+  @Override
+  public Object getSystemService(String name) {
+    if (delegate != null) {
+      Object service = delegate.getSystemService(name);
```
kageiit: do you want to check if name is not null here? given that we may not be adding the annotation?

runningcode: The name is `@NonNull` in `Context.getSystemService()`. I @rjrjr was referring to the return value for this method.

kageiit: ah ok :)

runningcode:
IMO, I think its a common enough pattern to warrant the change. For people using this pattern, having this in the project also lowers the barrier of entry for exopackage.
If not, I think @dreiss's alternative should be in the readme/documentation.

PR #1263, opened 2017-03-25
Add missing docs for prebuilt_apple_framework rule

aledalgrande:
Docs were missing but I learned by searching around the internets the construct was available. Added minimal docs.

### DIFF
--- a/docs/rule/prebuilt_apple_framework.soy
+++ b/docs/rule/prebuilt_apple_framework.soy
@@ -1,0 +1,71 @@
+{namespace prebuilt_apple_framework}
+
+/***/
+{template .soyweb}
+  {call buck.page}
+    {param title: 'prebuilt_apple_framework()' /}
+    {param navid: 'rule_prebuilt_apple_framework' /}
+    {param prettify: true /}
+    {param description}
+      A prebuilt_apple_framework() rule represents an Apple framework that has
+      already been compiled and exports a set of headers.
+    {/param}
+    {param content}
+      {call buck.rule}
+        {param status: 'FROZEN' /}
+        {param overview}
+          A <code>prebuilt_apple_framework()</code> rule represents a set of
+          Objective-C/C++ source files and is very similar to a <a href="{ROOT}rule/prebuilt_cxx_library.html">
+            <code>prebuilt_cxx_library()</code>
+          </a> rule.
+        {/param}
+
+        {param args}
+          {call buck.arg}
+            {param name: 'name' /}
+            {param desc}
+              The name of the rule.
+            {/param}
+          {/call}
+        {/param}
+
+        {param args}
+          {call buck.arg}
+            {param name: 'framework' /}
+            {param desc}
+              The path to the precompiled framework.
+            {/param}
+          {/call}
+        {/param}
+
+        {param args}
+          {call buck.arg}
+            {param name: 'preferred_linkage' /}
+            {param desc}
+              How to link to a binary: use <code>dynamic</code> for a dynamic
+              framework, and <code>static</code> for old universal static
+              frameworks manually lipo-ed together. <code>dynamic</code> will
+              copy the frameworks into the <code>Frameworks</code> directory
+              of an Apple bundle, and configure framework search paths and linker flags.
+              <code>static</code> will copy the resources of the framework into
+              an Apple bundle.
+            {/param}
+          {/call}
+        {/param}
+
+        {param examples}
+          {literal}<pre class="prettyprint lang-py">
+prebuilt_apple_framework(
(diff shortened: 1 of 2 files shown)


ryu2:
cc: @robbertvanginkel @nguyentruongtho

Coneko:
cc @yiding

docs/rule/prebuilt_apple_framework.soy
```diff
@@ -0,0 +1,38 @@
...
+  {call buck.page}
+    {param title: 'prebuilt_apple_framework()' /}
+    {param navid: 'rule_prebuilt_apple_framework' /}
+    {param prettify: true /}
+    {param description}
+      A prebuilt_apple_framework() rule represents an Apple bundle that has
+      already been compiled and exports a set of headers.
+    {/param}
+    {param content}
+      An <code>apple_library()</code> rule represents a set of Objective-C/C++
```
Coneko: You probably meant `prebuilt_apple_framework` here.

docs/rule/prebuilt_apple_framework.soy
```diff
@@ -0,0 +1,38 @@
...
+    {param navid: 'rule_prebuilt_apple_framework' /}
+    {param prettify: true /}
+    {param description}
+      A prebuilt_apple_framework() rule represents an Apple bundle that has
+      already been compiled and exports a set of headers.
+    {/param}
+    {param content}
+      An <code>apple_library()</code> rule represents a set of Objective-C/C++ 
+      source files and is very similar to a 
+      <a href="{ROOT}rule/cxx_library.html"><code>prebuilt_cxx_library()</code></a>
```
Coneko: The link should point to `prebuilt_cxx_library`.

docs/rule/prebuilt_apple_framework.soy
```diff
@@ -0,0 +1,38 @@
+{namespace prebuilt_apple_framework}
+
+/***/
+{template .soyweb}
+  {call buck.page}
+    {param title: 'prebuilt_apple_framework()' /}
+    {param navid: 'rule_prebuilt_apple_framework' /}
+    {param prettify: true /}
+    {param description}
+      A prebuilt_apple_framework() rule represents an Apple bundle that has
```
Coneko: An Apple framework, not a bundle.

aledalgrande:
Thanks for the review @Coneko.

yiding:
I think the 'preferred_linkage' flag is required, where "shared" means the usual framework, while "static" is the not-quite-a-framework-that's-packaged-like-a-framework.

aledalgrande:
I successfully used a "not-quite-a-framework-that's-packaged-like-a-framework" without the `preferred_linkage` flag, so don't think it's required. Is that what you meant?

robbertvanginkel:
I thought I added and made `preffered_linkage` required in #1064, maybe you are using an older buck binary?

The main reason this wasn't documented is because this rule type has no project generation support yet, and the copying behaviour isn't correct. For dynamic frameworks these should end up in an apple_bundle, for static ones their resources should. I believe both are not implemented, so might be good to note that in the docs.

aledalgrande:
Yeah, I have version `2016.11.11.01` (the one from Homebrew) and your PR was merged in after that release. Maybe time for a new one?

I will add a note about project generation support.

LegNeato:
@aledalgrande filed https://github.com/facebook/buck/issues/1268

aiked:
I updated the homebrew version to v2017.03.29.01.

yiding:
I did add some project generation support for the case of real frameworks, where they will be copied into Frameworks/ directory of the bundle, and set various search paths and linker flags, like what you'd get with Xcode "embedded frameworks" feature.

aledalgrande:
@yiding as soon as I can test the new release, I will add the project generation part.

aledalgrande:
The docs should now be complete, let me know if they make sense. Also, not sure why the specs are not passing.

docs/rule/prebuilt_apple_framework.soy
```diff
@@ -0,0 +1,71 @@
...
+    {param description}
+      A prebuilt_apple_framework() rule represents an Apple framework that has
+      already been compiled and exports a set of headers.
+    {/param}
+    {param content}
+      {call buck.rule}
+        {param status: 'FROZEN' /}
+        {param overview}
+          A <code>prebuilt_apple_framework()</code> rule represents a set of
+          Objective-C/C++ source files and is very similar to a <a href="{ROOT}rule/prebuilt_cxx_library.html">
```
sdwilsh: You should actually have used `{call buck.prebuilt_cxx_library /}` here.

PR #1270, opened 2017-03-29
Xcode Project Generation: framework headers related work

robbertvanginkel:
As discussed on the call today, some commits from our internal fork.


Coneko:
Merging directly without importing since it's not on master.

PR #1271, opened 2017-03-29
Make aapt2 configurable

kageiit:
This lets one customize the path to aapt2 similar to aapt which exists today


kageiit:
cc @dreiss

PR #1282, opened 2017-04-07
Let robolectric_manifest be specified as output of another rule

kageiit:
This lets the `robolectric_manifest` be set to the output of another rule like an `export_file` etc. similar to how `android_manifest` allows in the `skeleton` parameter


kageiit:
cc @dreiss

Issue #1037, opened 2016-12-05
Allow mixed java and kotlin sources in kotlin_library() target.


dsyang:
see comments in #948 

This would have to be a two step process. Run `kotlinc` on the kotlin sources first, then pass the result into `javac` and compile the java sources. https://discuss.kotlinlang.org/t/compiling-mixed-java-and-kotlin-files-on-the-command-line/1553

kageiit:
This is kinda related to how kapt works as well. kapt in the gradle plugin generates an abi jar of all the kotlin sources so javac annotation processing can run with that on the classpath.

It then runes kotlinc with the generated sources on the classpath so that kotlin code can actually refer to generated code from the javac annotation processor step.

brettwooldridge:
@dsyang @kageiit I have coincidentally started working on in-VM compilation of kotlin rather than shelling out, based on code hijacked from the [kotlin ant task](https://github.com/brettwooldridge/kotlin/blob/master/ant/src/org/jetbrains/kotlin/ant/KotlinCompilerAdapter.kt).  I can sweep up this issue at the same time.

dsyang:
That's awesome @brettwooldridge!

brettwooldridge:
@dsyang As I understand it, this is basically a sequential compilation currently, correct?  Meaning Kotlin and then Java.  So the following, in a single project would not be supported?
```
A.kt extends B.java
B.java references C.kt
```

brettwooldridge:
@dsyang I've got big plans for the coming weeks. 👍  Probably before this feature I will tackle ``autodeps`` support, which entails prying into ``kotlin-compiler.jar`` to get at the AST parser.

dsyang:
It is sequential but I believe `kotlinc` will load and Java classes
referenced by a kotlin file when compiling it (see the link above).

Your example should work as long as kotlinc can find B.java

dsyang:
@brettwooldridge can't wait!

kageiit:
@brettwooldridge the use case you mentioned should be supported. Kotlin adds all java sources to the class path when compiling and then the kotlin classes are added to the javac class path. So cross cutting dependencies between java and kotlin source should be fully supported. This is already possible in gradle and you can verify the same. It is still sequential.

With the availability of kapt3 in kotlin compiler, there are additional steps before where kotlin generates the java stubs from kotlin classes required for javac annotation processing to work.

kageiit:
[...]

Issue #956, opened 2016-10-17
Support for kapt (for improved Kotlin support)


MichaelEvans:
It was mentioned in #857 that `kapt` isn't supported. Opening this new issue to track support for it.

kageiit:
cc @dsyang

kageiit:
It seems like kapt3 in 1.0.6-eap77 ships with an `APT_ONLY` mode
https://github.com/JetBrains/kotlin/blob/8afac55a908c80feb7dd7bf7105e0557d34cacce/plugins/kapt3/src/org/jetbrains/kotlin/kapt3/Kapt3Plugin.kt#L88

This means any build system can call kotlinc with this option and pass in kotlin source files to generate the java stubs required for the javac annoation processor round and compilation

hzsweers:
Is there anything actually blocking this or does it just need someone to work on it now?

runningcode:
I don't think there's anything blocking it.
I filed this with the jetbrains team: https://youtrack.jetbrains.com/issue/KT-18072
Otherwise figuring out how to do this will require examining the source code of the gradle kapt plugin and seeing how it works.

I have some preliminary support for annotation processors in #1358. For example a kotlin class with an `@Inject` constructor seems to work fine when injected in to a java class. I haven't explored which parts of annotation processors aren't working.

runningcode:
I'll probably take a stab at this soon. I've been taking a look at how this is done in gradle land. This is how it is done in kotlin 1.1.3.

There's 4 steps.
1. kaptGenerateStubsKotlin: 
run kotlinc with `plugin:org.jetbrains.kotlin.kapt3:aptMode=stubs`
2. kaptKotlin
run kotlinc with `plugin:org.jetbrains.kotlin.kapt3:aptMode=apt`
3. compileKotlin
run kotlinc regularly
4. compileJava
run javac with `-proc:none` and pass the generated sources from step 2.

These steps are slightly different with each minor version of kotlin so this will be interesting. It seems there are 3 steps in kotlin 1.1.2:
1. kaptKotlin
run kotlinc with `plugin:org.jetbrains.kotlin.kapt3:aptOnly=true`
2. compileKotlin
run kotlinc regularly
3. compileJava
run javac with `-proc:none` and pass the generated sources from step 1.

cubuspl42:
For future reference: https://stackoverflow.com/questions/45217399/how-to-use-kapt-from-command-line-with-kotlinc

elihart:
[...]

Issue #395, opened 2015-08-13
Slow build-steps: ~2-5s per module on cache hit


bkase:
Thanks again, @dreiss for helping with my last issue.

The build is fast now, but there seems to a constant ~18sec overhead whenever `buck build app` is executed.

I attached traces, but it seems like there isn't much information on cache hit. Let me know if there is some way to produce more verbose profiling information.

[clean build trace](https://gist.github.com/bkase/7e60f2486acc8e39dbf3)
[no-op trace](https://gist.github.com/96dcc527e33bc745cecd)

The gen-rule to hack in kotlin takes almost no time, but the cache hit for the module seems way too long.

I don't quite understand how the `[cache]` field in .buckconfig is supposed to affect performance, but changing `mode` to `dir` under `[cache]` doesn't seem to noticeably affect build times. 

Our build time has already halved, but it would be really nice to shave off another 20s.

Thanks!

bkase:
I don't know why, but my cache-hit-check-time went down to `~100-200ms`!

The only thing I changed was this:

(in a file that declares a prebuilt jar for the kotlin runtime)
```
 genrule(
        name = 'kotlin-runtime-jar',
        out = 'kotlin-runtime.jar',
-       cmd = 'ln -s ~/kotlin2/dist/kotlinc/lib/kotlin-runtime.jar $OUT',
+       cmd = 'ln -s ' + os.environ["KOTLINPATH"] + '/dist/kotlinc/lib/kotlin-runtime.jar $OUT',
 )
```

and

(in a script that pretends to be `javac`)
```
-    exit(call(["/Users/highlight/kotlin2/dist/kotlinc/bin/kotlinc"] + params))
+    exit(call([os.environ["KOTLINPATH"] + "/dist/kotlinc/bin/kotlinc"] + params))
```

It's slow immediately before this commit, fast immediately after.

I'll leave this issue open (I'm happy to tweak my build settings to help you guys figure out why this was happening), but feel free to close it. Builds are fast.

k21:
From the traces it looks like almost all of the time in the no-op build is spent computing the rule keys, which usually means hashing the source files. If you are using `buckd`, then hashes of files in the repo should be cached, so the reason of the slow build is either that you are not using `buckd` or that the rules refer to files outside of the repository.

bkase:
I am using `buckd` (at least "Using buckd." is printed every time buck runs).

[...]

PR #1283, opened 2017-04-09
Kotlin in-memory compiler support.

brettwooldridge:
Changes to support #1037, #956, and #395.


src/com/facebook/buck/android/KotlinAndroidLibraryCompiler.java
```diff
@@ -43,8 +43,18 @@ public CompileToJarStepFactory compileToJar(
       JavacOptions javacOptions,
       BuildRuleResolver resolver) {
     return new KotlincToJarStepFactory(
-        kotlinBuckConfig.getKotlinCompiler().get(),
+        kotlinBuckConfig.getKotlinc(),
         ImmutableList.of(),
         ANDROID_CLASSPATH_FROM_CONTEXT);
   }
+
+//  @Override
```
kageiit: Can we clean up these comments?

buck.iml
```diff
@@ -92,5 +92,8 @@
     <orderEntry type="library" name="jna-4.2.0" level="project" />
     <orderEntry type="library" name="jna-platform-4.2.0" level="project" />
     <orderEntry type="library" name="rocksdbjni-5.1.2" level="project" />
+    <orderEntry type="library" name="kotlin-compiler-1.1" level="project" />
```
kageiit: Considering kotlin versions evolve rapidly, it may not make sense to bundle kotlin jars with buck for every release. Maybe its best if we let consumers point to the kotlin jars they want as `remote_file`s or jars etc.?

src/com/facebook/buck/jvm/kotlin/BUCK
```diff
@@ -1,9 +1,22 @@
+java_immutables_library(
+    name = 'support',
+    srcs = [
+        'AbstractKotlincVersion.java',
+    ],
+    immutable_types = [
+        'KotlincVersion',
+    ],
+    autodeps = True,
```
kageiit: I think autodeps was mostly removed by @Coneko recently

src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibraryBuilder.java
```diff
@@ -53,9 +55,42 @@ protected BuilderHelper newHelper() {
...
+              mavenCoords,
+              tests,
+              classesToRemoveFromJar);
+
+      return defaultKotlinLibrary;
+    }
+
+    @Override
     protected CompileToJarStepFactory buildCompileStepFactory() {
+//      new KotlincToJarStepFactory(
```
kageiit: cleanup comments

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
```diff
@@ -0,0 +1,237 @@
...
+          k -> loadCompilerShim(context));
+
+      Method compile = compilerShim
+          .getClass()
+          .getMethod("exec", new Class[] {PrintStream.class, String[].class});
+
+      Class<?> exitCodeClass = compilerShim
+          .getClass()
+          .getClassLoader()
+          .loadClass("org.jetbrains.kotlin.cli.common.ExitCode");
```
kageiit: move this to a constant

kageiit: We can also cache the class obtained this way to avoid needing to load the class for every compile invocation

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
```diff
@@ -0,0 +1,237 @@
...
+
+      ClassLoader classLoader = classLoaderCache.getClassLoaderForClassPath(
+          this.getClass().getClassLoader(),
+          ImmutableList.copyOf(
+          compilerClassPath
+            .stream()
+            .map(PATH_TO_URL)
+            .collect(Collectors.toList())));
+
+      return classLoader.loadClass("org.jetbrains.kotlin.cli.jvm.K2JVMCompiler").newInstance();
```
kageiit: constat

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
```diff
@@ -129,9 +129,25 @@ public Arg createUnpopulatedConstructorArg() {
...
   }
 
-
   @SuppressFieldNotInitialized
   public static class Arg extends JavaLibraryDescription.Arg {
-    public ImmutableList<String> extraKotlincArguments = ImmutableList.of();
+    ImmutableList<String> extraKotlincArguments = ImmutableList.of();
   }
+
+//  private ImmutableSortedSet<BuildRule> buildCompileTimeClasspathDeps(
```
kageiit: cleanup comments

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
```diff
@@ -90,7 +85,15 @@ public Arg createUnpopulatedConstructorArg() {
       return defaultJavaLibraryBuilder.buildAbi();
     }
 
-    JavaLibrary testsLibrary = resolver.addToIndex(defaultJavaLibraryBuilder.build());
+    DefaultKotlinLibrary testsLibrary = (DefaultKotlinLibrary) resolver.addToIndex(defaultJavaLibraryBuilder.build());
+
+//    Kotlinc kotlinc = kotlinBuckConfig.getKotlinc();
```
kageiit: cleanup comments

src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibrary.java
```diff
@@ -0,0 +1,84 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableSortedSet;
+
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class DefaultKotlinLibrary extends DefaultJavaLibrary {
```
runningcode: Is there a difference between this class and DefaultJavaLibrary? I'm new to this codebase and this seems like a common pattern here. I'm just thinking we should document things like this. :)

src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
```diff
@@ -0,0 +1,174 @@
...
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableCollection;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSortedSet;
+
+public class ExternalKotlinc implements Kotlinc {
+
+  private static final KotlincVersion DEFAULT_VERSION = KotlincVersion.of("unknown version");
```
runningcode: is this the default version on unknown version?
can we document the behavior around versions somewhere? how to set the version, what the default is, what happens if none is set?

brettwooldridge:
@kageiit I should have been clear.  This contains changes *toward* support of kapt, i.e. it lays the foundation for completion of #956.  We now invoke the ``K2JVMCompiler.exec()`` method, which winds up [here](https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/K2JVMCompiler.kt#L70) in the CLICompiler.  It is a short distance to go to thread the correct configuration through from buck.

brettwooldridge:
@kageiit Just to be clear, this pull request was not made because the code is ready to merge, it was made due to @dsyang request:
> Would you be comfortable putting your changes in kotlin-compiler up in a PR so we could start reviewing it?
>
> I know there's still some work that you'd like to do but this could make it easier to merge later down the line.

buck.iml
brettwooldridge: @kageiit This can be removed, because everything is done via reflection now.

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
brettwooldridge: Java reflection already performs caching, a subsequent call to ``loadClass()`` is basically a hash lookup.

kageiit:
> this pull request was not made because the code is ready to merge

Makes sense. I left a few comments in places but I guess there still needs to be clean up before merging. Looking forward to this being part of buck! :)

dsyang:
Thanks for putting this up @brettwooldridge! Main reason I suggested it is that it's one step closer to merging (and merging means no more rebases from hell).

kageiit:
@dsyang @brettwooldridge can we try to merge this or bits of it maybe? It makes it lot easier to do incremental changes on top then

brettwooldridge:
@kageiit @dsyang I keep rebasing it once a week or so.  I've got a tiny bit of cleanup now that I realized that it was not possible for an entry in the ``.buckconfig`` to reference a ``prebuilt_jar`` rule.  Give me 'til Monday or so before doing any hasty merging.

The good news is that our development team has been using this branch as our primary *buck* and compiling lots of Kotlin code for well over a month.

test/com/facebook/buck/jvm/kotlin/KotlinTestAssumptions.java
```diff
@@ -32,7 +32,7 @@ public static void assumeUnixLike() {
   public static void assumeCompilerAvailable() throws IOException {
     Throwable exception = null;
     try {
-      new KotlinBuckConfig(FakeBuckConfig.builder().build()).getKotlinCompiler();
+      new KotlinBuckConfig(FakeBuckConfig.builder().build()).getPathToCompilerJar(ruleFinder);
```
dsyang: ruleFinder doesn't exist.

dsyang:
Google just announced kotlin as a official language for Android. @brettwooldridge any chance we can get this PR updated and ready soon? ;)

src/com/facebook/buck/android/AndroidLibraryCompilerFactory.java
```diff
@@ -26,5 +26,6 @@
  */
 public interface AndroidLibraryCompilerFactory {
 
-  AndroidLibraryCompiler getCompiler(JvmLanguage language);
+  AndroidLibraryCompiler getCompiler(
+      JvmLanguage language);
 }
```
dsyang: unrelated change

src/com/facebook/buck/android/DefaultAndroidLibraryCompilerFactory.java
```diff
@@ -34,7 +34,8 @@ public DefaultAndroidLibraryCompilerFactory(
   }
 
   @Override
-  public AndroidLibraryCompiler getCompiler(AndroidLibraryDescription.JvmLanguage language) {
+  public AndroidLibraryCompiler getCompiler(
+      AndroidLibraryDescription.JvmLanguage language) {
```
dsyang: unrelated change. please revert.

src/com/facebook/buck/android/KotlinAndroidLibraryCompiler.java
```diff
@@ -27,7 +27,8 @@
 
   private final KotlinBuckConfig kotlinBuckConfig;
 
-  public KotlinAndroidLibraryCompiler(KotlinBuckConfig kotlinBuckConfig) {
+  public KotlinAndroidLibraryCompiler(
+      KotlinBuckConfig kotlinBuckConfig) {
```
dsyang: unrelated change

src/com/facebook/buck/jvm/kotlin/AbstractKotlincVersion.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package com.facebook.buck.jvm.kotlin;
+
+import org.immutables.value.Value;
+
+import com.facebook.buck.util.immutables.BuckStyleImmutable;
+
+@Value.Immutable
+@BuckStyleImmutable
+public abstract class AbstractKotlincVersion {
```
dsyang: javadoc

src/com/facebook/buck/jvm/kotlin/DefaultKotlinLibraryBuilder.java
```diff
@@ -52,9 +61,49 @@ protected BuilderHelper newHelper() {
...
+            buildRuleResolver.requireRule(btsp.getTarget());
+          } catch (NoSuchBuildTargetException e) {
+            throw new HumanReadableException("Rule for target '%s' could not be resolved.", btsp.getTarget());
+          }
+        }
+        System.err.println(ruleOptional);
+      }
       return new KotlincToJarStepFactory(
-          Preconditions.checkNotNull(kotlinBuckConfig).getKotlinCompiler().get(),
+          Preconditions.checkNotNull(kotlinBuckConfig).getKotlinc(),
```
dsyang: can you use the above `kotlinc` here?

Coneko:
You can run https://github.com/google/google-java-format on the files to format them automatically and remove spurious unrelated changes.

brettwooldridge:
I'll take a look.  I haven't gotten around to minimizing the diffs. Just working on the unit tests.  Once they're cleaned up and running I'll minimize the changes and the pull should be ready.

dsyang:
@brettwooldridge sounds good! Anything I can do to help? Happy to patch this PR and help test & fix unit tests.

brettwooldridge:
@dsyang I've rebased my changed on to master, but I'm getting a difficult to trace exception from the Kotlin compiler.  Working forward in steps from my April 27 commit, if I rebased onto a May 1 commit then everthing is still working, but if I rebase the same changed onto a May 10 commit I get the same failure.

It's 3:30am here, so I'm going to knock off for tonight.  Tomorrow I'll see if I can use git bisect to narrow down the issue, or do a manual bisect.  Seems classloader related, but I wonly't know until I actually find the breaking commit.  If I can resolve the issue on that commit, I suspect the rebase from there to the head of master will go smoothly.

dsyang:
Gotcha. I'll try and narrow it down in https://github.com/dsyang/buck/tree/kotlin-compiler2

Is the weird bug you're seeing this? https://gist.github.com/dsyang/9d5b7562ad676a10b18adc6d427cffda that's what I see when trying to use this branch.  

Thanks for all your hard work! Hope to have some good news for you tomorrow morning.

dsyang:
Not sure why I can't submit PRs to @brettwooldridge's buck repo but here's what I found:
 - there was a [trivial compilation error](https://github.com/dsyang/buck/commit/1c573c55d5fab90a118aac17d0271c85cfecb497) that prevented the kotlin tests from running. After fixing that tests would run but I would get a weird kotlin compiler exception ([stacktrace](https://gist.github.com/dsyang/9d5b7562ad676a10b18adc6d427cffda))
 - I traced this error to be related to the v1.1.x kotlin compilers.  If I tell the tests to use v1.0.7 by adding a .buckconfig like below, then all the tests pass:
```
# filename: test/com/facebook/buck/jvm/kotlin/testdata/kotlin_(library|test)_description/.buckconfig
[kotlin]
  compiler = "/Users/dsyang/Downloads/kotlinc-1.0.7/bin/kotlinc"
```
 - If I change the test's`.buckconfig` to use the compiler_jar, I run into another crash ([stacktrace](https://gist.github.com/dsyang/c966a74e5b47b1088c91342ec018b6fc)).  @brettwooldridge  was this the crash you were talking about?

The last crash is more tractable than the first one because it seems to be in the buck codebase.  I'll try and tackle that later today.

brettwooldridge:
@dsyang Yes, that is the error.

I have been running my ``kotlin`` branch (literally that is the name) with Kotlin v1.1.1 for some time without that error.  That is the branch from April 27.  When I rebase/merge that into master at May 10th, it fails with that error.  If I merge a bit further back, May 1, it still succeeds.  I feel like it is classloader related (the JarBackedKotlinCompiler uses a distinct classloader).


I have my rebased changes and cleanup on a branch called ``kotlin-merge`` -- failing with the same error.

brettwooldridge:
@dsyang Here's a thought.  There are significant changes in ``kotlin-merge``.  In Kotlin v1.1.1 it is no longer sufficient to specify a compiler jar and runtime jar, the kotlin compiler does it's own jar discovery based off of the location of the compiler jar, and pulls in *other* Kotlin jars.

I am wondering if the ``mapdb`` thirdparty component is somehow pulling in the thirdparty/java/kotlin jars -- which may be incompatible with the compiler.  Those are no longer needed or used by the Kotlin integration.

The Kotlin compiler integration required a true installation of Kotlin -- at least it needs all of the jars to be present.  The only .buckconfig variables are now ``kotlin_home`` and ``external`` (true for kotlinc shell out, false for in-process compilation).

dsyang:
Looking at `kotlin-merge` now. I think third-party/java/kotlin is unnecessary for `mapdb` to work.

Agreed that it looks like a true installation of kotlin is necessary.  I think the .buckconfig variables at `runtime_jar`, `compiler_jar`, and `compiler`, `KotlinBuckConfig` only checks `kotlin_home` as a environment variable.  Maybe we should replace `runtime_jar` and `compiler_jar` with `kotlin_home`.

brettwooldridge:
The KotlinBuckConfig in the kotlin-merge branch only looks for *kotlin_home* and *external*.  I'm pretty sure that the mapdb dependency is the cause.

dsyang:
Do you have a theory about why it would be causing this to fail?

brettwooldridge:
I'm away from my computer now, but maybe it is shading some Kotlin classes?  Maybe I shouldn't speculate... I'll have a chance to look in a few hours.

brettwooldridge:
@dsyang No clear explanation yet, but removing ``thirdparty/java/kotlin`` entirely, and deleting the dependency in ``thirdparty/java/mapdb/BUCK`` fixed the issue.  ``kotlin-merge`` now runs Kotlin builds against our production code without issue (unit tests may or may not all work, haven't checked them yet).

One possibility is, because there are two classloaders (at least), the main Java app classloader, and the one created by ``JarBackedReflectedKotlinc``, that somehow Kotlin stdlib classes were getting loaded in two separate classloaders.  If that proves to be the cause, then in *the future*, we might need to create a custom classloader for the kotlin compile that is non-delegating -- or at least *delegate last* rather than *delegate first*.

brettwooldridge:
I squashed and pushed ``kotlin-compiler2``.

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
```diff
@@ -0,0 +1,230 @@
...
+        .setReflectively("kotlinc.classpath", compilerClassPath.toString());
+  }
+
+  private Object loadCompilerShim(ExecutionContext context) {
+    try {
+      ClassLoaderCache classLoaderCache = context.getClassLoaderCache();
+      classLoaderCache.addRef();
+
+      ClassLoader classLoader = classLoaderCache.getClassLoaderForClassPath(
+          this.getClass().getClassLoader(),
```
dsyang: Found the culprit!

We're running the kotlin compiler with a classpath that includes all the classes loaded for buck.  That's not what we want. The only classes the compiler should know about / have access to are the ones specified in the `compilerClassPath`.

Changing this to `null` allows for successful compilation without needing to change `mapdb` or anything in `third-party`.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -58,67 +63,105 @@ public KotlinBuckConfig(BuckConfig delegate) {
...
+   * Find the Kotlin home (installation) directory by searching in this order: <br>
+   *  <ul>
+   *    <li>If the "kotlin_home" directory is specified in .buckconfig then use it.</li>
+   *    <li>Check the environment for a KOTLIN_HOME variable, if defined then use it.</li>
+   *    <li>Resolve "kotlinc" with an ExecutableFinder, and if found then deduce the
+   *        kotlin home directory from it.</li>
+   *  </ul>
+   *
+   * @return the Kotlin home path
+   */
```
dsyang: The types of directory structures we consider valid kotlin_home isn't quite clear.  If I have my kotlinc in `~/bin/kotlinc` then this considers `~/bin` to be my kotlin_home which seems like it'd be wrong (`~/bin/lib/kotlin-runtime.jar`? no thanks).

Can we add a verify method that makes sure `kotlin_home`s are either structured like:
```
kotlin_home_as_downloaded_and_unziped_from_website/
    bin/
    lib/
```
or 
```
kotlin_home_as_installed_from_homebrew/
    bin/
    libexec/
        lib/
```

src/com/facebook/buck/jvm/kotlin/Kotlinc.java
```diff
@@ -0,0 +1,53 @@
...
+      ProjectFilesystem fileSystem
+      ) throws InterruptedException;
+
+  String getDescription(
+      ImmutableList<String> options,
+      ImmutableSortedSet<Path> kotlinSourceFilePaths,
+      Path pathToSrcsList);
+
+  String getShortName();
+
```
dsyang: extra whitespace

src/com/facebook/buck/jvm/kotlin/KotlincStep.java
```diff
@@ -17,69 +17,129 @@
...
+          extraArguments,
+          declaredClassPathEntries,
+          Optional.empty(),
+          filesystem);
+
+      String firstOrderStdout = stdout.getContentsAsString(Charsets.UTF_8);
+      String firstOrderStderr = stderr.getContentsAsString(Charsets.UTF_8);
+      Optional<String> returnedStderr;
+      if (declaredDepsBuildResult != 0) {
+        returnedStderr = Optional.of(firstOrderStderr); //processBuildFailure(context, firstOrderStdout, firstOrderStderr);
```
dsyang: address comment or remove

src/com/facebook/buck/jvm/kotlin/KotlincStep.java
```diff
@@ -17,69 +17,129 @@
...
+
+  @VisibleForTesting
+  Kotlinc getKotlinc() {
+    return kotlinc;
+  }
+
+  @Override
+  public String getDescription(ExecutionContext context) {
+    return getKotlinc().getDescription(
+        ImmutableList.of(), // TODO getOptions(context, getClasspathEntries()),
```
dsyang: address comment or remove

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -68,24 +70,41 @@ public void createCompileStep(
...
             ImmutableSortedSet.<Path>naturalOrder()
                 .addAll(
-                    Optional.ofNullable(extraClassPath.apply(context)).orElse(ImmutableList.of()))
+                    Optional.ofNullable(extraClassPath.apply(buildContext)).orElse(ImmutableList.of()))
                 .addAll(declaredClasspathEntries)
                 .build(),
             filesystem));
   }
 
+//  @Override
```
dsyang: address comment or remove

src/com/facebook/buck/jvm/kotlin/LICENSE
```diff
@@ -0,0 +1,15 @@
+/*
```
dsyang: remove

test/com/facebook/buck/jvm/kotlin/KotlinBuckConfigTest.java
```diff
@@ -17,168 +17,190 @@
...
   private ProjectWorkspace workspace;
+  private Path testDataDirectory;
 
   @Before
   public void setUp() throws InterruptedException, IOException {
     KotlinTestAssumptions.assumeUnixLike();
-
-    workspace = TestDataHelper.createProjectWorkspaceForScenario(this, "kotlin_compiler_test", tmp);
-    workspace.setUp();
+    workspace = TestDataHelper
```
dsyang: I don't think we should create a new testdata directory for this buck config test.  That's pretty heavy when you can just use tmp to create temporary files/directory structures.

Check out how `DBuckConfigTest`,  does it.  We should do it that way and delete the `testdata/faux_kotlin_home` and `testdata/kotlin_compiler_test` directories.

test/com/facebook/buck/jvm/kotlin/KotlinBuckConfigTest.java
```diff
@@ -17,168 +17,190 @@
...
-      throws HumanReadableException, IOException {
+  public void testFindsKotlinStdlibJarInConfigWithAbsolutePath()
+      throws HumanReadableException, InterruptedException, IOException {
 
-    Path kotlinRuntime = workspace.resolve("lib").resolve("kotlin-runtime.jar");
+    Path kotlinRuntime = testDataDirectory
+        .resolve("faux_kotlin_home")
+        .resolve("libexec")
+        .resolve("lib")
+        .resolve("kotlin-stdlib.jar");
```
dsyang: i think tests are failing because of kotlin-stdlib / kotlin-runtime naming confusion

brettwooldridge:
@dsyang Thanks.  Fixed.  Also cleaned up code, minimized diffs, and fixed all unit tests.  Pending further review, I'm done on my end.

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
```diff
@@ -0,0 +1,230 @@
...
+import java.util.stream.Collectors;
+
+public class JarBackedReflectedKotlinc implements Kotlinc {
+
+  private static final String CLASSPATH_FLAG = "-cp";
+  private static final String DESTINATION_FLAG = "-d";
+  private static final String INCLUDE_RUNTIME_FLAG = "-include-runtime";
+  private static final String COMPILER_CLASS = "org.jetbrains.kotlin.cli.jvm.K2JVMCompiler";
+  private static final String EXIT_CODE_CLASS = "org.jetbrains.kotlin.cli.common.ExitCode";
+  private static final KotlincVersion VERSION = KotlincVersion.of("in memory");
```
brettwooldridge: I just basically mimicked what [JarBackedJavac](https://github.com/facebook/buck/blob/c112b5dc9330005cdfeb6c574e5f35b5ed2d6469/src/com/facebook/buck/jvm/java/JarBackedJavac.java#L67) was using.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -44,12 +41,20 @@ public KotlinBuckConfig(BuckConfig delegate) {
...
+      ImmutableSet<Path> classpathEntries = ImmutableSet.of(
+          getPathToStdlibJar(),
+          getPathToCompilerJar());
+
+      return new JarBackedReflectedKotlinc(classpathEntries);
+    }
+  }
+
+
+  Path getCompilerPath() {
```
brettwooldridge: Done.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
brettwooldridge: @dsyang I added additional tests to verify correct behavior.  The supported structures are...

For _external_ compilation:
 * *external = true* defined in ``.buckconfig``
 * *kotlin_home* pointing directly at directory containing ``kotlinc``
 * *kotlin_home* pointing at directory containing a ``bin`` directory that contains ``kotlinc``
 * No *kotlin_home* defined, ``kotlinc`` in environment ``PATH``

For _in-process_ compilation:
 * *kotlin_home* pointing directly at directory containing ``kotlin-*`` JARs <sup>1</sup>
 * *kotlin_home* pointing at directory containing a ``lib`` directory that contains ``kotlin-*`` JARs
 * *kotlin_home* pointing at directory containing a ``libexec`` directory, that contains a ``lib`` directory, that contains ``kotlin-*`` JARs
 * No *kotlin_home* defined, ``kotlinc`` in environment ``PATH``, assume ``kotlinc`` is in a ``bin`` directory, and the true "kotlin home" is one directory above that.

<sup>1</sup> This supports the case, actually my company's case, where Kotlin jars are present in the absence of an installation structure.  In our case, outside of buck, we essentially download the jars using maven:
```xml
<dependency>
    <groupId>org.jetbrains.kotlin</groupId>
    <artifactId>kotlin-*</artifactId>
    <version>${kotlin-version}</version>
</dependency>
```

src/com/facebook/buck/jvm/kotlin/Kotlinc.java
brettwooldridge: Removed.

src/com/facebook/buck/jvm/kotlin/KotlincStep.java
brettwooldridge: Addressed.

test/com/facebook/buck/jvm/kotlin/KotlinBuckConfigTest.java
brettwooldridge: All tests passing now.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
brettwooldridge: Done.

dsyang:
@brettwooldridge  I'm seeing a bunch of "done" / "addressed" comments without any changes.  Maybe one of your commits got lost in translation?

brettwooldridge:
A bunch?  The only missed ones I found was the change to ``getPathToCompilerBinary()`` and the comment on KotlincStep:117.  Both are pushed now.

dsyang:
what about https://github.com/facebook/buck/pull/1283#discussion_r117647080?

Or do you think that belongs in a separate PR?

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
kageiit: The usecase mentioned by @brettwooldridge in 1 will be what we would be using for okbuck as well. Thanks!

brettwooldridge:
I'm not sure what is particularly "heavy" about the testdata directory, given that it is merely a "structure" and the "binaries" and "jars" themselves are merely files of several dozen bytes.

I have complained before about *truly heavy* testdata contents like that used by the scala tests, but it seems to fall on deaf ears.  [Scala testdata](https://github.com/facebook/buck/tree/master/test/com/facebook/buck/jvm/scala/testdata) has the scala compiler (15MB) and various other scala jars (~10MB) checked in *multiple times* -- totaling over 100MB.  By comparison, I'm not particularly concerned about three folders with fake binary files totaling a 100 bytes or so.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
dsyang: Fair points, didn't think about maven or okBuck.  lets keep it this way then!

dsyang:
By "heavy" I meant cognitive load instead of file size load.  I'm also not worried about the size of those testdata directories.  Perhaps a better way to phrase it is: "Do we really need these to be testdata directories?" 

To me, testdata dirs are useful in places where file content is important (such as KotlinLibraryIntegrationTest where you actually compiling those sources).  In the case of KotlinBuckConfigTest we only care about the existence of those files in that hierarchy. That's exactly what `tmp.newExectuableFile()` and friends provide: the temporary existence of a file without caring about it's contents. 

The "cognitive load" part comes from locality.  To figure out what that test is doing I need to look at the directory structure, then look at the test.  With the way `DBuckConfigTest` does it, you only have to look at the test file to understand what's going on.

brettwooldridge:
Done.  I left the ``kotlin_compiler_test`` unchanged as that code was not touched by this PR.

src/com/facebook/buck/jvm/kotlin/AbstractKotlincVersion.java
dsyang: https://travis-ci.org/facebook/buck/jobs/235076076#L4351

src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
```diff
@@ -0,0 +1,158 @@
+package com.facebook.buck.jvm.kotlin;
```
dsyang: https://ci.appveyor.com/project/Facebook/buck/build/3716#L2559

src/com/facebook/buck/jvm/kotlin/KotlincStep.java
```diff
@@ -20,66 +20,174 @@
...
 import com.google.common.collect.ImmutableSortedSet;
+
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Optional;
 
-public class KotlincStep extends ShellStep {
-
+@SuppressWarnings("unused")
```
dsyang: are there warnings left to suppress? I thin we use everything

test/com/facebook/buck/jvm/kotlin/BUCK
```diff
@@ -137,8 +137,10 @@ java_library(
     ],
 )
 
-standard_java_test(
+java_test(
```
dsyang: this change should be reverted since we don't have a faux_kotlin_home directory anymore

brettwooldridge:
I see you started the merge, does that mean you are going to address the "nits", or do you still want me to change and squash them?  I'm out of the office right now but I could do it when I get in.

dsyang:
If you could squash them that'd be great.  I just imported the PR to run it in our internal CI servers in case there are more tests that fail (I don't expect them to)

src/com/facebook/buck/jvm/kotlin/BUCK
```diff
@@ -6,6 +6,8 @@ java_immutables_library(
     ],
     visibility = ["PUBLIC"],
     deps = [
+        "//src/com/facebook/buck/android:apkmodule",
+        "//src/com/facebook/buck/android:packageable",
```
brettwooldridge: When I take these out, I get this error:
```
buck/src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java:93: error: cannot access com.facebook.buck.android.AndroidPackageable
    DefaultJavaLibrary testsLibrary = resolver.addToIndex(defaultJavaLibraryBuilder.build());
                                              ^
  class file for com.facebook.buck.android.AndroidPackageable not found
Errors: 1. Warnings: 0.
```

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
```diff
@@ -0,0 +1,211 @@
...
+      Class<?> exitCodeClass = compilerShim
+          .getClass()
+          .getClassLoader()
+          .loadClass(EXIT_CODE_CLASS);
+
+      Method getCode = exitCodeClass.getMethod("getCode");
+
+      Object exitCode = compile.invoke(
+          compilerShim,
+          new UncloseablePrintStream(context.getStdErr()),
```
dsyang: Question: Why are we using a UncloseablePrintStream here? Curious what the problem is with just context.getStdErr().

brettwooldridge: The Kotlin compiler closes the stream and breaks the buck daemon.

dsyang: @brettwooldridge an Infer run is complaining about a [resource leak](http://fbinfer.com/docs/infer-bug-types.html#RESOURCE_LEAK) here:  Any ideas on the best way to fix this?

brettwooldridge:
@dsyang That push should fix the warning.

brettwooldridge:
Rebased onto master.

brettwooldridge:
@kageiit @dsyang Anything else waiting on me here?

dsyang:
Don't have my computer with me for the next couple of days will look at it
again on Thursday

kageiit:
@dsyang Ping on this :)

aww seems there are new merge conflicts :( @brettwooldridge

brettwooldridge:
@kageiit Jesus, do we really need all these refactors that apparently touch almost every class in the tree?

dsyang:
@brettwooldridge @kageiit  We're just waiting on some people from the buck team to do a quick once-over for this PR on our internal phabricator import, unfortunately some people have been OoO so it's not moving as fast as I'd like.  I've been rebasing and updating it internally but I don't know how to make it show up here in github. 

You guys shouldn't need to do anything more here.  It's my job now to push someone internally to look at it :/

brettwooldridge:
@dsyang @kageiit Just checking on this.  It is closed, but not merged?  A little concerned, because pull #1358 seems based off of the existing mainline but will surely create conficts if merged before this.

Coneko:
It was merged as 800d5a9669417d5a8b4bc455e34a629422ed1ff1.

sdwilsh:
"merge" means someone actually used the merge button in the GitHub UI. 
That will never happen with this project, so PRs always show up as "closed".

brettwooldridge:
@sdwilsh Sorry, I missed the ``facebook-github-bot closed this in 800d5a9 9 days ago`` message.  I guess the next question is, when might we see this is a release?

sdwilsh:
It looks like @aiked has been doing those for Buck.

dsyang:
@brettwooldridge  if you don't want to wait for a release you could also use a .buckversion file https://buckbuild.com/concept/buckversion.html

Issue #1285, opened 2017-04-11
BuckJavacTask breaks custom compilers


kageiit:
This commit introduced a regression in being able to consume custom compilers like error prone in buck

https://github.com/facebook/buck/commit/590b4a96c098f9b90a6f3ec42b761fc749a9a861

Specifically this line: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/plugin/api/BuckJavacTaskProxy.java#L42

Causes the build to insta fail with this message

```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at com.facebook.buck.jvm.java.plugin.api.BuckJavacTaskProxy.newInstance(BuckJavacTaskProxy.java:44)
	at com.facebook.buck.jvm.java.Jsr199Javac.buildWithClasspath(Jsr199Javac.java:262)
	at com.facebook.buck.jvm.java.Jsr199Javac.buildWithClasspath(Jsr199Javac.java:160)
	at com.facebook.buck.jvm.java.JavacStep.performBuild(JavacStep.java:154)
	at com.facebook.buck.jvm.java.JavacStep.tryBuildWithFirstOrderDeps(JavacStep.java:144)
	at com.facebook.buck.jvm.java.JavacStep.execute(JavacStep.java:106)
	at com.facebook.buck.jvm.java.JavacDirectToJarStep.execute(JavacDirectToJarStep.java:90)
	at com.facebook.buck.step.DefaultStepRunner.runStepForBuildTarget(DefaultStepRunner.java:47)
	at com.facebook.buck.rules.CachingBuildEngine.executeCommandsNowThatDepsAreBuilt(CachingBuildEngine.java:1538)
	at com.facebook.buck.rules.CachingBuildEngine.lambda$17(CachingBuildEngine.java:354)
	at com.facebook.buck.util.concurrent.WeightedListeningExecutorService.lambda$0(WeightedListeningExecutorService.java:81)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.doTransform(AbstractTransformFuture.java:211)
	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.doTransform(AbstractTransformFuture.java:200)
	at com.google.common.util.concurrent.AbstractTransformFuture.run(AbstractTransformFuture.java:130)
	at com.google.common.util.concurrent.MoreExecutors$5$1.run(MoreExecutors.java:988)
[...]

PR #1286, opened 2017-04-11
Add compilation task event mirror enum

kageiit:
Fixes compatibility with jdk 9 compilers. See https://github.com/facebook/buck/issues/1285#issuecomment-293146039


kageiit:
cc @jkeljo @msridhar

PR #1287, opened 2017-04-11
Fix persistent worker process bug when a process dies.

mikekap:
When a build is canceled, a SIGINT is sent to Buck's entire process tree. This kills any persistent workers. Unfortunately, this also hung the `WorkerShellStep`s for the next build.

This PR fixes the particular problem, but not the core issue of persistent worker death. I'm not sure how to tackle the core issue without either:
 - Adding a new message type to the protocol. (i.e. "ping")
 - Adding some amount of retries with persistent workers per step (that is >= the number of workers in the pool).

This PR solves the particular issue above (SIGINT & clean death) by killing workers that get interrupted in the middle of a build step. It also improves logging so worker tool debugging is a bit easier.

### DIFF
--- a/src/com/facebook/buck/shell/WorkerProcess.java
+++ b/src/com/facebook/buck/shell/WorkerProcess.java
@@ -82,6 +82,10 @@ public class WorkerProcess {
     this.tmpPath = tmpPath;
   }
 
+  public boolean isAlive() {
+    return launchedProcess != null && launchedProcess.isAlive();
+  }
+
   public synchronized void ensureLaunchAndHandshake() throws IOException {
     if (handshakePerformed) {
       return;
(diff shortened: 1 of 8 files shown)


mikekap:
Not sure why travis is choking on Go, but it's unrelated.

mikekap:
Ping

src/com/facebook/buck/shell/WorkerProcess.java
```diff
@@ -162,18 +166,22 @@ public void close() {
...
-          .replace("\n", "\nstderr: ");
-      LOG.error("stderr: %s", workerStderr);
+      try {
+        String workerStderr = MoreStrings
+            .truncatePretty(filesystem.readFileIfItExists(this.stdErr).orElse(""))
+            .trim()
+            .replace("\n", "\nstderr: ");
+        LOG.error("stderr: %s", workerStderr);
+      } catch (Throwable t) {
+        LOG.error(t, "Couldn't read stderr on failing close!");
```
Coneko: When does this happen?

src/com/facebook/buck/util/DefaultProcessExecutor.java
```diff
@@ -217,6 +217,13 @@ public Result waitForLaunchedProcessWithTimeout(
     );
   }
 
+  @Override
+  public boolean isAlive(LaunchedProcess launchedProcess) {
+    Preconditions.checkState(launchedProcess instanceof LaunchedProcessImpl);
+    final Process process = ((LaunchedProcessImpl) launchedProcess).process;
```
Coneko: Rather than accessing the field after the instanceof check add a `isAlive` method to `LaunchedProcess`.

It's possible the method won't be necessary on the executor at that point.

Coneko:
If you rebase on master Travis should go green.

src/com/facebook/buck/shell/WorkerProcess.java
mikekap: Various reasons. OOM is an easy one. Another is: when canceling a build, the Thread running `WorkerShellStep` gets interrupted, which means various system calls start failing with `InterruptedException` - reading files is one of them on SOME filesystems.

mikekap:
Looks like there's more failures now in travis related to timeouts.

PR #1296, opened 2017-04-19
entitlements file uses info_plist_substitutions for variable substitution

ProfessaA:
### background ###

Currently, `apple_bundle`s can take `info_plist_substitutions` that are used to replace variables in the specified `info_plist`.  However, variable substitution does not occur for the entitlements file.  This makes it tricky to share things like `GROUP_ID` or `AppIdentifierPrefix` that need to match across the plist and entitlements files.  It also diverges from xcode's behavior where variables set in build settings can be accessed in both the info plist and entitlements files.

### change ###

This PR makes it so that `info_plist_substitutions` are also applied to the entitlements file specified via `CODE_SIGN_ENTITLEMENTS`.

### verification ###

- [x] add `AppleBundleIntegrationTest.infoPlistSubstitutionsAreAppliedToEntitlements`
- [x] verify that app with entitlements variables can be correctly code signed and launched on device


ryu2:
@facebook-github-bot shipit

PR #1298, opened 2017-04-20
Update the SimpleConsoleEventBusListener output once every 15s

shs96c:
Without this, systems that timeout after too long
without output on stdout/stderr time out long
running buck builds and tasks (eg. Travis)


shs96c:
Test failures look unrelated.

PR #1299, opened 2017-04-20
Introduce rules for compiling closure javascript

shs96c:
This includes:

- `closure_library`: no concrete output, but a
  useful way of collecting functionality into
  modules.

- `closure_binary`: compile a set of
  `closure_library` instances into a single
  minified output.

- `closure_fragment`: takes a single method
  from a `closure_library` and exposes it as
  a minified set of JS, suitable for injeting
  into a web page.

The last rule is used extensively within Selenium
to implement shared functionality which is best
done using JS.

### DIFF
--- a/src/org/openqa/selenium/buck/javascript/ClosureLibraryDescription.java
+++ b/src/org/openqa/selenium/buck/javascript/ClosureLibraryDescription.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2013-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.openqa.selenium.buck.javascript;
+
+import com.facebook.buck.model.BuildTarget;
+import com.facebook.buck.parser.NoSuchBuildTargetException;
+import com.facebook.buck.rules.AbstractDescriptionArg;
+import com.facebook.buck.rules.BuildRule;
+import com.facebook.buck.rules.BuildRuleParams;
+import com.facebook.buck.rules.BuildRuleResolver;
+import com.facebook.buck.rules.CellPathResolver;
+import com.facebook.buck.rules.Description;
+import com.facebook.buck.rules.SourcePath;
+import com.facebook.buck.rules.TargetGraph;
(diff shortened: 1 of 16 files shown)


shs96c:
I kind of expect a few rounds of review on this one, but let's kick off the process. This is the largest delta between Selenium's Buck version and Facebook's. It'd be nice to get us back to using the official Buck release.

Coneko:
There's no tests and the package name is weird, but we can fix both up after the initial merge if you're up for that.

I'll be working under the assumption that if they languish I'll revert the whole thing.

shs96c:
The package name is weird because I wrote these for the fork of Buck used by Selenium (`org.openqa.selenium` is our normal package). I'd be okay changing the package name if it'd make life easy.

The only tests that ever existed were the build for selenium.... Writing some sanity checks wouldn't be too hard.

src/org/openqa/selenium/buck/javascript/ClosureLibraryDescription.java
```diff
@@ -0,0 +1,55 @@
...
+import com.facebook.buck.rules.BuildRule;
+import com.facebook.buck.rules.BuildRuleParams;
+import com.facebook.buck.rules.BuildRuleResolver;
+import com.facebook.buck.rules.CellPathResolver;
+import com.facebook.buck.rules.Description;
+import com.facebook.buck.rules.SourcePath;
+import com.facebook.buck.rules.TargetGraph;
+import com.google.common.collect.ImmutableSortedSet;
+
+public class ClosureLibraryDescription implements Description<ClosureLibraryDescription.Arg> {
```
ttsugriy: buck has moved over to using immutables for arguments. Please update your code to use them as well.

src/org/openqa/selenium/buck/javascript/JavascriptSource.java
```diff
@@ -0,0 +1,140 @@
...
+//        "^goog.addDependency\\s*\\(\\s*",
+//        "['\"]([^'\"]+)['\"]",     // Relative path from Closure's base.js
+//        "\\s*,\\s*",
+//        "\\[([^\\]]+)?\\]",        // Provided symbols
+//        "\\s*,\\s*",
+//        "\\[([^\\]]+)?\\]",        // Required symbols
+//        "\\s*",
+//        "(?:,\\s*(true|false))?",  // Module flag.
+//        "\\s*\\)"
+//    ));
```
ttsugriy: why is this code commented out?

src/org/openqa/selenium/buck/javascript/JavascriptDependencies.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+
+class JavascriptDependencies {
+
+  private Map<String, JavascriptSource> provide2Source = Maps.newHashMap();
+
+  public void add(JavascriptSource source) {
```
ttsugriy: nit: passing mutable objects like this around makes reasoning about invariants harder, so I would recommend considering separating a mutable builder from final javascript deps representation.

PR #1308, opened 2017-04-24
No logical change: reformat implemented interfaces for DefaultJavaLibrary

shs96c:
This class appears to be attempting to implement
every interface that Buck has, so make it simple
to read them all (nb: clarity is the real reason
for this diff.


Coneko:
This took so long to try to land that it got subsumed by the automated refactor.

PR #1310, opened 2017-04-24
add important JVM arguments for buck launched with NO_BUCKD

msridhar:
Without `-XX:SoftRefLRUPolicyMSPerMB=0`, running Buck can have really bad performance with `NO_BUCKD`, as the JVM is not aggressive enough about clearing `SoftReference`s.  It seemed best to also enable the G1 GC in `NO_BUCKD` mode but this may not be as critical.  The other JVM args passed when using the Buck daemon seemed less relevant without the daemon.

### DIFF
--- a/programs/buck_tool.py
+++ b/programs/buck_tool.py
@@ -210,7 +210,9 @@ class BuckTool(object):
 
             command = ["buck"]
             extra_default_options = [
-                "-Djava.io.tmpdir={0}".format(self._tmp_dir)
+                "-Djava.io.tmpdir={0}".format(self._tmp_dir),
+                "-XX:SoftRefLRUPolicyMSPerMB=0",
+                "-XX:+UseG1GC",
             ]
             command.extend(self._get_java_args(buck_version_uid, extra_default_options))
             command.append("com.facebook.buck.cli.bootstrapper.ClassLoaderBootstrapper")


kageiit:
cc @dreiss @Coneko can we please get this reviewed?

Coneko:
Sorry I'm waiting for some extra telemetry to land internally so we can benchmark this change easily.

Coneko:
Didn't see any big changes on our side, would this address only certain builds?

Also I notice there's other JVM args we pass to the buckd invocation but not the invocation without the daemon, should we apply those as well by factoring them out?

msridhar:
@Coneko the main thing we observed were additional OOMs without these flags with `NO_BUCKD`.  Buck makes use of `SoftReference`s in certain places (I forget where), and without the `-XX:SoftRefLRUPolicyMSPerMB=0` flag the JVM waits a very long time before collecting them.  Beyond solving that issue, I don't think we saw much running time performance improvement.

As far as the other JVM flags passed to buckd [here](https://github.com/msridhar/buck/blob/5948dacfd47ed4dc57e21823ab0b6d66f4ab6af6/programs/buck_tool.py#L269), they seemed to me to be primarily daemon-specific.  E.g., `-XX:GuaranteedSafepointInterval=5000` seems designed to prevent the daemon from waking up too much in the background.  But I don't have a complete understanding of all the flags, and probably they won't hurt anything if passed in non-daemon mode if consistency is desired.

PR #1313, opened 2017-04-27
Fix docs and don't escape literal braces

cpick:
`get_cell_name()`s documentation was missing the required `args` param
causing soy to error out and fail to build the documentation.

Add the param.

Escaped braces (eg `{lb}` and `{rb}`) should not be used within
`{literal}` blocks as that will literally print the strings "{lb}" or
"{rb}".

Fix the tutorial page.

When the Tutorial was split out of Getting Started in 71f8452,
the 'Path Setup' section was garbled with a copy-paste error from
a later section.

Fix tutorial's 'Path Setup' section

### DIFF
--- a/docs/function/get_cell_name.soy
+++ b/docs/function/get_cell_name.soy
@@ -18,6 +18,16 @@ The <code>get_cell_name()</code> function is used the canonical cell name of the
 from within a build file.
 {/param}
 
+{param args}
+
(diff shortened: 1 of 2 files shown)


Coneko:
cc @mzlee

mzlee:
Thank you!

PR #1315, opened 2017-05-01
Xcode project generator expands location macros in linker flags

ProfessaA:
### background ###
Previously, `ProjectGenerator` did not expand location macros used in linker flags.  As a result, buck builds for targets with linker flags using location macros would succeed but building a project generated from the same target would fail.

### change ###
`ProjectGenerator` now expands location macros to their absolute path and adds the macro's build target to the `ProjectGenerator`'s list of required targets.

### verification ###
- [x] add `ProjectGeneratorTest.testAppleLibraryLinkerFlagsWithLocationMacrosAreExpanded`
- [x] `buck test test/com/facebook/buck/apple/project_generator:project_generator`
- [x] verify that `buck project` generates a project with linker flag location macros correctly expanded, and that the targets of those macros were built as part of project generation

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -113,9 +113,8 @@ import com.facebook.buck.rules.TargetNode;
 import com.facebook.buck.rules.args.StringWithMacrosArg;
 import com.facebook.buck.rules.coercer.FrameworkPath;
 import com.facebook.buck.rules.coercer.SourceList;
-import com.facebook.buck.rules.macros.AbstractMacroExpander;
 import com.facebook.buck.rules.macros.LocationMacro;
-import com.facebook.buck.rules.macros.Macro;
+import com.facebook.buck.rules.macros.LocationMacroExpander;
 import com.facebook.buck.rules.macros.StringWithMacros;
 import com.facebook.buck.shell.AbstractGenruleDescription;
 import com.facebook.buck.shell.ExportFileDescriptionArg;
@@ -310,7 +309,7 @@ public class ProjectGenerator {
     this.defaultCxxPlatform = defaultCxxPlatform;
     this.buildRuleResolverForNode = buildRuleResolverForNode;
     this.defaultBuildRuleResolver =
-        new BuildRuleResolver(TargetGraph.EMPTY, new DefaultTargetNodeToBuildRuleTransformer());
+        new BuildRuleResolver(targetGraph, new DefaultTargetNodeToBuildRuleTransformer());
     this.defaultPathResolver =
         new SourcePathResolver(new SourcePathRuleFinder(this.defaultBuildRuleResolver));
     this.buckEventBus = buckEventBus;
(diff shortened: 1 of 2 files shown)


ryu2:
The high level approach seems OK, but the action graph creation due to build rule resolution causes various issues with things like `CxxLibrary` rules, since the setup we do for project generation differs in subtle ways from builds.

Let me play around with things and see if this can be fixed easily.

ryu2:
@ProfessaA This seems to work with the below changes (basically using `defaultBuildRuleResolver` which doesn't instantiate the AG).  Can you patch this and see if this still works for your scenario?

```
diff --git a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
index d7e7073fd4..1b7fff6ddd 100644
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -301,7 +301,7 @@ public class ProjectGenerator {
     this.defaultCxxPlatform = defaultCxxPlatform;
     this.buildRuleResolverForNode = buildRuleResolverForNode;
     this.defaultBuildRuleResolver =
-        new BuildRuleResolver(TargetGraph.EMPTY, new DefaultTargetNodeToBuildRuleTransformer());
+        new BuildRuleResolver(targetGraph, new DefaultTargetNodeToBuildRuleTransformer());
     this.defaultPathResolver =
         new SourcePathResolver(new SourcePathRuleFinder(this.defaultBuildRuleResolver));
     this.buckEventBus = buckEventBus;
@@ -860,7 +860,7 @@ public class ProjectGenerator {
               ImmutableList.of(locationExpander),
               node.getBuildTarget(),
               node.getCellNames(),
-              buildRuleResolverForNode.apply(node))
+              defaultBuildRuleResolver)
           .appendToCommandLine(result, defaultPathResolver);
     }
     return result.build();
diff --git a/src/com/facebook/buck/rules/macros/BuildTargetMacroExpander.java b/src/com/facebook/buck/rules/macros/BuildTargetMacroExpander.java
index d508c4f944..16bd04bbff 100644
--- a/src/com/facebook/buck/rules/macros/BuildTargetMacroExpander.java
+++ b/src/com/facebook/buck/rules/macros/BuildTargetMacroExpander.java
@@ -21,14 +21,15 @@ import com.facebook.buck.model.MacroException;
 import com.facebook.buck.parser.BuildTargetParseException;
 import com.facebook.buck.parser.BuildTargetParser;
 import com.facebook.buck.parser.BuildTargetPatternParser;
+import com.facebook.buck.parser.NoSuchBuildTargetException;
 import com.facebook.buck.rules.BuildRule;
 import com.facebook.buck.rules.BuildRuleResolver;
 import com.facebook.buck.rules.CellPathResolver;
 import com.facebook.buck.rules.SourcePathResolver;
 import com.facebook.buck.rules.SourcePathRuleFinder;
+import com.facebook.buck.rules.TargetGraph;
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
-import java.util.Optional;
 
 /**
@@ -44,11 +45,11 @@ public abstract class BuildTargetMacroExpander<M extends BuildTargetMacro>
       throws MacroException;
 
   protected BuildRule resolve(BuildRuleResolver resolver, M input) throws MacroException {
-    Optional<BuildRule> rule = resolver.getRuleOptional(input.getTarget());
-    if (!rule.isPresent()) {
+    try {
+      return resolver.requireRule(input.getTarget());
+    } catch (NoSuchBuildTargetException | TargetGraph.NoSuchNodeException e) {
       throw new MacroException(String.format("no rule %s", input.getTarget()));
     }
-    return rule.get();
   }
 
   BuildTarget parseBuildTarget(
```

ProfessaA:
@ryu2 yup with that patch the location macros are still expand properly.  thanks for the help!

ryu2:
LGTM.  cc: @illicitonion @Coneko for a quick sanity check on the `BuildTargetMacroExpander` changes.

illicitonion:
I believe this is a significant change to how the macro works here. The old code said "Any build target you reference must be a (transitive) dependency of the rule with the macro". The new code says "Any build target you reference must be in the same target graph as the rule with the macro". I *think* that that means this change may break cache invalidation in some subtle way, but I don't know for sure...

@andrewjcg you know this corner a lot better than I do, WDYT?

ProfessaA:
@ryu2 @illicitonion In that case, would it make sense to leave the `BuildTargetMacroExpander` alone and in the project generator first try to use the `LocationMacroExpander`, and if that fails fall back to the `AsIsLocationMacroExpander`?

ProfessaA:
@ryu2 as an alternative how about https://github.com/ProfessaA/buck/commit/eb8d7c59f2a611244854c71f9ede5aaca07980a4 ? It's the same approach except that the logic added to `BuildTargetMacroExpander` is moved to the `ProjectGenerator`'s anonymous `LocationMacroExpander` instance so that the macro expander is left alone.  If it looks good, I'll target that branch instead for this PR.

ryu2:
@ProfessaA -- that sounds like a better approach.  Does it also work with `export_file` rules as well?  That sounds like a good use case for things like custom order files, etc.

ProfessaA:
@ryu2 yup project generation works with `export_file` rules like:
```
apple_library(
  ...
  exported_linker_flags = ['-force_load', '$(location :libExported.a)'],
)

export_file( name = 'libExported.a' )
```

I also updated the `ProjectGeneratorTest` to make sure that linker flags with location macros referencing `export_file` as well as `genrule` targets get properly expanded.

ryu2:
@facebook-github-bot shipit

ProfessaA:
@ryu2 hey, do you know what the status is for this PR?

ryu2:
This PR is good to merge.  We were experiencing some issues with our github tooling which was recently resolved, so currently working through our backlog of PRs.

ProfessaA:
awesome, thanks!

ProfessaA:
@ryu2 the commit that landed (https://github.com/facebook/buck/commit/4bd64527b095f53f609a1f80001dea9f66a55ffd) doesn't have all the changes that were made in this PR - I think maybe the initial version of the PR got pulled in instead.  What's the best way to fix this, should I submit another PR?

ryu2:
Yeah, that's probably the best thing to do.

ryu2:
Actually, no need.  We had to revert this PR due to some build errors we found.  They're on our side, so I will fix forward and re-import the latest version of this PR.

So unless you have further amendments to make, no need for any action from you.

ProfessaA:
nope no further amendments, the latest version of this PR should be good.  thanks!

PR #1317, opened 2017-05-03
actool uses correct value for '--minimum-deployment-target' flag

ProfessaA:
### background ###

Previously, the `actool` command to compile asset catalogs passed a hard-coded `7.0` to the `--minimum-deployment-target` flag.

### change ###

Now the correct value, taken from `AppleCxxPlatform#getMinVersion`, is passed to the `--minimum-deployment-target` flag.

### verification ###

- [x] `AppleBundleIntegrationTest#appleAssetCatalogsAreIncludedInBundle` passes
- [x] `buck build` project with deployment target of 9.0 and verify that the `Assets.car` file is correct

### DIFF
--- a/src/com/facebook/buck/apple/ActoolStep.java
+++ b/src/com/facebook/buck/apple/ActoolStep.java
@@ -29,6 +29,7 @@ import java.util.SortedSet;
 class ActoolStep extends ShellStep {
 
   private final String applePlatformName;
+  private final String targetSDKVersion;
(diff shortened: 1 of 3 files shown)


ProfessaA:
no problem.  I've updated the variable naming to be consistent with the buck config `*_target_sdk_version` parameter.

ryu2:
@facebook-github-bot shipit

PR #1318, opened 2017-05-03
Fix incorrect RDotTextEntry sorting

paveldudka:
It turns out that currently `RDotTxtEntry` sorting logic doesn't handle one of the corner cases which produces invalid R.java.
That's basically the reason why we had bunch of test failures after we integrated #879

Consider following example:

```
<declare-styleable name="AnchorLayout">
  <attr name="anchorInner" format="reference" />
</declare-styleable>

<declare-styleable name="AnchorLayout_Layout">
  <attr name="somethingElse"/>
</declare-styleable>
```

This should result in something like this:

```
   int[] styleable AnchorLayout { 0x7F010101 }
   int styleable AnchorLayout_anchorInner 0

   int[] styleable AnchorLayout_Layout { 0x7F010102 }
   int styleable AnchorLayout_Layout_somethingElse 0
```

However, current sorting send `AnchorLayout_anchorInner` all the way to the botton (since `AnchorLayout_anchorInner` > `AnchorLayout_Layout`) which breaks assumptions made in `MergeAndroidResourcesStep`, so here is what we currently get:

```
   int[] styleable AnchorLayout { }

   int[] styleable AnchorLayout_Layout { 0x7F010102 }
   int styleable AnchorLayout_Layout_somethingElse 0

   ....
   int styleable AnchorLayout_anchorInner 0x7F010101	
```

which is a complete nonesence.

Approach:

Since we need to group our styleable attributes based on their parent and it is impossible to correctly detect parent name from styleable attr name (consider something like this `int styleable A_B_C 0` where it can be attr `B_C` which belongs to styleable `A` or attr `C` which belongs to styleable  `A_B` :)), we decided to introduce "parent" field to styleable attributes, so it carries information about its parent, so we can properly group them.

@raviagarwal7 , since you last touched this logic, it does make sense if you are up-to-speed with these changes


src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -339,7 +339,7 @@ void writePerPackageRDotJava(
     SortedSetMultimap<RDotTxtEntry, Path> bannedDuplicateResourceToSymbolsFiles =
         TreeMultimap.create();
 
-    HashMap<RDotTxtEntry, String> resourceToIdValuesMap = new HashMap<>();
+    HashMap<Integer, RDotTxtEntry> resourceToIdValuesMap = new HashMap<>();
```
raviagarwal7: why do we need `HashMap<Integer, RDotTxtEntry>`? we are only using the idValue below. 

I think it should atleast be `HashMap<RDotTxtEntry, RDotTxtEntry>` so that `equals()` on `RDotTxtEntry` is also called.

paveldudka: we also pass this map to `getStyleableResources()` method where we retrieve actual RDotTxtEntry instance (which should contain proper "parent" information). Otherwise parent info will be lost here.

Changing to Map<RDotTxtEntry, RDotTxtEntry> makes sense - Ill make proper changes

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -376,15 +376,16 @@ void writePerPackageRDotJava(
         } else if (useOldStyleableFormat) { // NOPMD  more readable this way, IMO.
           // Nothing extra to do in this case.
 
-        } else if (resourceToIdValuesMap.get(resource) != null) {
-          resource = resource.copyWithNewIdValue(resourceToIdValuesMap.get(resource));
+        } else if (resourceToIdValuesMap.containsKey(resource)) {
+          resource = resourceToIdValuesMap.get(resource);
+          resource = resource.copyWithNewIdValue(resource.idValue);
```
raviagarwal7: nit: don't really like setting resource from the map to resource and then replacing it in the next line
would prefer `resource = resource.copyWithNewIdValue(resourceToIdValuesMap.get(resource).idValue);` or using some local variable

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -461,12 +462,17 @@ void writePerPackageRDotJava(
           && styleableResource.type == RType.STYLEABLE
           && styleableResource.name.startsWith(styleablePrefix)) {
 
+        styleableResource = styleableResource.copyWithNewParent(resource.name);
```
raviagarwal7: nit: move it up on line 457

```
RDotTxtEntry styleableResource = 
     getResourceAtIndex(linesInSymbolsFile, styleableIndex + index).copyWithNewParent(resource.name);
```

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -495,11 +501,13 @@ void writePerPackageRDotJava(
...
           // Add resource to cache so that the id value is consistent across all R.txt
-          resourceToIdValuesMap.put(attrResource.copyWithNewIdValue(attrIdValue), attrIdValue);
+          attrResource = attrResource.copyWithNewIdValue(attrIdValue);
+          resourceToIdValuesMap.put(attrResource, attrResource);
         }
 
         styleableResourceMap.put(
-            styleableResource.copyWithNewIdValue(String.valueOf(styleableIndex)), attrIdValue);
+            styleableResource.copyWithNewIdValue(String.valueOf(styleableIndex)),
+            attrResource.idValue);
```
raviagarwal7: indentation looks incorrect

check if `ant lint` passes locally

Coneko: Sorry `ant lint` doesn't run google-java-format.

paveldudka: yeah, our IntelliJ runs google-java-format automatically now, so should be good

Issue #1325, opened 2017-05-04
Changes to a layout resource is not getting reflected in the built APK


raviagarwal7:
__Initial Layout:__
```lang=java
cp__app_app_layout.xml
<merge
  xmlns:android="http://schemas.android.com/apk/res/android">
    <android.support.v4.widget.Space android:id="@id/app_toolbar" android:visibility="gone" android:layout_width="fill_parent" android:layout_height="?actionBarSize" />
</merge>
```

__Updated Layout:__
```lang=java
cp__app_app_layout.xml
<merge
  xmlns:android="http://schemas.android.com/apk/res/android">
    <android.support.v4.widget.Space android:id="@id/app_toolbar_new" android:visibility="gone" android:layout_width="fill_parent" android:layout_height="?actionBarSize" />
</merge>
```

__On Launching the app it crashes with his error:__
```lang=java

No field app_toolbar_new

AndroidRuntime  E  FATAL EXCEPTION: main
               E  Process: com.comp.app, PID: 16487
               E  java.lang.NoSuchFieldError: No field app_toolbar_new of type I in class Lcom/comp/app/core/R$id; or its superclasses (declaration of 'com.com
                  p.app.core.R$id' appears in /data/app/com.comp.app-1/base.apk:classes10.dex)
               E      at com.comp.app.core.app.AppLayout_ViewBinding.<init>(AppLayout_ViewBinding.java:22)
```

__Checked that the UberRDotTxt is generated correctly:__
```lang=java
/Users/ravi/Comp/android/buck-out/bin/apps/app/app/__bin_debug#aapt_package_res_symbols__/R.java:
 11940:         public static final int app_toolbar_new=0x7f090264;

/Users/ravi/Comp/android/buck-out/bin/apps/app/app/__bin_debug#aapt_package_res_symbols__/R.txt:
 5937: int id app_toolbar_new 0x7f090264
```

__Each libraries R.java are also generated correctly using the UberRDotTxt:__
```lang=java
/Users/ravi/Comp/android/buck-out/bin/apps/app/app/__bin_debug#aapt_package_rdotjava_src__/com/comp/app/R.java:
 5949:     public static final int app_toolbar_new=0x7f090264;

/Users/ravi/Comp/android/buck-out/bin/apps/app/app/__bin_debug#aapt_package_rdotjava_src__/com/comp/app/core/R.java:
 2144:     public static final int app_toolbar_new=0x7f090264;

[...]

Issue #1252, opened 2017-03-16
Changing a resource id causes runtime crash


cwoodwar6:
We have a large, multidex enabled application. We have one module that exceeds the size of a single dex. When building with Buck, we must disable pre-dex because of this large module.

In our application, when a resource id is changed (for example when changing or adding a view id in a layout) the app will compile but crashes at runtime when trying to inflate the view where the id was changed. To get around this, we have to remove the contents of `buck-out` and build again. This appears to happen every time an id is changed.

We've recently tried moving a large chunk of code out of this module so we could enable pre-dex. Once we have done this, the issue seems to disappear.

I've attached the log from the runtime crash. It would be greatly appreciated if there is a way to work around this, we currently aren't able to keep pre-dex on.

[resource_id_log.txt](https://github.com/facebook/buck/files/849144/resource_id_log.txt)

The most interesting part of this log seems to be: `Caused by: java.lang.NoSuchFieldError: No static field my_view_id of type I in class Lcom/example/app/R$id; or its superclasses (declaration of 'com.example.app.R$id' appears in /data/app/com.example.app-1/base.apk:classes13.dex)`

raviagarwal7:
this seems to be a duplicate of https://github.com/facebook/buck/issues/1325.

@cwoodwar6 a fix has been landed on master, please verify and close this issue if it works.

cwoodwar6:
Thankfully, we no longer have to run with pre-dex disabled.

To verify, I disabled pre-dex, and was able to modify a resource id without a runtime crash. Looks resolved to me.

PR #1328, opened 2017-05-06
Use AccumulateClassNamesStep always to get classNamesToHashesSupplier

raviagarwal7:
Summary:
`packageableCollection.getClassNamesToHashesSupplier()` doesn’t gives hashes of 
class files like `R.class` causing R.java changes go undetected during smart dex
step.
This only affects apps with predexing off.

Resolves: https://github.com/facebook/buck/issues/1325, https://github.com/facebook/buck/issues/1252

Test Plan: Verified on the use case mentioned in https://github.com/facebook/buck/issues/1325 that R.java changes are detected now.


raviagarwal7:
@dreiss @cjhopman

PR #1331, opened 2017-05-08
Fix travis

LegNeato:
WIP, don't merge yet!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,9 @@ before_install:
   # Set up the Groovy environment
   - export GROOVY_HOME=/usr/share/groovy/
 
-cache:
-  directories:
-    - $HOME/android-ndk-linux
-    - $HOME/android-sdk-linux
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -43,10 +48,9 @@ before_install:
   # Set up the Groovy environment
   - export GROOVY_HOME=/usr/share/groovy/
 
-cache:
-  directories:
-    - $HOME/android-ndk-linux
-    - $HOME/android-sdk-linux
```
Coneko: I remember the reason we cached these is that it took so long to install them it would cause jobs to time out.

Coneko:
Update when ready.

.travis.yml
LegNeato: Doesn't seem to make it take too long:

https://travis-ci.org/facebook/buck/builds/229837080?utm_source=github_status&utm_medium=notification

PR #1338, opened 2017-05-10
Cache the ndk and sdk artifacts.

LegNeato:
Previously we were trying to cache the unzipped sdk and ndk, but it took
so long to tar them in the caching phase that it never worked (for an example see https://travis-ci.org/facebook/buck/builds/227525120).

Because Travis explicitly says to not cache android stuff (https://docs.travis-ci.com/user/caching/#Things-not-to-cache) I removed caching in a previous diff. Sadly, builds are still failing because sometimes we hit a slow mirror.

This turns caching back on, but only caches the compressed artifacts. Travis'
caching step should be materially faster (as it is only tarring one compressed file) and the cache should work.

If caching still times out, we can bump the limit...but I don't think we'll need to do so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,10 @@ before_install:
 
 # https://docs.travis-ci.com/user/caching#Things-not-to-cache
 # https://docs.travis-ci.com/user/caching#Explicitly-disabling-caching
-cache: false
+cache:
+  directories:
+    - $HOME/ndk_cache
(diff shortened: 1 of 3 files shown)


LegNeato:
You can see in https://travis-ci.org/LegNeato/buck/jobs/230848179 the caching now does not time out.

sdwilsh:
Oh man, this is sweet!

PR #1345, opened 2017-05-13
Add apple test's host app as a target dependency in generated xcode project

ProfessaA:
### background ###

Previously, the generated xcode project for `apple_test`s that specified a host app, did not add the app under the test's "Target Dependencies" build phase.  As a result, these tests failed to build if the host app hadn't already been built.

### change ###

Adds the host app under the test target's "Target Dependencies" build phase.

### verification ###

- [x] verify that the host app is a target dependency for unit tests with a specified host app
- [x] verify that the host app is a target dependency for UI tests
- [x] verify that logic tests without host apps continue to be generated correctly
- [x] verify that tests with host apps in a different project don't add the app as a target dependency
- [x] update `ProjectGeneratorTest`


ryu2:
cc: @yiding @Coneko --- can you guys review?

PR #1351, opened 2017-05-19
Bump kotlin versions to 1.1.2-2.

runningcode:


dsyang:
Looks ok to me. CI failures look unrelated

runningcode:
BTW, is there a better way to update these libraries other than a manual find and replace?

dsyang:
In retrospect, I don't think this is the right thing to do. 

third-party/java/kotlin is not used by buck itself but by another third party lib third-party/java/mapdb.

Looking at the mapdb BUCK file, the version we have in third-party/java/mapdb is 3.0.3. For that version, kotlin is pinned at 1.0.6 (https://github.com/jankotek/mapdb/commit/2fcb185f36ad60f115837538bb691a755845c39b#diff-600376dffeb79835ede4a0b285078036)

We probably shouldn't update this until mapdb is updated. cc @bertmaher since he added the mapdb dependency

dsyang:
Closing because of the above comment re: mapdb.  Before doing so though, I'd like to test something out with github's online editor if that's ok @runningcode

dsyang:
nvm experiment didn't work as I thought it would

Coneko:
Mapdb is being removed so we don't need to worry about that.

PR #1357, opened 2017-05-25
Allow not setting `exe` for worker_tools

mikekap:
This assumes the arguments will come via `args`, so you can have the first arg be e.g. `java`.

The only downside is folks that were previously abusing the `args` to pass multiple arguments in one entry will have their arguments properly quoted. Similarly for the react native worker rules w.r.t packager arguments.


mikekap:
Not sure what those failures are about, but they seem unrelated.

mikekap:
ping

PR #1364, opened 2017-06-02
Fix nailgun test on Windows on AppVeyor

LegNeato:
This test failed on the commit that introduced the test (https://github.com/facebook/buck/commit/5a79d2aeb14c87c1de8b8d17aee8f90009cfe76d) but passed on a subsequent commit (https://ci.appveyor.com/project/Facebook/buck/build/3554). It appears waiting for 3 seconds is not enough on AppVeyor (maybe due to noisy neighbors?), so let's double it.

### DIFF
--- a/third-party/nailgun/pynailgun/test_ng.py
+++ b/third-party/nailgun/pynailgun/test_ng.py
@@ -92,7 +92,7 @@ class TestNailgunConnection(unittest.TestCase):
         self.assertIsNone(self.ng_server_process.poll())
 
         # Give Java some time to create the listening socket.
-        for i in range(0, 300):
+        for i in range(0, 600):
             if not transport_exists(self.transport_file):
                 time.sleep(0.01)
 


LegNeato:
Hmmm, don't merge this as it didn't fix it. Something else is going on...

marcinkosiba:
I'm trying a bigger number in https://github.com/facebook/buck/pull/1364

maybe @ilya-klyuchnikov has some insight into the failing test

ilya-klyuchnikov:
I tracked the root cause: nailgun-server doesn't start correctly on appveyor (via named pipes). However, we never experience this when testing internally. A further investigation is needed.

ilya-klyuchnikov:
Update: the root cause is that appveyor has its own (incompatible with what we use) jna libraries (tracked by this run of nailgun: https://ci.appveyor.com/project/ilya-klyuchnikov/nailgun/build/b10/artifacts). 

This is the simple fix:  https://github.com/ilya-klyuchnikov/buck/commit/f1f7fa50af44ccddfb01ae48616be6ea18361c5d (and corresponding CI: https://ci.appveyor.com/project/ilya-klyuchnikov/buck-07nvh/build/6).

I will land next week.

sdwilsh:
Wow, that's subtle!  Good find!

LegNeato:
Nice!

ilya-klyuchnikov:
landed: f1f7fa50

PR #1365, opened 2017-06-02
Increase JVM startup timeout in nailgun python test.

marcinkosiba:
Let's try a more drastic number.

Closes: #1364


marcinkosiba:
60sec timeout doesn't help :(

PR #1380, opened 2017-06-13
Don't throw exception when we can't find meta-inf.

runningcode:
Gradle builds were missing this file and so the buck installer would
throw an exception when trying to install a buck build over a gradle
build. In this case it means that the apks are different so always
re-install.


runningcode:
cc @dreiss You mentioned I should change the native agent to match this behavior. It looks to me that the native agent doesn't throw an exception if it fails to find the manifest digest. But my C skills aren't very sharp so perhaps I was reading the code wrong.

dreiss:
It actually returns a failure code.

Wait, was the gradle build missing that file, or just missing the line in the file?

If it was just missing the line, did it have some other sort of manifest digest?

dreiss:
And if it was missing the file, did it have some other *.SF file in META-INF?

dreiss:
And if not, how did you install an app with no signature?

yiding:
@runningcode I think we are await your response regarding the aforementioned questions.

styurin:
Please reopen if you still have this problem.

runningcode:
Hi, sorry for not responding. I took a break from Android, but I'm back!

Anyways, I'm now working at a new company and working to transition gradle to buck and we encountered the same problem.

There's a `CERT.SF` and a `MANIFEST.MF` file.

The `MANIFEST.MF` file is the one that has a long list of contents like so:
```
Manifest-Version: 1.0
Built-By: Generated-by-ADT
Created-By: Android Gradle 3.1.0-alpha08

Name: AndroidManifest.xml
SHA-256-Digest: Un6YY/J2B4f4GI4xhC0m1a/EGClaOEdg7+ZZOLdcBz4=

Name: META-INF/android.arch.lifecycle_runtime.version
SHA-256-Digest: 9eXJU0UyhA/NRbsALmnthcYduidTRQ4mtEi33xSB4k0=

Name: META-INF/app_debug.kotlin_module
SHA-256-Digest: zwWKnIwxvaA649IRJxgIS4/epZhuAYgYvtDp5LcmSAk=
```

But it looks like buck is looking for the `SHA-256-Digest` in a file ending with `*.SF` which in this case is the `CERT.SF` which doesn't contain these `SHA-256-Digest` lines.

@dreiss How should we proceed?

PR #1383, opened 2017-06-15
Fix aapt2_compile across cells

seanabraham:
When referencing resources in a different cell, aapt2 compile used to fail with "No such file or directory" because it would search within the `buck-out` folder of the current cell. This fixes that with the same methodology seen in https://github.com/facebook/buck/commit/e803c7f082

I'm not sure if this is the proper way to do this but my testing showed that it worked as expected with this change.

cc @dreiss


dreiss:
Can you make it just pass the absolute path into the step's constructor?

seanabraham:
Not sure what's up with the test failures, they passed locally. Doesn't look like I have permissions to re-run so I'll rebase and push again

PR #1387, opened 2017-06-20
Make kotlinc always compile against full jars instead of abi

### LINKED ISSUE #1386: Kotlin rules cannot be compiled against abi jars
f43e2d20913443d7047fd4219e7559d9108a802a introduced a regression that puts abi jars on the kotlinc classpath instead of full jars. `kotlinc` does not like this and errors out on compilation when using methods from the kotlin standard library with the following stacktrace

```
Caused by: java.lang.IllegalStateException: Couldn't obtain compiled function body for public inline fun <T> mutableListOf(): kotlin.collections.MutableList<T> defined in kotlin.collections[DeserializedSimpleFunctionDescriptor@727380c2]
	at org.jetbrains.kotlin.codegen.inline.InlineCodegen$1.invoke(InlineCodegen.java:280)
	at org.jetbrains.kotlin.codegen.inline.InlineCodegen$1.invoke(InlineCodegen.java:275)
	at org.jetbrains.kotlin.codegen.inline.InlineCacheKt.getOrPut(InlineCache.kt:34)
	at org.jetbrains.kotlin.codegen.inline.InlineCodegen.createMethodNode(InlineCodegen.java:274)
	at org.jetbrains.kotlin.codegen.inline.InlineCodegen.genCallInner(InlineCodegen.java:180)
	... 69 more

    val words = mutableListOf<String>()
                ^
```

A repro case is available here: https://github.com/uber/okbuck/tree/gk_kotlin_abi
Command to run: `./buckw build //libraries/kotlinlibrary/...`
After removing the `compile_against_abis = true` change in `.buckconfig` in the repro, everything works as expected

Seems like this option is not safe for kotlin modules yet. The kotlin compiler is actually introspecting the classes to extract out and inline the method into the compiled code. If we compile against the abi jars, the compiled method body will not be found. I think as part of refactoring and reusing base jvm descriptions etc for f43e2d20913443d7047fd4219e7559d9108a802a , it inadvertently introduced kotlinc to be able to pickup abi jars for compilation

cc @dsyang @runningcode @brettwooldridge

kageiit:
Workaround for #1386

### DIFF
--- a/src/com/facebook/buck/jvm/kotlin/KotlinLibraryBuilder.java
+++ b/src/com/facebook/buck/jvm/kotlin/KotlinLibraryBuilder.java
@@ -43,6 +43,7 @@ public class KotlinLibraryBuilder extends DefaultJavaLibraryBuilder {
       JavaBuckConfig javaBuckConfig) {
     super(targetGraph, params, buildRuleResolver, cellRoots, javaBuckConfig);
     this.kotlinBuckConfig = kotlinBuckConfig;
+    setCompileAgainstAbis(false);
   }
 
   @Override


jkeljo:
I'm OK with this as a workaround, but let's not mark it as a fix...the real fix (at least for this specific issue) is to make the ABI generator aware of Kotlin's metadata for inline functions, and include the bodies of same. I'd like to keep the issue open until that can be done.

kageiit:
I marked it as a workaround. Seems like any code written with the `inline` modifier needs to be dealt with by including the body of the bytecode into the abi jar for kotlinc to work correctly

PR #1393, opened 2017-06-21
Fix cache key inconsistency in kotlin library rules

### LINKED ISSUE #1392: kotlin_library rules produce different cache/rulekeys on different machines with same settings
When using in memory kotlin compilation, absolute paths are being appended to the rulekey

https://github.com/facebook/buck/blob/00223c7b1e7c5c1544d1d4fc838b87745ffc4ffc/src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java#L68
https://github.com/facebook/buck/blob/00223c7b1e7c5c1544d1d4fc838b87745ffc4ffc/src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java#L165

A sourcepath should be used instead like how in memory java compiler rules work
https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/JarBackedJavac.java#L47
https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/JarBackedJavac.java#L69

This causes the rulekeys to be different on two different checkouts of the same codebase in different directories which basically completely breaks the network cache

kageiit:
Fixes #1392

Verified the rulekeys and target hashes are the same after this fix on two copies of the same codebase

Also verified that kotlin library rules can still build

### DIFF
--- a/src/com/facebook/buck/cli/BuckConfig.java
+++ b/src/com/facebook/buck/cli/BuckConfig.java
@@ -355,6 +355,18 @@ public class BuckConfig implements ConfigPathGetter {
   }
 
   /**
+   * @return a {@link SourcePath} identified by a {@link Path}.
+   */
+  public SourcePath getSourcePath(Path path) {
+    if (path == null) {
+      return null;
+    }
+    return new PathSourcePath(
+            projectFilesystem,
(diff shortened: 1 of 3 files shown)


brettwooldridge:
@kageiit LGTM

PR #1395, opened 2017-06-23
Fix groovy cross compile with empty options

kageiit:
When there are annotation processor deps but no annotation processors, the groovy cross compile step fails with

```
stderr: org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
Compile error during compilation with javac.
warning: [options] bootstrap class path not set in conjunction with -source 1.7
error: Annotation processor '' not found
1 error
1 warning
```

This changes the behavior by only adding a cross compile option if it is non empty


kageiit:
cc @andrewjcg @dsyang

PR #1405, opened 2017-06-28
haskell: add GHC to travis

andrewjcg:
Summary:
So Haskell tests actually run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ addons:
     sources:
       # For newer gcc than precise supports out of the box.
       - ubuntu-toolchain-r-test
+      # For newer ghc than precise supports out of the box.
+      - hvr-ghc
     packages:
       # Travis is on 64bit and there will be a cryptic aapt error w/o these libs.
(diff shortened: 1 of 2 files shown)


sdwilsh:
It doesn't look like this is passing travis.  Do we need to make changes to Buck?

andrewjcg:
Looks like the default ghc is 7.4, which might be too old.  Trying again with 7.6.

andrewjcg:
Used a PPA (https://launchpad.net/~hvr/+archive/ubuntu/ghc) which provides a new GHC (as per the pandoc example, https://github.com/jgm/pandoc/blob/master/.travis.yml, linked from https://docs.travis-ci.com/user/languages/haskell/).

andrewjcg:
If the tests pass, I'll update again to use GHC-8.

andrewjcg:
Ok, looks like the haskell tests run and pass now.

styurin:
Looks like it's working, do you want to merge it?

PR #1410, opened 2017-07-03
[IntelliJ] Handle resources_root in project generation

zhan-xiong:
IntelliJ project generation right now doesn't generate resource folders for Java code, instead relying directly on the generation of source folders. This results in resources being put in the wrong places if resources_root is different from the Java src_roots option. This change introduces a new type of IjFolder (JavaResourceFolder) which tracks resources_root if it is set and sets the relativeOutputPath of the IntelliJ folders to try and handle this correctly. 

Note that it probably can't be fully handled correctly if there are source files and resource files mixed in the same directories, because IntelliJ's granularity only goes to the directory level, while buck can set them at the file level.

I'm not entirely sure if this is the best approach to handling this, would appreciate any suggestions


kageiit:
@zhan-xiong Any updates on this?

zhan-xiong:
Having some delays with getting the CLA signed :( will update when it's settled (probably a few more days)

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -108,6 +111,36 @@ private void addDepsAndFolder(
       ModuleBuildContext context) {
     ImmutableMultimap<Path, Path> foldersToInputsIndex =
         getSourceFoldersToInputsIndex(targetNode.getInputs());
+
+    if (targetNode.getConstructorArg() instanceof JavaLibraryDescriptionArg) {
```
styurin: This needs to be in `JavaLibraryModuleRule`

zhan-xiong:
CLA should be accepted now, I think

zhan-xiong:
@styurin any updates about this?

styurin:
You need to reformat java files using google-java-format.

src/com/facebook/buck/ide/intellij/model/folders/AbstractIjSourceFolder.java
```diff
@@ -32,7 +32,12 @@
 
   public abstract boolean getIsTestSource();
```
styurin: I suggest simplifying design and be consistent with java source folders and java test folders.

You can rename this method to `getIsTesFolder` and use it together with `getIsResourceFolder` to detect the folder type. You would need to add `JavaTestResourceFolder` class to be consistent with `TestFolder` and `SourceFolder`.

src/com/facebook/buck/ide/intellij/model/folders/AbstractIjSourceFolder.java
```diff
@@ -32,7 +32,12 @@
...
 
-  public abstract boolean getIsAndroidResources();
+  public abstract boolean getIsResourceFolder();
+
+  public abstract ResourceFolderType getResourceFolderType();
```
styurin: Is this method used anywhere? Looks like it's not.

src/com/facebook/buck/ide/intellij/model/folders/AbstractIjSourceFolder.java
```diff
@@ -32,7 +32,12 @@
...
+
+  @Nullable
+  public abstract Path getRelativeOutputPath();
```
styurin: This method is also not used.

src/com/facebook/buck/ide/intellij/model/folders/AbstractIjSourceFolder.java
zhan-xiong: The fields are used by the template (ij-module.st); since IjSourceFolder is generated, I need the getter here to create the field.

src/com/facebook/buck/ide/intellij/model/folders/AbstractIjSourceFolder.java
zhan-xiong: Ditto above

src/com/facebook/buck/ide/intellij/model/folders/AbstractIjSourceFolder.java
zhan-xiong: Is there a way to make it more explicit?

styurin: oh, I missed that, probably not, may be add a comment?

styurin:
Please, reformat files with google-java-format: https://github.com/facebook/buck/blob/master/CONTRIBUTING.md#code-style

zhan-xiong:
Hmm thought I did o.o reformatting now

src/com/facebook/buck/ide/intellij/templates/ij-module.st
```diff
@@ -58,15 +58,21 @@
...
 }%%contentRoots:{contentRoot |
     <content url="%contentRoot.url%">
 %contentRoot.folders:{folder |
       <%folder.type% url="%folder.url%" %\\%
         isTestSource="%folder.isTestSource%" %\\%
         %if(folder.packagePrefix)%packagePrefix="%folder.packagePrefix%" %endif%%\\%
-        %if(folder.isAndroidResources)%type="java-resource" %endif%/>
+        %if(folder.isResourceFolder)%%\\%
+          type="%folder.resourceFolderType%" %\\%
+          %if(folder.relativeOutputPath)%relativeOutputPath="%folder.relativeOutputPath%" %endif%%\\%
```
styurin: Do we really need this relativeOutputPath? From https://www.jetbrains.com/help/idea/configuring-content-roots.html:

>  During the build process, by default, the resources are copied into the root of the compilation output folder. If necessary, you can specify a different folder within that output folder.

It seems that this is only useful if we use IntelliJ build. `buck project` can generate a project that works with IntelliJ, but it doesn't guarantee such compatibility. Do you have a legitimate use case to provide this option?

zhan-xiong: Being able to build and run in IntelliJ improves our iteration speed. Specifically, a lot of built-in features (e.g. automatically attaching debugger, more informative testing UI, easier-to-customize run/debug configurations) in IntelliJ are better for our workflows than command-line buck.

The IntelliJ buck plugin is an okay alternative, but for now doesn't have enough features compared to native IntelliJ for our liking (though I'm also adding a couple of changes there, hopefully we can start using it soon), so being able to support this natively in IntelliJ in addition would be really nice for us.

src/com/facebook/buck/ide/intellij/ModuleBuildContext.java
```diff
@@ -153,23 +155,43 @@ public void addSourceFolder(IjFolder folder) {
...
+      Path resourcesRoot =
+          (to instanceof JavaResourceFolder)
+              ? ((JavaResourceFolder) to).getResourcesRoot()
+              : ((JavaTestResourceFolder) to).getResourcesRoot();
+      return new JavaResourceFolder(to.getPath(), IjFolder.combineInputs(from, to), resourcesRoot);
+    }
 
-    return from.merge(to);
+    // If we're not sure what to do otherwise, use SourceFolder.
+    return new SourceFolder(
```
styurin: what about this comment?

styurin:
Please, add integration tests

src/com/facebook/buck/ide/intellij/model/folders/ResourceFolderType.java
```diff
@@ -0,0 +1,65 @@
...
+    if (this.equals(JAVA_RESOURCE)) {
+      return JavaResourceFolder.getFactoryWithResourcesRoot(resourcesRoot);
+    } else if (this.equals(JAVA_TEST_RESOURCE)) {
+      return JavaTestResourceFolder.getFactoryWithResourcesRoot(resourcesRoot);
+    } else {
+      throw new IllegalArgumentException("Invalid resource type");
+    }
+  }
+
+  public boolean isIjFolderInstance(IjFolder folder) {
```
styurin: the same for this method.

src/com/facebook/buck/ide/intellij/model/folders/ResourceFolderType.java
```diff
@@ -0,0 +1,65 @@
...
+    if (this.equals(JAVA_RESOURCE)) {
+      return folder instanceof JavaResourceFolder;
+    } else if (this.equals(JAVA_TEST_RESOURCE)) {
+      return folder instanceof JavaTestResourceFolder;
+    } else {
+      return false;
+    }
+  }
+
+  public Path getResourcesRootFromFolder(IjFolder folder) {
```
styurin: The same for this method.

src/com/facebook/buck/ide/intellij/model/folders/ResourceFolderType.java
```diff
@@ -0,0 +1,65 @@
...
+  ResourceFolderType(String resourceType) {
+    this.resourceType = resourceType;
+  }
+
+  @Override
+  public String toString() {
+    return resourceType;
+  }
+
+  public IJFolderFactory getFactoryWithResourcesRoot(Path resourcesRoot) {
```
styurin: JavaResourceFolder

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,75 @@
...
+  public Path getResourcesRoot() {
+    return resourcesRoot;
+  }
+
+  @Override
+  protected IJFolderFactory getFactory() {
+    return getFactoryWithResourcesRoot(resourcesRoot);
+  }
+
+  public IJFolderFactory getFactoryWithSameResourcesRoot() {
```
styurin: this is not used

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,75 @@
...
+    return getFactory();
+  }
+
+  public static final IJFolderFactory getFactoryWithResourcesRoot(Path resourcesRoot) {
```
styurin: no need for final, it's static. Also, this needs to have `@Nullable` because `resourcesRoot` can be a null.

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,75 @@
...
+package com.facebook.buck.ide.intellij.model.folders;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSortedSet;
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.Optional;
+
+public class JavaTestResourceFolder extends InclusiveFolder {
+  private Path resourcesRoot;
```
styurin: This needs to have `@Nullable`.

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,75 @@
...
+
+  public JavaTestResourceFolder(Path path, ImmutableSortedSet<Path> inputs, Path resourcesRoot) {
+    super(path, false, inputs);
+    Preconditions.checkNotNull(resourcesRoot);
```
styurin: this variable can be null, right?

src/com/facebook/buck/ide/intellij/model/folders/ResourceFolderType.java
zhan-xiong: I'm not exactly sure what you mean here?

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
zhan-xiong: Done

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
zhan-xiong: Removed

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
zhan-xiong: Removed check

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
zhan-xiong: Done

src/com/facebook/buck/ide/intellij/ModuleBuildContext.java
zhan-xiong: Changed, though actually I think regardless of whether it's set to SourceFolder or TestFolder in this case, it's not going to work correctly, but TestFolder is probably a better attempt in that case

src/com/facebook/buck/ide/intellij/ModuleBuildContext.java
```diff
@@ -153,23 +155,43 @@ public void addSourceFolder(IjFolder folder) {
...
 
-  private IjFolder mergeAllowingTestToBePromotedToSource(IjFolder from, IjFolder to) {
-    if ((from instanceof TestFolder && to instanceof SourceFolder)
-        || (to instanceof TestFolder && from instanceof SourceFolder)) {
-      return new SourceFolder(
-          to.getPath(),
-          from.getWantsPackagePrefix() || to.getWantsPackagePrefix(),
-          IjFolder.combineInputs(from, to));
+  /**
+   * Merges the two folders according to the following rules: - Folders of the same type merge
```
zhan-xiong: Done

src/com/facebook/buck/ide/intellij/model/folders/ResourceFolderType.java
styurin: Weird, I had different comment here.

You can declare an abstract method here and then override it in each enum object. This would remove the need of IllegalArgumentException

styurin: Note sure if you saw my comment, but I mean something like this:

```
public interface HasResourceRoot {
  public Path getResourcesRoot();
}

public class JavaResourceFolder extends InclusiveFolder implements HasResourceRoot {
  ..
}

public class JavaTestResourceFolder extends InclusiveFolder implements HasResourceRoot {
  ..
}

public enum ResourceFolderType {
  JAVA_RESOURCE("java-resource", JavaResourceFolder.class) {
    public IJFolderFactory getFactoryWithResourcesRoot(Path resourcesRoot) {
      return JavaResourceFolder.getFactoryWithResourcesRoot(resourcesRoot);
    }
  },
  JAVA_TEST_RESOURCE("java-test-resource", JavaTestResourceFolder.class) {
    public IJFolderFactory getFactoryWithResourcesRoot(Path resourcesRoot) {
      return JavaTestResourceFolder.getFactoryWithResourcesRoot(resourcesRoot);
    }
  };

  private final String resourceType;
  private final Class<? extends HasResourceRoot> folderClass;

  ResourceFolderType(String resourceType, Class<? extends HasResourceRoot> folderClass) {
    this.resourceType = resourceType;
    this.folderType = folderClass;
  }

  @Override
  public String toString() {
    return resourceType;
  }

  public abstract IJFolderFactory getFactoryWithResourcesRoot(Path resourcesRoot);

  public abstact boolean isIjFolderInstance(IjFolder folder) {
    return folderClass.isInstance(folder);
  }

  public Path getResourcesRootFromFolder(IjFolder folder) {
    return folderClass.cast(folder).getResourcesRoot();
  }
}

```

zhan-xiong: Sorry for the delay, was traveling for the past few days; will update today

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,71 @@
...
+package com.facebook.buck.ide.intellij.model.folders;
+
+import com.google.common.collect.ImmutableSortedSet;
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+public class JavaTestResourceFolder extends InclusiveFolder implements HasResourcesRoot {
+  @Nullable private Path resourcesRoot;
```
styurin: final

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public JavaTestResourceFolder(Path path, ImmutableSortedSet<Path> inputs, Path resourcesRoot) {
```
styurin: `@Nullable` Path resourcesRoot

src/com/facebook/buck/ide/intellij/model/folders/JavaTestResourceFolder.java
```diff
@@ -0,0 +1,71 @@
...
+  public Optional<Path> getRelativeOutputPath() {
+    if (resourcesRoot == null || !getPath().startsWith(resourcesRoot)) {
+      return Optional.empty();
+    } else {
+      return Optional.of(resourcesRoot.relativize(getPath()));
+    }
+  }
+
+  @Override
+  public Path getResourcesRoot() {
```
styurin: `@Nullable`

src/com/facebook/buck/ide/intellij/model/folders/HasResourcesRoot.java
```diff
@@ -0,0 +1,23 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.ide.intellij.model.folders;
+
+import java.nio.file.Path;
+
+public interface HasResourcesRoot {
```
styurin: On the second thought it's worth moving common functionality that deals with resources root to this class:

```
public class ResourceFolder {
  @Nullable private Path resourcesRoot;

  public Optional<Path> getRelativeOutputPath() {
    ...
  }
}
```

Then you can just derive `JavaResourceFolder` and `JavaTestResourceFolder` from this class.

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -118,18 +134,87 @@ private void addDepsAndFolder(
...
+  // the default folders generated should work. On the other hand, if there is a resources_root,
+  // then for resources under this root, we need to create java-resource folders with the
+  // correct relativeOutputPath set. We also return a filter that removes the resources that we've
+  // added.
+  protected Predicate<Map.Entry<Path, Path>> addResourcesAndGetFilter(
+      ResourceFolderType resourceFolderType,
+      Collection<SourcePath> resources,
+      Path resourcesRoot,
+      ModuleBuildContext context) {
+    ImmutableSet<Path> resourcePaths =
```
styurin:
```
    ImmutableSet<Path> resourcePaths =
        resources
            .stream()
            .filter(PathSourcePath.class::isInstance)
            .map(PathSourcePath.class::cast)
            .map(PathSourcePath::getRelativePath)
            .filter(path -> path.startsWith(resourcesRoot))
            .collect(MoreCollectors.toImmutableSet());
```

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -118,18 +134,87 @@ private void addDepsAndFolder(
...
+            .map(
+                sourcePath ->
+                    (sourcePath instanceof PathSourcePath)
+                        ? ((PathSourcePath) sourcePath).getRelativePath()
+                        : null)
+            .filter(Objects::nonNull)
+            .filter(path -> path.startsWith(resourcesRoot))
+            .collect(MoreCollectors.toImmutableSet());
+
+    addSourceFolders(
```
styurin: This is incorrect, it's adding resource folders, not source folders. I recommend adding a separate interface:

```
public interface ResourceFolderFactory {
  IjFolder create(Path path, Path resourceRoot, ImmutableSortedSet<Path> inputs);
}
```

Then add `addResourceFolders` method:

```
addResourceFolders(
  resourceFolderType.getFactory(),
  getSourceFoldersToInputsIndex(resourcePaths),
  resourcesRoot,
  context);
```

zhan-xiong:
I've made the `ResourceFolderFactory` change but I'm not entirely sure if it makes things neater, since `ResourceFolder`s have to eventually inherit from `IjFolder`, which requires an `IjFolderFactory getFactory()` method for merging. This means we have two kinds of factory for each resource folder, which seems confusing.
Also, I think that source folders in the context of the `addSourceFolders` etc methods actually refer to `IjFolder`s (which are given as <sourceFolder> in the module xml), perhaps keeping the old factories and renaming the methods to `addIjFolders` might make more sense?

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -100,14 +106,39 @@ protected void addSourceFolders(
...
+    for (Map.Entry<Path, Collection<Path>> entry : foldersToInputsIndex.asMap().entrySet()) {
+      context.addSourceFolder(
+          factory.create(
+              entry.getKey(),
+              resourcesRoot,
+              ImmutableSortedSet.copyOf(Ordering.natural(), entry.getValue())));
+    }
+  }
+
+  private void addDepsAndFolderWithFolderInputIndexFilter(
```
styurin: Better name would be `addDepsAndFoldersWithFilteringIfNeeded` or just `addDepsAndFoldersWithFiltering`

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -118,18 +149,84 @@ private void addDepsAndFolder(
...
+        folderInputIndexFilter);
+  }
+
   protected void addDepsAndTestSources(
       TargetNode<T, ?> targetNode, boolean wantsPackagePrefix, ModuleBuildContext context) {
     addDepsAndFolder(
         TestFolder.FACTORY, DependencyType.TEST, targetNode, wantsPackagePrefix, context);
   }
 
+  // If there is no resources_root, then we use the java src_roots option from .buckconfig, so
```
styurin: I don't see how this comment is related to the code, specifically the part about `src_roots`

zhan-xiong: Rewrote the comment to give a bit more context for why the method exists and has the given return type. If it still isn't too helpful, then I'll just remove it?

styurin:
I thought it would be faster for us if I show you my suggestions in code, here it is: https://github.com/styurin/buck/commit/c5c84de4e66275ea9a9b3f0d6d1c49d92744b0aa

k21:
The tests are now failing: https://travis-ci.org/facebook/buck/jobs/266322233#L3651. It looks like the `JavaResourceFolder` will either need to implement `merge(IjFolder)` without using `getFactory()` or it will have to overload `canMergeWith(IjFolder)` so that the root simplifier does not try to merge it. I am not completely familiar with how this works, so I am not sure which option is the right one.

k21:
Please apply https://github.com/k21/buck/commit/e96c7f234d3046acb3027256806c2017c82407af to address Infer warnings.

src/com/facebook/buck/ide/intellij/IjSourceRootSimplifier.java
```diff
@@ -330,6 +340,60 @@ private FolderTypeWithPackageInfo findBestFolderType(ImmutableCollection<IjFolde
...
+
+      if (childrenToMerge.isEmpty()) {
+        return Optional.empty();
+      }
+
+      final Path resourcesRoot =
+          ijResourceFolderType.getResourcesRootFromFolder(childrenToMerge.get(0));
+
+      // If merging would recurse above resources_root, don't merge.
+      if (!currentPath.startsWith(resourcesRoot)) {
```
k21: It looks like `resourcesRoot` can be `null`, which would make this throw a `NullPointerException`, please address.

styurin:
Sorry for the delay. I was trying this fix and it seems it doesn't handle some situations, for example, where there is no `resources_root`. Here's my test: https://github.com/styurin/buck/commit/4f300dc3623327156f07f271e70a9d2eaf9febdc

zhan-xiong:
Yeah, I realize that I'm wrong about not needing to handle relativeOutputPath if the resources root isn't set, since source folders can have a packagePrefix that will still change where resources end up away from the regular one. Will put up a fix soon.

zhan-xiong:
Here's a fix that tries to detect the effective resource root for each file and uses that to compute the relative output path. A few questions though:
1) I had a couple of bugs where calling Path::startsWith() on an empty path returns false, which seems unintuitive to me. Is there a better way of handling this rather than checking against an empty path in addition to using startsWith()?
2) After thinking about it some more, it doesn't make sense for there not to be a resources root ever (at worst, buck uses the project root path as the "resources root"), so I removed Nullable from it. I'm not sure if I'm missing any corner cases though

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -118,18 +151,110 @@ private void addDepsAndFolder(
...
+      ImmutableSet<Path> resourcePaths) {
+    final JavaPackageFinder packageFinder =
+        projectConfig.getJavaBuckConfig().createDefaultJavaPackageFinder();
+    return resourcePaths
+        .stream()
+        .collect(
+            MoreCollectors.toImmutableMultimap(
+                path ->
+                    MorePaths.stripCommonSuffix(
+                            path.getParent(), packageFinder.findJavaPackageFolder(path))
```
styurin: I doubt `src_root` is used to detect the relative location of resources in the final jar. That's the purpose of `resources_root`, right? If `resources_root` not present the resources are placed in the same location in a jar as they are relatively to target's base path. I think this logic should be reproduced here, otherwise IntelliJ and buck will have resources in different locations.

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
```diff
@@ -118,18 +151,110 @@ private void addDepsAndFolder(
...
+                        .getFirst(),
+                Function.identity()));
+  }
+
+  // This function should only be called if resources_root is present. If there is no
```
styurin: you need to update this comment if my assumption is valid

src/com/facebook/buck/ide/intellij/BaseIjModuleRule.java
zhan-xiong: Actually, based on https://buckbuild.com/rule/java_library.html (and on my own testing), if resources_root is not set, then it basically uses the corresponding src_root as a resources_root

PR #1411, opened 2017-07-05
[IntelliJ plugin] Test run configuration producer

zhan-xiong:
Currently, creating a buck test configuration must be done manually. This PR handles automatically filling out the test configuration when right-clicking on a test method or test class. This allows "Run" and "Debug" from right-clicking on the test method/class (as well as shortcuts like ctrl-shift-R) to work with buck.


zhan-xiong:
Just realized part of it is implemented [here](https://github.com/facebook/buck/commit/76b93f93c125c5edeecc2b0538bdb39f8405057d) two days ago, will rework to use some of the features there

kageiit:
@zhan-xiong Any updates on this?

zhan-xiong:
Having some delays with getting the CLA signed :( will update when it's settled (probably a few more days)

zhan-xiong:
CLA should be accepted now, I think

src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedTestAction.java
```diff
@@ -41,11 +40,10 @@
 import com.intellij.psi.PsiClass;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiMethod;
-import java.util.List;
 import java.util.Optional;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import javax.swing.Icon;
+import javax.swing.*;
```
styurin: Please, replace this wildcard with individual imports

styurin:
Also, could update the description? This PR doesn't add support for creating configurations from current file, right?

styurin:
By the way, travis is failing because of your change, you need to run `ant lint`

src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedTestAction.java
zhan-xiong: Done

zhan-xiong:
* Fixed ant lint
* The PR can save configurations from the current file when right-clicking on a test method/class, updated description to be clearer

styurin:
Can you please rebase and format Java files using google-java-format? See https://github.com/facebook/buck/blob/master/CONTRIBUTING.md#code-style for instructions

src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/configurations/TestConfigurationUtil.java
```diff
@@ -0,0 +1,116 @@
...
+                  testConfiguration.data.target = strings.get(0);
+                  testConfiguration.data.testSelectors = testSelectors.orElse("");
+                }
+                if (testConfigurationProcessor != null) {
+                  testConfigurationProcessor.apply(testConfiguration);
+                }
+                return null;
+              }
+            });
+    handler.command().addParameter("kind('java_test', owner(" + containingFile.getPath() + "))");
```
styurin: Why did you change it to filter out non-java_test targets? What if it's robolectric_test or something else?

zhan-xiong: Good point, would it be fine to change it to ".*_test"? Or are there other kinds that it might not be a good idea to filter out?

styurin: Yeah, I think that should work. Ideally we need to have a way to indicate that we need only test targets, since there is no way right now, let's use `*_java`.

styurin:
1. You need to update the title, it's not only java_test

2. I was testing this locally and noticed that when a cursor is in a method name, Buck tab shows up every time there is an updated even if I opened a different tab. Also, I noticed that right-mouse clicking has a very long and noticeable delay (in order of 2-3 seconds). It's probably calling `buck query` when it's not needed. I don't notice such a delay when I use line actions.

zhan-xiong:
1. Filtered out non-test files from the test config producer. Still kind of slow (~1-2 seconds lag for me) when right-clicking in test files (I guess buck query just isn't fast enough..?). I can't think of a good way around it; line actions isn't affected because it runs in the background, so all that happens is that it takes a good number of seconds on startup to generate the button.

2. Not sure if it's caused by this change, but the first run of a buck test when starting up (regardless of the method used) doesn't seem to process the buck output correctly, so doesn't notice the number of tests that have actually passed.

I think I'm going to need to do a lot more testing for this change first, but any other comments would be welcome.

PR #1412, opened 2017-07-10
Add compiler output property to JVM-language module IML files

jbarr21:
This diff sets `inherit-compiler-output` property to false and sets the compiler output path to the default Java classes dir for JVM-language modules so that IntelliJ's Android Support's Layout Preview knows where to find module classes.

### DIFF
--- a/src/com/facebook/buck/ide/intellij/DefaultIjModuleFactoryResolver.java
+++ b/src/com/facebook/buck/ide/intellij/DefaultIjModuleFactoryResolver.java
@@ -24,6 +24,7 @@ import com.facebook.buck.android.DummyRDotJava;
 import com.facebook.buck.ide.intellij.model.IjModuleFactoryResolver;
 import com.facebook.buck.io.ProjectFilesystem;
 import com.facebook.buck.jvm.java.AnnotationProcessingParams;
+import com.facebook.buck.jvm.java.DefaultJavaLibrary;
 import com.facebook.buck.jvm.java.JavaLibraryRules;
 import com.facebook.buck.jvm.java.JvmLibraryArg;
 import com.facebook.buck.model.BuildTarget;
@@ -121,6 +122,13 @@ class DefaultIjModuleFactoryResolver implements IjModuleFactoryResolver {
     }
 
     return JavaLibraryRules.getAnnotationPath(projectFilesystem, targetNode.getBuildTarget());
+  }
+
+  @Override
+  public Optional<Path> getCompilerOutputPath(TargetNode<? extends JvmLibraryArg, ?> targetNode) {
+    BuildTarget buildTarget = targetNode.getBuildTarget();
+    Path compilerOutputPath = DefaultJavaLibrary.getClassesDir(buildTarget, projectFilesystem);
+    return Optional.ofNullable(compilerOutputPath);
   }
 
   private Path getRelativePathAndRecordRule(SourcePath sourcePath) {
--- a/src/com/facebook/buck/ide/intellij/DefaultIjModuleFactoryResolver.java
+++ b/src/com/facebook/buck/ide/intellij/DefaultIjModuleFactoryResolver.java
@@ -128,7 +128,7 @@ class DefaultIjModuleFactoryResolver implements IjModuleFactoryResolver {
   public Optional<Path> getCompilerOutputPath(TargetNode<? extends JvmLibraryArg, ?> targetNode) {
     BuildTarget buildTarget = targetNode.getBuildTarget();
     Path compilerOutputPath = DefaultJavaLibrary.getClassesDir(buildTarget, projectFilesystem);
-    return Optional.ofNullable(compilerOutputPath);
+    return Optional.of(compilerOutputPath);
   }
 
   private Path getRelativePathAndRecordRule(SourcePath sourcePath) {
--- a/src/com/facebook/buck/ide/intellij/IjProject.java
+++ b/src/com/facebook/buck/ide/intellij/IjProject.java
@@ -104,7 +104,11 @@ public class IjProject {
             javaPackageFinder);
     IjProjectTemplateDataPreparer templateDataPreparer =
         new IjProjectTemplateDataPreparer(
-            parsingJavaPackageFinder, moduleGraph, projectFilesystem, projectConfig);
+            parsingJavaPackageFinder,
+            moduleGraph,
+            projectFilesystem,
+            projectConfig,
+            buildRuleResolver);
     IjProjectWriter writer =
         new IjProjectWriter(templateDataPreparer, projectConfig, projectFilesystem);
 
--- a/src/com/facebook/buck/ide/intellij/IjProject.java
+++ b/src/com/facebook/buck/ide/intellij/IjProject.java
@@ -107,8 +107,7 @@ public class IjProject {
             parsingJavaPackageFinder,
             moduleGraph,
             projectFilesystem,
-            projectConfig,
-            buildRuleResolver);
+            projectConfig);
     IjProjectWriter writer =
         new IjProjectWriter(templateDataPreparer, projectConfig, projectFilesystem);
 
--- a/src/com/facebook/buck/ide/intellij/IjProject.java
+++ b/src/com/facebook/buck/ide/intellij/IjProject.java
@@ -104,10 +104,7 @@ public class IjProject {
             javaPackageFinder);
     IjProjectTemplateDataPreparer templateDataPreparer =
         new IjProjectTemplateDataPreparer(
-            parsingJavaPackageFinder,
-            moduleGraph,
-            projectFilesystem,
-            projectConfig);
+            parsingJavaPackageFinder, moduleGraph, projectFilesystem, projectConfig);
     IjProjectWriter writer =
         new IjProjectWriter(templateDataPreparer, projectConfig, projectFilesystem);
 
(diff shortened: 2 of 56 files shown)


src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -478,6 +479,30 @@ private void addAndroidResourcePaths(
     }
   }
 
+  private void addAndroidCompilerOutputPath(
+      Map<String, Object> androidProperties,
+      Path moduleBasePath,
+      IjModuleAndroidFacet androidFacet) {
+    String variantName = "debug";
```
kageiit: why are we hardcoding debug?

jbarr21: it was a default fallback incase the manifest wasn't present (since it is optional - but it shouldn't happen)

kageiit: Can you add a comment about this so we know why it is defaulted to "debug"

styurin:
Can you add more information about variant? What is it?

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -478,6 +479,30 @@ private void addAndroidResourcePaths(
...
+      variantName = manifestParentPath.getName(manifestParentPath.getNameCount() - 1).toString();
+    }
+
+    Path rootModulePath = Paths.get(projectConfig.getProjectRoot());
+    Path moduleRelativePath = rootModulePath.relativize(moduleBasePath);
+    Path moduleLibClassesPath = Paths.get(
+        rootModulePath.toString(),
+        "buck-out/bin",
+        moduleRelativePath.toString(),
+        "lib__src_" + variantName + "__classes");
```
kageiit: We need to add a comment that this may not work if buck spools directly to jar

styurin:
Could you add a test with non-default variant?

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -503,6 +513,41 @@ private void addAndroidResourcePaths(
     }
   }
 
+  private void addAndroidCompilerOutputPath(
+      Map<String, Object> androidProperties,
+      IjModule module,
+      Path moduleBasePath) {
+    // This is the fallback default in case we can't find a valid build target name
+    String targetName = "src_release";
```
kageiit: I do not think we should have a default. There can be resource only modules that have no sources which will not have a compiler output path and they will be fine inheriting from the root project compiler output path

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -503,6 +513,41 @@ private void addAndroidResourcePaths(
...
+
+    if (module.getTargets() != null) {
+      for (BuildTarget target : module.getTargets()) {
+        BuildRule buildRule = buildRuleResolver.getRule(target);
+        if ((buildRule instanceof AndroidLibrary
```
kageiit: Can we just do this for any JvmLibrary. That way it will work for android,java,kotlin,groovy rules etc without any extra configuration

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -503,6 +513,41 @@ private void addAndroidResourcePaths(
...
+            || buildRule instanceof JavaLibrary) && !target.getShortName().contains("aidl")) {
```
styurin: `target.getShortName().contains("aidl")` what is that?

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -503,6 +513,41 @@ private void addAndroidResourcePaths(
...
+    Path rootModulePath = Paths.get(projectConfig.getProjectRoot());
+    Path moduleRelativePath = rootModulePath.relativize(moduleBasePath);
+
+    // IntelliJ may not be able to find classes on the compiler output path
+    // if the jar_spool_mode is set to direct_to_jar.
+    Path moduleLibClassesPath = Paths.get(
+        rootModulePath.toString(),
+        "buck-out/bin",
+        moduleRelativePath.toString(),
+        "lib__" + targetName + "__classes");
```
styurin: Use this: https://github.com/facebook/buck/blob/44be031266fc118cb712802f1d67489291cde8d0/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java#L280-L282

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
jbarr21: [AIDL](https://developer.android.com/guide/components/aidl.html) is an Android-specific interface. since a target can have multiple source-generating rules, we want to prefer the Android/Java-Library one instead of the AIDL since it will be less important to the layout preview

if we only want to set the compiler output for android & java libs but can't deal with particular target types here, then where should that be done? (we're already setting android properties in this part of the class) here, we just want to pick the scratch folder for the module but only for android/java projects.

src/com/facebook/buck/ide/intellij/lang/java/JavaLibraryModuleRule.java
```diff
@@ -48,6 +48,7 @@ public void apply(
     addDepsAndSources(target, true /* wantsPackagePrefix */, context);
     JavaLibraryRuleHelper.addCompiledShadowIfNeeded(projectConfig, target, context);
     context.setJavaLanguageLevel(JavaLibraryRuleHelper.getLanguageLevel(projectConfig, target));
+    context.setCompilerOutputPath(moduleFactoryResolver.getCompilerOutputPath(target));
```
kageiit: Can you please add this one line change to kotlin and groovy module rules as well

src/com/facebook/buck/ide/intellij/IjProject.java
```diff
@@ -104,7 +104,10 @@ public IjProject(
             javaPackageFinder);
     IjProjectTemplateDataPreparer templateDataPreparer =
         new IjProjectTemplateDataPreparer(
-            parsingJavaPackageFinder, moduleGraph, projectFilesystem, projectConfig);
```
styurin: looks like this change is unnecessary

test/com/facebook/buck/ide/intellij/IjProjectDataPreparerTest.java
```diff
@@ -391,10 +391,7 @@ public void testExcludePaths() throws Exception {
 
     IjProjectTemplateDataPreparer dataPreparer =
         new IjProjectTemplateDataPreparer(
-            javaPackageFinder,
```
styurin: the same, can you revert changes in this file?

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -503,6 +505,28 @@ private void addAndroidResourcePaths(
...
+      Map<String, Object> androidProperties,
+      IjModule module,
+      Path moduleBasePath) {
+    // The compiler output path is relative to the project root
+    Optional<Path> compilerOutputPath = module.getCompilerOutputPath();
+    if (compilerOutputPath.isPresent()) {
+      Path rootModulePath = Paths.get(projectConfig.getProjectRoot());
+      Path moduleToRootPath = moduleBasePath.relativize(rootModulePath);
+      Path moduleRelativecompilerOutputPath = Paths.get(
+          moduleToRootPath.toString(), compilerOutputPath.get().toString());
```
styurin: all these lines can be replaced with `moduleBasePath.relativize(compilerOutputPath.get())`

src/com/facebook/buck/ide/intellij/DefaultIjModuleFactoryResolver.java
```diff
@@ -123,6 +124,13 @@ public Path getAndroidManifestPath(TargetNode<AndroidBinaryDescriptionArg, ?> ta
     return JavaLibraryRules.getAnnotationPath(projectFilesystem, targetNode.getBuildTarget());
   }
 
+  @Override
+  public Optional<Path> getCompilerOutputPath(TargetNode<? extends JvmLibraryArg, ?> targetNode) {
+    BuildTarget buildTarget = targetNode.getBuildTarget();
+    Path compilerOutputPath = DefaultJavaLibrary.getClassesDir(buildTarget, projectFilesystem);
+    return Optional.ofNullable(compilerOutputPath);
```
styurin: This should be `Optional.of()`

test/com/facebook/buck/ide/intellij/IjProjectDataPreparerTest.java
```diff
@@ -52,11 +52,11 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
+import org.junit.Before;
+import org.junit.Test;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Optional;
-import org.junit.Before;
-import org.junit.Test;
```
styurin: Revert these changes

PR #1414, opened 2017-07-11
Fix split_zip step exception

kageiit:
I am trying out release builds with buck and am seeing a weird issue during proguard.
Without obfuscation, everything works fine and dandy. With obfuscation, I see stuff like

 ```failed on step split_zip with an exception:
classpath android/support/v7/widget/ListPopupWindow$1 is contained in multiple dex stores:
```
It is usually some layout's anonymous inner classes.

Upon debugging further it seems like there are duplicate entries in the list of dex stores that cause the exception to be thrown unnecessarily.

screen shot 2017-07-10 at 10 55 48 pm

This change removes the source of the duplication without affecting the check. Verified that the `split_zip` step works correctly after this change when proguard obfuscation is enabled for a test build

### DIFF
--- a/src/com/facebook/buck/android/APKModuleGraph.java
+++ b/src/com/facebook/buck/android/APKModuleGraph.java
@@ -35,6 +35,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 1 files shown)


kageiit:
cc @dreiss

joelmccall:
Interesting, I expect this might happen if you are adding those classes in from two different locations, such as from a prebuilt jar file as well as java_library rules

kageiit:
There is actually only a single location for the class being added and that is a prebuilt_jar . We do not have support library in source

PR #1417, opened 2017-07-12
Add a project integration test for groovy and kotlin rules

kageiit:


kageiit:
cc @styurin Here is the repro case we talked about

styurin:
It doesn't work because `buck project` doesn't parse Kotlin and Groovy files to determine package name.

kageiit:
Do we need to parse files to figure out the source roots? Can we not give a hint to the project generation via an argument to the rule instead?

PR #1422, opened 2017-07-14
Kotlin documentation enhancements.

brettwooldridge:


mzlee:
Err, sorry. It looks like something was broken here and the diff was reverted. /cc @styurin

PR #1428, opened 2017-07-19
[TestNG] Support Guice injection in test classes

zhan-xiong:
Formerly, the TestNG runner would skip over test classes that were initialized using guice/javax injection, since they might not have no-arg constructor. This PR makes it so that these classes are now correctly detected and run as test classes. This PR also adds simple tests for both this change and for the TestNG runner in general.


styurin:
could you rebase it, please?

zhan-xiong:
Any updates about this one?

test/com/facebook/buck/testrunner/testdata/simple_testng/test/BUCK.fixture
```diff
@@ -1,7 +1,25 @@
 java_test(
-  name = 'test',
+  name = 'SimpleTest',
```
styurin: This name doesn't follow naming convention we usually use, it should be lower case and words separated by `-`.

test/com/facebook/buck/testrunner/testdata/simple_testng/test/SimpleFailingTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class SimpleFailingTest {
+
+  @Test(expectedExceptions = RuntimeException.class)
+  public void victory() {
+    System.out.println("all i do is win");
+  }
+
+  @Test
+  public void defeat() {
```
styurin: why do you need two tests here?

zhan-xiong:
I think what happens is that when annotated with @Guice, then guice handles all of the dependency injection, so that anything in https://github.com/google/guice/wiki/Injections will work. We use injection via constructor rather frequently throughout our unit tests; it works correctly with this fix applied.

test/com/facebook/buck/testrunner/testdata/simple_testng/test/SimpleFailingTest.java
zhan-xiong: Good point, removed

test/com/facebook/buck/testrunner/testdata/simple_testng/test/BUCK.fixture
zhan-xiong: Changed

styurin:
I'm trying to run these tests and `testThatInjectionWorks` is failing on my laptop with this message:

```
PARSING BUCK FILES... 0.1s (100%)
CREATING ACTION GRAPH: FINISHED IN 0.0s
BUILT 1/14 JOBS 0.0s //lib:aopalliance
BUILT 2/14 JOBS 0.0s //lib:aopalliance#class-abi
BUILT 3/14 JOBS 0.0s //lib:guice
BUILT 4/14 JOBS 0.0s //lib:hamcrest-core#class-abi
BUILT 5/14 JOBS 0.0s //lib:hamcrest-library#class-abi
BUILT 6/14 JOBS 0.0s //lib:hamcrest-core
BUILT 7/14 JOBS 0.0s //lib:hamcrest-library
BUILT 8/14 JOBS 0.0s //lib:javax-inject
BUILT 9/14 JOBS 0.0s //lib:javax-inject#class-abi
BUILT 10/14 JOBS 0.1s //lib:guice#class-abi
BUILT 11/14 JOBS 0.1s //lib:testng
BUILT 12/14 JOBS 0.1s //lib:testng#class-abi
BUILT 13/14 JOBS 0.1s //test:injection-test#testsjar
DOWNLOADED 0.00 BYTES/SEC AVG, 0 ARTIFACTS, 0.00 BYTES
BUILDING: FINISHED IN 0.4s (100%) 13/14 JOBS, 13 UPDATED, 92.9% CACHE MISS
BUILD SUCCEEDED
TESTING //test:injection-test
BUILT 14/14 JOBS 0.0s //test:injection-test
NO TESTS RAN
```

It passes in CI though. Any idea what can cause this?

zhan-xiong:
@styurin Sorry for the delay, was busier than expected for the past few weeks. I can't seem to reproduce the test failure locally either, do you have more details maybe?

PR #1429, opened 2017-07-20
Support creation of extra dependency modules for IntelliJ

jbarr21:
Since the Layout Preview uses compiler output of modules, not the module dependencies, this diff adds the ability to generate extra modules. This is useful for generating an extra module for each Android Library module so that the generated R.java class files can be added to the compiler output path. Doing this allows IntelliJ's Android Layout Preview to resolve the R file classes and properly load views in the Preview tab.

This approach was necessary since Buck stores compiled module class files in a different directory than the compiled dummy R.java class files and IntelliJ only support one compiler output path per module. This functionality is put behind a `.buckconfig` flag.

### DIFF
--- a/docs/concept/buckconfig.soy
+++ b/docs/concept/buckconfig.soy
@@ -2078,8 +2078,9 @@ cxx_library(
   {param name: 'extra_compiler_output_modules_path' /}
   {param example_value: 'buck-out/extra-intellij-output' /}
   {param description}
-    This flag decides whether or not extra library modules will be generated for Android library
-    modules and specifies the root location of those extra modules.
+    This option specifies the location of additional modules for code generated outside of buck
+    graph. For example, it can be used to specify the location of R.java classes generated for
+    Android plugin to help Layout Preview with resolving references to resources.
   {/param}
 {/call}
 
--- a/docs/concept/buckconfig.soy
+++ b/docs/concept/buckconfig.soy
@@ -2073,6 +2073,16 @@ cxx_library(
   {/param}
 {/call}
 
+{call buckconfig.entry}
+  {param section: 'intellij' /}
+  {param name: 'extra_compiler_output_modules_path' /}
+  {param example_value: 'buck-out/extra-intellij-output' /}
+  {param description}
+    This flag decides whether or not extra library modules will be generated for Android library
+    modules and specifies the root location of those extra modules.
+  {/param}
+{/call}
+
 {call buckconfig.section}
   {param name: 'java' /}
   {param description}
--- a/src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
+++ b/src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
@@ -36,7 +36,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
@@ -307,7 +306,7 @@ public final class IjModuleGraphFactory {
           && !module.getExtraModuleDependencies().isEmpty()) {
         IjModule extraModule =
             createExtraModuleForCompilerOutput(
-                module, projectFilesystem, extraCompileOutputRootPath.get());
+                module, extraCompileOutputRootPath.get());
         moduleDeps.put(extraModule, DependencyType.PROD);
         depsBuilder.put(extraModule, ImmutableMap.of());
       }
@@ -329,21 +328,9 @@ public final class IjModuleGraphFactory {
 
   private static IjModule createExtraModuleForCompilerOutput(
       IjModule module,
-      ProjectFilesystem projectFilesystem,
       Path extraCompileOutputRootPath) {
-    Path extraModuleRelativePath = extraCompileOutputRootPath.resolve(module.getModuleBasePath());
-    if (!projectFilesystem.exists(extraModuleRelativePath)) {
-      try {
-        projectFilesystem.mkdirs(extraModuleRelativePath);
-      } catch (IOException e) {
-        throw new AssertionError(
-            "Could not make directories for extra module for compiler output module: "
-                + extraModuleRelativePath.toString());
-      }
-    }
-
     return IjModule.builder()
-        .setModuleBasePath(extraModuleRelativePath)
+        .setModuleBasePath(extraCompileOutputRootPath.resolve(module.getModuleBasePath()))
         .setTargets(ImmutableSet.of())
         .addAllFolders(ImmutableSet.of())
         .putAllDependencies(ImmutableMap.of())
--- a/src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
+++ b/src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
@@ -307,7 +307,7 @@ public final class IjModuleGraphFactory {
           && !module.getExtraModuleDependencies().isEmpty()) {
         IjModule extraModule =
             createExtraModuleForCompilerOutput(
-                module, projectConfig, projectFilesystem, extraCompileOutputRootPath.get());
+                module, projectFilesystem, extraCompileOutputRootPath.get());
         moduleDeps.put(extraModule, DependencyType.PROD);
         depsBuilder.put(extraModule, ImmutableMap.of());
       }
@@ -329,15 +329,9 @@ public final class IjModuleGraphFactory {
 
   private static IjModule createExtraModuleForCompilerOutput(
       IjModule module,
-      IjProjectConfig projectConfig,
       ProjectFilesystem projectFilesystem,
       Path extraCompileOutputRootPath) {
-    Path projectRootPath = Paths.get(projectConfig.getProjectRoot());
-    Path extraModuleBasePath =
-        Paths.get(projectRootPath.toString(), extraCompileOutputRootPath.toString())
-            .resolve(projectRootPath.relativize(module.getModuleBasePath()));
-
-    Path extraModuleRelativePath = projectRootPath.relativize(extraModuleBasePath.normalize());
+    Path extraModuleRelativePath = extraCompileOutputRootPath.resolve(module.getModuleBasePath());
     if (!projectFilesystem.exists(extraModuleRelativePath)) {
       try {
         projectFilesystem.mkdirs(extraModuleRelativePath);
@@ -349,7 +343,7 @@ public final class IjModuleGraphFactory {
     }
 
     return IjModule.builder()
-        .setModuleBasePath(extraModuleBasePath)
+        .setModuleBasePath(extraModuleRelativePath)
         .setTargets(ImmutableSet.of())
         .addAllFolders(ImmutableSet.of())
         .putAllDependencies(ImmutableMap.of())
--- a/src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
+++ b/src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
@@ -23,6 +23,7 @@ import com.facebook.buck.ide.intellij.model.IjLibrary;
 import com.facebook.buck.ide.intellij.model.IjLibraryFactory;
 import com.facebook.buck.ide.intellij.model.IjModule;
 import com.facebook.buck.ide.intellij.model.IjModuleFactory;
+import com.facebook.buck.ide.intellij.model.IjModuleType;
 import com.facebook.buck.ide.intellij.model.IjProjectConfig;
 import com.facebook.buck.ide.intellij.model.IjProjectElement;
 import com.facebook.buck.io.ProjectFilesystem;
@@ -35,6 +36,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
@@ -229,6 +231,7 @@ public final class IjModuleGraphFactory {
     ImmutableMap.Builder<IjProjectElement, ImmutableMap<IjProjectElement, DependencyType>>
         depsBuilder = ImmutableMap.builder();
     final Set<IjLibrary> referencedLibraries = new HashSet<>();
+    Optional<Path> extraCompileOutputRootPath = projectConfig.getExtraCompilerOutputModulesPath();
 
     for (final IjModule module : ImmutableSet.copyOf(rulesToModules.values())) {
       Map<IjProjectElement, DependencyType> moduleDeps = new HashMap<>();
@@ -300,6 +303,15 @@ public final class IjModuleGraphFactory {
         moduleDeps.put(extraClassPathLibrary, DependencyType.PROD);
       }
 
+      if (extraCompileOutputRootPath.isPresent()
+          && !module.getExtraModuleDependencies().isEmpty()) {
+        IjModule extraModule =
+            createExtraModuleForCompilerOutput(
+                module, projectConfig, projectFilesystem, extraCompileOutputRootPath.get());
+        moduleDeps.put(extraModule, DependencyType.PROD);
+        depsBuilder.put(extraModule, ImmutableMap.of());
+      }
+
       moduleDeps
           .keySet()
           .stream()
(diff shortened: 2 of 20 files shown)


styurin:
Please, re-format java files using google-java-format

src/com/facebook/buck/cli/ProjectCommand.java
```diff
@@ -176,6 +176,12 @@ public static Ide fromString(String string) {
   )
   private boolean skipBuild = false;
 
+  @Option(
+      name = "--enable-extra-compiler-output-modules",
+      usage = "Allow generation of extra modules that provide extra compiler output to IntelliJ."
+  )
+  private boolean enableExtraCompilerOutputModules = false;
```
styurin: Do you really need a command line option? Isn't buck config enough?

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
```diff
@@ -315,6 +325,36 @@ public static IjModuleGraph from(
     return new IjModuleGraph(depsBuilder.build());
   }
 
+  private static IjModule createExtraModuleForCompilerOutput(
+      IjModule module, IjProjectConfig projectConfig, ProjectFilesystem projectFilesystem) {
+    Path projectRootPath = Paths.get(projectConfig.getProjectRoot());
+    Path extraModuleBasePath = projectRootPath
+        .resolve("buck-out")
+        .resolve("extra-intellij-output")
```
styurin: where did you get that?

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
```diff
@@ -315,6 +325,36 @@ public static IjModuleGraph from(
...
+        .resolve(projectRootPath.relativize(module.getModuleBasePath()));
+
+    Path extraModuleRelativePath = projectRootPath.relativize(extraModuleBasePath.normalize());
+    if (!projectFilesystem.exists(extraModuleRelativePath)) {
+      try {
+        projectFilesystem.mkdirs(extraModuleRelativePath);
```
styurin: Why do you need to create a directory here?

src/com/facebook/buck/cli/ProjectCommand.java
jbarr21: buck config should be enough, i was just following what seemed like convention to do both. i can remove this

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
jbarr21: I created it. This is a new folder that we use to store our synthetic library modules so that we don't clutter our project view

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
jbarr21: we are creating a synthetic library module so that we can use it's compiler output path to put the r class files on the class path for the Layout Preview. the module is created based off of the base path, which, might not already exist for a clean build. if it doesn't, then the `IjProjectTemplateDataPreparer` runs into a `NoSuchFileException` when it expects the synthetic library module's base path to exist. this prevents that from occurring

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
```diff
@@ -315,6 +325,37 @@ public static IjModuleGraph from(
     return new IjModuleGraph(depsBuilder.build());
   }
 
+  private static IjModule createExtraModuleForCompilerOutput(
+      IjModule module, IjProjectConfig projectConfig, ProjectFilesystem projectFilesystem) {
+    Path projectRootPath = Paths.get(projectConfig.getProjectRoot());
+    Path extraModuleBasePath =
+        projectRootPath
+            .resolve("buck-out")
+            .resolve("extra-intellij-output")
```
styurin: Where this path is coming from?

jbarr21: I created it. This is a new folder that we use to store our synthetic library modules so that we don't clutter our project view. The path is arbitrary and can go somewhere else if you prefer

styurin: I suppose you can move this location to a config parameter and change logic to enable with extra dependencies if path is not empty.

jbarr21: the logic is already gated on https://github.com/facebook/buck/pull/1429/files/3c9b7a377ec8cbaf2a2d847ca19765603d2440a5#diff-4b53a690d0d3ac329da62c36fcacf204R306 by the buck flag `enable_extra_compiler_output_modules`. is that still necessary?

styurin: You can replace `enable_extra_compiler_output_modules` with a path and enable this logic when this path is not empty. No need to add two options.

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
styurin: The purpose of `IjModuleGraphFactory` is to create `IjModuleGraph`. Creating a directory has nothing to do with that. It's better either adjust `IjProjectTemplateDataPreparer` or create these directories before main logic of `buck project`.

jbarr21: We are adding synthetic modules to the module graph here. In order for the module to be properly created and added to the graph, it's module root must be created (since there is no actual code there already). This depends upon knowledge of each module, so it can't happen before the main logic of buck project. Attached is what the module structure would look like for the uber/okbuck sample app. The directories that it would create are listed below. They would only be created for Android libraries and if the project configuration flag enabled them.
- buck-out/extra-intellij-output/another-app
- buck-out/extra-intellij-output/libraries/common
- buck-out/extra-intellij-output/libraries/empty-library

screen shot 2017-08-08 at 2 51 11 pm

styurin:
you need to re-format java files

src/com/facebook/buck/ide/intellij/lang/android/AndroidLibraryModuleRule.java
```diff
@@ -53,6 +53,7 @@ public void apply(
     Optional<Path> dummyRDotJavaClassPath = moduleFactoryResolver.getDummyRDotJavaPath(target);
     if (dummyRDotJavaClassPath.isPresent()) {
       context.addExtraClassPathDependency(dummyRDotJavaClassPath.get());
```
styurin: Why this logic doesn't work for you? This should create a library with extra class path which should point to R.java.

jbarr21:
> you need to re-format java files

sorry, i installed the pre-commit hook after last time. will double check it

src/com/facebook/buck/ide/intellij/lang/android/AndroidLibraryModuleRule.java
jbarr21: this creates an extra library, but we need to create an extra module so that we can specify the compiler output (since that's where the Layout Preview loads classes from)

jbarr21:
Rebased and squashed. All feedback should be addressed

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
```diff
@@ -315,6 +327,38 @@ public static IjModuleGraph from(
...
+      IjModule module,
+      IjProjectConfig projectConfig,
+      ProjectFilesystem projectFilesystem,
+      Path extraCompileOutputRootPath) {
+    Path projectRootPath = Paths.get(projectConfig.getProjectRoot());
+    Path extraModuleBasePath =
+        Paths.get(projectRootPath.toString(), extraCompileOutputRootPath.toString())
+            .resolve(projectRootPath.relativize(module.getModuleBasePath()));
+
+    Path extraModuleRelativePath = projectRootPath.relativize(extraModuleBasePath.normalize());
```
styurin: It looks a bit complex, how about this:

```
  private static IjModule createExtraModuleForCompilerOutput(
      IjModule module,
      ProjectFilesystem projectFilesystem,
      Path extraCompileOutputRootPath) {
    Path extraModuleRelativePath = extraCompileOutputRootPath.resolve(module.getModuleBasePath());
    if (!projectFilesystem.exists(extraModuleRelativePath)) {
      try {
        projectFilesystem.mkdirs(extraModuleRelativePath);
      } catch (IOException e) {
        throw new AssertionError(
            "Could not make directories for extra module for compiler output module: "
                + extraModuleRelativePath.toString());
      }
    }

    return IjModule.builder()
        .setModuleBasePath(extraModuleRelativePath)
        .setTargets(ImmutableSet.of())
        .addAllFolders(ImmutableSet.of())
        .putAllDependencies(ImmutableMap.of())
        .setLanguageLevel(module.getLanguageLevel())
        .setModuleType(IjModuleType.ANDROID_MODULE)
        .setCompilerOutputPath(module.getExtraModuleDependencies().asList().get(0))
        .build();
  }

File: src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java (hunk 3 above)
Comment (styurin) [inline src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java]: When I removed this logic, your test passes. What are the conditions for this logic to be not working?

File: docs/concept/buckconfig.soy [hunk 7]
```diff
@@ -2073,6 +2073,16 @@ cxx_library(
   {/param}
 {/call}
 
+{call buckconfig.entry}
+  {param section: 'intellij' /}
+  {param name: 'extra_compiler_output_modules_path' /}
+  {param example_value: 'buck-out/extra-intellij-output' /}
+  {param description}
+    This flag decides whether or not extra library modules will be generated for Android library
```
styurin: This is not a flag, right? Also, it seems generic enough, not Android specific.

How about something like this:

> This option specifies the location of additional modules for code generated outside of buck graph. For example, it can be used to specify the location of R.java classes generated for Android plugin to help Layout Preview with resolving references to resources.

src/com/facebook/buck/ide/intellij/lang/android/AndroidLibraryModuleRule.java
styurin: Can you update the description with this information? I mean the fact that Layout Preview uses compiler output, not the module dependencies.

src/com/facebook/buck/ide/intellij/ModuleBuildContext.java
```diff
@@ -44,6 +44,7 @@
 
   private Optional<IjModuleAndroidFacet.Builder> androidFacetBuilder;
   private ImmutableSet.Builder<Path> extraClassPathDependenciesBuilder;
+  private ImmutableSet.Builder<Path> extraModuleDependenciesBuilder;
```
jbarr21: I was trying to keep it generalized so that other build steps could leverage this synthetic modules functionality. If you prefer, I can just use a single Path.

styurin: Either is fine, I suppose

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -363,6 +363,16 @@ private boolean isRootAndroidResourceDirectory(IjModule module, Path dir) {
 
     boolean isAndroidFacetPresent = androidFacetOptional.isPresent();
     androidProperties.put("enabled", isAndroidFacetPresent);
+
+    Path basePath = module.getModuleBasePath();
+
+    Optional<Path> extraCompilerOutputPath = projectConfig.getExtraCompilerOutputModulesPath();
+    if (isAndroidFacetPresent
```
jbarr21: My mistake. I will remove the 2nd call. In a prior PR (https://github.com/facebook/buck/pull/1412/files#diff-3edfe4c74fe850fc67a4d5eb7422c512R388), I already added the compiler output property to the IML for every Android module. The synthetic modules don't have the Android facet, which is why the check has been updated to accommodate both scenarios

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
jbarr21: When generating a real project (I used uber/okbuck sample app), without this logic, I get the exception below. It looks like the `FakeProjectFilesystem` used in the test code does not throw the same exception since it's `.exists()` method only checks whether its a file or directory and not the presence of the actual file on disk
```
[2017-08-11 11:56:17.363][error][command:null][tid:106][com.facebook.buck.cli.Main] /Users/jbarr/Uber/okbuck/buck-out/extra-intellij-output/app
java.nio.file.NoSuchFileException: /Users/jbarr/Uber/okbuck/buck-out/extra-intellij-output/app
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:144)
	at java.nio.file.Files.readAttributes(Files.java:1737)
	at com.facebook.buck.io.ProjectFilesystem$FileTreeWalker.getAttributes(ProjectFilesystem.java:1190)
	at com.facebook.buck.io.ProjectFilesystem$FileTreeWalker.walk(ProjectFilesystem.java:1107)
	at com.facebook.buck.io.ProjectFilesystem$FileTreeWalker.access$2(ProjectFilesystem.java:1106)
	at com.facebook.buck.io.ProjectFilesystem.walkFileTree(ProjectFilesystem.java:535)
	at com.facebook.buck.io.ProjectFilesystem.walkRelativeFileTree(ProjectFilesystem.java:524)
	at com.facebook.buck.io.ProjectFilesystem.walkRelativeFileTree(ProjectFilesystem.java:481)
	at com.facebook.buck.ide.intellij.IjProjectTemplateDataPreparer.createExcludes(IjProjectTemplateDataPreparer.java:228)
	at com.facebook.buck.ide.intellij.IjProjectTemplateDataPreparer.getContentRoots(IjProjectTemplateDataPreparer.java:291)
	at com.facebook.buck.ide.intellij.IjProjectWriter.write(IjProjectWriter.java:59)
	at com.facebook.buck.ide.intellij.IjProject.write(IjProject.java:113)
	at com.facebook.buck.ide.intellij.IjProjectCommandHelper.writeProjectAndGetRequiredBuildTargets(IjProjectCommandHelper.java:269)
	at com.facebook.buck.ide.intellij.IjProjectCommandHelper.runIntellijProjectGenerator(IjProjectCommandHelper.java:233)
	at com.facebook.buck.ide.intellij.IjProjectCommandHelper.parseTargetsAndRunProjectGenerator(IjProjectCommandHelper.java:196)
	at com.facebook.buck.cli.ProjectCommand.runWithoutHelp(ProjectCommand.java:317)
	at com.facebook.buck.cli.AbstractCommand.run(AbstractCommand.java:227)
	at com.facebook.buck.cli.AbstractContainerCommand.run(AbstractContainerCommand.java:67)
	at com.facebook.buck.cli.BuckCommand.run(BuckCommand.java:81)
	at com.facebook.buck.cli.Main.runMainWithExitCode(Main.java:983)
	at com.facebook.buck.cli.Main.runMainThenExit(Main.java:372)
	at com.facebook.buck.cli.Main.nailMain(Main.java:1725)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.martiansoftware.nailgun.NGSession.run(NGSession.java:329)
```

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
```diff
@@ -315,6 +327,32 @@ public static IjModuleGraph from(
...
   }
 
+  private static IjModule createExtraModuleForCompilerOutput(
+      IjModule module,
+      ProjectFilesystem projectFilesystem,
+      Path extraCompileOutputRootPath) {
+    Path extraModuleRelativePath = extraCompileOutputRootPath.resolve(module.getModuleBasePath());
+    if (!projectFilesystem.exists(extraModuleRelativePath)) {
+      try {
+        projectFilesystem.mkdirs(extraModuleRelativePath);
```
styurin: > When generating a real project (I used uber/okbuck sample app), without this logic, I get the exception below. It looks like the FakeProjectFilesystem used in the test code does not throw the same exception since it's .exists() method only checks whether its a file or directory and not the presence of the actual file on disk

Let's fix `IjProjectTemplateDataPreparer.createExcludes` instead of hacking in this class:

```
  public ImmutableCollection<IjFolder> createExcludes(final IjModule module) throws IOException {
    final Path moduleBasePath = module.getModuleBasePath();
    if (!projectFilesystem.exists(moduleBasePath)) {
      return ImmutableList.of();
    }
    final ImmutableList.Builder<IjFolder> excludesBuilder = ImmutableList.builder();
    projectFilesystem.walkRelativeFileTree(
```

src/com/facebook/buck/ide/intellij/IjModuleGraphFactory.java
```diff
@@ -23,6 +23,7 @@
 import com.facebook.buck.ide.intellij.model.IjLibraryFactory;
 import com.facebook.buck.ide.intellij.model.IjModule;
 import com.facebook.buck.ide.intellij.model.IjModuleFactory;
+import com.facebook.buck.ide.intellij.model.IjModuleType;
```
styurin: Please, re-format this file with google-java-format

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -225,6 +225,9 @@ private static void addPathAndParents(Set<Path> pathSet, Path path) {
   public ImmutableCollection<IjFolder> createExcludes(final IjModule module) throws IOException {
     final ImmutableList.Builder<IjFolder> excludesBuilder = ImmutableList.builder();
```
styurin: Please, move this declaration after `if`

PR #1434, opened 2017-07-21
Create a FeatureEnableService so Layout Preview tab appears in IntelliJ

jbarr21:
This diffs adds a `FeatureEnableService` to the IdeaBuck plugin so that IntelliJ displays the Layout Preview "Design" tab for Android layout XML files.

Without "Design" tab below XML editor:
layout-xml-without-tabs

With "Design" tab below XML editor that provides access to Android Layout Preview:
layout-xml-with-tabs

### DIFF
--- a/src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
+++ b/src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
@@ -146,6 +146,8 @@
   <depends>com.intellij.modules.platform</depends>
   <depends>com.intellij.modules.lang</depends>
 
+  <depends optional="true">org.jetbrains.android</depends>
+
   <extensions defaultExtensionNs="com.intellij">
     <fileTypeFactory implementation="com.facebook.buck.intellij.ideabuck.file.BuckFileTypeFactory"/>
     <lang.parserDefinition
@@ -205,6 +207,10 @@
     <programRunner
             implementation="com.facebook.buck.intellij.ideabuck.configurations.TestProgramRunner"/>
     <runLineMarkerContributor language="JAVA" implementationClass="com.facebook.buck.intellij.ideabuck.actions.select.SelectedTestRunLineMarkerContributor"/>
+  </extensions>
+
+  <extensions defaultExtensionNs="com.android.project">
+    <featureEnableService implementation="com.facebook.buck.intellij.ideabuck.proj.BuckFeatureEnableService"/>
   </extensions>
 
   <actions>
--- a/third-party/java/intellij/BUCK
+++ b/third-party/java/intellij/BUCK
@@ -1,6 +1,7 @@
 java_library(
     name = "intellij-plugin-sdk",
     exported_deps = [
+        ":android",
         ":annotations",
         ":extensions",
         ":idea",
@@ -47,4 +48,9 @@ prebuilt_jar(
 prebuilt_jar(
     name = "jdom",
     binary_jar = "jdom.jar",
+)
+
+prebuilt_jar(
+    name = "android",
+    binary_jar = "android.jar",
 )
--- a/src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/proj/BuckFeatureEnableService.java
+++ b/src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/proj/BuckFeatureEnableService.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2015-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.intellij.ideabuck.proj;
+
+import com.android.tools.idea.project.FeatureEnableService;
+import com.intellij.openapi.project.Project;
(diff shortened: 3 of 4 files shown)


jbarr21:
> I'm trying to test this change, but I can't see which feature it enables. Can you provide some screenshots, please?

this re-enables the extra "Design" tab when editing Android Layout XML files. as of IntelliJ 2017.2, XML layouts won't display the tab without this change. screenshots added to the description

third-party/java/intellij/BUCK
```diff
@@ -48,3 +49,8 @@ prebuilt_jar(
     name = "jdom",
     binary_jar = "jdom.jar",
 )
+
+prebuilt_jar(
+    name = "android",
+    binary_jar = "android.jar",
```
jbarr21: I got this JAR from IntelliJ installation. it was needed so that we could find the `FeatureEnableService` symbol

jbarr21: yes, it's from an IntelliJ installation. it does not need to be a specific version.

@styurin which piece of that do you want in a separate PR?

styurin: > which piece of that do you want in a separate PR?

Updating `get_jars.sh` and adding a new jar.

styurin:
> this re-enables the extra "Design" tab when editing Android Layout XML files. as of IntelliJ 2017.2, XML layouts won't display the tab without this change. screenshots added to the description

Do you use aggregation of modules with Android resources (option `intellij.aggregate_android_resource_modules` in `.buckconfig`)? This is the only reason I can think of why IntelliJ isn't showing that tab.

If you don't use this option, can you provide a sample project where this tab isn't shown?

jbarr21:
>> this re-enables the extra "Design" tab when editing Android Layout XML files. as of IntelliJ 2017.2, XML layouts won't display the tab without this change. screenshots added to the description

> Do you use aggregation of modules with Android resources (option intellij.aggregate_android_resource_modules in .buckconfig)? This is the only reason I can think of why IntelliJ isn't showing that tab.
>
> If you don't use this option, can you provide a sample project where this tab isn't shown?

The .buckconfig option has no effect on the issue.

Any project (I used the [OkBuck sample app](https://github.com/uber/okbuck)), when opened with the IntelliJ 2017.2 build, does not display the editor tab. This is because the [condition for the Layout Preview Editor](https://github.com/JetBrains/android/blob/master/designer/src/com/android/tools/idea/uibuilder/editor/NlEditorProvider.java#L46) to be displayed fails since the [GradleFeatureEnableService](https://github.com/JetBrains/android/blob/master/android/src/com/android/tools/idea/gradle/project/GradleFeatureEnableService.java#L25) requires the project to be [built with Gradle](https://github.com/JetBrains/android/blob/master/android/src/com/android/tools/idea/gradle/project/GradleProjectInfo.java#L68).

This PR adds a FeatureEnableService extension that checks if the project was built with Buck, and enables the Layout Preview editor tab.

third-party/java/intellij/BUCK
jbarr21: ok, will do, but then this PR won't compile (since no android.jar), is that OK?

jbarr21: Added in #1448

styurin:
This is failing to build because `FeatureEnableService` was added in 2017.2 and I updated libraries from 2017.1. This merge this PR we would have to bump ideabuck's min version to 2017.2, but right now we are using 2017.1 internally. I think the only way to proceed here is to wait for us to update to 2017.2 and then we can merge this fix.

cc @mkillianey

mkillianey:
I'm working on a change to our chef scripts so that we can install the old
version in 2017.1 and the new version in 2017.2 on the same machine.  I
hope to have that change ready by next week.

styurin:
@jbarr21 could you update this PR with a new android.jar from IntelliJ 2017.2? I want to review this code while Mick is working on his fix.

styurin:
Can you also update IntelliJ configuration to allow to build the plugin from IntelliJ?

styurin:
OMG, it works with merging android resources! Thanks a lot :)

styurin:
By the way, android.jar should be added as a provided dependency in ideabuck.iml to avoid pulling stripped jar.

jbarr21:
@styurin can you please clarify the android.jar via ideabuck.iml comment? shouldn't the dependency be provided via the buck configuration, i'll just update the script to not stub the android.jar?

jbarr21:
@styurin just wanted to check on the status of your IntelliJ installs. is this PR still blocked on that?

as for building the plugin from IntelliJ, that sounds outside of the scope of this PR

styurin:
@mkillianey can you please comment if we are ready to bump the minimum version for buck plugin to 2017.2?

> can you please clarify the android.jar via ideabuck.iml comment?

When building and testing plugin inside IntelliJ (using IntelliJ build-in functionality, not buck build), the plugin jar contains stripped version of android.jar and, since the content of the stripped jar is not working, the plugin is not working. If dependency type on android.jar is "provided" then android.jar is not included in the plugin jar and the plugin is using classes provided by InteliiJ.

jbarr21:
IntelliJ 2017.2.4 picks up this fix for the issue, though I think it's still a good idea to continue this PR in case their legacy logic changes in the future https://github.com/JetBrains/android/commit/09d47f3a608659723b98f46505b2c35d32436048

src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
```diff
@@ -207,6 +209,10 @@
     <runLineMarkerContributor language="JAVA" implementationClass="com.facebook.buck.intellij.ideabuck.actions.select.SelectedTestRunLineMarkerContributor"/>
   </extensions>
 
+  <extensions defaultExtensionNs="com.android.project">
+    <featureEnableService implementation="com.facebook.buck.intellij.ideabuck.proj.BuckFeatureEnableService"/>
```
mkillianey: Q:  Does using this mean that you need to update the `since-build` on line 144?

styurin:
I think this change is required for a project generated by `buck project` when Android resource module aggregation is enabled because such projects are not legacy.

What this PR needs at this point:

- bump minimal version to 2017.2 (a version where `FeatureEnableService` first appeared
- upload a version of `android.jar` from 2017.2. The version in buck repo is from 2017.1

PR #1435, opened 2017-07-21
Add support to buck for unbundled resources

### LINKED ISSUE #699: Enhancement Request: Allow usage of unbundled resources in java and scala tests
I'd like to request that Buck be enhanced to allow for usage of unbundled resources in java and scala tests I'd like to migrate a fairly large gradle project (30+ sub-projects, 2000+ tests) to Buck, however, many of our tests rely on the directory from the getResource() call in java to return a local file reference, not a jar. 

So for a specific implementation idea, something like external_resources and external_resource_root would be great; they would for resources to be accessible by tests, but located in the file system, not the jar.

kageiit:
Fixes #699

Verified this by adding the unbundled_resources_root and running tests that access resources via `getResource` which relies on the file path.

### DIFF
--- a/src/com/facebook/buck/android/RobolectricTest.java
+++ b/src/com/facebook/buck/android/RobolectricTest.java
@@ -92,6 +92,7 @@ public class RobolectricTest extends JavaTest {
       ForkMode forkMode,
       Optional<Level> stdOutLogLevel,
       Optional<Level> stdErrLogLevel,
+      Optional<SourcePath> unbundledResourcesRoot,
       Optional<String> robolectricRuntimeDependency,
       Optional<SourcePath> robolectricManifest) {
     super(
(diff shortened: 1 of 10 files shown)


kageiit:
I can add a test for this logic if the approach looks ok.

ttsugriy:
Thank you for your contribution @kageiit! Looks like your change breaks the [build](https://travis-ci.org/facebook/buck/jobs/256249368#L2944) though:
```
stderr: /home/travis/build/facebook/buck/test/com/facebook/buck/jvm/java/JavaLibraryBuilder.java:119: error: no suitable method found for setResourcesRoot(java.util.Optional<java.nio.file.Path>)
    getArgForPopulating().setResourcesRoot(Optional.of(root));
```

could you please address build breakage before we move forward?

kageiit:
I have fixed the build breakage in the tests. @ttsugriy

yiding:
Still has test failures.

kageiit:
After chatting with @dreiss , we decided to make this a new option instead

PR #1437, opened 2017-07-24
Set output url of intellij project generation

zhan-xiong:
This sets the project compiler output directory in IntelliJ using the output_url option in the [intellij] section of the .buckconfig file.


src/com/facebook/buck/ide/intellij/IjProjectBuckConfig.java
```diff
@@ -121,6 +121,7 @@ public static IjProjectConfig create(
         .setGeneratingAndroidManifestEnabled(
             buckConfig.getBooleanValue(
                 INTELLIJ_BUCK_CONFIG_SECTION, "generate_android_manifest", false))
+        .setOutputUrl(buckConfig.getValue(INTELLIJ_BUCK_CONFIG_SECTION, "output_url"))
```
styurin: Can you rename this option to be something like `project_compiler_output_url` instead? Right now it's too generic.

styurin:
Can you also add this new option to docs in https://buckbuild.com/concept/buckconfig.html#intellij ?

styurin:
Also, re-format files using google-java-format

jiangty-addepar:
Related to this PR: I've been confused about this for a while, since the Buck fork we're using is well behind master: with Buck IntelliJ integration, are developers intended to build/test/run code through IntelliJ, or should they be only using the Buck plugin?

The reason I'm asking this is that my original impression of "new project generation" was for Buck and IntelliJ to build in the same way (i.e. only through Buck) instead of having IntelliJ and Buck both build things, possibly in slightly different ways. Pardon me if I'm misunderstanding something.

styurin:
> are developers intended to build/test/run code through IntelliJ, or should they be only using the Buck plugin?

It depends on your project complexity. Simple projects can be built using IntelliJ, but IntelliJ model is less flexible and it doesn't allow certain configurations (like when two targets are present in one directory). In such cases a project generated by `buck project` wouldn't be compileable by IntelliJ and the only way to build it is to use ideabuck plugin.

"new project generation" still tries to convert buck dependency graph into a form that be understood by IntelliJ. Since IntelliJ has restrictions, the conversion cannot be exact.

PR #1439, opened 2017-07-26
Enable path detection when running executables using buckd

zhan-xiong:
After [this commit](https://github.com/facebook/buck/commit/a271a23a24a63e88decba55e176d623717628719), running executables using buckd doesn't detect the PATH environment variable correctly, (e.g. `bin/buck run //src/com/facebook/buck/jvm/java/abi:api-stubber` doesn't find the `java` executable correctly). This uses execvpe instead of execve, which does respect the PATH variable.

### DIFF
--- a/programs/buck.py
+++ b/programs/buck.py
@@ -50,7 +50,7 @@ if __name__ == "__main__":
         setup_logging()
         propagate_failure(main(sys.argv))
     except ExecuteTarget as e:
-        e.execve()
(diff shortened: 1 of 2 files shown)


kageiit:
Seems like it is fixed in https://github.com/facebook/buck/commit/cba47adc4eb2fd150f8e5d669e221c63f5a446e6

Issue #1371, opened 2017-06-09
Un-ignore mixed sources tests when kotlin 1.1.3 is released.


runningcode:
kotlin 1.1.3 includes support for passing java files. We can un-ignore the ignored mixed sources tests when the final version of the compiler is released.

PR #1441, opened 2017-07-27
Add support for extra kotlinc args

cwoodwar6:
This provides the ability to add kotlinc plugins such as kotlin android extensions. Both `robolectric_test` rule and `android_library` rules needed support.

Since `extra_arguments` only applies to the java compiler, we needed a separate argument for the kotlin compiler.

Example of its use (for kotlin android extensions):

```
extra_kotlinc_arguments = [
   '-Xplugin=<kotlin_home>/kotlin-android-extensions.jar',
   '-P',
   'plugin:org.jetbrains.kotlin.android:variant=main;<project-root>/app/src 
    /main/res,plugin:org.jetbrains.kotlin.android:package=com.example.app',
],
```

### DIFF
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -42,8 +42,8 @@ import com.facebook.buck.rules.TargetGraph;
 import com.facebook.buck.util.immutables.BuckStyleImmutable;
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableSet;
-import java.util.Optional;
 import org.immutables.value.Value;
+import java.util.Optional;
 
 public class AndroidLibraryDescription
     implements Description<AndroidLibraryDescriptionArg>,
@@ -143,14 +143,12 @@ public class AndroidLibraryDescription
   }
 
   public interface CoreArg
-      extends JavaLibraryDescription.CoreArg, HasDepsQuery, HasProvidedDepsQuery {
+      extends JavaLibraryDescription.CoreArg, AndroidKotlinCoreArg, HasDepsQuery, HasProvidedDepsQuery {
     Optional<SourcePath> getManifest();
 
     Optional<String> getResourceUnionPackage();
 
     Optional<String> getFinalRName();
-
-    Optional<JvmLanguage> getLanguage();
   }
 
   @BuckStyleImmutable
--- a/docs/rule/android_library.soy
+++ b/docs/rule/android_library.soy
@@ -141,6 +141,14 @@ compiled class files and resources.
 {/call}
 
 {call buck.arg}
+  {param name: 'extra_kotlinc_arguments' /}
+  {param default : '[]' /}
+  {param desc}
+  List of additional arguments to pass into the Kotlin compiler.
+  {/param}
+{/call}
+
+{call buck.arg}
   {param name: 'exported_deps' /}
   {param default : '[]' /}
   {param desc}
--- a/docs/rule/robolectric_test.soy
+++ b/docs/rule/robolectric_test.soy
@@ -55,6 +55,14 @@ with Robolectric test runner. It extends from <code>java_test()</code> rule.
   {/param}
 {/call}
 
+{call buck.arg}
+  {param name: 'extra_kotlinc_arguments' /}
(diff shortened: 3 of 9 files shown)


src/com/facebook/buck/android/AndroidLibraryDescription.java
```diff
@@ -153,6 +154,8 @@ public void findDepsForTargetFromConstructorArgs(
     Optional<Query> getDepsQuery();
 
     Optional<Query> getProvidedDepsQuery();
+
+    ImmutableList<String> getExtraKotlincArguments();
```
kageiit: We should add this to kotlin library description as well

cwoodwar6: Kotlin library already has `extra_arguments` which pass arguments to the kotlin compiler. Android library is different since it invokes two compilers.

kageiit: ah cool

src/com/facebook/buck/jvm/java/AndroidCoreArg.java
```diff
@@ -0,0 +1,26 @@
...
+package com.facebook.buck.jvm.java;
+
+import com.facebook.buck.android.AndroidLibraryDescription;
+import com.google.common.collect.ImmutableList;
+import java.util.Optional;
+
+public interface AndroidCoreArg {
+  Optional<AndroidLibraryDescription.JvmLanguage> getLanguage();
+
+  ImmutableList<String> getExtraKotlincArguments();
```
kageiit: Can we make this `Optional` instead?

cwoodwar6:
@dsyang - Could you take a look at this one too please?

src/com/facebook/buck/jvm/java/AndroidCoreArg.java
kageiit: @cwoodwar6 not sure if you saw this comment but we should make this optional

cwoodwar6: Yep, I'll update it.

src/com/facebook/buck/jvm/java/AndroidCoreArg.java
```diff
@@ -0,0 +1,26 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.facebook.buck.jvm.java;
+
+import com.facebook.buck.android.AndroidLibraryDescription;
+import com.google.common.collect.ImmutableList;
+import java.util.Optional;
+
+public interface AndroidCoreArg {
```
dsyang: buck.jvm.java seems like the wrong place to put Android stuff. Can this be moved to com.facebook.buck.android?

cwoodwar6: Yep, I'll move it.

src/com/facebook/buck/android/AndroidCoreArg.java
```diff
@@ -0,0 +1,25 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.facebook.buck.android;
+
+import com.google.common.collect.ImmutableList;
+import java.util.Optional;
+
+public interface AndroidCoreArg {
```
dsyang: Last small nit: these fields don't seem related to "AndroidCore". `AndroidLibraryDescription.CoreArg` is more "core" than this. Maybe rename to AndroidKotlinCoreArg?

cwoodwar6:
I added a test, but it must intentionally fails until we bump the kotlin version in https://github.com/facebook/buck/issues/1371. Then we can test how to use the plugin appropriately.

src/com/facebook/buck/android/AndroidCoreArg.java
cwoodwar6: Good suggestion.

cwoodwar6:
@dsyang - Are we all good here?

dsyang:
@cwoodwar6  I'm having trouble importing this PR. when you get the chance can you rebase?

cwoodwar6:
Just rebased, how about now?

PR #1448, opened 2017-08-03
Update get_jars.sh with stubbed android.jar for FeatureEnableService

jbarr21:
This updates the `get_jars.sh` script to pull and stub the android.jar from an IntelliJ instance. The ` /Applications/IntelliJ IDEA.app/Contents/plugins/android/lib/android.jar` is needed as a provided dependency for #1434 but will ultimately be supplied by the IntelliJ Android plugin dependency. It does not need to be a specific version. I used IJ 2017.2

### DIFF
--- a/third-party/java/intellij/get_jars.sh
+++ b/third-party/java/intellij/get_jars.sh
@@ -118,11 +118,15 @@ fi
 #  Copy and/or stub jars
 # ====================================================================
 printf "Copying IntelliJ jars from %q\n" "$SRC_DIR"
-copy_or_stub_jars "$SRC_DIR" ${#INTELLIJ_LIBS_TO_COPY[@]} ${INTELLIJ_LIBS_TO_COPY[@]} ${#INTELLIJ_LIBS_TO_STUB[@]} ${INTELLIJ_LIBS_TO_STUB[@]}
+copy_or_stub_jars "$SRC_DIR" \
+    ${#INTELLIJ_LIBS_TO_COPY[@]} ${INTELLIJ_LIBS_TO_COPY[@]} \
+    ${#INTELLIJ_LIBS_TO_STUB[@]} ${INTELLIJ_LIBS_TO_STUB[@]}
 
 ANDROID_SRC_DIR="$SRC_DIR/../plugins/android/lib"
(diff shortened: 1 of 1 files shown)


third-party/java/intellij/get_jars.sh
```diff
@@ -59,43 +114,15 @@ if [[ -z "$SRC_DIR" ]]; then
...
-    SHOULD_FAIL=true
-  fi
-done
-if $SHOULD_FAIL ; then
-  echo "Aborting."
-  exit 1
-fi
+ANDROID_SRC_DIR="$SRC_DIR/../plugins/android/lib"
+printf "Copying IntelliJ Android jars from %q\n" "$ANDROID_SRC_DIR"
+copy_or_stub_jars "$ANDROID_SRC_DIR" ${#ANDROID_LIBS_TO_COPY[@]} ${#ANDROID_LIBS_TO_STUB[@]} ${ANDROID_LIBS_TO_COPY[@]} ${ANDROID_LIBS_TO_STUB[@]}
```
styurin: It looks like the arguments are wrong:

> copy_or_stub_jars "$ANDROID_SRC_DIR" ${#ANDROID_LIBS_TO_COPY[@]} ${#ANDROID_LIBS_TO_STUB[@]} ${ANDROID_LIBS_TO_COPY[@]} ${ANDROID_LIBS_TO_STUB[@]}

should be

> copy_or_stub_jars "$ANDROID_SRC_DIR" ${#ANDROID_LIBS_TO_COPY[@]} ${ANDROID_LIBS_TO_COPY[@]} ${#ANDROID_LIBS_TO_STUB[@]} ${ANDROID_LIBS_TO_STUB[@]}

The format is: number of libs, libs, number of stubs, stubs. But in your case it's: number of libs, number of stubs, libs, stubs.

Ideally, to catch these kinds of errors you would need to have some kind of separators in the list of arguments, but it's up to you to use that.

jbarr21: good catch, updated this w/ some separators.

styurin:
I'm going to merge this with a jar taken from 2017.1 because that's the version we currently support in ideabuck plugin.

PR #1449, opened 2017-08-04
Don't invoke kotlinc if we don't have any kotlin files

cwoodwar6:
Don't bother invoking the kotlin compiler if you don't actually pass it any kotlin files. It will result in `error: no source file`.

### DIFF
--- a/src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
+++ b/src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
@@ -42,6 +42,7 @@ import java.util.stream.Collectors;
 public class KotlincToJarStepFactory extends BaseCompileToJarStepFactory {
 
   private static final PathOrGlobMatcher JAVA_PATH_MATCHER = new PathOrGlobMatcher("**.java");
+  private static final PathOrGlobMatcher KOTLIN_PATH_MATCHER = new PathOrGlobMatcher("**.kt");
 
   private final Kotlinc kotlinc;
   private final ImmutableList<String> extraArguments;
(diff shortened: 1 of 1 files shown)


kageiit:
@dsyang can we get this simple change merged?

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -82,21 +83,32 @@ public void createCompileStep(
...
-                .build(),
-            kotlinc,
-            extraArguments,
-            filesystem));
+    ImmutableSortedSet<Path> kotlinSourceFiles =
+        ImmutableSortedSet.copyOf(
+            sourceFilePaths
+                .stream()
+                .filter(KOTLIN_PATH_MATCHER::matches)
+                .collect(Collectors.toSet()));
```
dsyang: Can you just use `anyMatch` here instead of building a new set? (https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate-)

cwoodwar6: Done.

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -82,21 +83,24 @@ public void createCompileStep(
...
+                      Optional.ofNullable(extraClassPath.apply(buildContext))
+                          .orElse(ImmutableList.of()))
+                  .addAll(declaredClasspathEntries)
+                  .build(),
+              kotlinc,
+              extraArguments,
+              filesystem));
+    }
 
     ImmutableSortedSet<Path> javaSourceFiles =
```
kageiit: Maybe opportunistically fix this as well to not make a copy?

cwoodwar6: I thought about that as well but I realized that it actually needs the set to provide only the java source files to the java compiler. Kolinc actually needs both sets of source files so we didn't need a set of only kotlin files.

kageiit: ah gotcha

PR #1450, opened 2017-08-05
Use java test args for jvm based test rules

kageiit:


kageiit:
cc @dreiss @styurin

styurin:
build is broken

kageiit:
Ant build works fine, but buck build is giving a cryptic error that looks like a jdk bug (https://bugs.openjdk.java.net/browse/JDK-8144101) apparently fixed in jdk 9. cc @jkeljo

src/com/facebook/buck/jvm/kotlin/KotlinTestDescription.java
```diff
@@ -168,27 +164,5 @@ public void findDepsForTargetFromConstructorArgs(
...
-    default boolean getRunTestSeparately() {
-      return false;
-    }
-
-    @Value.Default
-    default ForkMode getForkMode() {
-      return ForkMode.NONE;
-    }
-  }
+      extends KotlinLibraryDescription.CoreArg, JavaTestDescription.CoreArg {}
```
styurin: by the way, you are bringing more by using JavaTestDescription.CoreArg, it inherits `JavaLibraryDescription.CoreArg`. The proper way is to create an abstract JvmBasedTestArg<T> where T is the type of library. You would be able to use it like this: 

    extends KotlinLibraryDescription.CoreArg, JvmBasedTestArg<KotlinLibraryDescription.CoreArg> {}

kageiit: `KotlinLibraryDescription.CoreArg` already inherits from `JavaLibraryDescription.CoreArg`. So this change doe snot bring in anything more

kageiit: @styurin any updates on this?

PR #1455, opened 2017-08-08
Decouple proguard application logic from package type

kageiit:
Decoupling proguard application from the package type as there are cases when one would want to apply proguard on a debug build for debugging purposes with a different configuration than release

### DIFF
--- a/src/com/facebook/buck/android/AndroidBinaryDescription.java
+++ b/src/com/facebook/buck/android/AndroidBinaryDescription.java
@@ -207,7 +207,7 @@ public class AndroidBinaryDescription
       }
 
       ProGuardObfuscateStep.SdkProguardType androidSdkProguardConfig =
-          args.getAndroidSdkProguardConfig().orElse(ProGuardObfuscateStep.SdkProguardType.DEFAULT);
+          args.getAndroidSdkProguardConfig().orElse(ProGuardObfuscateStep.SdkProguardType.NONE);
 
       // If the old boolean version of this argument was specified, make sure the new form
       // was not specified, and allow the old form to override the default.
@@ -223,7 +223,7 @@ public class AndroidBinaryDescription
         androidSdkProguardConfig =
             args.getUseAndroidProguardConfigWithOptimizations().orElse(false)
                 ? ProGuardObfuscateStep.SdkProguardType.OPTIMIZED
-                : ProGuardObfuscateStep.SdkProguardType.DEFAULT;
+                : ProGuardObfuscateStep.SdkProguardType.NONE;
       }
 
       EnumSet<ExopackageMode> exopackageModes = EnumSet.noneOf(ExopackageMode.class);
@@ -239,9 +239,15 @@ public class AndroidBinaryDescription
       DexSplitMode dexSplitMode = createDexSplitMode(args, exopackageModes);
 
       PackageType packageType = getPackageType(args);
+      boolean shouldProguard =
+          args.getProguardConfig().isPresent()
+              || (args.getAndroidSdkProguardConfig().isPresent()
+                  && !ProGuardObfuscateStep.SdkProguardType.NONE.equals(
+                      args.getAndroidSdkProguardConfig().get()));
+
       boolean shouldPreDex =
           !args.getDisablePreDex()
-              && PackageType.DEBUG.equals(packageType)
+              && !shouldProguard
               && !args.getPreprocessJavaClassesBash().isPresent();
 
       ResourceFilter resourceFilter = new ResourceFilter(args.getResourceFilter());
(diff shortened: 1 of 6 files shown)


kageiit:
cc @dreiss

bertmaher:
Seems like a reasonable idea to me.  I'd like a like a little time to fix up our build configs at FB before merging this, so I don't accidentally start making all our devs run PG :).  Shouldn't take too long (I hope).

kageiit:
This seems to break some AndroidBinaryIntegrationTests. @bertmaher @dreiss please comment if the code looks good and I can fixup the tests

bertmaher:
The code lgtm, thanks!

kageiit:
Fixed the tests. Should be good to go

bertmaher:
Looks like there are still a few failed tests, can you take a look:
TESTS FAILED: 13 FAILURES
Failed target: //test/com/facebook/buck/android:aar-integration
FAIL com.facebook.buck.android.AarWithLibsIntegrationTest
FAIL com.facebook.buck.android.AndroidPrebuiltAarIntegrationTest
FAIL com.facebook.buck.android.AndroidBinaryIntegrationTest

kageiit:
Sure. Those did not show up on the previous run. I will take a look

kageiit:
Tests should be fixed now

kageiit:
Also tested this PR against our internal apps. Works great!

bertmaher:
One more test - can you take a look at AndroidBinaryIntegrationTest.testProguardDontObfuscateGeneratesMappingFile?

After that this looks good to go.

kageiit:
Updated this PR to fix tests that caught a couple of issues. The default proguard config is NONE now as we do not check for the release packageType anymore.

We only try to proguard if the default proguard config is not NONE or if there is an explicit proguard config specified

kageiit:
Hmm. Something is up with this change. I think it is not working correctly. Ill report back. Please hold off on importing

kageiit:
Ok PR updated. It should be good now. I forgot to change a place where we were checking for whether we should preDex or not before

kageiit:
Need to tweak one more place to decouple `NO_LOCALS` from the package type in dexing

kageiit:
Ok fixed the last place around dexing as well

kageiit:
@bertmaher This should be good to go for import now :)

kageiit:
@bertmaher any update on this? :)

kageiit:
@bertmaher Fixed the unused field usage. Ran `ant lint` to confirm no more issues

kageiit:
I could have sworn `ant` worked for me locally. I'll try to build clean this time :|

kageiit:
Ok `ant clean default` works now. @bertmaher

bertmaher:
Argh.  Once I got this running on our CI, I see that we have a number of builds that appear to care about how this setting gets used.  I really want to get those fixed before merging this, but I'm going to be out for a few days.  Lemme see if I can get someone else to take this PR so we can get it in quicker.

kageiit:
@bertmaher any update on this?

src/com/facebook/buck/android/AndroidBinaryDescription.java
```diff
@@ -239,9 +239,15 @@ public BuildRule createBuildRule(
       DexSplitMode dexSplitMode = createDexSplitMode(args, exopackageModes);
 
       PackageType packageType = getPackageType(args);
+      boolean shouldProguard =
```
bertmaher: Is it possible to unify this with AndroidBinaryBuildable.shouldProguard()

kageiit: Just noticed this. It might be possible, but the logic is fairly trivial and not sure if its worth DRY'ing up

bertmaher:
Hey @kageiit, as I was fixing up some of our targets I noticed that this diff makes android_sdk_proguard_config default to "none" instead of "default".  I hadn't thought through the implications until I got bit by it, and I'm worried about other OSS users suddenly getting sub-optimal or wrong results by not using the default PG options.  What do you think?  Would it be reasonable to leave "default" alone, but allow a target to disable PG by explicitly setting it to "none?

kageiit:
> Hey @kageiit, as I was fixing up some of our targets I noticed that this diff makes android_sdk_proguard_config default to "none" instead of "default". I hadn't thought through the implications until I got bit by it, and I'm worried about other OSS users suddenly getting sub-optimal or wrong results by not using the default PG options. What do you think? Would it be reasonable to leave "default" alone, but allow a target to disable PG by explicitly setting it to "none?

Hmm that seems fine, but the current checks for `shouldProguard` will not work then. If the default is always going to be the standard android skd rules, how do we know if the binary needs to be proguarded or not? It would require another flag like `proguard = True` which makes it a weirder api

bertmaher:
Ah, I see your point.  I was thinking that for non-PG builds you'd add android_sdk_proguard_config = "none" to your android_binary rule, and leave well enough alone for "normal" builds.  But it's probably better to explicitly turn PG on, rather than turn it off :).

Anyways, I'm almost done updating our internals... should be later today (I hope).

PR #1464, opened 2017-08-15
Add support to merge service Descriptor files in java_binary rules

### LINKED ISSUE #1460: Add support to merge service Descriptor files in java_binary rules
Buck currently provides an option to merge manifests when building java_binary rules, but I am not seeing an option to merge service manifest files as described in http://imperceptiblethoughts.com/shadow/#controlling_jar_content_merging

```
Java libraries often contain service descriptors files in the META-INF/services directory of the JAR. A service descriptor typically contains a line delimited list of classes that are supported for a particular service. At runtime, this file is read and used to configure library or application behavior.

Multiple dependencies may use the same service descriptor file name. In this case, it is generally desired to merge the content of each instance of the file into a single output file.
```

kageiit:
Fixes #1460

### DIFF
--- a/src/com/facebook/buck/zip/JarBuilder.java
+++ b/src/com/facebook/buck/zip/JarBuilder.java
@@ -20,15 +20,21 @@ import static com.facebook.buck.zip.ZipOutputStreams.HandleDuplicates.APPEND_TO_
 
 import com.facebook.buck.util.HumanReadableException;
 import com.facebook.buck.util.RichStream;
+import com.google.common.base.Charsets;
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.io.ByteStreams;
+import com.google.common.io.CharStreams;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


kageiit:
cc @jkeljo

src/com/facebook/buck/zip/JarBuilder.java
```diff
@@ -235,6 +254,20 @@ private void addEntryToJar(JarEntrySupplier entrySupplier, CustomJarOutputStream
       return;
     }
 
+    // Collect all services together for later merging and addition to the output jar
+    if (isService(entryName)) {
+      try (InputStream entryInputStream = entrySupplier.getInputStreamSupplier().get()) {
+        if (entryInputStream != null) {
+          // Null stream means a directory
+          Set<String> existingServices = services.getOrDefault(entryName, new LinkedHashSet<>());
```
jkeljo: Use of `getOrDefault` here is going to create a garbage `LinkedHashSet` every time. `computeIfAbsent` is better. You also won't need the `putIfAbsent` afterwards.

src/com/facebook/buck/zip/JarBuilder.java
```diff
@@ -235,6 +254,20 @@ private void addEntryToJar(JarEntrySupplier entrySupplier, CustomJarOutputStream
       return;
     }
 
+    // Collect all services together for later merging and addition to the output jar
+    if (isService(entryName)) {
+      try (InputStream entryInputStream = entrySupplier.getInputStreamSupplier().get()) {
+        if (entryInputStream != null) {
+          // Null stream means a directory
```
jkeljo: Minor: If you have `isService` also confirm that the name does not end in `/` (i.e., is not a directory) you can `Preconditions.checkNotNull` in the initializer rather than having this check.

kageiit:
@jkeljo I addressed your comments and added more test cases

Issue #1465, opened 2017-08-17
debian packaging


nikhedonia:
Would be nice if you would provide a selfcontained debian package for buck.

Many users of Buckaroo complain that its difficult to install Buck on debian jessie as it requires to install openjdk-8-jre which is not included in the default repository.

ttsugriy:
@nikhedonia, I just realized that your PR also packages buckd pex which in effect increases deb package size by 2x. Is this intentional? Do your users use buckd directly?

nikhedonia:
@ttsugriy no I don't know if anyone does. I just wanted to make sure that everything is included.
How usable is buck without buckd ?
It is definetly possible to split the debian package into two...

ttsugriy:
I see, @nikhedonia. Oh well, let's give an existing [release](https://github.com/facebook/buck/releases/tag/v2017.09.04.02) a try and see how it goes :) You can now download deb files from there.

PR #1466, opened 2017-08-17
adds genrule for packaging debian

nikhedonia:
#1465 adds genrule to create a debian package.
TODO: Edit Changelog
TODO: Automate generation of bucks version number in debian/buck.equivs
TODO: bundle with openjdk (see https://github.com/LoopPerfect/buckaroo/blob/master/debian/buckaroo-with-jdk.equivs)


programs/BUCK
```diff
@@ -25,6 +25,7 @@ python_binary(
     deps = [
         ":buck_version",
     ],
+    visibility = ['PUBLIC'],
```
ttsugriy: please use a more specific include here and below.

BUCK
```diff
@@ -0,0 +1,22 @@
+genrule(
```
ttsugriy: please move this and everything relating to debian packaging into `scripts/packages/debian`

nikhedonia: how can I access files in the project-root like LICENSE from scripts/packages/debian ?

ttsugriy: @nikhedonia, you should be able to use [export_file](https://buckbuild.com/rule/export_file.html) to export license file and add a dependency on it just like on any other rule.

PR #1467, opened 2017-08-18
Change Android compiler output from classes dir to output jar

jbarr21:
This diff is a followup to #1412 since the IntelliJ Layout Preview can load classes from the compiler output using the output jar path. This makes it so that you do not necessarily need to change the `spool_to_jar` buckconfig option to `intermediates_to_disk` in order to get Layout Preview working.

### DIFF
--- a/src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
+++ b/src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
@@ -74,6 +74,11 @@ public class IjProjectTemplateDataPreparer {
   private static final String ASSETS_FOLDER_TEMPLATE_PARAMETER = "asset_folder";
   private static final String PROGUARD_CONFIG_TEMPLATE_PARAMETER = "proguard_config";
   private static final String RESOURCES_RELATIVE_PATH_TEMPLATE_PARAMETER = "res";
+  private static final String JAR_FILE_EXTENSION = ".jar";
+  private static final String MODULE_DIR = "$MODULE_DIR$/";
+  private static final String FILE_PREFIX = "file://";
+  private static final String JAR_PREFIX = "jar://";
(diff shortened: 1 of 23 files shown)


src/com/facebook/buck/ide/intellij/templates/ij-module.st
```diff
@@ -45,8 +45,12 @@
     %if(androidFacet.compiler_output_path)%inherit-compiler-output="false"%else%inherit-compiler-output="true"%endif%>
 %endif%
     %if(androidFacet.compiler_output_path)%
+    %if(androidFacet.is_compiler_output_path_jar)%
+    <output url="jar://$MODULE_DIR$%androidFacet.compiler_output_path%!/" />
```
styurin: How about instead of increasing complexity of the template move URL construction to java code and pass one parameter here:

```
    <output url="%androidFacet.compiler_output_path%" />
```

src/com/facebook/buck/ide/intellij/IjProjectTemplateDataPreparer.java
```diff
@@ -515,18 +520,27 @@ private void addAndroidResourcePaths(
...
       Map<String, Object> androidProperties, IjModule module, Path moduleBasePath) {
     // The compiler output path is relative to the project root
     Optional<Path> compilerOutputPath = module.getCompilerOutputPath();
     if (compilerOutputPath.isPresent()) {
-      Path relativeCompilerOutputPath = moduleBasePath.relativize(compilerOutputPath.get());
-      androidProperties.put(
-          "compiler_output_path",
-          "/" + MorePaths.pathWithUnixSeparators(relativeCompilerOutputPath));
+      String fullCompilerOutputPath;
+      String relativeOutputPath =
```
styurin: Can you move this functionality to `IjProjectPaths`?

PR #1486, opened 2017-08-28
Fix shouldRefuseToRunAppleTestIfXctestNotPresent

robbertvanginkel:
This currently fails if the xctool is available in the PATH, because
it auto resolves. Making the path fake makes this test consistently
fail regardless of environment.


robbertvanginkel:
Verified this works locally. Put this in the execution command instead of the `.buckconfig` is because the test folder is shared across more tests.

The import re-ordering was part of some project style.

ttsugriy:
thank you for your contribution @robbertvanginkel!

PR #1490, opened 2017-08-30
Add an option to disallow all duplicates in java_binary

kageiit:
When processing jars output from `java_binary` rules with tools like jarjar, it is not possible to remove all the entries from the jar (non class files) that buck added duplicate entries for.

This adds an option for consumers to treat duplicate files the same (`.class` and non class files) and not re-add them if they are already added

### DIFF
--- a/src/com/facebook/buck/jvm/java/AbstractJarParameters.java
+++ b/src/com/facebook/buck/jvm/java/AbstractJarParameters.java
@@ -36,6 +36,11 @@ abstract class AbstractJarParameters {
     return false;
   }
 
+  @Value.Default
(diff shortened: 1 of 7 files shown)


kageiit:
> should this be a configuration option in .buckconfig instead of an attribute of the java_library?

I am open to making it an option in `.buckconfig`. I felt doing it per target was more flexible. Let me know if you feel otherwise and I can change it

ttsugriy:
I guess I'm fine either way, but please add a test :)

kageiit:
@ttsugriy I added tests and rebased

PR #1491, opened 2017-08-31
Add res_folder_relative_path property to Android IML file

jbarr21:
This diff sets the RES_FOLDER_RELATIVE_PATH property of the Android Facet. The resource folder path is used by the Android Layout Preview to load themes and styles in order to properly visualize and load views into the Preview window. The logic for which path to use for the module is as follows:
- if only one path, use it
- if a module's resource rule is labeled as preferred, then use that one
- otherwise, choose the first one

### DIFF
--- a/src/com/facebook/buck/ide/intellij/IjProjectPaths.java
+++ b/src/com/facebook/buck/ide/intellij/IjProjectPaths.java
@@ -55,6 +55,21 @@ public class IjProjectPaths {
     }
   }
 
+  /**
+   * @param path path to folder.
+   * @param moduleLocationBasePath path to the location of the .iml file.
+   * @return a path, relative to the module .iml file location describing a folder without the
+   *     IntelliJ format.
+   */
+  static String toRelativeString(Path path, Path moduleLocationBasePath) {
+    String moduleRelativePath = moduleLocationBasePath.relativize(path).toString();
+    if (moduleRelativePath.isEmpty()) {
+      return "";
+    } else {
+      return "/" + MorePaths.pathWithUnixSeparators(moduleRelativePath);
+    }
+  }
+
   String toProjectDirRelativeString(Path repoRelativePath) {
     return MorePaths.pathWithUnixSeparators(
         projectRootPath.toAbsolutePath().relativize(repoRelativePath.toAbsolutePath()));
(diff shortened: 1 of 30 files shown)


src/com/facebook/buck/ide/intellij/IjProjectPaths.java
```diff
@@ -55,6 +55,21 @@ static String toModuleDirRelativeString(Path path, Path moduleLocationBasePath)
...
+   * @param moduleLocationBasePath path to the location of the .iml file.
+   * @return a path, relative to the module .iml file location describing a folder without the
+   *     IntelliJ format.
+   */
+  static String toRelativeString(Path path, Path moduleLocationBasePath) {
+    String moduleRelativePath = moduleLocationBasePath.relativize(path).toString();
+    if (moduleRelativePath.isEmpty()) {
+      return "";
+    } else {
+      return "/" + MorePaths.pathWithUnixSeparators(moduleRelativePath);
```
kageiit: Use File separator instead of `/`

src/com/facebook/buck/ide/intellij/lang/android/AndroidResourceModuleRule.java
```diff
@@ -30,11 +30,14 @@
 import com.facebook.buck.rules.TargetNode;
 import com.google.common.collect.ImmutableSet;
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
 public class AndroidResourceModuleRule extends AndroidModuleRule<AndroidResourceDescriptionArg> {
```
kageiit: We need a preconditions check to ensure there is only one resource rule labelled `preferred_res_folder` per `BUCK` file/iml

src/com/facebook/buck/ide/intellij/IjProjectPaths.java
jbarr21: I used what is used by ProjectPaths.toModuleDirRelativeString(), which this method is derived from

src/com/facebook/buck/ide/intellij/lang/android/AndroidResourceModuleRule.java
kageiit: Can you also add a test that we throw an error in this case please?

styurin:
Could you add a stand-alone test to demonstrate the changes instead of modifying a test responsible for aggregation?

jbarr21:
Abandoning for now in favor of waiting for https://android-review.googlesource.com/c/platform/tools/adt/idea/+/474539 to get merged and picked up by a new Jetbrains release.

PR #1503, opened 2017-09-05
Optimize apple imports

robbertvanginkel:
Run intellij's 'Optimize imports' on `test/com/facebook/buck/apple/` and `src/com/facebook/buck/apple/`.

The default import order seems to have changed for a lot of these files. Auto formatting them all to make future changes less noisy.

### DIFF
--- a/src/com/facebook/buck/apple/AbstractAppleBundleDestinations.java
+++ b/src/com/facebook/buck/apple/AbstractAppleBundleDestinations.java
@@ -19,9 +19,9 @@ package com.facebook.buck.apple;
 import com.facebook.buck.rules.RuleKeyAppendable;
 import com.facebook.buck.rules.RuleKeyObjectSink;
 import com.facebook.buck.util.immutables.BuckStyleImmutable;
+import org.immutables.value.Value;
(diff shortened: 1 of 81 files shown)


robbertvanginkel:
Not sure if you take PR's like this, but I've been working on some PR's related to apple things and I find the import re-ordering quite noisy when reviewing the changes. This is an attempt at auto formatting them to conform to the project's defaults, so that later changes are cleaner.

Let me know if this is an acceptable change.

yiding:
We use google-java-format, which sometimes organize imports differently vs intellij. I noticed that we don't have a google-java-format jar in the open source repo (it's stripped out), so I'm going to move that somewhere where people can run it.

If you want to get a head start, we are using version 1.3, which you can grab from here: https://github.com/google/google-java-format/releases/tag/google-java-format-1.3

robbertvanginkel:
Ah, I was incorrectly assuming that the file at [`.idea/codeStyleSettings.xml`](https://github.com/facebook/buck/blame/master/) was up to date, but it seems like the thing I was using is 4 years old.

I'll start using `google-java-format` then/

PR #1507, opened 2017-09-05
Add an argument for scheme pre and post actions in Xcode

robbertvanginkel:
Xcode currently allows you to execute a shell step when you start an stop executing in a scheme. I would like buck to be able to generate a project with this configured.

I have an implementation of this on an internal fork, it works but the interface isn't very satisfying. Although the rule configuration seems reasonable:
```
xcode_workspace_config(
    ...
    additional_scheme_actions = {
      'Build': {
          'Pre': ['echo start'],
          'Post': ['echo stop'],
      }
    }
    ...
)
```
The current argument type that is coerced seems too long
```
Optional<ImmutableMap<SchemeActionType, ImmutableMap<XCScheme.AdditionalActions, ImmutableList<String>>>>
```
Any suggestions on how to handle this? Is this acceptable in java? Should I try to achieve this outside of buck?

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
@@ -25,6 +25,7 @@ import com.facebook.buck.log.Logger;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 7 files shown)


ttsugriy:
@dinhviethoa, do you happen to have some thoughts about this?

src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
```diff
@@ -107,6 +115,7 @@ public SchemeGenerator(
     this.remoteRunnablePath = remoteRunnablePath;
     this.actionConfigNames = actionConfigNames;
     this.targetToProjectPathMap = targetToProjectPathMap;
+    this.additionalSchemeActions = additionalSchemeActions;
```
dinhviethoa: You're not using this value to write data to the file yet?

src/com/facebook/buck/apple/xcode/XCScheme.java
```diff
@@ -75,6 +75,12 @@ public String getName() {
     return archiveAction;
   }
 
+  public enum AdditionalActions {
+    PRE,
```
dinhviethoa: It looks too short :/
Maybe we could use PRE_ACTION nad POST_ACTION.

dinhviethoa:
Is it a work in progress?

dinhviethoa:
About the argument type, it's probably fine as is.

src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
robbertvanginkel: No not yet, wanted to get feedback first on if the type for `additionalSchemeActions` would be acceptable.

robbertvanginkel:
I have some other commits that make the full implementation, but I wanted to share this first to get feedback on the long argument type. I'll update with the full implementation.

robbertvanginkel:
@dinhviethoa its been a while but I finally found time to rebase this. I incorporated the feedback on the short `AdditionalActions` enum names. It now also includes the serialisation code and appropriate tests.

milend:
LGTM

Issue #1292, opened 2017-04-14
Does NdkCxxPlatforms need to statically define versions?


hzsweers:
This is a recurring issue for us, as [we have to manually update every time there's a new major NDK release](https://github.com/facebook/buck/pulls?q=is%3Apr+author%3Akageiit+ndk+is%3Aclosed) and it blocks local development until we update buck or go through the tedium of downgrading the NDK version on the system. Can we make the number matching accept any upper bound with the caveat of being unsupported/YMMV?

CC @kageiit 

https://github.com/facebook/buck/blob/master/src/com/facebook/buck/android/NdkCxxPlatforms.java#L102-L111

Coneko:
I believe this is related to the fact google has changed how the NDK and SDK identify versions several times: both in how to find the version string, and how the version string is to be interpreted.

Ideally we'd have the users specify what version is installed where in the config, rather than doing the whole find installed versions ourselves voodoo, but that is probably not happening anytime soon.

cc @aiked, @marcinkosiba

hzsweers:
This broke again on NDK 16

```
Generating project: finished in 26.6 sec
[2017-09-06 17:43:59.012][error][command:null][tid:15][com.facebook.buck.cli.Main] Uncaught exception at top level
java.lang.AssertionError: Unknown ndk version: 16.0.4293906-beta1
	at com.facebook.infer.annotation.Assertions.assertCondition(Assertions.java:72)
	at com.facebook.buck.android.NdkCxxPlatforms$NdkCxxToolchainPaths.<init>(NdkCxxPlatforms.java:1128)
	at com.facebook.buck.android.NdkCxxPlatforms$NdkCxxToolchainPaths.<init>(NdkCxxPlatforms.java:1106)
	at com.facebook.buck.android.NdkCxxPlatforms.build(NdkCxxPlatforms.java:572)
	at com.facebook.buck.android.NdkCxxPlatforms.getPlatforms(NdkCxxPlatforms.java:279)
	at com.facebook.buck.android.NdkCxxPlatforms.getPlatforms(NdkCxxPlatforms.java:244)
	at com.facebook.buck.android.NdkCxxPlatforms.getPlatforms(NdkCxxPlatforms.java:221)
	at com.facebook.buck.android.toolchain.NdkCxxPlatformsProvider.create(NdkCxxPlatformsProvider.java:58)
	at com.facebook.buck.rules.KnownBuildRuleTypes.createBuilder(KnownBuildRuleTypes.java:280)
[...]

PR #1511, opened 2017-09-07
Update for NDK 16

hzsweers:
Repeat of #1292


hzsweers:
I thought about adding 17-20 while I was at it :)

ttsugriy:
Well, hopefully by version 17, someone would generalize this method a bit, so that we don't run out of horizontal space :) Thank you for adding version 16 support, @hzsweers!

PR #1517, opened 2017-09-09
Updated code-coverage-format in test task to support multiple formats…

romanoid:
Updated code-coverage-format in test task to support multiple formats.

- Change made in the way existing users will see no difference.
- Verified change locally.

### DIFF
--- a/src/com/facebook/buck/cli/TestCommand.java
+++ b/src/com/facebook/buck/cli/TestCommand.java
@@ -67,26 +67,37 @@ import com.facebook.buck.util.concurrent.ConcurrencyLimit;
 import com.facebook.buck.versions.VersionException;
 import com.facebook.infer.annotation.SuppressFieldNotInitialized;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Iterables;
+import org.kohsuke.args4j.CmdLineException;
+import org.kohsuke.args4j.CmdLineParser;
+import org.kohsuke.args4j.Option;
+import org.kohsuke.args4j.OptionDef;
+import org.kohsuke.args4j.spi.Messages;
+import org.kohsuke.args4j.spi.OptionHandler;
+import org.kohsuke.args4j.spi.Parameters;
+import org.kohsuke.args4j.spi.Setter;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Comparator;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import javax.annotation.Nullable;
-import org.kohsuke.args4j.Option;
 
 public class TestCommand extends BuildCommand {
 
@@ -101,8 +112,13 @@ public class TestCommand extends BuildCommand {
   @Option(name = "--code-coverage", usage = "Whether code coverage information will be generated.")
   private boolean isCodeCoverageEnabled = false;
 
-  @Option(name = "--code-coverage-format", usage = "Format to be used for coverage")
-  private CoverageReportFormat coverageReportFormat = CoverageReportFormat.HTML;
+  @Option(
+    name = "--code-coverage-format",
+    usage = "Comma separated Formats to be used for coverage",
+    handler = CoverageReportFormatsHandler.class
+  )
+  private CoverageReportFormat[] coverageReportFormats =
+      new CoverageReportFormat[] {CoverageReportFormat.HTML};
 
   @Option(name = "--code-coverage-title", usage = "Title used for coverage")
   private String coverageReportTitle = "Code-Coverage Analysis";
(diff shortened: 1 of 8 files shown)


src/com/facebook/buck/cli/TestCommand.java
```diff
@@ -101,8 +111,13 @@
...
   private boolean isCodeCoverageEnabled = false;
 
-  @Option(name = "--code-coverage-format", usage = "Format to be used for coverage")
-  private CoverageReportFormat coverageReportFormat = CoverageReportFormat.HTML;
+  @Option(
+    name = "--code-coverage-format",
+    usage = "Comma separated Formats to be used for coverage",
+    handler = CoverageReportFormatsHandler.class
+  )
+  private CoverageReportFormat[] coverageReportFormats =
```
ttsugriy: when working with enums it's much more convenient to use `EnumSet`

romanoid: Thanks, switched to EnumSet where possible. unfortunately I were not able to make args4j option to work with anything but array though.

romanoid:
Can you please suggest if this travis failure is expected or it is something I need to take care of: 

"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received"

src/com/facebook/buck/cli/TestCommand.java
```diff
@@ -252,7 +268,7 @@ private TestRunningOptions getTestRunningOptions(CommandRunnerParams params) {
             .setShufflingTests(isShufflingTests)
             .setPathToXmlTestOutput(Optional.ofNullable(pathToXmlTestOutput))
             .setPathToJavaAgent(Optional.ofNullable(pathToJavaAgent))
-            .setCoverageReportFormat(coverageReportFormat)
+            .setCoverageReportFormats(EnumSet.copyOf(Arrays.asList(coverageReportFormats)))
```
ttsugriy: this will blow up if coverageReportFormats is an empty array. You can refer to https://stackoverflow.com/questions/22886287/enumset-from-array-shortest-variant for examples of how to avoid it.

src/com/facebook/buck/cli/TestCommand.java
```diff
@@ -691,4 +707,60 @@ public void printUsage(PrintStream stream) {
   public String getShortDescription() {
     return "builds and runs the tests for the specified target";
   }
+
+  /**
+   * args4j does not support parsing repeated (or delimiter separated) Enums by default.
```
ttsugriy: is it possible to use `multiValued` attribute instead? I personally would prefer having something like
`--code-coverage-format=HTML --code-coverage-format=XML` to a comma delimited string.

ttsugriy: I was referring to http://args4j.kohsuke.org/apidocs/org/kohsuke/args4j/Argument.html#multiValued()

ttsugriy:
@romanoid, I'm not sure if the issue is related to your changes, but I'll test it internally and will let you know how it goes.

ttsugriy:
@romanoid, our internal testing didn't reveal any issues, so please consider using multiValued argument instead of option to avoid potential exceptions and make the API easier to use and you'll be all set :)

src/com/facebook/buck/cli/TestCommand.java
romanoid: Incidentally the way code written right now array will never be empty, but extra safety measure will never hurt, updated the code according to your suggestion.

romanoid:
@ttsugriy Can you please elaborate on suggestion about "multiValued argument instead of option"?

src/com/facebook/buck/cli/TestCommand.java
romanoid: multiValued attribute support was removed while ago: https://github.com/kohsuke/args4j/issues/24

Let me try updating CoverageReportFormatsHandler to parse multiple values

ttsugriy:
oh, I'm sorry, you're right. But https://github.com/kohsuke/args4j/issues/24#issuecomment-102079165 suggested using `StringArrayOptionHandler`,  but in your case it should probably be http://args4j.kohsuke.org/apidocs/org/kohsuke/args4j/spi/EnumOptionHandler.html Would that work?

src/com/facebook/buck/cli/TestCommand.java
```diff
@@ -259,6 +259,11 @@ public int getNumTestManagedThreads(BuckConfig buckConfig) {
   }
 
   private TestRunningOptions getTestRunningOptions(CommandRunnerParams params) {
+    // this.coverageReportFormats should never be empty, but doing this to avoid problems with
```
ttsugriy: if it should never be empty, please add a check and throw a human readable exception telling users that they should pass the parameter :) Previously it would just throw an enumset exception and now it's going to be completely silent :)

romanoid: CoverageReportFormatsHandler will throw the human readable option exception in case of it being empty, that's why it will never be empty by this point.

src/com/facebook/buck/cli/TestCommand.java
romanoid: All right, I experimented a little, "--code-coverage-format=HTML --code-coverage-format=XML" is not supported, what you can do though is "--code-coverage-format HTML XML", but I'd argue it is even worse than csv.

romanoid:
http://args4j.kohsuke.org/apidocs/org/kohsuke/args4j/spi/EnumOptionHandler.html works only for single Enum argument and it's implemented in a way of framework supplying it special value, I attempted to implement generic repeated enum handler, but current args4j does not support it.

src/com/facebook/buck/cli/TestCommand.java
romanoid: Also maybe I'm not configuring output verbosity correctly, but if I throw any exception from this context, buck just silently exits, that's why I Option handler returning user facing is preferable

romanoid:
Rebased + squished the commits addressing suggestion to original commit

romanoid:
Actually I found 1 side issue with the parameter handling, please give me a second fixing it

src/com/facebook/buck/cli/TestCommand.java
```diff
@@ -691,4 +712,60 @@ public void printUsage(PrintStream stream) {
...
+        } else {
+          throw new CmdLineException(
+              owner, Messages.ILLEGAL_OPERAND, params.getParameter(-1), invalidFormats);
+        }
+      }
+
+      for (CoverageReportFormat format : formats) {
+        setter.addValue(format);
+      }
+      return 1;
```
romanoid: this was formats.size() in one of the previous revisions, which is incorrect. For csv arg we always consume only single one.

src/com/facebook/buck/cli/TestCommand.java
ttsugriy: awesome, that sounds good!

PR #1521, opened 2017-09-11
Make scala libraries not build against abi jars

kageiit:
Similar to kotlin


kageiit:
cc @jkeljo seems like kotlin has the similar issues to kotlin

src/com/facebook/buck/jvm/scala/ScalaLibraryBuilder.java
```diff
@@ -51,6 +51,7 @@
         cellRoots,
         javaBuckConfig);
     this.scalaBuckConfig = scalaBuckConfig;
+    setCompileAgainstAbis(false);
```
ttsugriy: please add a comment explaining why this is necessary

kageiit: Added a comment to point to https://github.com/facebook/buck/issues/1386

jkeljo:
Not terribly surprising. Thanks for the PR, we'll get it in.

Issue #1520, opened 2017-09-11
genrule's query_outputs macro should allow inputs to be found


shs96c:
Using the `inputs` function. This would allow one to find all the source files of a target without needing to do anything crazy.

asp2insp:
Can you give an example of how you would like this to work and what you would like to accomplish? Right now, `query_outputs` and `query_targets` explicitly only support fetching lists of build targets, in order to build a dependency graph, so anything which is not a build target cannot be returned.

Internally, we've often solved problems related to sharing sub-structure of rules by defining a macro. Would this work in your case?

shs96c:
Sure. I want to build a source jar containing all the sources of files within a given part of my source tree (basically, third party deps are in a `//third_party` folder and source in `//java`). The jar crosses maven boundaries, so `#maven,src` won't work, but I can find them using:

`buck query "filter(\.java, inputs(filter(//java/, deps(//java/client/src/org/openqa/selenium:client-combined))))"`

It'd be easy to put that into a genrule that then strips the prefix of the root of the project (possibly using the `$(location)` macro, and magically have everything work.

asp2insp:
That makes sense. I don't think `query_outputs` is the correct vessel for this, but perhaps a new query macro, ie `query_attributes` or `query_inputs` macro would serve. Happy to help with a PR. It should be straightforward to implement a new macro which extends: [QueryMacroExpander](https://github.com/facebook/buck/blob/master/src/com/facebook/buck/rules/macros/QueryMacroExpander.java)

PR #1523, opened 2017-09-12
Add a `query_paths` macro

shs96c:
This allows a genrule to find paths from source rules.

### DIFF
--- a/src/com/facebook/buck/rules/BUCK
+++ b/src/com/facebook/buck/rules/BUCK
@@ -47,11 +47,6 @@ java_immutables_library(
     ],
 )
 
-genrule(name = "foo",
-  out = "foo.txt",
-  cmd = "echo $(query_paths 'inputs(:source_path)') > $OUT",
-)
-
 java_immutables_library(
     name = "archive_member_source_path",
     srcs = [
@@ -588,18 +583,5 @@ java_library(
         "//src/com/facebook/buck/model:model",
         "//src/com/facebook/buck/parser:rule_pattern",
         "//third-party/java/guava:guava",
-    ],
-)
-
-genrule(
-    name = 'file',
-    out = "Foo.java",
-    cmd = "touch $OUT",
-)
-
-java_library(
-    name = 'with-deps',
-    srcs = [
-        ':file',
     ],
 )
--- a/src/com/facebook/buck/rules/BUCK
+++ b/src/com/facebook/buck/rules/BUCK
@@ -47,6 +47,11 @@ java_immutables_library(
     ],
 )
 
+genrule(name = "foo",
+  out = "foo.txt",
+  cmd = "echo $(query_paths 'inputs(:source_path)') > $OUT",
+)
+
 java_immutables_library(
     name = "archive_member_source_path",
     srcs = [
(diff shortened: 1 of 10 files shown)


shs96c:
Closes https://github.com/facebook/buck/issues/1520

src/com/facebook/buck/rules/BUCK
```diff
@@ -47,6 +47,11 @@ java_immutables_library(
     ],
 )
 
+genrule(name = "foo",
```
ttsugriy: changes to this build file should be part of an integration test - not production code ;)

src/com/facebook/buck/rules/macros/AbstractQueryPathsMacro.java
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.
+ */
+
+package com.facebook.buck.rules.macros;
+
+import com.facebook.buck.util.immutables.BuckStyleTuple;
+import org.immutables.value.Value;
+
+/**
+ * Class providing the type for macros used in `$(query_outputs ...)` macro strings. The
```
ttsugriy: did you mean `$(query_paths ...)`?

ttsugriy:
also added @asp2insp to take a look, since he's our query wizard :)

src/com/facebook/buck/rules/BUCK
asp2insp: If you want an existing test to model, here's where most of the genrule query tests live so far: https://github.com/facebook/buck/blob/master/test/com/facebook/buck/shell/GenruleDescriptionIntegrationTest.java and here's where their buck files are: https://github.com/facebook/buck/blob/master/test/com/facebook/buck/shell/testdata/genrule_classpath_filtering/BUCK.fixture

src/com/facebook/buck/rules/macros/QueryPathsMacroExpander.java
```diff
@@ -0,0 +1,131 @@
...
+
+    return precomputedWork
+        .results
+        .stream()
+        .map(
+            queryTarget -> {
+              // What we do depends on the input.
+              if (QueryBuildTarget.class.isAssignableFrom(queryTarget.getClass())) {
+                BuildRule rule = resolver.getRule(((QueryBuildTarget) queryTarget).getBuildTarget());
+                return Optional.ofNullable(rule.getSourcePathToOutput())
```
asp2insp: What's the use case for the mixed-mode outputs/inputs here?

src/com/facebook/buck/rules/BUCK
shs96c: Ha! Ooops. :) Rolled back to master's

src/com/facebook/buck/rules/macros/AbstractQueryPathsMacro.java
shs96c: Yes. Fixed.

src/com/facebook/buck/rules/macros/QueryPathsMacroExpander.java
shs96c: You want people to be able to use a query string that returns either rules or files and have it work in the same way, which makes it easier to iteratively futz with the query string as you tweak a build to get it Just Right.

Ideally, there would be just a single `query` macro that works no matter what it is you're trying to do. That's not where we are :)

styurin:
Please, re-format Java files with google-java-format: https://github.com/facebook/buck/blob/master/CONTRIBUTING.md#code-style

PR #1524, opened 2017-09-12
Allow zips to flatten their inputs to a single directory

shs96c:


test/com/facebook/buck/zip/testdata/zip-flatten/example/BUCK
```diff
@@ -0,0 +1,8 @@
+zip_file(
```
ttsugriy: in testdata `BUCK` files should not be used. Instead it has to be `BUCK.fixture`. I fixed it, but please keep this in mind in future.

PR #1525, opened 2017-09-12
Allow services to be merged when disallowing all duplicates

kageiit:
Services are treated specially just like manifest during merging and should be excluded from duplicate disallow logic

### DIFF
--- a/test/com/facebook/buck/jvm/java/JarBuilderTest.java
+++ b/test/com/facebook/buck/jvm/java/JarBuilderTest.java
@@ -32,6 +32,7 @@ import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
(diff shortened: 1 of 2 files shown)


test/com/facebook/buck/jvm/java/JarBuilderTest.java
```diff
@@ -85,53 +86,57 @@ public void testSortsEntriesFromAllContainers() throws IOException {
...
-                  .addEntry("META-INF/services/com.example.Foo2", "com.example.Bar4"))
-          .addEntryContainer(
-              container3
-                  .addEntry("META-INF/services/com.example.Foo2", "com.example.Bar3")
-                  .addEntry("META-INF/services/foo/bar", "bar"))
-          .createJarFile(tempFile.toPath());
-    }
 
-    try (JarFile jarFile = new JarFile(tempFile)) {
+    for (boolean shouldDisallowAllDuplicates : Arrays.asList(true, false)) {
```
ttsugriy: nit: how about parameterized tests? https://github.com/junit-team/junit4/wiki/parameterized-tests :)

kageiit: Sure. I can make the change

ttsugriy:
sorry about the delay. Looks like ant pmd is failing with
```
test/com/facebook/buck/jvm/java/JarBuilderTest.java:51:	This class name ends with Test but contains no test cases
```

You should be able to reproduce it with `ant pmd`. The problem is that this seems to be a bug in PMD, so I was trying to update PMD and check if it helps, but unfortunately even the latest version of PMD fails with the same error :(

ttsugriy:
ok, I gave up and reverted to using plain for loop instead of parameterized tests :( With JUnit5 it would be so much nicer... Sigh...

kageiit:
> ok, I gave up and reverted to using plain for loop instead of parameterized tests :( With JUnit5 it would be so much nicer... Sigh...

ouch. sorry to hear. Thanks for working around it though

ttsugriy:
no worries, @kageiit. I just felt bad for asking you to convert from for loop to parameterized test just to discover that it does not work :)

PR #1528, opened 2017-09-13
Make `QueryFileTarget` take a `SourcePath`

### LINKED ISSUE #1527: QueryFileTarget isn't ProjectFileSystem aware
Consequently, paths are relative to their cell roots, and aren't very helpful. They should be SourcePaths.

shs96c:
This allows them to be aware of cells, and to work when the
base directory is redirected.

Closes #1527


styurin:
Please, re-format java files with google-java-format.

shs96c:
So here's a question for you folks: how do we want to represent paths when we stringify them?

**Always absolute paths**
This has the advantage that you can always find the file, so you could perform a query within your cell to output a list of paths, and be able to use that list in another cell. It's also consistent (yay!). The downside is that unless folks are careful you can break hermeticity pretty easily. 

**Relative in the same cell, absolute otherwise**
In the common case, this leaves paths exactly as they are when printed to the console, but has the downside that the any paths returned by `query_paths` become unusable since there's no environment variable saying where the root of the project is.

**Relative to the cell**
I guess this looks nice when printed?

**What I'm going to do**
I've chosen to use "always absolute" in this PR, but to reformat the paths when printing to the console as part of `buck query`. This allows the two major use cases ("I want to use buck query to figure out which files have changed so I can run a targeted build", and "I want to build a source jar") to work.

asp2insp:
Cells are a pain. We'd like to eventually get rid of them, and in the magical world where they don't exist, the easiest transition would be from **Relative in the same cell, absolute otherwise**. However, it's also pretty easy to get from **absolute everywhere** to that point, so this course of action looks good to me :)

shs96c:
I wrote the initial cells code in buck. The idea was to support the ability to fragment a monorepo into meaningful chunks, each with their own `KnownBuildRuleTypes` (so fbandroid could target one kind of java_library, fbjava another, and so on) yet still retain all the advantages of a single tree. I left FB before we were able to get a plugin mechanism sorted out, or to really make use of cells.

The other idea behind cells was to allow folks to export parts of the tree, which would make Open Sourcing parts of the repo easy. Again, I didn't stay long enough to build the tooling to make that happen.

sdwilsh:
Facebook also uses cells pretty heavily now as part of BuckIt (using it for what @shs96c original meant it for), so @asp2insp, I suggest chatting with @philipjameson :D

LegNeato:
What's buckit? :-)

sdwilsh:
A poorly documented experimental thing, but it should let us start to 
build with Buck in more of our open source projects: 
https://github.com/facebookexperimental/buckit/

shs96c:
@asp2insp I fixed the tests :)

PR #1529, opened 2017-09-13
Allow src jars to be added to zips

shs96c:
This allows one to create traditional release zips for java projects, containing both the library code and the source code. Which is nice.


shs96c:
Not sure why this couldn't happen automatically as part of the pre-commit workflow if the process is automated by Google's tool, but I've reformatted it anyway.

ttsugriy:
@shs96c, it should be @styurin is investigating why it doesn't work :( Sorry for the inconveniences.

PR #1531, opened 2017-09-15
Add an support for environment variables in xcode_workspace_config

robbertvanginkel:
This allows users to set environment variables for a scheme using a new argument `environment_variables` in `xcode_workspace_config`. These can be set per scheme action and show up in generated xcode projects.

Example usage:
```
xcode_workspace_config(
    ...
    environment_variables = {
      'Launch': {
          'OS_ACTIVITY_MODE': 'disable',
      }
    }
    ...
)
```


asp2insp:
@dinhviethoa do you want to take a look at this?

src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
```diff
@@ -301,6 +327,13 @@ public static Element serializeTestAction(Document doc, XCScheme.TestAction test
       testableElem.appendChild(refElem);
     }
 
+    if (testAction.getEnvironmentVariables().isPresent()) {
+      testActionElem.setAttribute("shouldUseLaunchSchemeArgsEnv", "NO");
```
dinhviethoa: Is there any reason why you disable it here?
Maybe you could add a comment?

src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
```diff
@@ -336,18 +369,32 @@ public static Element serializeLaunchAction(Document doc, XCScheme.LaunchAction
...
+          serializeEnvironmentVariables(doc, launchAction.getEnvironmentVariables().get());
+      launchActionElem.appendChild(environmentVariablesElement);
+    }
+
     return launchActionElem;
   }
 
   public static Element serializeProfileAction(Document doc, XCScheme.ProfileAction profileAction) {
     Element profileActionElem = doc.createElement("ProfileAction");
+    profileActionElem.setAttribute("shouldUseLaunchSchemeArgsEnv", "YES");
```
dinhviethoa: Why do you have to set it explicitly here?

dinhviethoa:
Thanks for helping with it!

src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
robbertvanginkel: This is already present in the Test action. This is to mimic Xcode's default settings of having the same variables on test/profile as launch unless you explicitly specify otherwise.

src/com/facebook/buck/apple/project_generator/SchemeGenerator.java
robbertvanginkel: On line 316, this is set to yes by default. So by default the test action will have the Launch actions environment variables. If however, you specify custom variables for the test action, you should not be using the launch action's variables here.

In xcode, shouldUseLaunchSchemeArgsEnv controls a checkbox that greys out the custom environment variables. This is an override that is enabled by default, so to have the custom variables we need to make sure the checkbox is not enabled.

dinhviethoa:
All good! I think it's just a matter of adding comments now.

dinhviethoa:
It looks like coupleof tests in //test/com/facebook/buck/apple/xcode:xcode fail.

robbertvanginkel:
You can use `$ git diff HEAD~1 --color-words='\\w+|[^[:space:]]'` to get a better format of what changed in these one line xcschemes. In all cases it added `shouldUseLaunchSchemeArgsEnv="YES"` as an attribute to `ProfileAction`.

dinhviethoa:
Alright, last one request and I think that will be it: could you rebase your code?
We need all the fixes for Xcode 9 that are already in the repo before committing the code in the repo.

PR #1399, opened 2017-06-27
Enables mixed kotlin + java integration tests.

dsyang:
This also bumps the kotlin version to 1.1.3.
Fixes #1371

PR #1541, opened 2017-09-27
Bump the kotlin version to the latest (1.1.50)

cwoodwar6:
This does not yet cover #1371 since the first attempt at that seemed to
have issues: https://github.com/facebook/buck/pull/1399.

This is the first step towards getting kapt support.


kageiit:
:shipit:

cwoodwar6:
I actually expected this to fail the test `AndroidLibraryIntegrationTest#testAndroidKotlinLibraryExtraArgumentsCompilation` because the previous version did not have support for `-Xplugin` args.

kageiit:
@dsyang can we merge this please?

dsyang:
@cwoodwar6 it looks like you're right, appveyor's tests didn't fully run.  When running these tests locally there's I see an error message like: https://gist.github.com/dsyang/9e5dec03e5d1f5494b695fc805bceb48

cwoodwar6:
I'll get the test to pass. But are there any changes I can make to get the tests to run in appveyor?

sdwilsh:
I'm guessing we don't have the right dependencies installed there.

dsyang:
Right dependencies for appveyor not running the tests? I think it's not just that. Appveyor only runs one test: https://ci.appveyor.com/project/Facebook/buck/build/4833#L2387

sdwilsh:
Wow.  That is broken I think.  @ilya-klyuchnikov might know what's up with that.

dsyang:
Looking at the output of the stack trace, it appears theres a problem with how we're using reflection in `JarBackedReflectedKotlinc.java:130-154`. Specifically we're missing definitions in the `kotlin.reflect` package.

From googling the error to digging in other github issues, https://github.com/spring-projects/spring-boot/issues/9984#issuecomment-321195996 makes me think we need to also include the kotlin-reflect.jar in this commit since it's needed by kotlinc.

cwoodwar6:
@dsyang You were right about `kotlin-reflect.jar`. I also had to add `kotlin-stdlib.jar`.

ttsugriy:
also, please use google formatter to format your code. Looks like you've broken formatting for `test/com/facebook/buck/android/AndroidLibraryIntegrationTest.java`. In particular import line has moved to a wrong position.

PR #1543, opened 2017-09-27
Xcode project set LD_RUNPATH_SEARCH_PATHS with Swift

robbertvanginkel:
Xcode copies the `libswift*.dylib` into the `Frameworks` folder of the binary bundle when you have a swift dependency. To be able to load these libraries, this requires the `LC_RPATH` (controlled in Xcode by`LD_RUNPATH_SEARCH_PATHS`) to be set to that folder.

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -1483,6 +1483,8 @@ public class ProjectGenerator {
     HashSet<String> librarySearchPaths = new HashSet<>();
     librarySearchPaths.add("$BUILT_PRODUCTS_DIR");
     HashSet<String> ldRunpathSearchPaths = new HashSet<>();
+    ImmutableSet<PBXFileReference> swiftDeps =
+        collectRecursiveLibraryDependenciesWithSwiftSources(node);
 
     Stream.concat(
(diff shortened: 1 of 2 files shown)


dinhviethoa:
@milend could you take a look at this one?

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1560,13 +1562,16 @@ private void addPBXTargetDependency(PBXNativeTarget pbxTarget, BuildTarget depen
...
       // transitively.
-      ImmutableSet<PBXFileReference> swiftDeps =
-          collectRecursiveLibraryDependenciesWithSwiftSources(node);
       if (swiftDeps.size() > 0) {
         librarySearchPaths.add("$DT_TOOLCHAIN_DIR/usr/lib/swift/$PLATFORM_NAME");
       }
     }
 
+    if (swiftDeps.size() > 0) {
+      ldRunpathSearchPaths.add("@executable_path/Frameworks");
```
milend: @robbertvanginkel This would not work for macOS, we will need to use `@executable_path/../Frameworks` (same for the line below).

robbertvanginkel: Pushed a fix for this.

PR #1546, opened 2017-09-28
 Xcode project gen add OTHER_SWIFT_FLAGS

robbertvanginkel:
This takes the flags from `swift.compiler_flags` from `.buckconfig` and target specific flags and adds them to the `OTHER_SWIFT_FLAGS` in the xcconfig files.

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -1311,7 +1311,7 @@ public class ProjectGenerator {
 
         Iterable<String> otherSwiftFlags =
             Iterables.concat(
-                swiftBuckConfig.getFlags("compiler_flags").orElse(DEFAULT_SWIFTFLAGS),
+                swiftBuckConfig.getCompilerFlags().orElse(DEFAULT_SWIFTFLAGS),
                 targetSpecificSwiftFlags.build());
         Iterable<String> otherCFlags =
             Iterables.concat(
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -186,6 +186,7 @@ public class ProjectGenerator {
   private static final Logger LOG = Logger.get(ProjectGenerator.class);
   private static final ImmutableList<String> DEFAULT_CFLAGS = ImmutableList.of();
   private static final ImmutableList<String> DEFAULT_CXXFLAGS = ImmutableList.of();
+  private static final ImmutableList<String> DEFAULT_SWIFTFLAGS = ImmutableList.of();
   private static final String PRODUCT_NAME = "PRODUCT_NAME";
 
   private static final ImmutableSet<Class<? extends Description<?>>>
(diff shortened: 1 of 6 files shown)


src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1300,6 +1301,18 @@ private PBXNativeTarget generateBinaryTarget(
           appendConfigsBuilder.put("SWIFT_INCLUDE_PATHS", "$BUILT_PRODUCTS_DIR");
         }
 
+        ImmutableList.Builder<String> targetSpecificSwiftFlags = ImmutableList.builder();
+        if (targetNode.getConstructorArg() instanceof SwiftCommonArg) {
```
milend: I'd use `targetNode.castArg` here instead.

src/com/facebook/buck/swift/SwiftBuckConfig.java
```diff
@@ -36,6 +37,18 @@ public SwiftBuckConfig(BuckConfig delegate) {
     return value.map(input -> Splitter.on(" ").split(input.trim()));
   }
 
+  public Optional<ImmutableList<String>> getFlags(String field) {
```
milend: The method above (`getFlags`) should just be implemented by calling `getFlags(COMPILER_FLAGS_NAME)`.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1300,6 +1301,18 @@ private PBXNativeTarget generateBinaryTarget(
...
+          targetSpecificSwiftFlags.addAll(
+              convertStringWithMacros(
+                  targetNode,
+                  ((SwiftCommonArg) targetNode.getConstructorArg()).getSwiftCompilerFlags()));
+        }
+
+        Iterable<String> otherSwiftFlags =
+            Iterables.concat(
+                swiftBuckConfig.getFlags("compiler_flags").orElse(DEFAULT_SWIFTFLAGS),
```
milend: Rather than hardcoding `compiler_flags` here, we should just use `swiftBuckConfig.getFlags()` (which already reads `compiler_flags`).

`SwiftCompile` already uses `getFlags()`, so it's better to be consistent here.

src/com/facebook/buck/swift/SwiftBuckConfig.java
```diff
@@ -32,21 +31,13 @@ public SwiftBuckConfig(BuckConfig delegate) {
     this.delegate = delegate;
   }
 
-  public Optional<Iterable<String>> getFlags() {
-    Optional<String> value = delegate.getValue(SECTION_NAME, COMPILER_FLAGS_NAME);
+  public Optional<Iterable<String>> getFlags(String field) {
```
milend: We should mark this as `private`.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
milend: Let's use `swiftBuckConfig.getFlags()` instead because that's what's used by `SwiftCompile`, so that we pass the same sets of flags in Xcode + `buck build`.

robbertvanginkel: I did it like this to keep it consistent with the rest of the file, which all has getFlags with a parameter. With a parameter seems better than just `getFlags`, as that is a bit generic. If you have `swiftBuckConfig.getFlags()` which flags are you getting? Now its compiler flags, but that seems arbitrary.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: 👍

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
milend: My main concern here is the hardcoded string value which can get out of sync with other places and end up in different flags being passed in Xcode vs buck build. Can we instead expose enum in `SwiftBuckconfig`that defines all valid flags?

PR #1560, opened 2017-10-04
Xcode project gen add import-underlying-module

robbertvanginkel:
When a modular `apple_library` contains swift code, as well as public headers those headers should be compiled into a module and made accessible to the swift code through `-import-underlying-module`. Swift compilation will fail if the underlying module can't be found.


dinhviethoa:
@milend you should take a look at it.

PR #1561, opened 2017-10-04
Add generator for clang Modulemap, Umbrella Header and VFSOverlay

robbertvanginkel:
This extends the clang package with classes for generating some clang related build concepts:
- Modulemaps to support modular imports.
- Umbrella headers for use within module maps.
- VFSOverlays to be able to use module maps in virtual places, like header maps can do for headers.

### DIFF
--- a/src/com/facebook/buck/apple/clang/ModuleMap.java
+++ b/src/com/facebook/buck/apple/clang/ModuleMap.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2014-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.facebook.buck.apple.clang;
+
+import org.stringtemplate.v4.ST;
+
+/**
+ * Generate a modulemap, with optional references to a swift generated swift header.
+ *
+ * <p>Note: currently only supports modulemaps for modules with an umbrella header, as otherwise the
+ * inferred submodule functionality does not work.
+ */
+public class ModuleMap {
+
+  public enum SwiftMode {
+    NO_SWIFT,
+    EXCLUDE_SWIFT_HEADER,
+    INCLUDE_SWIFT_HEADER
+  }
+
+  private final SwiftMode swiftMode;
+  private final String moduleName;
+
+  private String generatedModule;
+  private static final String template =
+      "module <module_name> {\n"
+          + "    umbrella header \"<module_name>.h\"\n"
+          + "\n"
+          + "    export *\n"
+          + "    module * { export * }\n"
+          + "}\n"
+          + "\n"
+          + "<if(include_swift_header)>"
+          + "module <module_name>.Swift {\n"
+          + "    header \"<module_name>-Swift.h\"\n"
+          + "    requires objc\n"
+          + "}\n"
+          + "<endif>"
+          + "\n"
+          + "<if(exclude_swift_header)>"
+          + "module <module_name>.__Swift {\n"
+          + "    exclude header \"<module_name>-Swift.h\"\n"
+          + "}\n"
+          + "<endif>"
+          + "\n";
+
+  public ModuleMap(String moduleName, SwiftMode swiftMode) {
+    this.moduleName = moduleName;
+    this.swiftMode = swiftMode;
+  }
+
+  public String render() {
+    if (this.generatedModule == null) {
+      ST st =
+          new ST(template)
+              .add("module_name", moduleName)
+              .add("include_swift_header", false)
+              .add("exclude_swift_header", false);
+      switch (swiftMode) {
+        case INCLUDE_SWIFT_HEADER:
+          st.add("include_swift_header", true);
+          break;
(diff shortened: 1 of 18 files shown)


robbertvanginkel:
Let me know if I should split this into smaller PR's

src/com/facebook/buck/apple/clang/ModuleMap.java
```diff
@@ -0,0 +1,87 @@
...
+    INCLUDE_SWIFT_HEADER
+  }
+
+  private final SwiftMode swiftMode;
+  private final String moduleName;
+
+  private String generatedModule;
+  private static final String template =
+      "module <module_name> {\n"
+          + "    umbrella header \"<module_name>.h\"\n"
```
milend: I think we should also be passing the umbrella header name explicitly as an input. I'm mainly thinking about clashes with existing `apple_library` which have headers with the same name as the module.

test/com/facebook/buck/apple/clang/UmbrellaHeaderTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.google.common.collect.ImmutableList;
+import org.junit.Test;
+
+public class UmbrellaHeaderTest {
+
+  @Test
+  public void testUmbrellaHeader() {
+    UmbrellaHeader header = new UmbrellaHeader(ImmutableList.of("header1.h", "header2.h"));
+    assertEquals("#import \"header1.h\"\n" + "#import \"header2.h\"\n", header.render());
```
milend: This implies that umbrella headers are not exported as those imports would not work from other libraries. Do we plan to expose those umbrella headers, so that non-modular libraries can do `#import <Module/Module.h>`?

src/com/facebook/buck/apple/clang/VFSOverlay.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @JsonProperty("contents")
+    private ImmutableList<File> computeContents() {
+      return this.contents
+          .stream()
+          .map(x -> new File(x.getFirst(), x.getSecond()))
+          .collect(MoreCollectors.toImmutableList());
+    }
+
+    public Directory(Path root, Collection<Pair<Path, Path>> contents) {
```
milend: I'd try to clarify what `contents` is. Given that it's a collection of pairs of filename and real path, it might be good to change type of `name` to `String` so that it's clearer it's just a filename.

src/com/facebook/buck/apple/clang/VFSOverlay.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @JsonSerialize(as = File.class)
+  private class File {
+    @JsonProperty private final String type = "file";
+    @JsonProperty private final Path name;
+
+    @JsonProperty("external-contents")
+    private final Path realPath;
+
+    public File(Path name, Path realPath) {
```
milend: Do we want to enforce any properties of `realPath`, i.e., does it have to be absolute or can it be relative?

src/com/facebook/buck/apple/clang/VFSOverlay.java
```diff
@@ -0,0 +1,92 @@
...
+  public VFSOverlay(ImmutableSortedMap<Path, Path> overlays) {
+    this.overlays = overlays;
+  }
+
+  public String render() throws IOException {
+    return ObjectMappers.WRITER.withDefaultPrettyPrinter().writeValueAsString(this);
+  }
+
+  @JsonSerialize(as = Directory.class)
+  private class Directory {
```
milend: - For clarity, I feel like we should name this `VirtualDirectory`.
- My other observation is that `Directory` does not support nested `Directory` even though the Clang spec says it's supported. I think that's fine for now but we should add a test case for this.

src/com/facebook/buck/apple/clang/VFSOverlay.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public Directory(Path root, Collection<Pair<Path, Path>> contents) {
+      this.name = root;
+      this.contents = contents;
+    }
+  }
+
+  @JsonSerialize(as = File.class)
+  private class File {
```
milend: Would `VirtualFile` be more descriptive?

src/com/facebook/buck/apple/clang/ModuleMap.java
robbertvanginkel: Apple's convention is to have a master header which has the same name as the target. This is convention for regular modules/frameworks and required for umbrella frameworks  [1]. When there is a pre-existing header with the same name, I think it is expected that that is the umbrella header.

I can add an optional parameter to override it, but I don't think there's a strong usecase. It would also require us to add an extra field `umbrella_header` on `apple_library` to get user input for it, so maybe we should move the implementation to when that is needed?

[1] https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Tasks/IncludingFrameworks.html

test/com/facebook/buck/apple/clang/UmbrellaHeaderTest.java
robbertvanginkel: My bad, I'll fix it to take a name and fix the headers.

src/com/facebook/buck/apple/clang/VFSOverlay.java
robbertvanginkel: Should be absolute at this time. VFS' have an option for having the realpaths relative to the vfs file by specifying the option `overlay-relative`, but I haven't found a use case for that.

src/com/facebook/buck/apple/clang/VFSOverlay.java
robbertvanginkel: I named it when I was still thinking about the nested directories, I'll update it with something more appropriate

src/com/facebook/buck/apple/clang/VFSOverlay.java
milend: Let's add a `Precondition` in that case to make sure the inputs are valid.

Issue #1290, opened 2017-04-12
platform_linker_flags not working correctly?


nikhedonia:
I'm trying to customize my build using platform specific libraries.
Unfortunately i can't get it working on debian nor osx. 
For some reason `platform_linker_flags` are only matching on `default`.

```python
#BUCK 
cxx_binary(
  name = 'foo',
  header_namespace = 'foo',
  srcs = glob([
    'src/**/*.cpp',
  ]),
  headers = subdir_glob([
    ('include', '**/*.hpp'),
  ]),
  platform_linker_flags = [
    ('^linux.*', ['-lbar'])
  ],
)
```

buck version: adc0f78a68db45334098ac083171c7c2ebc32a8c (v2017.03.29)
`uname -s` returns Linux
in java: `System.getProperty("os.name")` returns 'Linux' as well.

njlr:
Can confirm the same issue on macOS - the platform is "default"

voznesenskym:
Confirmed - Ditto for us. Let me know if you need an example project created.

Coneko:
What is the command you are running? This might only work if you explicitly run `buck build target#linux-x86_64`.

njlr:
I get the error when running just `buck build :my-target`. Do we have to specify the target for every build? Seems like it would be more streamlined if the platform macro expanded to the actual host platform by default.

Coneko:
I don't know, depends on whether it works if you expand it manually. If it doesn't it could be another issue.

njlr:
I have made a small C++ project here: [https://github.com/njlr/buck-platforms](https://github.com/njlr/buck-platforms). 

Here are my results on macOS: 

```
  buck run :hello
  Hello from default

  buck run :hello#macosx-x86_64
  Hello from macOS

  buck run :hello#linux-x86_64
  Unrecognized flavor in target //:hello#linux-x86_64 while parsing //BUCK
```

nikhedonia:
I can confirm. On Linux i have the same behavior but the other way around:

```
  buck run :hello
  Hello from default

  buck run :hello#linux-x86_64
  Hello from linux

  buck run :hello#macosx-x86_64
  Unrecognized flavor in target //:hello#macosx-x86_64 while parsing //BUCK

```

Coneko:
Yep, then it's what I mentioned above.

cc @andrewjcg, @Ktwu was either of you two looking at this?

Coneko:
cc @mzlee as well.

[...]

PR #1562, opened 2017-10-04
CXX targets default to the actual platform

njlr:
This patch causes Buck to choose the correct flavor for `cxx_*` targets when no build flavour is specified.

In other words, on macOS, these become equivalent: 

```
buck build :app#macosx-x86_64
buck build :app
```

https://github.com/facebook/buck/issues/1290

### DIFF
--- a/src/com/facebook/buck/cxx/toolchain/CxxPlatformsProvider.java
+++ b/src/com/facebook/buck/cxx/toolchain/CxxPlatformsProvider.java
@@ -95,9 +95,6 @@ public class CxxPlatformsProvider {
     }
 
     // Finalize our "default" host.
-    // TODO(kelliem) The host flavor should default to a concrete flavor
-    // like "linux-x86_64", not "default".
-    hostFlavor = DefaultCxxPlatforms.FLAVOR;
     Optional<String> hostCxxPlatformOverride = cxxBuckConfig.getHostPlatform();
     if (hostCxxPlatformOverride.isPresent()) {
       Flavor overrideFlavor = InternalFlavor.of(hostCxxPlatformOverride.get());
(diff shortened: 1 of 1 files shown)


mzlee:
The idea is sound, but I think just changing this directly is going to break a lot of assumptions people are making. Would it be possible to control this behind a buck configuration parameter? I'm happy to tackle that.

njlr:
My understanding is that the Buck team can make breaking changes because the version of Buck can be locked down inside of a project. Perhaps this isn't the case though? 

I think in the long run it is easier to fix the behaviour than to add a switch. If the switch is _off_ by default, then users will still have to perform an arbitrary action to build cross-platform projects (flipping the switch instead of adding the platform flavor). If the switch is _on_ by default, then we would be breaking some builds anyway, so why not fix just the build? After all, the `default` platform is a forgotten `TODO`, rather than a conscious design decision. 

The migration procedure is quite easy. 

For example, this:

```
platform_preprocessor_flags = [
  ('default', [ '-DFOO=1' ]),
  ('^macos.*', [ '-DFOO=2' ]),
],
```

... might become... 

```
platform_preprocessor_flags = [
  ('^macos.*', [ '-DFOO=2' ]),
  ('*', [ '-DFOO=1' ]),
],
```

What are your thoughts on this?

mzlee:
I totally agree that there are breaking changes all the time, but I'm just thinking about all of the hardcoded places where this might come up. The platform_preprocessor_flags you've shown is actually unsound, it went from `-DFOO=1` in MacOS to `-DFOO=2` (I'm assuming you meant `'.*'` for the default argument).

The switch is to be able to stage the rollout and fix things or quickly revert if necessary. I know some places where I hardcode `default`, but I'm not sure where else they might be lurking.

njlr:
Yes, perhaps not the best example. We are placing generic settings in `default` and then _also_ setting the `^linux.*` and `^macos.*` entries (even if they are the same), so for us it is not much of change. 

Either way, we would be really happy to see this get included, even if it is behind a flag for a while. 👍

mzlee:
I guarded this change with `cxx. should_remap_host_platform` (a bit verbose, I know..) Does that work for you? You can set it in the buckconfig

    [cxx]
        should_remap_host_platform = true

njlr:
Sounds reasonable 👍

njlr:
Closing because this feature was added by another PR. https://buckbuild.com/files-and-dirs/buckconfig.html#cxx.should_remap_host_platform

PR #1568, opened 2017-10-07
Michaelv/fix1538

### LINKED ISSUE #1538: New cxx_library 'static_pic_lib' option not properly supported
I am working on upgrading our BUCK to the latest version, and when making a prebuilt_cxx_library, the static_pic_lib command does not work properly.

Example:
`
prebuilt_cxx_library (
    name = 'icu',
    static_pic_lib = 'linux/x86/libicu/libicuuc.a',
    exported_deps = [
        ':icuuc_headers',
    ],
    header_namespace = '',
    visibility = [
        'PUBLIC'
    ],
)
`

**Context**
This lib is included with other libs into a top level static pic library. However, when I attempt to link it into a shared library with -fPIC flags (for JNI), it tells me that the generated library is invalid and that I should recompile with -fPIC.

**Investigation**
This means that the static_pic_lib command does not actually do anything, it just makes a static_lib.

I decided to dig into the BUCK source code to find the issue, and it appears as though everything is wired up under the hood, except the getStaticPicLibrary method in AbstractNewPrebuiltCxxLibraryPaths never gets called. 

This is because of line 740 in CxxDescriptionEnhancer, `args.getLinkStyle().orElse(Linker.LinkableDepType.STATIC),` basically forces any cxx_prebuilt_library to be static. The args.getLinkStyle() it checks (CxxBinaryDescription.CommonArg) always seems to return null and does not even appear to be implemented.  

**Solution?**
Changing line 740 mentioned above to force a Linker.LinkableDepType.STATIC_PIC does work. This means everything is indeed wired up properly under the hood, there just needs to be some resolution at wherever we parse the BUCK files. 

I plan to have a fix up in the next few days, but wanted to open this issue up as soon as I found it.

voznesenskym:
Closes #1538


voznesenskym:
@asp2insp There are not actually conflicts

```
<<<<<<< michaelv/fix1538

=======
>>>>>>> master
```

As I fixed them by hand. Let me know what next steps are here!

src/com/facebook/buck/cxx/CxxLibraryDescription.java
```diff
@@ -440,6 +440,7 @@ private static CxxLink createSharedLibrary(
       ImmutableSet<BuildTarget> blacklist,
       TransitiveCxxPreprocessorInputFunction transitiveCxxPreprocessorInputFunction,
       Optional<CxxLibraryDescriptionDelegate> delegate) {
+
```
asp2insp: Looks like the code formatter you used added some whitespace. Please make sure to run Google's Java formatter on the PR (see Contributing.md)

src/com/facebook/buck/cxx/CxxDescriptionEnhancer.java
```diff
@@ -753,6 +753,11 @@ public static CxxLinkAndCompileRules createBuildRulesForCxxBinaryDescriptionArg(
     extraDeps.stream().map(resolver::getRule).forEach(depsBuilder::add);
     ImmutableSortedSet<BuildRule> deps = depsBuilder.build();
 
+    Linker.LinkableDepType linkStyle = args.getLinkStyle().orElse(
```
asp2insp: This looks good to me! @andrewjcg would you mind taking a quick look?

src/com/facebook/buck/cxx/LinkableCxxConstructorArg.java
```diff
@@ -17,13 +17,17 @@
...
 import com.facebook.buck.cxx.toolchain.linker.Linker;
-import java.util.Optional;
+import com.facebook.buck.rules.SourcePath;
 import org.immutables.value.Value;
+import java.util.Optional;
 
 public interface LinkableCxxConstructorArg extends CxxConstructorArg {
   Optional<Linker.LinkableDepType> getLinkStyle();
 
   @Value.Default
```
asp2insp: This should be `@Value` instead of `@Value.Default` since `Optional` has a sane default

asp2insp:
Fantastic, thanks!

asp2insp:
Hm, the lint is still complaining about the formatting on the java files. Could you run it one more time? Thanks!

voznesenskym:
I am running it manually w/

./scripts/check-java-file-format ~/my_path/google-java-format-1.4-all-deps.jar src/com/facebook/buck/cxx/CxxLibraryDescription.java 

And it produces no delta. What I am missing?

src/com/facebook/buck/cxx/LinkableCxxConstructorArg.java
```diff
@@ -17,13 +17,17 @@
...
+  Optional<SourcePath> getStaticPicLib();
```
andrewjcg: Sorry, not sure I follow this.  Why are we adding a `static_pic_lib` parameter to *all* C/C++ rules (via `LinkableCxxConstructorArg`)?  Shouldn't this change only affect `prebuilt_cxx_library`?

PR #1570, opened 2017-10-09
Added details on unsupported flavors to flavor-related exceptions

shybovycha:
This PR is aimed to provide more detailed feedback on why the build failed.

Currently, `apple_bundle` targets require the flavor to be provided. But if one is not provided, the error does not point to why the build really failed and what options to fix that failure are. For more details see the example below.

Before:

```
$ buck run //irrlicht-newton-game:bundle-osx1
Parsing buck files: finished in 2.3 sec (100%)
Creating action graph: finished in 0.7 sec (100%)
Building: finished in 1.1 sec
  Total time: 4.2 sec
BUILD FAILED: //irrlicht-newton-game:bundle-osx#dwarf-and-dsym,no-include-frameworks: Apple bundle requires an Apple platform, found 'default'

A common cause of this error is that the required SDK is missing.
Please check whether it's installed and retry.
```

After:

```
$ ../buck/bin/buck run //irrlicht-newton-game:bundle-osx
Parsing buck files: finished in 2.8 sec (100%)
Creating action graph: finished in 0.7 sec (100%)
Building: finished in 0.8 sec
  Total time: 4.4 sec
BUILD FAILED: //irrlicht-newton-game:bundle-osx#dwarf-and-dsym,no-include-frameworks: Apple bundle requires an Apple platform, found 'default'

A common cause of this error is that the required SDK is missing.
Please check whether it's installed and retry.
Original exception: "Apple C++ Platform" has no flavor "default"; available ones: appletvos11.0-arm64, appletvsimulator11.0-x86_64, iphoneos11.0-armv7, iphoneos11.0-arm64, iphonesimulator11.0-i386, iphonesimulator11.0-x86_64, macosx10.13-i386, macosx10.13-x86_64, watchos4.0-armv7k, watchsimulator4.0-i386, appletvos-arm64, appletvsimulator-x86_64, iphoneos-armv7, iphoneos-arm64, iphonesimulator-i386, iphonesimulator-x86_64, macosx-i386, macosx-x86_64, watchos-armv7k, watchsimulator-i386
```

### DIFF
--- a/src/com/facebook/buck/parser/UnexpectedFlavorException.java
+++ b/src/com/facebook/buck/parser/UnexpectedFlavorException.java
@@ -26,6 +26,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import java.util.Optional;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 public class UnexpectedFlavorException extends HumanReadableException {
 
@@ -64,8 +65,9 @@ public class UnexpectedFlavorException extends HumanReadableException {
   public static UnexpectedFlavorException createWithSuggestions(
       Flavored flavored, Cell cell, BuildTarget target) {
     ImmutableSet<Flavor> invalidFlavors = getInvalidFlavors(flavored, target);
+    ImmutableSet<Flavor> validFlavors = getValidFlavors(flavored, target);
     // Get the specific message
-    String exceptionMessage = createDefaultMessage(target, invalidFlavors);
+    String exceptionMessage = createDefaultMessage(target, invalidFlavors, validFlavors);
     // Get some suggestions on how to solve it.
     Optional<ImmutableSet<PatternAndMessage>> configMessagesForFlavors =
         cell.getBuckConfig().getUnexpectedFlavorsMessages();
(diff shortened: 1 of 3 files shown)


src/com/facebook/buck/parser/UnexpectedFlavorException.java
```diff
@@ -114,13 +116,24 @@ public static UnexpectedFlavorException createWithSuggestions(
     return builder.build();
   }
 
+  private static ImmutableSet<Flavor> getValidFlavors(Flavored flavored, BuildTarget target) {
```
ttsugriy: looks like this can be replaced with a stream filter operation

src/com/facebook/buck/parser/UnexpectedFlavorException.java
```diff
@@ -114,13 +116,24 @@ public static UnexpectedFlavorException createWithSuggestions(
...
-      BuildTarget target, ImmutableSet<Flavor> invalidFlavors) {
+      BuildTarget target, ImmutableSet<Flavor> invalidFlavors, ImmutableSet<Flavor> validFlavors) {
     ImmutableSet<String> invalidFlavorsStr =
         invalidFlavors.stream().map(Flavor::toString).collect(MoreCollectors.toImmutableSet());
     String invalidFlavorsDisplayStr = String.join(", ", invalidFlavorsStr);
     return String.format(
-        "The following flavor(s) are not supported on target %s:\n%s.",
-        target, invalidFlavorsDisplayStr);
+        "The following flavor(s) are not supported on target %s:\n%s. Available flavors are: %s",
+        target, invalidFlavorsDisplayStr, validFlavors.stream().map(Flavor::getName).collect(
```
ttsugriy: nit: please extract the last expression into a local variable, since it's not on the same semantical level as the other argument statements

ttsugriy:
thank you for your contribution @shybovycha! It definitely looks like an improvement. Please update address the comments and you'll be good to go.

src/com/facebook/buck/parser/UnexpectedFlavorException.java
shybovycha: I just wanted to keep consistency across the project - other methods in the same file are using plain-old `for` loops. But that doesn't mean it's not worth refactoring, right? ;)

src/com/facebook/buck/parser/UnexpectedFlavorException.java
```diff
@@ -114,13 +118,25 @@ public static UnexpectedFlavorException createWithSuggestions(
     return builder.build();
   }
 
+  private static ImmutableSet<Flavor> getValidFlavors(Flavored flavored, BuildTarget target) {
+    return target.getFlavors().stream()
+        .filter(flavor -> flavored.hasFlavors(ImmutableSet.of(flavor)))
+        .collect(collectingAndThen(toList(), ImmutableSet::copyOf));
```
ttsugriy: pleas use `MoreCollectors.toImmutableSet()` directly

shybovycha: Awesome! Did not know this collector exists =(

shybovycha: And again, to make code more consistent I've changed `getInvalidFlavors` to the same approach, since both methods are private and are used in the same context.

src/com/facebook/buck/parser/UnexpectedFlavorException.java
```diff
@@ -105,22 +107,30 @@ public static UnexpectedFlavorException createWithSuggestions(
...
+
+    String validFlavorsStr = validFlavors.stream()
+        .map(Flavor::getName)
+        .collect(Collectors.joining(", "));
+
     String invalidFlavorsDisplayStr = String.join(", ", invalidFlavorsStr);
     return String.format(
-        "The following flavor(s) are not supported on target %s:\n%s.",
-        target, invalidFlavorsDisplayStr);
+        "The following flavor(s) are not supported on target %s:\n%s. Available flavors are: %s",
```
ttsugriy: this line also got a following comment from one of our engineers:
```
In the test plan, there are lots of similar platforms. Maybe try to de-dedup as much as possible.
In all cases, if the number of items in the list is larger than 4, there should be one item per line.
```

shybovycha: Hmm... What should I dedup, given both methods use Sets?

ttsugriy: I think comment author meant that `macosx-i386, macosx-x86_64` can be grouped into a single category, but I don't think it's a good thing to do, but writing flavors on separate lines makes sense to me.

shybovycha: Well, that could be a separate feature - grouping flavors and platforms, but it also has an edge case: what if some flavors do not support some specific platforms?

Will split flavors by a newline and update this PR.

src/com/facebook/buck/parser/UnexpectedFlavorException.java
```diff
@@ -105,22 +107,32 @@ public static UnexpectedFlavorException createWithSuggestions(
...
+        .collect(MoreCollectors.toImmutableSet());
   }
 
   private static String createDefaultMessage(
-      BuildTarget target, ImmutableSet<Flavor> invalidFlavors) {
-    ImmutableSet<String> invalidFlavorsStr =
-        invalidFlavors.stream().map(Flavor::toString).collect(MoreCollectors.toImmutableSet());
+      BuildTarget target, ImmutableSet<Flavor> invalidFlavors, ImmutableSet<Flavor> validFlavors) {
+    String invalidFlavorsStr =
+        invalidFlavors.stream().map(Flavor::toString).collect(Collectors.joining("\n"));
```
ttsugriy: please use a cross platform way of generating new lines `System.lineSeparator()`

shybovycha:
@ttsugriy done

PR #1582, opened 2017-10-13
Xcode project support modular apple_library

robbertvanginkel:
When the `modular = True` flag is set on `apple_library`, create a `module.modulemap` for modular imports of Objective C code into into swift. Support for both cross target imports and mixed target imports.

Cross targets the `module.modulemap` file will be auto detected in a header search path, which will allow modular import of a target into swift and objective-c. Currently this only works when `apple.use_header_maps_in_xcode` is set to `false`.

Mixed modular apple_libraries supported through the `-import-underlying-module` flag. In the case of a mixed (swift source + at least one header file) `apple_library`, a `objc.modulemap` is created that references the defined target headers excluding the generated swift header. When compiling the swift part of the target this partial modulemap then overrides the complete modulemap through the use of a vfsoverlay.

Also adds an optional config `generate_missing_umbrella_headers` to the `[apple]` section, which generates an umbrella header if no umbrella header is present.

(Note that this is stacked on PR #1561, but I can't seem to change that base so until that lands you can review this without looking at 951871c but 951871c...ea3f297)


src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1443,18 +1446,23 @@ private PBXNativeTarget generateBinaryTarget(
...
     createHeaderSymlinkTree(
         publicCxxHeaders,
         getSwiftPublicHeaderMapEntriesForTarget(targetNode, isFocusedOnTarget, hasSwiftVersionArg),
+        moduleName,
         getPathToHeaderSymlinkTree(targetNode, HeaderVisibility.PUBLIC),
         arg.getXcodePublicHeadersSymlinks().orElse(cxxBuckConfig.getPublicHeadersSymlinksEnabled())
             || isHeaderMapDisabled(),
         !shouldMergeHeaderMaps());
     if (isFocusedOnTarget) {
+      /**/
```
milend: Nit: seems like leftover temporary code.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1443,18 +1446,23 @@ private PBXNativeTarget generateBinaryTarget(
       }
     }
 
+    Optional<String> moduleName =
+        isModularAppleLibrary ? Optional.of(getProductName(targetNode)) : Optional.empty();
```
milend: `SwiftCompile` actually gets `args.getModuleName().orElse(buildTarget.getShortName())` as the module name, I think we should be consistent here. It might be best to centralise this logic in a static method on `SwiftLibraryDescription`.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -2110,6 +2110,25 @@ private void createHeaderSymlinkTree(
     headerSymlinkTrees.add(headerSymlinkTreeRoot);
   }
 
+  private void writeUmbrellaHeaderIfNeeded(
+      String moduleName,
+      Path headerSymlinkTreeRoot,
+      ImmutableSortedMap<Path, Path> resolvedContents)
```
milend: Nit: I'd give `resolvedContents` a more descriptive name.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -396,6 +399,10 @@ private boolean shouldGenerateHeaderSymlinkTreesOnly() {
     return options.contains(Option.GENERATE_HEADERS_SYMLINK_TREES_ONLY);
   }
 
+  private boolean shouldGenerateMissingUmbrellaHeaders() {
+    return options.contains(Option.GENERATE_MISSING_UMBRELLA_HEADER);
```
milend: My only comment here is that this setting might have to be on a per-target basis. But I think it's okay to cross that bridge when needed as we can just change this method to accept the `TargetNode`.

robbertvanginkel: Changing this on a per target basis can be achieved by putting in an empty header with the module name now. We could make this more explicit but I agree that can be left for later.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: Will change

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: I'll consolidate this a little bit better in project gen so it is the same. As for consolidating between project gen and buck build I think we should change this by having a `modular` interface, so this logic can be shared between CxxLibrary and SwiftLibrary. This will allow us to implement it into `SwiftDescriptions.populateSwiftLibraryDescriptionArg` properly.

PR #1583, opened 2017-10-13
Add ability to right click on files and folders to generate project.

clonetwin26:
This diff seeks to mitigate an issue where our devs are opening too many modules at once. With this add-on you can right click the directory you want to get ready, and run project on it. It is very useful in large mono repos, for generating projects that would be pesky to do on the command line. I tested it on the buck, okbuck and our internal mono repo.

**Creating projects from project tool:**
screen shot 2017-10-16 at 11 18 20 am
screen shot 2017-10-16 at 11 18 37 am

**Creating projects from file:**
screen shot 2017-10-16 at 11 14 59 am
screen shot 2017-10-16 at 11 15 25 am

From my instrumentation testing:
Average indexing time before 8599.29ms
Average indexing time after 444.786ms

**Indexing time in ms before and after plugin:**
screen shot 2017-10-13 at 2 47 24 pm
Indexing time in ms with plugin:
screen shot 2017-10-13 at 2 47 02 pm

### DIFF
--- a/src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
+++ b/src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
@@ -239,6 +239,16 @@
   </extensions>
 
   <actions>
+    <group description="Generate projects below" id="GenerateProjects" text="GenProjects" type="">
+      <action
+              class="com.facebook.buck.intellij.ideabuck.actions.select.RunSelectedProjectsAction"
+              description="Generates project below."
+              icon="/icons/actions/Project.png"
+              id="GenerateProject"
+              text="Generate Buck Project below">
+        <add-to-group group-id="ProjectViewPopupMenu" anchor="after" relative-to-action="ReplaceInPath"/>
+      </action>
+    </group>
     <group id="buck.Menu">
       <action
           id="buck.GoToBuckFile"
--- a/src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
+++ b/src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
@@ -239,13 +239,13 @@
   </extensions>
 
   <actions>
-    <group description="Generate projects below" id="GenerateProjects" text="GenProjects" type="">
+    <group description="Generate buck projects" id="buck.GenerateProjects" text="GenProjects" type="">
       <action
-              class="com.facebook.buck.intellij.ideabuck.actions.select.RunSelectedProjectsAction"
-              description="Generates project below."
+              class="com.facebook.buck.intellij.ideabuck.actions.select.RunSelectedProjectsFromFolderAction"
+              description="Generate buck project"
               icon="/icons/actions/Project.png"
-              id="GenerateProject"
-              text="Generate Buck Project below">
+              id="buck.GenerateProjectFromFolder"
+              text="Generate Buck Project">
         <add-to-group group-id="ProjectViewPopupMenu" anchor="after" relative-to-action="ReplaceInPath"/>
       </action>
     </group>
@@ -256,6 +256,13 @@
           text="Go to Buck file"
           description="Go to Buck file for this source file.">
         <keyboard-shortcut keymap="$default" first-keystroke="ctrl shift P"/>
+      </action>
+      <action
+        id="buck.GenerateProjectFromFile"
+        class="com.facebook.buck.intellij.ideabuck.actions.select.RunSelectedProjectFromFileAction"
+        text="Generate Buck Project"
+        description="Generate Buck Project">
+        <keyboard-shortcut keymap="$default" first-keystroke="ctrl shift L"/>
       </action>
       <add-to-group anchor="first" group-id="BuildMenu"/>
       <add-to-group anchor="first" group-id="EditorPopupMenu"/>
--- a/src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedProjectFromFileAction.java
+++ b/src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedProjectFromFileAction.java
@@ -1,0 +1,87 @@
+package com.facebook.buck.intellij.ideabuck.actions.select;
+
+import com.facebook.buck.intellij.ideabuck.build.BuckBuildCommandHandler;
+import com.facebook.buck.intellij.ideabuck.build.BuckBuildManager;
+import com.facebook.buck.intellij.ideabuck.build.BuckCommand;
+import com.facebook.buck.intellij.ideabuck.build.BuckQueryCommandHandler;
+import com.facebook.buck.intellij.ideabuck.config.BuckModule;
+import com.facebook.buck.intellij.ideabuck.file.BuckFileUtil;
+import com.facebook.buck.intellij.ideabuck.icons.BuckIcons;
+import com.google.common.base.Function;
+import com.intellij.openapi.actionSystem.AnActionEvent;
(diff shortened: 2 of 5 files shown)


src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
```diff
@@ -239,6 +239,16 @@
   </extensions>
 
   <actions>
+    <group description="Generate projects below" id="GenerateProjects" text="GenProjects" type="">
+      <action
+              class="com.facebook.buck.intellij.ideabuck.actions.select.RunSelectedProjectsAction"
+              description="Generates project below."
```
kageiit: Rename to  "Generate Buck Project"

src/com/facebook/buck/intellij/ideabuck/META-INF/plugin.xml
```diff
@@ -239,6 +239,16 @@
...
+              icon="/icons/actions/Project.png"
+              id="GenerateProject"
+              text="Generate Buck Project below">
```
kageiit: Rename to  "Generate Buck Project"

src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedProjectsAction.java
```diff
@@ -0,0 +1,60 @@
...
+
+    BuckModule buckModule = project.getComponent(BuckModule.class);
+    DataContext dataContext = anActionEvent.getDataContext();
+    final IdeView view = LangDataKeys.IDE_VIEW.getData(dataContext);
+    if (view == null) {
+      return;
+    }
+
+    final PsiDirectory directory = view.getOrChooseDirectory();
+    if (directory == null) {
```
kageiit: if its done on a file, we can use buck query to see which targets own the file as well

src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedProjectFromFileAction.java
```diff
@@ -0,0 +1,87 @@
...
+                      new BuckBuildCommandHandler(project, project.getBaseDir(), BuckCommand.PROJECT);
+                  handler.command().addParameter(target);
+                  handler.command().addParameter("--ide");
+                  handler.command().addParameter("INTELLIJ");
+
+                  BuckModule buckModule = project.getComponent(BuckModule.class);
+                  BuckBuildManager buildManager = BuckBuildManager.getInstance(project);
+                  buildManager.setBuilding(project, true);
+                  buckModule.attach(target);
+                  buildManager.runBuckCommand(handler, "Running buck project for selected file.");
```
kageiit: Instead of "selected file", we can show what file and target got invoked

src/com/facebook/buck/intellij/ideabuck/src/com/facebook/buck/intellij/ideabuck/actions/select/RunSelectedProjectsFromFolderAction.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    String relative = directoryString.replace(basepath, "");
+    String target = "/" + relative + "/...";
+
+    BuckBuildCommandHandler handler =
+        new BuckBuildCommandHandler(project, project.getBaseDir(), BuckCommand.PROJECT);
+    handler.command().addParameter(target);
+    handler.command().addParameter("--ide");
+    handler.command().addParameter("INTELLIJ");
```
kageiit: lot of common code between this and the above that can be DRY'ed

styurin:
Have you considered using an external tool for this? You can write simple script that accepts a file path or folder path and generates a project.

clonetwin26:
Considered it. My fear with scripts is my users have no way of guessing what the right thing to do on the command line is. I am worried they will pass either too large things or not realize the tool is there in the first place. Plus right clicking around a large file system, feels much better and faster than paying attention to the paths on the command line even with auto complete.

styurin:
I meant an external tool in IntelliJ:

screen shot 2017-10-16 at 3 42 10 pm

screen shot 2017-10-16 at 3 42 23 pm

clonetwin26:
derp. Let me take a look lol

kageiit:
External tool is nested one level below and needs to be setup by every user. Any objections on adding this in the buck plugin instead? It's much easier for discover ability and will benefit more than just us as it is a better us for oss users.

styurin:
I'm not really following the developer's workflow here.

Do developers open old project and then run `buck project` on a new package?

> External tool is nested one level below

An advantage of this approach is that you can put external tools in different categories and can have a category related to Buck only. If the problem with having it in a nested menu is the faster access to this action users always can use keyboard shortcuts.

> needs to be setup by every user.

Yeah, that might be a problem, but I don't think we should use Buck plugin as a solution to deliver functionality to developers.

> Any objections on adding this in the buck plugin instead? It's much easier for discover ability and will benefit more than just us as it is a better us for oss users.

I don't think we want to expose this option to developers, IntelliJ probably can handle changes in the project on a fly, but I'd prefer not to rely on this behavior.

clonetwin26:
Almost done writing this as an external tool. On the one hand chaining the buck owner query and project query becomes much simpler. Strangely the utilities to find a Buck File make generating projects from a directory more complicated in the scripts than the action. I don't think there will be a lot of discoverability behind external tools, which is tough as this issue hits our newest devs the worst, and as such I prefer the implementation inside the plugin. 

> Do developers open old project and then run buck project on a new package?
Yes, and the projects they are generating are much too large.

> I don't think we want to expose this option to developers, IntelliJ probably can handle changes in the project on a fly, but I'd prefer not to rely on this behavior.
This is kind of already exposed in the buck tool window as you can buck project the selected target, this just makes the user experience much friendlier.

clonetwin26:
Built most of it as an external tool.

clonetwin26:
Here is an implementation of this as an external tool. The install is pretty goofy:
screen shot 2017-10-17 at 2 26 03 pm

`#!/usr/bin/env bash

projectRoot=$2
fileName=$4
if [[ !  -z  $5  ]]; then
    echo "buck query \"owner('$projectRoot/$fileName')\" | xargs buck project"
    buck query "owner('$projectRoot/$fileName')" | xargs buck project
else
    directory=$1
    relativeFilePath=$3
    relativeModulePath=${directory#$projectRoot}
    relativeModulePath=${relativeModulePath:1}
    if [ -e "$directory/BUCK" ] > /dev/null; then
        :
    elif [ "." != "$relativeFilePath" ]; then
       relativeModulePath=${relativeModulePath%$relativeFilePath}
       if [[ "$relativeModulePath" =~ '/'$ ]]; then
          relativeModulePath=${relativeModulePath%?}
       fi
       # The module is not necessarily loaded so relative paths can be broken. Look for a Buck file if we don't have.
       while [[ ! -e "$relativeModulePath/BUCK" && "." != "$relativeModulePath" ]]; do
          relativeModulePath=$(dirname "$relativeModulePath")
       done
       # No Buck file so just generate projects for everything below.
       if [ $relativeModulePath == "." ]; then
          directory=$1
          relativeFilePath=$3
          relativeModulePath=${directory#$projectRoot}
          relativeModulePath=${relativeModulePath:1}
       fi
    fi
    echo "buck project $relativeModulePath/..."
    buck project $relativeModulePath/...
fi
`

kageiit:
@ryanson let's build it into our custom intellij plugin instead. We can discuss offline

clonetwin26:
Okay. Only mystery is registering an external tool via the plugin to fix the install flow.

PR #1584, opened 2017-10-13
#956 Add kotlin annotation processor (kapt) support.

cwoodwar6:
This is still a major work in progress but I wanted to get the ball rolling. Currently this is building but fails at runtime to find the generated classes. I have only tested running it on an Android Project.

A huge shoutout to my former colleague @runningcode who helped tremendously on this. #956 outlines the steps that he discovered when running kapt via Gradle.

I've created a simple [tool](https://github.com/cwoodwar6/kapt-cli) to better understand what steps are required when kapt is plugged into the compiler. I'll continue to update this PR and the tool as I make more progress. Feedback is welcome, looking for suggestions early so we can get this added as soon as possible.


src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -132,6 +135,49 @@ Path getPathToCompilerJar() {
...
+            + ").");
+  }
+
+  /**
+   * Get the path to the java tools jar.
+   *
+   * @return the java tools jar path
+   */
+  Path getPathToTools() {
+    Path compiler = getJavaHome().resolve("lib/tools.jar");
```
kageiit: This is no longer present in jdk 9. Can we not need this or atleast let users specify a location for this jar?

cwoodwar6: Currently it is required when building with JDK 7 or JDK 8. 

According to https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-A78CC891-701D-4549-AA4E-B8DD90228B4B the tools jar may be instead replaced by a "more efficient format in implementation-specific files in the lib directory." so we could potentially depend on whatever in tools.jar kapt requires when building with JDK 9.

I can change the implementation for now to allow users to specify the location. But I'll keep digging to see what kapt in Gradle will do when building with JDK 9 if it is supported yet.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -25,11 +25,13 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Map;
 import java.util.Optional;
 import javax.annotation.Nullable;
 
 public class KotlinBuckConfig {
   private static final String SECTION = "kotlin";
+  private static final String JAVA_SECTION = "java";
```
jkeljo: Er, don't do that. Can you just use a `JavaBuckConfig`?

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
jkeljo: It looks like this path is used to initialize a `ClassLoader` that is used to load the compiler? If that's the only usage, `ToolProvider.getSystemToolClassLoader` will get you a `ClassLoader` that is equivalent to including `tools.jar`. You can use that as the parent `ClassLoader`.

Note that `ToolProvider` is not actually thread safe (yay!) so we need to ensure all our usages of it are synchronized. Right now the only usage is in `JdkProvidedInMemoryJavac`; you'll need to factor out some new synchronized wrapper class to use both here and there.

That will allow you to get rid of the scary `java_home` finding logic too right?

thalescm: I've found a mention, and explains where tools.jar is used on kapt [here](https://github.com/JetBrains/kotlin/blob/0b37c9e83cd09008db5908fd47583cd62e9fc17b/plugins/kapt3/kapt3-compiler/src/org/jetbrains/kotlin/kapt3/Kapt3Plugin.kt#L210)

thalescm:
Well, I've talked a lot with @yanex about how kapt works and what does it need / the options, plugins and etc...
It is possible to to use kapt with java 9, though now JRE will warn about reflection access of JDK private declarations.
With this in mind, and considering the comments of @kageiit and @jkeljo, I was thinking about the following points:

As @runningcode pointed out (and @yanex confirmed), that are four steps:
1. Generate sutbs
2. Run apt
3. CompileKotlin
4. CompileJava

Today's kotlin rule already does the steps 3 and 4.
So we could make a custom rule, called `kotlin-annotation-processor` which would be responsible for the steps 1 and 2. 
This rule would give you all the options you need to get kapt to work properly, for example, as `tools.jar` is a plugin, and the ap jar is a plugin as well, you would have an array of plugins (then on Java9, you simply don't pass the tools.jar), and you could stablish the output for the generated sources, which is basically an entry called `kapt.kotlin.generated` with the value as the desired path on the `apoptions` plugin parameter.
And them on the kotlin rule you set that path as one of the sources to look at.   

WDYT?

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
jkeljo: That reference doesn't contradict what I'm saying...it's expecting to be able to use its default `ClassLoader` to load certain classes that live in `tools.jar` on Java 8-. Controlling the `ClassLoader` as I describe above should work for that without having to directly reference `tools.jar`.

jkeljo:
I think I might need to just see a proof of concept to fully understand what you're suggesting, but here are my thoughts based on what I think you're saying.

> So we could make a custom rule, called kotlin-annotation-processor which would be responsible for the steps 1 and 2.

I would suggest that you have separate rules for 1 and 2. That way if Buck detects nothing has changed about the stubs, it can just skip running the annotation processors. This does not mean that the user has to manually write these rules. Buck would simply create them during action graph creation, in much the same way as it already expands a `java_library` into `DefaultJavaLibrary` and `CalculateAbiFromSource` and a few others.

> This rule would give you all the options you need to get kapt to work properly, for example, as tools.jar is a plugin, and the ap jar is a plugin as well, you would have an array of plugins (then on Java9, you simply don't pass the tools.jar)

I want to reiterate that I think the right way to give kapt access to the stuff that lives in `tools.jar` on Java 8 and elsewhere on Java 9 is by controlling which class loader you use to load the compiler.

Related, since we're running `kapt` in process...does it have an API that lets the caller instantiate the annotation processors, like `javac` does? If so, might as well give Kotlin the annotation processor `ClassLoader` sharing that Java has had with Buck for a while now. It's a huge perf win for people who use a lot of APs.

> you could stablish the output for the generated sources, which is basically an entry called kapt.kotlin.generated with the value as the desired path on the apoptions plugin parameter.
And them on the kotlin rule you set that path as one of the sources to look at.

This makes sense. From a Buck standpoint you'll want to make sure that the generated sources are considered the default output of the rule that runs annotation processors, and that you use a `BuildTargetSourcePath` to represent that when giving it to the kotlin rule.

Sayary:
Hi @jkeljo !~ I happen to see your last comment about running kapt in-process, is there any approach that we change it to out-of-process? There's a known issue about running ButterKnife with kotlinc in-process config, see https://github.com/JakeWharton/butterknife/issues/1148

kageiit:
> There's a known issue about running ButterKnife with kotlinc in-process config

@thalescm I dont think we should make kapt run out of process because of some issue with a thrid party library. That issue should be fixed by the library maintainers. Running out of process incurs significant overhead for process creation and not being able to take advantage of JIT.

Sayary:
@kageiit I agree in-process has lots of benefits about taking advantage of JIT, but could we make it an option like what kotlin gradle plugin does https://github.com/JetBrains/kotlin/blob/149b197b2491be178f9e46b09559d7e9e6f5867a/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt#L43

At least this will unblock users who use butterknife & kotlin and want to adopt BUCK.

kageiit:
Is it not possible to fix butter knife to deal with this? It seems like a problem with the library than the tool and it should be solved in the right place.

kageiit:
I wrote the original logic in butterknife that uses the sun trees api to make it support library projects. It should be possible to fix the issue in butterknife to use the available trees class that is available. I think k the fix should reside in butterknife as an out or process compiler does not make sense and can complicate the implementation of kept in buck more.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
thalescm: I really don't know, the compile expects `tools.jar` as a plugin, and use it internally. I'll have to see if there is a way to provide this ClassLoader directly.

kageiit: @thalescm The clue is in https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java#L160

This is where the `Kotlinc` is loaded by creating a new classloader with the kotlin jars passed in via buckconfig. you should change the parent classloader from null to `ToolProvider.getSystemToolClassLoader`

This way, classloader hierarchy will set setup such that you dont have to add the tools jar again when running the `KotlincStep`

brettwooldridge:
@Sayary I’m late to the party here, but I just wanted to note that the current kotlin_library rule *does* support out of process execution.

thalescm:
@brettwooldridge does this mean that this implementation break this current behavior of `kotlin_library` rule? Or is just something you'd want to see on kapt?

thalescm:
Would someone be able to take a look in this PR?

src/com/facebook/buck/util/PathMatchers.java
```diff
@@ -0,0 +1,11 @@
+package com.facebook.buck.util;
```
jkeljo: This feels like too broad a location as we try to modularize Buck further. Should probably be at least `com.facebook.buck.jvm`. @styurin probably has the most skin in the game here, though, so I defer to him.

src/com/facebook/buck/jvm/kotlin/Kotlinc.java
```diff
@@ -46,4 +46,8 @@ String getDescription(
       Path pathToSrcsList);
 
   String getShortName();
+
+  Path getAPPaths();
```
jkeljo: No abbreviations, please.

src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
```diff
@@ -166,23 +143,30 @@ Path getPathToScriptRuntimeJar() {
...
 
-    compiler = getKotlinHome().resolve(Paths.get("libexec", "lib", "kotlin-compiler.jar"));
-    if (Files.isRegularFile(compiler)) {
-      return compiler.normalize();
+  /**
+   * Get the path to the Kotlin annotation processing jar.
+   *
+   * @return the Kotlin annotation processing jar path
+   */
+  Path getPathToAP() {
```
jkeljo: No abbreviations, please. By the name I would have expected this to be getting the path to the annotation processor itself, and that's not what it's doing.

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
```diff
@@ -157,7 +175,7 @@ private Object loadCompilerShim(ExecutionContext context) {
 
       ClassLoader classLoader =
           classLoaderCache.getClassLoaderForClassPath(
-              null /* parent classloader */,
+              ToolProvider.getSystemToolClassLoader(),
```
jkeljo: Because `ToolProvider` itself is not thread safe, you'll need to synchronize this with the (currently single) other usage of `ToolProvider`, or else there may be strange and intermittent failures when running under `buckd`.

src/com/facebook/buck/jvm/java/JavacPipelineState.java
```diff
@@ -125,25 +127,37 @@ public boolean isRunning() {
...
-                  libraryJarParameters,
-                  compilerParameters.getAbiGenerationMode(),
-                  compilerParameters.getAbiCompatibilityMode(),
-                  compilerParameters.getSourceOnlyAbiRuleInfo());
-
-      closeables.add(invocation);
+      /* When compiling a kotlin project, we can't know if there will or not have annotation
+         processors that generates java source files to compile. So we pass the directories
+         where potencially generated files will be, and here, which runs after all kotlin steps
+         we filter for only java files, or don't invoke the compiler, if there aren't any */
```
jkeljo: This is another reason to do these as rules rather than steps -- you won't need to special case this since a `java_library` rule with empty sources already does the right thing.

src/com/facebook/buck/util/PathMatchers.java
thalescm: I'll wait for his response, then. Tks!

thalescm:
@jkeljo I strongly agree with the proposed changes.
About the rules, today java's AP is not a rule itself either. Though that shouldn't be an argument to not make the best of the way kotlin works, and as kotlin requires more than one call to `kotlinc` to make apt work when `javac` doesn't, I've preferred to do this way cause it was the way the PR started, so it was way more intuitive than doing from scratch.

If facebook core team decide it's best to do your proposed way, I can probably take a look in the future, but as this works for me right now, I probably won't have work time to allocate on that, so it would be slower.

About the locally scoped requested changes, I'll try to clean them by tomorrow :) 
Tks for the review

runningcode:
@jkeljo @thalescm Is it possible to have the AP as a separate rule? I think the reason that it is a step instead of a rule is that you can have human-written code call directly in to the generated code.
If the AP fails to generate code, that would mean that the compile "rule" failed, but you would only know until you ran the AP.
AP Generated code and human written code can call each other.

thalescm:
@runningcode @jkeljo The main possibility of failure, if split into two rules, is not providing the same sources on both rules (which is totally possible), thus it could fail drastically when compiling the source (as it would probably depend on code that was not generated).
Having human-written code calling directly in to the generated code would not be a problem, if you correctly pass the generated files as sources to be compiled. (note that they need to be sources, only putting their package on the classpath wouldn't solve. 
E.g You could generate classes that relies on the source code to compile (dagger, for example, does that for generating member injectors)
You could generate code that the human-written code would directly call (dagger does that too on components)
For a good integration with gradle, one should just put as `srcDirs` the locations where bucks stores the generated files (at current impl is at `buck-out/annotation/<module>/__<rule_name>_gen__` and `buck-out/annotation/<module>/__<rule_name>_kapt_generated__`

src/com/facebook/buck/util/PathMatchers.java
thalescm: But my point here is that this could be a location to store matchers for any kind of file, not only `jvm`s.
If this should be moved to `jvm` package, I'd say best is to separate into two classes, both inside their on language package, `JavaPathMatcher` and `KotlinPathMatcher`

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
thalescm: Do you think that wrapping this inside a `synchronized (ToolProvider.class) { ... }` as `JdkProvidedInMemoryJavac` does is enough?

thalescm:
I'm just not sure how to add the output of a rule as the src (not as dependency) to another rule. @jkeljo do you have an example?

src/com/facebook/buck/util/PathMatchers.java
jkeljo: Yep, I get that. I don't know what the policy is on "core" pieces of Buck knowing about the various plugins though.

src/com/facebook/buck/jvm/kotlin/JarBackedReflectedKotlinc.java
jkeljo: That should do it. If you wouldn't mind, though, factor that out into a helper class somewhere so that we don't have to repeat the comment every time.

jkeljo:
@thalescm, each src is just a `SourcePath`, so you should be able to use `getSourcePathToOutput` and just add that to the sources. It can even be a `.src.zip` (at least for Java).

docs/concept/buckconfig.soy
```diff
@@ -2379,7 +2379,7 @@ cxx_library(
 {call buckconfig.entry}
   {param section: 'kotlin' /}
   {param name: 'kotlin_home' /}
-  {param example_value: '/usr/local/Cellar/kotlin/1.1.1' /}
+  {param example_value: '/usr/local/Cellar/kotlin/1.2.21' /}
```
styurin: This change doesn't bring any value, does it?

thalescm: It actually does, instigating people to use most recent versions is very good. `1.1.1` is very outdated already :)

.gitignore
```diff
@@ -39,6 +39,7 @@
 /.idea/workspace.xml
 /.idea/misc.xml
 /.idea/copyright/profiles_settings.xml
+/.idea/markdown*
```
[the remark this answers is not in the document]
thalescm: It's a plugin I have on my IDEA, probably other people have it also, I didn't see any damage on adding this and thus not needing to keep filtering which files to stage

src/com/facebook/buck/util/PathHelper.java
```diff
@@ -0,0 +1,49 @@
...
+    ImmutableMultiset.Builder<Path> builder = new ImmutableMultiset.Builder<>();
+    getInnerFiles(filesystem, sourcePath, builder);
+    return builder.build().elementSet();
+  }
+
+  private static void getInnerFiles(ProjectFilesystem filesystem, Path sourcePath, ImmutableMultiset.Builder<Path> builder)
+      throws IOException {
+    if (filesystem.isDirectory(sourcePath)) {
+      for (Path child : filesystem.getDirectoryContents(sourcePath)) {
+        getInnerFiles(filesystem, child, builder);
```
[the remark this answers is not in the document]
thalescm: I'll check on how that works and change, tks for the tip!

.gitignore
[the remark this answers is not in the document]
styurin: Could you move it to a separate PR?

docs/concept/buckconfig.soy
styurin: For the purpose of documentation it doesn't matter

thalescm:
If you guys can take another look now :)

jkeljo:
I'm happy with the changes you made for me. I'll leave it to @styurin

thalescm:
Please don't merge it yet. I've found issues running android tests that do not happen on the current master version. I'm trying to understand what caused it and will probably make changes to fix it

thalescm:
Solved :) all right to merge (if no further changes are requested)

thalescm:
Thanks so much for the comments @yanex, I'll take a deeper look!

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -83,51 +122,232 @@ public void createCompileStep(
...
+                .build(),
+            kotlinc,
+            ImmutableList.<String>builder()
+                .addAll(extraArguments)
+                .add(MODULE_NAME)
+                .add(invokingRule.getShortNameAndFlavorPostfix())
+                .add(COMPILER_BUILTINS)
+                .add(LOAD_BUILTINS_FROM)
+                .add(PLUGIN)
+                .add(KAPT3_PLUGIN + APT_MODE + "stubs," + join)
```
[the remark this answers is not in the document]
thalescm: I've tested a lot, and the `,` are to keep passing more options together instead of having to provide `-P` in each of them

docs/rule/kotlin_library.soy
```diff
@@ -34,7 +34,12 @@ of the compiled class files and resources.
   {param name: 'srcs' /}
   {param default: '[]' /}
   {param desc}
-  The set of <code>.kt</code> or <code>.kts</code> files to compile for this rule.
+  The set of <code>.kt</code>, <code>.java</code> or <code>.kts</code> files to compile for this rule.
+  If any of the files in this list end in <code>.src.zip</code>,
+  then the entries in the ZIP file that end in <code>.java</code> or <code>.kt</code> will be
+  included as ordinary inputs to compilation. This is common when using
+  a {call buck.java_annotation_processor /} to auto-generate some Java/Kotlin source code that
+  needs to be compiled with some hand-written Java/Kotlin code.
```
jkeljo: `This is common when...` is all an implementation detail; someone writing `BUCK` files would never need to use this capability for that purpose.

src/com/facebook/buck/jvm/java/javax/BUCK
```diff
@@ -0,0 +1,7 @@
+java_library(
+    name = "javax",
+    srcs = [
+        "SynchronizedToolProvider.java",
```
jkeljo: `glob('*.java')`

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -139,4 +377,35 @@ public void createCompileStep(
...
+    steps.addAll(
+        MakeCleanDirectoryStep.of(
+            BuildCellRelativePath.fromCellRelativePath(
+                buildContext.getBuildCellRootPath(), filesystem, location)));
+    buildableContext.recordArtifact(location);
+  }
+
+  private String encodeOptions(Map<String, String> options) {
+    try {
+      ByteArrayOutputStream os = new ByteArrayOutputStream();
```
runningcode: try with resources?

test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/AnnotationProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.annotation.processing.SupportedOptions;
+import javax.annotation.processing.SupportedSourceVersion;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.TypeElement;
+import javax.tools.JavaFileObject;
+
+@SupportedAnnotationTypes("*")
+@SupportedSourceVersion(SourceVersion.RELEASE_8)
+@SupportedOptions({})
+public class AnnotationProcessor extends AbstractProcessor {
```
runningcode: would be cool to add a test that outputs kotlin files as well ;)

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -83,51 +128,244 @@ public void createCompileStep(
...
+                    .build())
+            .setSourceFilePaths(javaSourceFiles)
+            .build();
+
+    new JavacToJarStepFactory(
+        resolver,
+        ruleFinder,
+        projectFilesystem,
+        javac,
+        javacOptions.withAnnotationProcessingParams(AnnotationProcessingParams.EMPTY),
```
jkeljo: Internal CI is failing because annotation processors are not run on the Java parts of a mixed Java & Kotlin module, presumably because of this line.

src/com/facebook/buck/jvm/java/JdkProvidedInMemoryJavac.java
```diff
@@ -16,22 +16,16 @@
...
 import javax.tools.JavaCompiler;
-import javax.tools.ToolProvider;
 
 public class JdkProvidedInMemoryJavac extends Jsr199Javac {
   @Override
   protected JavaCompiler createCompiler(
       JavacExecutionContext context, SourcePathResolver pathResolver) {
-    JavaCompiler compiler;
-    synchronized (ToolProvider.class) {
-      // ToolProvider has no synchronization internally, so if we don't synchronize from the
```
dmitry-zaitsev: I think it would make sense to move it to `SynchronizedToolProvider` as it is not exactly obvious.

thalescm: I didn't get exactly what do you mean, could you explain better, @dmitry-zaitsev ?

dmitry-zaitsev: This synchronized block was moved into `SynchronizedToolProvider`

```
synchronized (ToolProvider.class) {
     compiler = ToolProvider.getSystemJavaCompiler();
}
```

It is not really obvious why this synchronization is performed, so there was a comment left to explain that. Now that this logic is moved into a new class the comment is removed and the purpose of `SynchronizedToolProvider` is not documented anywhere. Would be great to move the contents of this comment into the JavaDoc of `SynchronizedToolProvider`.

thalescm: Done 👍

src/com/facebook/buck/jvm/java/javax/SynchronizedToolProvider.java
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.tools.JavaCompiler;
+import javax.tools.ToolProvider;
+
+/**
+ * ToolProvider has no synchronization internally, so if we don't synchronize from the
+ * outside we could wind up loading the compiler classes multiple times from different
+ * class loaders.
+ */
+public class SynchronizedToolProvider {
```
styurin: If this PR is going to be in review for some other fixes you may want to move change related to this class to a separate PR

thalescm: This change was requested from @jkeljo in order to get this PR done. There was no reason for this class prior to the changes related to this PR

PR #1587, opened 2017-10-16
Xcode project support testing modular apple_library

robbertvanginkel:
If we are testing a modular apple_library, we expose it non-modular. This allows the
testing target to see both the public and private interfaces of the tested target
without triggering header errors related to modules. We hide the module definition by
using a filesystem overlay that overrides the module.modulemap with an empty file.

Note: stacked on top of #1561 and #1582.

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -1430,6 +1430,7 @@ public class ProjectGenerator {
           targetSpecificSwiftFlags.addAll(collectModularTargetSpecificSwiftFlags(targetNode));
         }
 
+        ImmutableList<String> testingOverlay = getExcludesForModulesUnderTests(targetNode);
         Iterable<String> otherSwiftFlags =
             Iterables.concat(
                 swiftBuckConfig.getCompilerFlags().orElse(DEFAULT_SWIFTFLAGS),
@@ -1447,6 +1448,7 @@ public class ProjectGenerator {
                 .addAll(
                     convertStringWithMacros(
                         targetNode, targetNode.getConstructorArg().getPreprocessorFlags()))
+                .addAll(testingOverlay)
                 .build();
         Iterable<String> otherCxxFlags =
             ImmutableList.<String>builder()
@@ -1460,6 +1462,7 @@ public class ProjectGenerator {
                 .addAll(
                     convertStringWithMacros(
                         targetNode, targetNode.getConstructorArg().getPreprocessorFlags()))
+                .addAll(testingOverlay)
                 .build();
 
         Iterable<String> otherLdFlags =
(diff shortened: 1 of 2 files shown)


src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -2492,6 +2532,11 @@ private Path getObjcModulemapVFSOverlayLocationFromSymlinkTreeRoot(Path headerSy
     return headerSymlinkTreeRoot.resolve("objc-module-overlay.yaml");
   }
 
+  private Path getTestingModulemapVFSOverlayLocationFromSymlinkTreeRoot(
+      Path headerSymlinkTreeRoot) {
+    return headerSymlinkTreeRoot.resolve("testing.yaml");
```
milend: I'd be consistent with the naming of `objc-module-overlay.yaml` and name this `testing-overlay.yaml`.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1503,6 +1506,27 @@ private PBXNativeTarget generateBinaryTarget(
     return target;
   }
 
+  private ImmutableList<String> getExcludesForModulesUnderTests(
```
milend: Nit: I'd name this `getFlagsForExcludesForModulesUnderTests` so that it's clear that it returns flags.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1503,6 +1506,27 @@ private PBXNativeTarget generateBinaryTarget(
...
+      TargetNode<? extends CxxLibraryDescription.CommonArg, ?> targetNode) {
```
milend: Nit: Let's clarify that `targetNode` is actually the `testingTarget`.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1503,6 +1506,27 @@ private PBXNativeTarget generateBinaryTarget(
...
+    ImmutableList.Builder<String> testingOverlayBuilder = new ImmutableList.Builder<>();
+    visitRecursivePrivateHeaderSymlinkTreesForTests(
+        targetNode,
+        (nativeNode, headerVisibility) -> {
```
milend: Nit: I think it would be more descriptive for `nativeNode` to be named `libBeingTested`.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1503,6 +1506,27 @@ private PBXNativeTarget generateBinaryTarget(
...
+          // If we are testing a modular apple_library, we expose it non-modular. This allows the
```
milend: If I understand correctly, this implies that testing targets written in Obj-C cannot `@import` the libs under testing and have to resort to `#import`. Is that right?

robbertvanginkel: Yes, that is correct. It seems non-ideal but after some investigation into how Xcode handles this this seems like the way to go. I couldn't find any configuration of settings in a hand made xcode project which would allow you to use both import styles.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: 👍

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: 👍

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
robbertvanginkel: 👍

robbertvanginkel:
@milend I rebased and updated this. Can you give it another look?

milend:
@robbertvanginkel LGTM, apologies for the delayed review. Do you mind rebasing on top of the latest master so I can land it? Thanks!

Issue #1547, opened 2017-09-28
Watch apps fail to link due to missing path for bitcode symbol map


rmaz:
Currently it is not possible to link watch apps built with buck. This is due to the default link options being specified in https://github.com/facebook/buck/blob/6f7961687cafc29d25230507a539fd17d4c05895/src/com/facebook/buck/apple/AppleCxxPlatforms.java#L201 

The `bitcode_symbol_map` flag requires a path:

```
     -bitcode_symbol_map path
                 Specifies the output for bitcode symbol reverse mapping (.bcsymbolmap). If path is an existing directory,
                 UUID.bcsymbolmap will be written to that directory.  Otherwise, the reverse map will be written to a file at
                 path.
```

This causes linker errors as the next arg in the link list is interpreted as a path and fails the link.

Ideally we would pass in the gen folder for the linked target in at this point, but that is not possible from this method. This should be removed from the platform and added in some helper class further up IMO. In the short term we can just remove these flags and add them to the specific targets that require them. Another simple alternative is to specify a single folder path here, but that makes it difficult to know what symbolmap belongs to which target.

rmaz:
I had a quick look to see if there was a simple way to resolve this, but I could not find a decent integration point to fetch the output location to put the symbol map in. It feels like bitcode enabling and hiding should be handled by flavours, as it has to handle both compile flag additions and output location aware link flags which cannot be handled in BUCK files directly.

PR #1588, opened 2017-10-16
don't hide bitcode symbols

rmaz:
Currently it is not possible to link watch apps as the `-bitcode_symbol_map` flag requires a path. I took a look for a good place to put a path for this, but it seems better handled in the BUCK files for now.

#1547


ttsugriy:
@dinhviethoa could you please take a look with your expert iOS knowledge

rmaz:
any movement on this one?

yiding:
@ryu2 do you recall why you set these flags? your commit did not provide any rationale for these flags.

ryu2:
IIRC, this was what Xcode adds automatically in its WatchOS builds, or was at the time of the original commit.  Since Apple's App Store validation is a bit of a black box, I wanted to play it safe and replicate Xcode's flags as much as possible.

@rmaz If you've built and successfully submitted an app to Apple with this change, then sure.

rmaz:
Thanks for merging this one in.

> IIRC, this was what Xcode adds automatically in its WatchOS builds

This is what is added automatically by Xcode when archiving, but it also adds a path for the bitcode symbol map to be placed at, which is required. Its equivalent to stripping more or less, but has to happen at link time, so unfortunately is not a great fit for bucks existing strip flavour approach. I could not see an obvious place to add this without adding a whole new flavour, so it seemed more sensible to just omit the flags for now and let users add them for their own release builds. This at least makes it possible to link a watch binary, and matches what users have to do for any iOS app with bitcode enabled.

> However, if you need to add those flags to the BUCK file in order for a app to pass validation, then we will have to make sure this change doesn't break releases.

I don't believe this is required to pass validation, but desirable for sure. Open to suggestions for the best way of getting this implemented. The same issue applies to all app store apps, not just watch apps, so we should come up with a clean way of handling this in general. IMO the only way to get this done is by adding a `bitcode` flavour, which we could use to apply the correct compiler and linker flags and output the symbol map next to the dSYM. Then the flavour could be applied separately for release builds, avoiding slowing down debug builds and so on. Thoughts?

dinhviethoa:
I think you need to rebase and then we can land it.

rmaz:
Oh, misunderstood the phab import message, rebased now

dinhviethoa:
Could you fix the following unit test?
//test/com/facebook/buck/apple:apple_test_integration - watchApplicationBundle (com.facebook.buck.apple.AppleBundleIntegrationTest) (platform: macosx)

To reproduce:
buck test //test/com/facebook/buck/apple:apple_test_integration --filter AppleBundleIntegrationTest#watchApplicationBundle

rmaz:
Tricky, looks like all tests requiring code signing have been disabled. I'll just pull out this test for now, doesn't seem possible to validate as this only applies to device builds.

PR #1596, opened 2017-10-19
Use filelists in SwiftCompile

robbertvanginkel:
Introduces a new config option `use_filelist` on the `[swift]` section. If enabled it will pass a `-filelist` parameter to swiftc instead of passing all file names in argv.

This can make compile commands significantly shorter when having a large amount of source files.

### DIFF
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -53,6 +53,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -203,9 +209,7 @@ private SwiftCompileStep makeCompileStep(SourcePathResolver resolver) {
         });
 
     compilerCommand.addAll(Arg.stringify(compilerFlags, resolver));
-    for (SourcePath sourcePath : srcs) {
-      compilerCommand.add(resolver.getRelativePath(sourcePath).toString());
-    }
+    compilerCommand.add("-filelist", swiftFileListPath.toString());
```
milend: @robbertvanginkel Can we make this a config option? I think it will be useful for debugging and logging.

PR #1505, opened 2017-09-05
Don't use test_host as bundle_loader for UI tests

robbertvanginkel:
This allows `apple_test` bundles with `is_ui_test` to be correctly build
with buck, so that they can be used with an external test runner.

Currently, the `test_host_app` is used as the `bundle_loader` when
linking an apple_test. Since the uitest bundle isn't injected into the
app it is testing, but into a stub app when building a uitest we should
not use it as the bundle_loader.

PR #1615, opened 2017-11-10
Fix building of apple_test with is_ui_test

robbertvanginkel:
`apple_test` targets have a `test_host` if the tests test an application, used to set the `-bundle_loader` argument during linking to provide the test bundle with the symbols that would be available when the test bundle gets injected into the app.

With #1505, I introduced a change that when a test target is marked as `is_ui_test`, it would not set this bundle loader flag. This is because XCUITests don't run injected into the app, but in a separate process. The implementation in that PR made sure the test_host_app wouldn't be passed when creating a test target, having the unintended side effect of removing the runtime dependency of test host app. This resulted in the test host app not being build when doing `buck build //test:target`.

This PR changes it so that the TestHostInfo can be created and passes the test_host_app. This way the runtime dependency stays in place while the host app is not used as bundle_loader.


robbertvanginkel:
@milend can you review this one?

milend:
@robbertvanginkel Yup, I'll take a look tomorrow.

milend:
LGTM.

PR #1619, opened 2017-11-11
Minor fix

kant:


ttsugriy:
thank you for making buck's documentation better, @kant!

ttsugriy:
@kant, I just realized that this is a change in a third-party code which should only be updated once upstream version is updated. Please consider contributing your changes to the argparse project after which we'll be glad to accept an updated version of argparse library.

PR #1627, opened 2017-11-16
Add new kotlin jars required during compilation from Kotlin 1.1.60

kageiit:
Compilation fails otherwise with class not found exceptions from kotlin 1.1.60 onwards

### DIFF
--- a/src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
+++ b/src/com/facebook/buck/jvm/kotlin/KotlinBuckConfig.java
@@ -47,6 +47,8 @@ public class KotlinBuckConfig {
       ImmutableSet<SourcePath> classpathEntries =
           ImmutableSet.of(
               delegate.getSourcePath(getPathToStdlibJar()),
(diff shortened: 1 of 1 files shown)


kageiit:
cc @hzsweers @cwoodwar6 @artem-zinnatullin

hzsweers:
Did you not need to handle the annotation processing jar?

kageiit:
> Did you not need to handle the annotation processing jar?

Its not required for compilation. kapt support is being handled separately by @cwoodwar6 's kapt PR that will attach the jar as a compiler argument instead

hzsweers:
The reason I ask is because our internal build failed due to okbuck trying
to download it, so I wasn't sure where in the pipeline that was happening

kageiit:
> The reason I ask is because our internal build failed due to okbuck trying
> to download it, so I wasn't sure where in the pipeline that was happening

Thats handled separately in https://github.com/uber/okbuck/commit/00b96385c4879cace766894d20820866cc74d182

hzsweers:
:+1: LGTM then

PR #1639, opened 2017-11-22
Add ability to skip creating intermediate R.java

raviagarwal7:
of dependent libraries/aar for android_library target via skip_prebuilt_r_dot_java

- add `skip_non_union_r_dot_java` arg to `android_library` target.
- add `resource_union_package` arg to `robolectric_test` target.
- Only return RDotJavaFiles of union package when `skip_non_union_r_dot_java` is true.
- Only create union package’s R.java file and skip other packages when `skip_non_union_r_dot_java` is true.

This helps in improving the build times as only `n` R.java will be compiled now (1 for each library in the dependency graph), which earlier it used to be `n * log(n)`.

### DIFF
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -155,6 +155,11 @@ public class AndroidLibraryDescription
 
     Optional<String> getResourceUnionPackage();
 
+    @Value.Default
+    default boolean isSkipPrebuiltRDotJava() {
+      return false;
+    }
+
     Optional<String> getFinalRName();
   }
 
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -95,7 +95,7 @@ public class AndroidLibraryDescription
           buildTarget, projectFilesystem, params, args.getSrcs(), args.getMavenCoords());
     }
 
-    if (args.isSkipPrebuiltRDotJava()) {
+    if (args.isSkipNonUnionRDotJava()) {
       Preconditions.checkArgument(
           args.getResourceUnionPackage().isPresent(),
           "union_package should be specified if skip_prebuilt_r_dot_java is set");
@@ -163,7 +163,7 @@ public class AndroidLibraryDescription
     Optional<String> getResourceUnionPackage();
 
     @Value.Default
-    default boolean isSkipPrebuiltRDotJava() {
+    default boolean isSkipNonUnionRDotJava() {
       return false;
     }
 
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -41,6 +41,7 @@ import com.facebook.buck.rules.SourcePath;
 import com.facebook.buck.rules.TargetGraph;
 import com.facebook.buck.toolchain.ToolchainProvider;
 import com.facebook.buck.util.immutables.BuckStyleImmutable;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableSet;
 import java.util.Optional;
@@ -92,6 +93,12 @@ public class AndroidLibraryDescription
     if (buildTarget.getFlavors().contains(JavaLibrary.SRC_JAR)) {
       return new JavaSourceJar(
           buildTarget, projectFilesystem, params, args.getSrcs(), args.getMavenCoords());
+    }
+
+    if (args.isSkipPrebuiltRDotJava()) {
+      Preconditions.checkArgument(
+          args.getResourceUnionPackage().isPresent(),
+          "union_package should be specified if skip_prebuilt_r_dot_java is set");
     }
 
     boolean hasDummyRDotJavaFlavor = buildTarget.getFlavors().contains(DUMMY_R_DOT_JAVA_FLAVOR);
--- a/src/com/facebook/buck/android/AndroidLibraryDescription.java
+++ b/src/com/facebook/buck/android/AndroidLibraryDescription.java
@@ -98,7 +98,7 @@ public class AndroidLibraryDescription
     if (args.isSkipNonUnionRDotJava()) {
       Preconditions.checkArgument(
           args.getResourceUnionPackage().isPresent(),
-          "union_package should be specified if skip_prebuilt_r_dot_java is set");
+          "union_package should be specified if skip_non_union_r_dot_java is set");
     }
 
     boolean hasDummyRDotJavaFlavor = buildTarget.getFlavors().contains(DUMMY_R_DOT_JAVA_FLAVOR);
--- a/src/com/facebook/buck/android/MergeAndroidResourcesStep.java
+++ b/src/com/facebook/buck/android/MergeAndroidResourcesStep.java
@@ -254,9 +254,9 @@ public class MergeAndroidResourcesStep implements Step {
       if (skipNonUnionRDotJava) {
         Preconditions.checkArgument(
             unionPackage.isPresent(),
-            "union_package should be specified if skip_prebuilt_r_dot_java is set");
+            "union_package should be specified if skip_non_union_r_dot_java is set");
 
-        // If skip_prebuilt_r_dot_java is true remove all packages except union package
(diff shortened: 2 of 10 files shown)


raviagarwal7:
cc: @dreiss @kageiit

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -152,15 +157,30 @@ public static MergeAndroidResourcesStep createStepForUberRDotJava(
...
+          .toSortedSet(natural());
+
+    } else {
+      return FluentIterable.from(androidResourceDeps)
+          .transform(HasAndroidResourceDeps::getRDotJavaPackage)
+          .append(unionPackage
+              .map(Collections::singletonList)
+              .orElse(Collections.emptyList())
+          )
+          .transform(this::getPathToRDotJava)
```
jkeljo: Rather than duplicating code between the branches, do something like the below. It makes the function of the switch clearer and eliminates the possibility of drift in the duplicated code over time.
```
FluentIterable<String> packages = FluentIterable.from(
    unionPackage.map(Collections::singletonList).orElse(Collections.emptyList());

if (!skipPrebuiltRDotJava) {
    packages = packages.append(
        FluentIterable.from(androidResourceDeps)
            .transform(HasAndroidResourceDeps::getRDotJavaPackage));
}

return packages
    .transform(this::getPathToRDotJava)
    .toSortedSet(natural());
```

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -216,29 +236,55 @@ public StepExecutionResult execute(ExecutionContext context)
...
+
+      if (skipPrebuiltRDotJava) {
+        // If a resource_union_package was specified and skip_prebuilt_r_dot_java is true,
+        // copy all resource into that package and remove all other packages.
+        Preconditions.checkArgument(
+            unionPackage.isPresent(),
+            "union_package should be specified if skip_prebuilt_r_dot_java is set");
+
         String unionPackageName = unionPackage.get();
-        // Create a temporary list to avoid concurrent modification problems.
```
jkeljo: The change loses this comment, and it's an important one

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -216,29 +236,55 @@ public StepExecutionResult execute(ExecutionContext context)
               filesystem,
               useOldStyleableFormat);
 
-      // If a resource_union_package was specified, copy all resource into that package,
-      // unless they are already present.
-      if (unionPackage.isPresent()) {
+      ImmutableSet.Builder<String> requiredPackages = ImmutableSet.<String>builder();
+
+      if (skipPrebuiltRDotJava) {
```
jkeljo: Similar feedback here -- there's a lot of code duplication here that should be eliminated.

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
raviagarwal7: good point, removed duplicate as you suggested.

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
raviagarwal7: remove code duplication

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
raviagarwal7: added the comment back

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -216,29 +230,47 @@ public StepExecutionResult execute(ExecutionContext context)
...
 
+      ImmutableSet.Builder<String> requiredPackages = ImmutableSet.<String>builder();
+
+      // Create a temporary list to avoid concurrent modification problems.
+      ArrayList<Entry<String, RDotTxtEntry>> entries =
+          new ArrayList<>(rDotJavaPackageToResources.entries());
+
+      if (skipPrebuiltRDotJava) {
+        // If skip_prebuilt_r_dot_java is true remove all packages except union package
+        Preconditions.checkArgument(
```
dreiss: This should also be checked in the Description (or the rule constructor) to ensure that we catch this early.

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -216,29 +230,47 @@ public StepExecutionResult execute(ExecutionContext context)
...
+        requiredPackages.addAll(symbolsFileToRDotJavaPackage.values());
+      }
+
       // If a resource_union_package was specified, copy all resource into that package,
       // unless they are already present.
       if (unionPackage.isPresent()) {
         String unionPackageName = unionPackage.get();
+        requiredPackages.add(unionPackageName);
+
         // Create a temporary list to avoid concurrent modification problems.
```
dreiss: Looks like this comment is no longer relevant?

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
```diff
@@ -216,29 +230,47 @@ public StepExecutionResult execute(ExecutionContext context)
               filesystem,
               useOldStyleableFormat);
 
+      ImmutableSet.Builder<String> requiredPackages = ImmutableSet.<String>builder();
+
+      // Create a temporary list to avoid concurrent modification problems.
```
dreiss: Not sure this comment makes sense anymore.  Maybe comment why this has to be created before you reassign rDotJavaPackageToResources below?

src/com/facebook/buck/android/AndroidLibraryDescription.java
```diff
@@ -145,6 +145,11 @@ public void findDepsForTargetFromConstructorArgs(
 
     Optional<String> getResourceUnionPackage();
 
+    @Value.Default
+    default boolean isSkipPrebuiltRDotJava() {
```
dreiss: Why "prebuilt"?

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
raviagarwal7: yes removed the comment.

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
raviagarwal7: added a check in the rule constructor.

src/com/facebook/buck/android/MergeAndroidResourcesStep.java
raviagarwal7: improved the comment

src/com/facebook/buck/android/AndroidLibraryDescription.java
raviagarwal7: using `prebuilt` as this scenario is only applicable with union package enabled which in turn only leaves us with prebuilt's R Dot Java. setting this would skip the prebuilt's r dot java.

raviagarwal7:
@dreiss - please review, have made the requested code review changes.

src/com/facebook/buck/android/AndroidLibraryDescription.java
dreiss: I still don't understand what "prebuilt" means in this context.  That's the term we generally use for binaries that are checked into the repository.  Maybe just name this `isSkipNonUnionRDotJava`?  I don't know of any other name for these packages.  `isSkipRDotJavaForThePackageThatDefinesEachResource` doesn't seem quite right. :)

Also, can you document this parameter in the docs subdir?

raviagarwal7: renamed to `isSkipNonUnionRDotJava` and documented the param in docs.

kageiit:
@jkeljo @dreiss Ping :)

dreiss:
Sorry, looks like the patch doesn't apply anymore?

raviagarwal7:
just rebased to latest master although didn't encounter any merged conflicts.

dreiss:
Hrm.  Might be a conflict with our internal tree.  Sorry.  I'm going to attempt to land manually.

raviagarwal7:
nw, thanks for landing!

raviagarwal7:
closing as it's merged into master now.

PR #1640, opened 2017-11-22
Verify bundle resources do not conflict

rmaz:
Make it a build error when an apple bundle has multiple resources with the same path that will end up overwriting each other.


robbertvanginkel:
@milend can you take a look at this?

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -426,6 +427,7 @@ public boolean isLegacyWatchApp() {
             resources.getResourceDirs(),
             resources.getDirsContainingResourceDirs(),
             resources.getResourceFiles()))) {
+      verifyResourceConflicts(resources, context.getSourcePathResolver());
```
milend: We should introduce this behaviour behind a config option as this has the potential to break existing apps which already include duplicate resources (which means we cannot update Buck without fixing any offending apps which is impractical).

The config for this should be off by default.

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -670,6 +672,24 @@ public boolean isLegacyWatchApp() {
     return stepsBuilder.build();
   }
 
+  private void verifyResourceConflicts(AppleBundleResources resources,
+                                       SourcePathResolver resolver) {
+    // Ensure there are no resources that will overwrite each other
+    Set<Path> filenames = new HashSet<>();
+    for (SourcePath path : Iterables.concat(
+        resources.getResourceDirs(),
```
milend: It seems that this is duplicating the resource paths form lines 427-429:

    resources.getResourceDirs(),
    resources.getDirsContainingResourceDirs(),
    resources.getResourceFiles()

We should extract those and pass them into the method instead.

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -670,6 +672,24 @@ public boolean isLegacyWatchApp() {
...
+        resources.getDirsContainingResourceDirs(),
+        resources.getResourceFiles())) {
+      Path filename = resolver.getRelativePath(path).getFileName();
+      if (filenames.contains(filename)) {
+        throw new HumanReadableException(
+            "Bundle contains multiple resources with filename %s", filename);
```
milend: Nit: `filename` can be inaccurate as it can be directories, too. Can we improve the naming?

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -670,6 +672,24 @@ public boolean isLegacyWatchApp() {
     return stepsBuilder.build();
   }
 
+  private void verifyResourceConflicts(AppleBundleResources resources,
+                                       SourcePathResolver resolver) {
+    // Ensure there are no resources that will overwrite each other
+    Set<Path> filenames = new HashSet<>();
+    for (SourcePath path : Iterables.concat(
+        resources.getResourceDirs(),
+        resources.getDirsContainingResourceDirs(),
```
milend: This check won't work correctly because it's the contents *inside* that get copied to the root of the resources dir, not the dirs themselves.

The deeper issue here is that those dirs will usually not have the contents at this point but only be generated later (e.g., a `genrule`). So you won't be able to perform the check here.

Given the above, you'd have to move the check to be part of the `CopyStep` itself.

test/com/facebook/buck/apple/AppleBundleIntegrationTest.java
```diff
@@ -622,6 +622,16 @@ public void appBundleWithResources() throws Exception {
             "%s"));
   }
 
+  @Test
+  public void appBundleWithConflictingResources() throws Exception {
```
milend: Can you add another test where we have duplicate nested filenames, so something like:

    FolderA/Image.png
    FolderB/Image.png

rmaz: How can I add a test to cover `DirsContainingResourceDirs` too?

src/com/facebook/buck/apple/AppleBundle.java
rmaz: I was a bit confused about `DirsContainingResourceDirs`, is this only the output of genrules or also other stuff? How can I set up a BUCK config to check this?

rmaz: If we move this to `CopyStep`, do you want this to be enabled globally or just for apple bundle resources?

milend: @rmaz I had to look into it and it seems it's only used for `js_bundle` (see `JsBundleDescription::addAppleBundleResources()`). I couldn't find any integration tests for `js_bundle`, so I think we can ignore `DirsContainingResourceDirs` for now (but please put a `TODO`).

test/com/facebook/buck/apple/AppleBundleIntegrationTest.java
milend: @rmaz As per my above comment, I don't think we can easily test this as `js_bundle` doesn't seem to have integration tests.

src/com/facebook/buck/apple/AppleBundle.java
rmaz: done

rmaz:
Updated per comments, if you have some preferred naming let me know

test/com/facebook/buck/apple/AppleBundleIntegrationTest.java
```diff
@@ -632,6 +632,16 @@ public void appBundleWithConflictingResources() throws Exception {
...
   }
 
+  @Test
+  public void appBundleWithConflictingNestedFolderResources() throws Exception {
+    ProjectWorkspace workspace =
+        TestDataHelper.createProjectWorkspaceForScenario(this, "app_bundle_with_conflicting_nested_resources", tmp);
+    workspace.setUp();
+
+    BuildTarget target = workspace.newBuildTarget("//:DemoApp#iphonesimulator-x86_64,no-debug");
+    workspace.runBuckCommand("build", target.getFullyQualifiedName()).assertFailure();
```
milend: @rmaz We should add another test that actually tests for success but has files with the same name inside the folders. We can just copy `app_bundle_with_conflicting_nested_resources` and change:

    files = glob(["**/*.png"]),

to:

    dirs = ["ImagesA", "ImagesB"],

Alternatively, we can just skip the images as the localizable strings all have the same file names inside.

milend:
@rmaz LGTM, I'd just want to add one more test that asserts success (in addition to the other two tests that assert failure).

rmaz:
done, is this what you were after?

milend:
@rmaz Yup!

PR #1648, opened 2017-11-28
Use plutil to compile strings files

robbertvanginkel:
Compiling strings files to binary plists makes their size smaller, partially due to removing the comments.


src/com/facebook/buck/apple/AppleCxxPlatforms.java
```diff
@@ -301,6 +301,12 @@ static AppleCxxPlatform buildWithExecutableChecker(
     Tool lldb =
         VersionedTool.of(getToolPath("lldb", toolSearchPaths, xcodeToolFinder), "lldb", version);
 
+    Tool plutil =
+        VersionedTool.of(
+            getToolPath("plutil", ImmutableList.of(Paths.get("/", USR_BIN)), xcodeToolFinder),
```
robbertvanginkel: `plutil` is a util that has been on osx since 10.2, but unlike all these other tools it is not shipped in Xcode. This seemed like the preferred way of getting it here.

ryu2:
@robbertvanginkel Have you done comparisons across a variety of languages?  Binary plists use UTF-16 encoding for text, and we found (at least for us) having UTF-8 encoded strings files saved overall app size.  (We don't have comments in our .strings files though.)

Perhaps make this configurable?

robbertvanginkel:
Hmm, interesting! Should definitely make this configurable then.

src/com/facebook/buck/apple/AppleBundle.java
```diff
@@ -1043,15 +1046,19 @@ private void addResourceProcessingSteps(
...
+              new PlutilConvertStep(
+                  getBuildTarget(),
+                  getProjectFilesystem(),
+                  plutil.getEnvironment(resolver),
+                  plutil.getCommandPrefix(resolver),
+                  PlutilConvertStep.Format.binary1,
+                  sourcePath,
+                  destinationPath));
+        } else {
+          stepsBuilder.add(CopyStep.forFile(getProjectFilesystem(), sourcePath, destinationPath));
```
yiding: Do you think it would make sense to run the the plist through plutil either way? This way any comments can still be stripped.

yiding:
Have you looked at using the ddplist library, which buck already uses to write plists for the xcode project generator, to read and write the plist files? It should support binary plists.

I would rather use that instead of a system utility.

robbertvanginkel:
I did look at the ddplist, the reason this seems tricky to use is that the `.strings` files we use for translation are ascii/NeXTSTEP plists with a dictionary, but without the enclosing `{ ... }` indicating the top level dict. Because of this, ddplist doesn't read and can't convert the strings files even though plutil could.

We could try to read the file and add the curly brackets before passing it to the ddplist parser, that way we could write the binary plists without comments.

yiding:
How about you add the braces hack as described, and open a github issue on https://github.com/3breadt/dd-plist, so that we can update our dd-plist library if/when this is supported?

robbertvanginkel:
There was an open issue regarding this that was closed as a won'tfix: https://github.com/3breadt/dd-plist/issues/29. The suggested workaround is wrapping the file in the extra chars.

We could just do the wrapping in buck, and not attempt to add it to ddplist? Are we okay with this hack?

yiding:
I'm okay with this hack.

PR #1651, opened 2017-11-28
Improve project documentation and fix typo.

clonetwin26:
This diff adds docs for the parameters we have found very useful in a large Mono repo project. It also fixes a type in the description for exclude-artifacts.

### DIFF
--- a/docs/command/project.soy
+++ b/docs/command/project.soy
@@ -65,6 +65,26 @@
 {/call}
 </ul>
 
+<h2>Uncommon Parameters</h2>
+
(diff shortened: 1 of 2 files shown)


docs/command/project.soy
```diff
@@ -65,6 +65,26 @@
 {/call}
 </ul>
 
+<h2>Uncommon Parameters</h2>
```
styurin: why do we need this? Looks like we don't use h2 in other place. I'd just add these parameters without putting them into a separate category

PR #1653, opened 2017-11-29
delete remnants of buck test --dry-run

jiangty-addepar:
This PR deletes the remaining dead code relating to `buck test --dry-run`.

https://github.com/facebook/buck/commit/4ea890754b5612c41cacf6a94d089a3960282e40 was an incomplete un-revert of https://github.com/facebook/buck/commit/b5f4e756a922c74b1e3d8fad7ba9d704dfd2f771.

Test Plan: `buck test`.

### DIFF
--- a/src/com/facebook/buck/event/listener/SuperConsoleEventBusListener.java
+++ b/src/com/facebook/buck/event/listener/SuperConsoleEventBusListener.java
@@ -123,7 +123,6 @@ public class SuperConsoleEventBusListener extends AbstractConsoleEventBusListene
   private final AtomicInteger numExcludedTests = new AtomicInteger(0);
   private final AtomicInteger numDisabledTests = new AtomicInteger(0);
   private final AtomicInteger numAssumptionViolationTests = new AtomicInteger(0);
-  private final AtomicInteger numDryRunTests = new AtomicInteger(0);
 
   private final AtomicReference<TestRunEvent.Started> testRunStarted;
   private final AtomicReference<TestRunEvent.Finished> testRunFinished;
@@ -731,10 +730,8 @@ public class SuperConsoleEventBusListener extends AbstractConsoleEventBusListene
     int testFailuresVal = numFailingTests.get();
     int testSkipsVal =
         numDisabledTests.get()
-            + numAssumptionViolationTests.get()
-            +
             // don't count: numExcludedTests.get() +
-            numDryRunTests.get();
+            + numAssumptionViolationTests.get();
     if (testSkipsVal > 0) {
       return Optional.of(
           String.format(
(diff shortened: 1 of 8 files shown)


styurin:
I don't build. Please, submit working code :)

jiangty-addepar:
I ran both `ant` and `./bin/buck test` successfully locally...

The build failures 
```
FAILURE: Build failed with an exception.
* What went wrong:
Task 'assemble' not found in root project 'buck'.
```

```
W: Failed to fetch http://ppa.launchpad.net/chris-lea/redis-server/ubuntu/dists/precise/main/binary-amd64/Packages  Unable to connect to ppa.launchpad.net:http:
```

jiangty-addepar:
How do I trigger a rebuild?

ttsugriy:
@jiangty-addepar, there is an issue with our internal build, which we are going to take care of very soon. ETA - EOD.

jiangty-addepar:
Thanks @ttsugriy !

ttsugriy:
for some reason `//test/com/facebook/buck/android:binary-integration-2` is timing out with this commit. I'm investigating...

ttsugriy:
Update: I'm still trying to figure out why this test is timing out. The weird thing is that failed test runner invocation seems to contain `--dry-run` flag which I have no idea where it comes from...

ttsugriy:
So I finally found what the issue is - we use external test runner for running buck tests and it uses `--dry-run` to list tests before running them. Now that `--dry-run` command line option is removed test runner tries to interpret `--dry-run` as a class and fails to do so. I'll have to see if we don't actually need `--dry-run` internally. @jiangty-addepar, are you sure nobody else is relying on `--dry-run`'s existence?

ttsugriy:
so far I don't see a quick way for us to remove a dependency on `--dry-run` :(

jiangty-addepar:
@ttsugriy The year-old commit I linked in my PR description actually deleted the `--dry-run` command-line option; in fact, on master, when I run 
```bash
./bin/buck test --dry-run
````
it tells me `"--dry-run" is not a valid option`.

This PR should only delete some internal references to `--dry-run`, so what you're saying is quite surprising!

ttsugriy:
@jiangty-addepar, I'm referring to https://github.com/facebook/buck/blob/master/src/com/facebook/buck/testrunner/BaseRunner.java#L235-L237 which you haven't removed in your change, but I believe it's the one that stopped working because of other changes :(

jiangty-addepar:
@ttsugriy wow, I guess I missed that! I'm still confused though: are you saying that "if we delete the references in `BaseRunner`, things will break"? Currently, does Buck only support `--dry-run` for Java tests?

ttsugriy:
@jiangty-addepar, tests are failing with and without those lines removed - just in different ways - when that line is removed the failure says that it cannot interpret `--dry-run` as a class name, but with that option tests are simply timing out :(

jiangty-addepar:
To summarize: facebook internally uses `--dry-run`, so we shouldn't delete it?

ttsugriy:
yes, looks like our internal test runner is relying on `--dry-run` to list all test cases :( I'll try to follow-up with someone working on the test runner to see if there is a plan to change this.

v-jizhang:
`--dry-run` is still used internally.

PR #1656, opened 2017-11-30
Fix modulars swift project umbrella header generation issue

robbertvanginkel:
In a previous refactor, accidentally we check the wrong variable to determine if the user supplied an umbrella header. The result is if the user sets `apple.generate_missing_umbrella_headers`, buck project would overwrite the user's symlinked header. This fixes it and adds a unittest.


robbertvanginkel:
@milend

milend:
@robbertvanginkel LGTM.

milend:
@robbertvanginkel Land failed due to merge conflict, can you please rebase?

PR #1657, opened 2017-12-01
Improvements to TestNG test runner

jiangty-addepar:
(This PR is on top of https://github.com/facebook/buck/pull/1653, so ignore the first commit for now.)

This is a collection of TestNG test improvements from our fork of Buck. Changes include:

- `Minor fixes to testng runner.`
    - Mark skipped tests as failed
    - No hard time limit per test suite
- `Respect ClassName and MethodName filters in TestNG runner.`
    - Make ClassName / MethodName filters work with TestNG runner
    - Remove TestNGWrapper
    - Remove "TmpSuite" xml suite
- `TestNG results now show method parameters.`
    - Show method parameters in TestNG test results (for example, for tests that use DataProviders)
- `Support Stdout and Exception Details when Test Setup Fails`
    - Fail test and print exceptions/stdout when test setup fails
    - Record all test failures in JUnit test report, including those from test setup

The remaining changes are to tests. I did some refactoring to use the `buck` cell (reducing jar duplication) for an existing test as well.

Test plan: `buck test`; added a number of integration tests.

PR #1662, opened 2017-12-04
[TestNG] Mark skipped tests as ASSUME

jiangty-addepar:
1st part of https://github.com/facebook/buck/pull/1657.

- Mark skipped tests as ASSUME
    - Test plan: added test in TestNGIntegrationTest
- Remove hard timeout on all tests
    - Test plan: run `buck test`. Not sure it's worth it to write a test that just sleeps for 10s


jiangty-addepar:
`SuperConsoleBusListenerTest` seems to have failed, but I'm pretty sure I don't touch it?

styurin:
Why skipped tests need to marked as failed?

jiangty-addepar:
Test setup failures cause the test to be skipped. It's possible that in combination with https://github.com/facebook/buck/pull/1665 , we might be able to mark skipped tests as skipped instead ("assumption violation", perhaps? It seems there isn't a ResultType.SKIPPED)

jiangty-addepar:
I can change this behavior in #1665 if desired, but before that goes in, it's better to be stricter with potentially failing tests.

jiangty-addepar:
Any other comments on this PR? Did `SuperConsoleBusListenerTest` get fixed / un-flaked?

jiangty-addepar:
Hello?

mkillianey:
It seems to me that [SkipException#isSkip](https://github.com/cbeust/testng/blob/17185dd3b6fc93df2efdec91582c0ca7b8d0fc8f/src/main/java/org/testng/SkipException.java#L26-L37) suggests that a skipped test isn't a failure unless you explicitly override the class to say that it is.

jiangty-addepar:
Ok, I can do something like:

- if a test was skipped because setup failed, mark it as `ResultType.ASSUMPTION_VIOLATION`
- if a test is skipped otherwise, mark it as `ResultType.DISABLED`

However, this change will require combining this PR with #1665 (to properly handle test setup failures), and there are a bunch of intermediate changes between them. I could go back to just combining all these PRs (#1662 , #1663 , #1664 , #1665), or I could merge this PR as a temporary state before going to the "desired state" in #1665 .

jiangty-addepar:
bump again.

Another thing I could do is open this as a PR into #1663, #1663 into #1664, #1664 into #1665 and then merge #1665 as one big PR.

styurin:
Please, split changes into small independent modifications so that it's easier to understand what you are trying to do. This would speed up the review process.

jiangty-addepar:
@styurin isn't that what these PRs are?

jiangty-addepar:
I can try rewriting them again to concentrate the refactors in one place, I guess?

src/com/facebook/buck/testrunner/TestNGRunner.java
```diff
@@ -92,7 +92,6 @@ public void run() throws Throwable {
   private XmlSuite createXmlSuite(Class<?> c) {
     XmlSuite xmlSuite = new XmlSuite();
     xmlSuite.setName("TmpSuite");
-    xmlSuite.setTimeOut(String.valueOf(defaultTestTimeoutMillis));
```
styurin: why this change is part of this PR? Is that necessary?

jiangty-addepar: It's a separate change, and I can make a separate 1-line PR instead? We originally had these together in 1 commit, which is why they're together.

jiangty-addepar: Or, could I combine this timeout removal with some of the "structural refactors" from #1663 ?

styurin: Yes, please, move it to a separate change. In case it causes any problems we can revert this change only without touching logic around skipped tests

jiangty-addepar:
@styurin Ok, I removed the change to test timeouts in this PR.

PR #1663, opened 2017-12-04
[TestNG] Correctly handle test selectors

jiangty-addepar:
On top of #1662 and 2nd part of #1657 . Ignore commits before `Respect ClassName and MethodName filters in TestNG runner.`. (Didn't want to make a PR into that branch because the semantics become confusing, and no one will see a PR from our fork into our fork anyway.)

We now correctly handle test selectors in TestNG. 
Also, we do some minor refactoring, removing `TestNGWrapper` and the "TmpSuite" xml suite.

Test plan: test in TestNGIntegrationTest.


jiangty-addepar:
Going to replace this with some smaller PRs.

PR #1664, opened 2017-12-04
[TestNG] Log method parameters

jiangty-addepar:
On top of #1663 and 3rd part of #1657 . Ignore commits before `TestNG results now show method parameters`.

TestNG has features like `@DataProvider` which involve passing in parameters to tests. This changes Buck to log those test parameters.

Test Plan: `TestNGLoggingTest`


jiangty-addepar:
Replaced by #1722

PR #1665, opened 2017-12-04
[TestNG] Mark tests as failed if test setup fails; log setup failures

jiangty-addepar:
On top of #1664 and 4th part of #1657 . Ignore commits before `Support Stdout and Exception Details when Test Setup Fails`.

- If test setup fails in a TestNG test:
    - Mark test as failed
    - Print stdout and exception stack traces from setup methods
    - Record all test failures in JUnit test report, including those from test setup

Todo: figure out how to not also mark the "setup" method as a failing "test"? It's not a big deal because the goal of "tell the developer when test setup is broken" is still being achieved.

Test Plan: `SetupTestNGTest`


styurin:
Is that still relevant?

jiangty-addepar:
Might be able to close this for now. I haven't had time to try to fix the TODO.

PR #1666, opened 2017-12-05
Update D8 to latest to handle a bug in 5.1 devices

kageiit:
Picks up fix for https://issuetracker.google.com/issues/69364976


kageiit:
@dreiss there seems to be an issue with this. Integration test should have failed. Fixing it. Please dont import just yet

kageiit:
Ok should be good now

kageiit:
@dreiss hoping to get another fix in - https://issuetracker.google.com/issues/70348575

if we can wait a bit

kageiit:
Seems like this merged. Will open another PR for the fix from  https://issuetracker.google.com/issues/70348575

PR #1671, opened 2017-12-07
Don't version gate watch stub copying

### LINKED ISSUE #1670: Watch apps for SDK version > 2.0 do not contain a second stub app
For mysterious reasons, watch apps have 2 copies of the watch stub app inside the bundle. The existing code gates copying this in based on the version number, explicitly checking for OS2.0. The version check is redundant, we should always copy in the stub app regardless of version.

rmaz:
Fixes #1670. Always copy the watch app stub twice, this is still required for OS 3.0 and up.


rmaz:
@ryu2 might be another one for you, looks good?

ryu2:
@facebook-github-bot shipit

Issue #1548, opened 2017-09-28
Buck does not compile with Java 11


shs96c:
Set your JAVA_HOME to point to an official Java 9 JDK. Run `ant clean default`. Watch the catastrophic failure of the build.

Fortunately, once you have a compiled version of buck, it _runs_ under Java 9, which is nice.

jkeljo:
> it _runs_ under Java 9

Well, sort of. Some codepaths in the Java compilation support don't like it much. :-)

ttsugriy:
known problem :( Internally we haven't embraced Java 9 yet, but it's on our radar to fix Java 9 builds :)

epkugelmass:
I've submitted a patch that mostly gets Buck built under JDK9. There are some remaining issues because of the Skylark dependency, which uses an old version of AutoValue that is not JDK9 friendly.

https://github.com/facebook/buck/pull/1672

ttsugriy:
Thank you, @epkugelmass! This is pretty awesome. I'm going to work on resolving the old AutoValue issue.

ttsugriy:
Looks like even the latest version of AutoValue still has issues, so I'm going to have to wait until https://github.com/google/auto/pull/563 is merged, new version of AutoValue is released and it's used in Bazel :(

epkugelmass:
Awesome! You'll probably also want to bump ASM for the RELEASE_9 Opcodes. Or I can do it. Was feeling a bit lazy.

epkugelmass:
Nevermind -- buck is already using 6.0. I'll update the patch to take advantage of that.

davido:
@ttsugriy I released a private version of auto-value with this PR included https://github.com/google/auto/pull/563 here: [1]. I needed it to build gerrit code review with bazel, with Java 9: [2].

* [1] https://github.com/davido/auto/releases/tag/auto-value-1.6
* [2] https://gerrit-review.googlesource.com/#/c/gerrit/+/147233/

TarasMazepa:
Actually google/auto#563 is merged so @ttsugriy it is good time to revisit this issue ;)

ttsugriy:
Java 9 is no longer supported by Oracle, so we are unlikely to be supporting it. We may repurposes this for Java 10 support though.

ttsugriy:
@davido, looks like Bazel is still using autovalue 1.4 https://source.bazel.build/bazel/+/master:third_party/auto/auto-value-1.4.jar :(

davido:
[...]

PR #1672, opened 2017-12-11
Ant build completes under 9.0.1

epkugelmass:
This is a demonstration patch and is not meant to be merged.

Ant build completes under jdk 9.0.1
but with AutoValue annotation processor failures
the resulting buck binaries are unable to build buck

for autovalue annotation processor failures -- see https://github.com/google/auto/issues/560 and related tickets

Question for the buck maintainers -- clearly this patch is not backward compatible with java 8. what's the policy for this project?

referencing issue https://github.com/facebook/buck/issues/1548


styurin:
@epkugelmass have you tried to build Java 9 code with buck using out of the process compilation?

bobyangyf:
Unfortunately buck doesn't work with Java 9.
There's a number of issues in terms of compatibility that even if we were to get it to build, buck would not run properly.

It is on the backlog to modernize it newer standards of java.

bobyangyf:
Unfortunately buck doesn't work with Java 9.
There's a number of issues in terms of compatibility that even if we were to get it to build, buck would not run properly.

It is on the backlog to modernize it newer standards of java.

#1548 tracks this

PR #1682, opened 2017-12-15
Fix chrome_trace test target name

robbertvanginkel:
In `src/com/facebook/buck/event/chrome_trace/BUCK` this is referred to as:
```
java_library(
...
    tests = [
        "//test/com/facebook/buck/event/chrome_trace:chrome_trace",
    ],
)
```


robbertvanginkel:
Suspicously, this target also doesn't have any source files. Is this intended? Maybe this was missed in the move in 8e20872a792a262c2f5c5287df74a9d5331fee16?

robbertvanginkel:
This fails before `buck query 'deps(testsof(//src/com/facebook/buck/event/chrome_trace/...))'`

PR #1691, opened 2017-12-21
Whitelist profile entitlements

### LINKED ISSUE #1690: Buck is merging invalid iCloud entitlements from profiles
87eea587a2227a5fd55b924f49d4df2a5f37619b removed some entitlements keys from the `getMergeableEntitlements` method. This now merges the following keys when signing builds:

```
            "com.apple.developer.icloud-container-environment",
            "com.apple.developer.icloud-container-identifiers",
            "com.apple.developer.icloud-services",
            "com.apple.developer.restricted-resource-mode",
            "com.apple.developer.ubiquity-container-identifiers",
            "com.apple.developer.ubiquity-kvstore-identifier",
```

These keys are not present in an equivalent build made via Xcode, they are only added if explicitly present in the apps entitlements. The profile values for these keys contain wildcards, which will cause app store rejection:

> Error Domain=ITunesConnectionOperationErrorDomain Code=1091 \"Invalid Code Signing Entitlements. The signature for your app bundle contains entitlement values that are not supported. For the com.apple.developer.ubiquity-kvstore-identifier entitlement, the value must start with the prefix provided by Apple in the provisioning profile, followed by characters that are uppercase or lowercase Roman letters [A-Z, a-z], the digits 0 through 9, dot ['.'], or hyphen ['-'], and not contain any wildcard characters.

These keys should be added back to the method, they should not be merged in from the provisioning profile. This list was determined from examining the Xcode entitlements merging code, we should be mirroring the Xcode behaviour here.

rmaz:
Fixes #1690. Don't use a blacklist for provisioning profile entitlements to merge, use a whitelist instead. Taken from `+[IDEEntitlementsMerger keysToAlwaysCopyFromProfile]` and ` +[IDEEntitlementsMerger genericallyDefinedAppKeysToCopyFromProfile]`. Prevents any profile entitlements with wildcards from being merged in to a signed app.

### DIFF
--- a/src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
+++ b/src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
@@ -18,6 +18,7 @@ package com.facebook.buck.apple;
 
 import com.dd.plist.NSDictionary;
 import com.dd.plist.NSObject;
+import com.dd.plist.NSString;
 import com.dd.plist.PropertyListFormatException;
 import com.dd.plist.PropertyListParser;
 import com.facebook.buck.apple.toolchain.ApplePlatform;
@@ -192,20 +193,23 @@ class ProvisioningProfileCopyStep implements Step {
     // Copy the actual .mobileprovision.
     filesystem.copy(provisioningProfileSource, provisioningProfileDestination, CopySourceMode.FILE);
 
+    // Create the application identifier from the app ID and bundle ID
+    String appID = bestProfile.get().getAppID().getFirst() + "." + bundleID;
+
     // Merge the entitlements with the profile, and write out.
     if (entitlementsPlist.isPresent()) {
+
       return (new PlistProcessStep(
               filesystem,
               entitlementsPlist.get(),
               Optional.empty(),
               signingEntitlementsTempPath,
               bestProfile.get().getMergeableEntitlements(),
-              ImmutableMap.of(),
+              ImmutableMap.of(APPLICATION_IDENTIFIER, new NSString(appID)),
               PlistProcessStep.OutputFormat.XML))
           .execute(context);
     } else {
       // No entitlements.plist explicitly specified; write out the minimal entitlements needed.
-      String appID = bestProfile.get().getAppID().getFirst() + "." + bundleID;
       NSDictionary entitlementsPlist = new NSDictionary();
       entitlementsPlist.putAll(bestProfile.get().getMergeableEntitlements());
       entitlementsPlist.put(APPLICATION_IDENTIFIER, appID);
--- a/src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
+++ b/src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
@@ -18,7 +18,6 @@ package com.facebook.buck.apple;
 
 import com.dd.plist.NSDictionary;
 import com.dd.plist.NSObject;
-import com.dd.plist.NSString;
 import com.dd.plist.PropertyListFormatException;
 import com.dd.plist.PropertyListParser;
 import com.facebook.buck.apple.toolchain.ApplePlatform;
@@ -193,8 +192,6 @@ class ProvisioningProfileCopyStep implements Step {
     // Copy the actual .mobileprovision.
     filesystem.copy(provisioningProfileSource, provisioningProfileDestination, CopySourceMode.FILE);
 
-    String appID = bestProfile.get().getAppID().getFirst() + "." + bundleID;
-
     // Merge the entitlements with the profile, and write out.
     if (entitlementsPlist.isPresent()) {
       return (new PlistProcessStep(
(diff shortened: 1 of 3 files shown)


rmaz:
@milend @ryu2 this was a required change to get our app to pass app store validation. We have numerous wildcard entitlements in our production profile that we do not use any more, and merging them fails app store validation. I took a look at the current Xcode entitlements merging code and from what I could see it has changed from a blacklist to a whitelist with only the 4 keys added in this PR merged in. This produced identical signing entitlements between Xcode and buck for our app.

kapfab:
That seems indeed a much better way to merge entitlements than the current incomplete list of excluded entitlements.

Though, is the specific `com.apple.developer.icloud-container-environment` entitlement correctly addressed? It is dynamically added by Xcode during export and is not meant to be part of App Entitlements (or potentially populated with a `Development` value).

rmaz:
> Though, is the specific com.apple.developer.icloud-container-environment entitlement correctly addressed? It is dynamically added by Xcode during export and is not meant to be part of App Entitlements (or potentially populated with a Development value).

We have it set to `Production` in our app store entitlements file and have been signing fine with Xcode with no warnings. 

Unless this key is set, app store submission would fail in either case. In the current behaviour the following would be merged in to the xcent file:

```
<key>com.apple.developer.icloud-container-environment</key>
<array>
<string>Development</string>
<string>Production</string>
</array>
```

This is not valid and would fail submission. In the new behaviour the key would simply not be present, which is also invalid.

kapfab:
OK, so you expect it to be forced in the app entitlements.

The approach we took in a recent PR made for the fastlane project is to add this entitlement dynamically, depending on the type of Provisioning Profile being used for signing (`Production` for Distribution PP, `Development` otherwise). This prevents an app using `Development` to be pushed to Apple, but the downside is that it also prevents developers to release test apps using `Production` data… We'll probably need to add an optional parameter to allow that.

There’s no perfect way to do this, but I’m pleased to discuss the different options with someone else!

rmaz:
You are right that it is basically impossible to mimic what Xcode does as it is a moving target with no command line tool we can use to merge entitlements (feel free to file a radar around this to dupe mine). I think the most sensible path here is the most conservative, do the absolute minimum of magic and allow the users to set the entitlements as much as possible. Entitlements and code signing is already different enough in buck that it doesn't seem a big deal to add keys you wouldn't normally have to to get an app to pass validation, definitely preferred to the case where your entitlements are forcibly merged incorrectly making it impossible to submit an app to the store.

rmaz:
@milend any chance you could review this one? thanks

milend:
@rmaz I'm extremely busy at the moment but I'll try to set aside some time to look at this. Apologies in advance for the delay.

src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
```diff
@@ -192,20 +193,23 @@ public StepExecutionResult execute(ExecutionContext context)
...
     if (entitlementsPlist.isPresent()) {
+
       return (new PlistProcessStep(
               filesystem,
               entitlementsPlist.get(),
               Optional.empty(),
               signingEntitlementsTempPath,
               bestProfile.get().getMergeableEntitlements(),
-              ImmutableMap.of(),
+              ImmutableMap.of(APPLICATION_IDENTIFIER, new NSString(appID)),
```
milend: @rmaz Why do we now need to override the value for `APPLICATION_IDENTIFIER`? How did this work beforehand? Was it part of the value returned from `getMergeableEntitlements()`?

I think we should preserve the pre-existing behaviour and specify `APPLICATION_IDENTIFIER` as part of the `additionalKeys` parameter to `PlistProcessStep`, rather than as override. My concern here is that it might override a value already set in `entitlementsPlist`. On the other hand, the argument could be made that if those values did differ, the app will fail to launch due to a code signing issue. Still, it might be safer to err on the side of caution and preserve existing behaviour.

src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
```diff
@@ -192,20 +193,23 @@ public StepExecutionResult execute(ExecutionContext context)
     // Copy the actual .mobileprovision.
     filesystem.copy(provisioningProfileSource, provisioningProfileDestination, CopySourceMode.FILE);
 
+    // Create the application identifier from the app ID and bundle ID
+    String appID = bestProfile.get().getAppID().getFirst() + "." + bundleID;
```
milend: @rmaz To improve readability, I would actually write this out as:

```java
String appID = bestProfile.get().getAppID().getFirst() + "." + bundleID;
ImmutableMap<String, NSObject> overrideKeys = ImmutableMap.of(APPLICATION_IDENTIFIER, new NSString(appID));
```

Because the purpose of `ImmutableMap.of(APPLICATION_IDENTIFIER, new NSString(appID))` is not clear in the constructor of `PlistProcessStep`.

I'd also move this inside the `if` statement (`entitlementsPlist.isPresent()`) as it's only used inside that scope.

src/com/facebook/buck/apple/ProvisioningProfileCopyStep.java
rmaz: I'll wind this one back, I am inclined to agree. The current behaviour is to merge with the application-identifier in the entitlements, yes.

milend:
@rmaz Did you test your PR with your latest changes? Does it produce the same output as Xcode?

rmaz:
@milend thanks for pushing this through.

> Did you test your PR with your latest changes? Does it produce the same output as Xcode?

I just checked with the latest change, yes, we have the same entitlements for our app with embedded watch app and extensions.

PR #1694, opened 2017-12-22
Document provided_deps for android_library.

tinaroh:
Summary:
`provided_deps` is a valid attribute for android_library. Copied the format of `exported_deps`.

Test Plan:
- Run `docs/soyweb-local.sh`
- Open `http://localhost:9811/rule/android_library.html` and observe change

### DIFF
--- a/docs/rule/android_library.soy
+++ b/docs/rule/android_library.soy
@@ -148,22 +148,12 @@ compiled class files and resources.
   {/param}
 {/call}
 
-{call buck.arg}
-  {param name: 'exported_deps' /}
-  {param default : '[]' /}
-  {param desc}
-  This is the same as in <a href="{ROOT}rule/java_library.html">
-  <code>java_library</code></a>.
-  {/param}
+{call jvm_common.exported_deps}
+  {param library: 'android_library' /}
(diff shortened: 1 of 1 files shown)


docs/rule/android_library.soy
```diff
@@ -157,6 +157,15 @@ compiled class files and resources.
...
 {/call}
 
+{call buck.arg}
+  {param name: 'provided_deps' /}
+  {param default : '[]' /}
+  {param desc}
+  This is the same as in <a href="{ROOT}rule/java_library.html">
+  <code>java_library</code></a>.
+  {/param}
+{/call}
```
sdwilsh: It would be best here to do a `{call jvm_common.provided_deps}` here, like we do in `java_library.soy`.  I suspect, but haven't verified, that we want this to apply for `android_binary` and not `java_binary`, which is what `java_library` says.

tinaroh: Yes, thanks for the catch! This should be `android_binary` as an `android_library` cannot be packaged into a `java_library`. Updated, along with a similar change for `exported_deps`.

PR #1698, opened 2017-12-27
Xcode project add app_icon and launch_image from asset catalog

robbertvanginkel:
`apple_asset_catalogs` have settings for an app_icon and launch_image that were not used during project generation. This PR adds project generation support.

### DIFF
--- a/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
+++ b/src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
@@ -1208,6 +1208,11 @@ public class ProjectGenerator {
 
     extraSettingsBuilder.putAll(swiftDepsSettingsBuilder.build());
 
+    setAppIconSettings(
+        recursiveAssetCatalogs, directAssetCatalogs, buildTarget, defaultSettingsBuilder);
+    setLaunchImageSettings(
+        recursiveAssetCatalogs, directAssetCatalogs, buildTarget, defaultSettingsBuilder);
+
     ImmutableSortedMap<Path, SourcePath> publicCxxHeaders = getPublicCxxHeaders(targetNode);
     if (isModularAppleLibrary(targetNode) && isFrameworkProductType(productType)) {
       // Modular frameworks should not include Buck-generated hmaps as they break the VFS overlay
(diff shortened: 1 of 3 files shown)


src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1578,6 +1583,46 @@ private PBXNativeTarget generateBinaryTarget(
...
+      BuildTarget buildTarget,
+      ImmutableMap.Builder<String, String> defaultSettingsBuilder) {
+    ImmutableList<String> launchImage =
+        Stream.concat(directAssetCatalogs.stream(), recursiveAssetCatalogs.stream())
+            .map(x -> x.getLaunchImage())
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .collect(ImmutableList.toImmutableList());
+    if (launchImage.size() > 1) {
+      throw new HumanReadableException(
```
milend: Same comment as above.

src/com/facebook/buck/apple/project_generator/ProjectGenerator.java
```diff
@@ -1578,6 +1583,46 @@ private PBXNativeTarget generateBinaryTarget(
...
+      BuildTarget buildTarget,
+      ImmutableMap.Builder<String, String> defaultSettingsBuilder) {
+    ImmutableList<String> appIcon =
+        Stream.concat(directAssetCatalogs.stream(), recursiveAssetCatalogs.stream())
+            .map(x -> x.getAppIcon())
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .collect(ImmutableList.toImmutableList());
+    if (appIcon.size() > 1) {
+      throw new HumanReadableException(
```
milend: My only concern here is that we're adding a new constraint which has the potential to break project generation for existing codebases. Changing the behaviour of Buck in this way can be painful when deploying new builds at Facebook (which would lead to reverts and then having to do the whole commit dance again). Instead, I'd suggest one of the following (or a combination):

- Emit a warning rather than throw an exception.
- If we want to keep the ability to throw an exception, we can add a config option for it.

I'm wondering, what happens when Buck itself builds an app that has multiple catalogs with `app_icon` defined?

robbertvanginkel:
I understand the concert for the hard error. I copied this behaviour from buck build, see 

https://github.com/facebook/buck/blob/08f7f10f21be7759cc582dbca2b1827c514bb986/src/com/facebook/buck/apple/AppleDescriptions.java#L344-L370

Even though buck build already fails on this, it is possible that it'd cause issues in project generation. Happy to add something but considering the buck build I would try to argue we should keep them the same.

milend:
@robbertvanginkel:
> I copied this behaviour from buck build

I think it's fine in that case because we're preserving behaviour existing `buck build` behaviour.

PR #1700, opened 2017-12-28
Modular apple_library set -fmodule-name flag in populateCxxLibraryDescriptionArg

robbertvanginkel:
When building mixed swift/c modules, the c part of the code needs to be
compiled with the flag `-fmodule-name=` to correctly identify some
headers are part of the module it is building.

### DIFF
--- a/src/com/facebook/buck/apple/AppleDescriptions.java
+++ b/src/com/facebook/buck/apple/AppleDescriptions.java
@@ -59,6 +59,7 @@ import com.facebook.buck.rules.TargetNode;
 import com.facebook.buck.rules.Tool;
 import com.facebook.buck.rules.coercer.PatternMatchedCollection;
 import com.facebook.buck.rules.coercer.SourceList;
+import com.facebook.buck.rules.macros.StringWithMacros;
 import com.facebook.buck.shell.AbstractGenruleDescription;
 import com.facebook.buck.util.HumanReadableException;
 import com.facebook.buck.util.Optionals;
@@ -311,6 +312,14 @@ public class AppleDescriptions {
         SourceList.ofNamedSources(
             convertAppleHeadersToPublicCxxHeaders(
                 buildTarget, resolver::getRelativePath, headerPathPrefix, arg)));
+    if (arg.isModular()) {
+      output.addCompilerFlags(
+          StringWithMacros.of(
+              ImmutableList.of(
+                  Either.ofLeft(
+                      "-fmodule-name="
+                          + arg.getHeaderPathPrefix().orElse(buildTarget.getShortName())))));
+    }
   }
 
   public static Optional<AppleAssetCatalog> createBuildRuleForTransitiveAssetCatalogDependencies(
--- a/src/com/facebook/buck/apple/AppleLibraryDescription.java
+++ b/src/com/facebook/buck/apple/AppleLibraryDescription.java
@@ -819,11 +819,6 @@ public class AppleLibraryDescription
     Optional<SourcePath> getInfoPlist();
 
     ImmutableMap<String, String> getInfoPlistSubstitutions();
-
-    @Value.Default
-    default boolean isModular() {
-      return false;
-    }
   }
 
   // CxxLibraryDescriptionDelegate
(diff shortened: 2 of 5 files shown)


robbertvanginkel:
@milend do you have an idea who could help review this? I'm having a bit of trouble figuring out where this can go best. I'll put some comments in the code.

src/com/facebook/buck/cxx/CxxLibraryFactory.java
```diff
@@ -413,17 +413,23 @@ public BuildRule createBuildRule(
...
+                args.getCompilerFlags(),
+                args.getPlatformCompilerFlags(),
+                args.getLangCompilerFlags(),
+                cxxPlatform),
+            f ->
+                CxxDescriptionEnhancer.toStringWithMacrosArgs(
+                    buildTarget, cellRoots, ruleResolver, cxxPlatform, f)));
+    if (args.isModular()) {
+      allCompilerFlagsBuilder.putAll(
+          CxxFlags.toLanguageFlags(StringArg.from("-fmodule-name=" + args.getModuleName().get())));
```
robbertvanginkel: I'm not sure this is the correct place to put this. It seems like this might be better in the `CxxSourceRuleFactory`, but I couldn't find a simple place to inject it. This seemed the easiest place but I'm having a bit of trouble writing a reasonable test for it now, which would suggest that this isn't the best place.

The current suggestion I have now is passing along the modulename to `CxxSourceRuleFactory` if modular is set and it is present, to then add it to the flags through `private Iterable<Arg> getRuleCompileFlags(CxxSource.Type type)` there.

milend: @robbertvanginkel:
> I'm not sure this is the correct place to put this. It seems like this might be better in the CxxSourceRuleFactory, but I couldn't find a simple place to inject it.

There's at least one other place that might need this logic: `CxxDescriptionEnhancer::createBuildRulesForCxxBinary` which contains the following code (which is almost 1:1 duplicated):

```java
ImmutableListMultimap.copyOf(
  Multimaps.transformValues(
      CxxFlags.getLanguageFlagsWithMacros(
          preprocessorFlags,
          platformPreprocessorFlags,
          langPreprocessorFlags,
          cxxPlatform),
      f -> toStringWithMacrosArgs(target, cellRoots, resolver, cxxPlatform, f))),
```

Adding it as a `@Value.Parameter` to `AbstractCxxSourceRuleFactory` would make sure that we definitely add the module name parameter to all source rules rather than relying on hunting down all call sites.

Having said that, I'm not sure whether it's useful to pass `-fmodule-name` for rules like `apple_binary`. Do we know that we want to add the compiler flag everywhere? Another aspect that comes to mind is whether we want to have this flag added for (Objective-)C++ files where modules are not yet finalised.

robbertvanginkel:
I'll try to look at this a bit more in depth considering the comments. One think I can think of is we don't change the cxx classes but put this in `AppleDescription`'s `populateCxxConstructorArg`.

robbertvanginkel:
@milend I've revised this PR and found a better place to put this. By putting it in `populateCxxLibraryDescriptionArg` it doesn't have to go into the cxx part of buck and it keeps restricted to the apple part.

I can add an integration test later when this functionality is needed to not break building, but that requires some more PR's around modulemap generation with buck build to land.

src/com/facebook/buck/apple/AppleDescriptions.java
```diff
@@ -311,6 +312,14 @@ public static void populateCxxLibraryDescriptionArg(
         SourceList.ofNamedSources(
             convertAppleHeadersToPublicCxxHeaders(
                 buildTarget, resolver::getRelativePath, headerPathPrefix, arg)));
+    if (arg.isModular()) {
+      output.addCompilerFlags(
+          StringWithMacros.of(
+              ImmutableList.of(
+                  Either.ofLeft(
+                      "-fmodule-name="
+                          + arg.getHeaderPathPrefix().orElse(buildTarget.getShortName())))));
```
milend: @robbertvanginkel Doesn't `AppleNativeTargetDescriptionArg` have `moduleName`? I feel like this logic of determining an effective module name is duplicated in quite a few places.

PR #1706, opened 2018-01-10
Remove hard time limit on TestNG test suite

jiangty-addepar:
Having a time limit on the test suite is confusing; instead, stick to specifying time limits on tests directly, or putting time limits on test targets.

### DIFF
--- a/src/com/facebook/buck/testrunner/TestNGRunner.java
+++ b/src/com/facebook/buck/testrunner/TestNGRunner.java
@@ -92,7 +92,6 @@ public final class TestNGRunner extends BaseRunner {
   private XmlSuite createXmlSuite(Class<?> c) {
     XmlSuite xmlSuite = new XmlSuite();
     xmlSuite.setName("TmpSuite");
-    xmlSuite.setTimeOut(String.valueOf(defaultTestTimeoutMillis));
     XmlTest xmlTest = new XmlTest(xmlSuite);
     xmlTest.setName("TmpTest");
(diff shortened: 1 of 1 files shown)


jiangty-addepar:
```Caused by: java.io.IOException: No space left on device``` errors on TravisCI again

File: src/com/facebook/buck/testrunner/TestNGRunner.java [hunk 1]
```diff
@@ -92,7 +92,6 @@ public void run() throws Throwable {
   private XmlSuite createXmlSuite(Class<?> c) {
     XmlSuite xmlSuite = new XmlSuite();
     xmlSuite.setName("TmpSuite");
-    xmlSuite.setTimeOut(String.valueOf(defaultTestTimeoutMillis));
```
ttsugriy: while I agree using timeout on a suite is not idea, but we should always have timeouts for tests and as such please either add timeouts for test runs or better yet consider using DelegateRunnerWithTimeout

jiangty-addepar: I thought there is always a time limit per test target specified in `.buckconfig`?

jiangty-addepar: basically, I'm afraid of introducing "non-transparent" timeouts inside Buck; users should always know that their timeout is specified explicitly somewhere, imo

ttsugriy: hm, well, I'm fine with any timeouts as long as there are some :) If you're saying that test execution time is already limited - I'm fine with this then :)

PR #1712, opened 2018-01-17
Add support for downloading prebuilt binaries of buck via jitpack at any sha

kageiit:
- Uses Jitpack to build buck pex and serve them on demand. The very first time a version of buck is requested, it is built via [jitpack](https://jitpack.io/). Every subsequent request will just serve the built artifact directly.
- This enables anyone to just fetch `https://jitpack.io/com/github/facebook/buck/<sha>/buck-<sha>.pex` to download a prebuilt buck pex at any sha.
- Once this merges, this functionality will be available for any forks of buck as well, so one can fetch for example `https://jitpack.io/com/github/kageiit/buck/f0655d65907df1262b191339a51aa94a587f9f3c/buck-f0655d65907df1262b191339a51aa94a587f9f3c.pex`

screen shot 2018-01-18 at 12 36 57 am

- This will allow users of okbuck to not have to rely on checking out the entire buck repo and build it using ant before being able to use buck

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,9 +1,6 @@
-jdk:
-  - oraclejdk8
 install:
    - set -exo pipefail
    - ant
-   - bin/buck build buck
+   - PEX=$(bin/buck build buck --show-output | awk '{print $2}')
    - SHA=$(git rev-parse HEAD)
-   - PEX=$(bin/buck targets --show-output buck | awk '{print $2}')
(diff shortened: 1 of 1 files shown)


jitpack.yml
```diff
@@ -0,0 +1,9 @@
+jdk:
+  - oraclejdk8
+install:
+   - set -exo pipefail
+   - ant
+   - bin/buck build buck
```
ttsugriy: nit: you can avoid having to use `buck targets --show-output` and use `buck build --show-output` instead.

kageiit:
@ttsugriy done!

PR #1720, opened 2018-01-22
Fix test_host_app usage with swift

robbertvanginkel:
Tests with swift in both the test host and test target currently fail compilation with a cryptic error like 
```
ld: library not found for -lswiftMetal for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
``` 
This is because the `SwiftRuntimeNativeLinkable` is both dependency of the test_host_app and the tests. When linking the tests, the dependencies it shares with its test_host_app (`bundle_loader` in ld) are added to a blacklist so they are not linked into the tests to prevent duplicate symbols. 

However, if there's usage of swift in the tests their object files will contain linker directives to link against some wrappers around the standard library which require the flags added by `SwiftRuntimeNativeLinkable` to link correctly.

### DIFF
--- a/test/com/facebook/buck/swift/testdata/swift_test_with_host/.buckconfig
+++ b/test/com/facebook/buck/swift/testdata/swift_test_with_host/.buckconfig
@@ -1,0 +1,5 @@
+[cxx]
+default_platform = iphonesimulator-x86_64
+
+[swift]
+compiler_flags = -enable-testing
--- a/test/com/facebook/buck/swift/testdata/swift_test_with_host/BUCK.fixture
+++ b/test/com/facebook/buck/swift/testdata/swift_test_with_host/BUCK.fixture
@@ -1,22 +1,14 @@
-if read_config('testconfig', 'dep_type') == 'swift_library':
-  swift_library(
-      name = "dep",
-      srcs = [
-          "dep.swift",
-      ],
-  )
-elif read_config('testconfig', 'dep_type') == 'apple_library':
-  apple_library(
-      name = "dep",
-      srcs = [
-          "dep.swift",
-      ],
-  )
-
 if 'macosx' in read_config('cxx', 'default_platform'):
   default_frameworks = ["$SDKROOT/System/Library/Frameworks/AppKit.framework"]
 else:
   default_frameworks = ["$SDKROOT/System/Library/Frameworks/UIKit.framework"]
+
+apple_library(
+    name = "dep",
+    srcs = [
+        "dep.swift",
+    ],
+)
 
 apple_binary(
     name = "binary",
--- a/test/com/facebook/buck/swift/testdata/swift_test_with_host/BUCK.fixture
+++ b/test/com/facebook/buck/swift/testdata/swift_test_with_host/BUCK.fixture
@@ -13,6 +13,11 @@ elif read_config('testconfig', 'dep_type') == 'apple_library':
       ],
   )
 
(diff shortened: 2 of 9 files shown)


test/com/facebook/buck/swift/testdata/swift_test_with_host/BUCK.fixture
```diff
@@ -0,0 +1,49 @@
+swift_library(
```
milend: I think we should be testing `apple_library` in the very least, having a separate test for `swift_library` might be beneficial, too.

test/com/facebook/buck/swift/testdata/swift_test_with_host/.buckconfig
```diff
@@ -0,0 +1,5 @@
+[cxx]
+default_platform = iphonesimulator-x86_64
```
milend: We definitely want a separate test on macOS where we should force static linking of the Swift runtime (I think we can do that by setting `link_style = "static",` on the `apple_library` that contains Swift).

robbertvanginkel: When looking at this, it seems like buck currently doesn't support osx tests with a `bundle_loader`/`test_host_app`. 

When I make a test for this, everything builds but when running it runs something like `fbxctest -sdk macosx run-tests -appTest /path/to/test.xctest:/path/to/app.app`, which fails fbxctest with `UITests are not supported on OS X`. I tried updating `fbxctest` in buck using the script at `scripts/update-xctest.sh` to take in facebook/FBSimulatorControl@67bb15bd0d5eb58730b045fc6892902c2671dac0, but afterwards it tries to run as a uitest.

This leaves the question (I'm less familiar with tests on osx), does osx support the application tests in which you compile a test bundle with a `bundle_loader`? If not than this blacklisting code does not apply, and we should probably throw an error when trying to build a test a bundle with `test_host_app` for a macosx target.

test/com/facebook/buck/swift/testdata/swift_test_with_host/BUCK.fixture
robbertvanginkel: Agreed, I'll change it to test both.

robbertvanginkel:
I rebased and added a testcase for the apple_library, let me know what you think of the osx part.

test/com/facebook/buck/swift/testdata/swift_test_with_host/.buckconfig
milend: @robbertvanginkel I remember helping @marekcirkos with his efforts to add UI tests for macOS last year but my memory is a bit fuzzy. I know that we currently support logic tests without `test_host_app` and UI tests where both `test_host_app` and `ui_test_target_app` are specified (only on macOS 10.13+).

As long as we don't break logic tests with Swift on macOS, we should be good to go.

robbertvanginkel: Looked at this again, now it has a test on macos that validates the swift stdlib was linked statically.

robbertvanginkel:
Created a separate PR for the fbxctest update in #1769. Think we should be good with the extra tests now.

milend:
@robbertvanginkel LGTM, I'm landing #1769 first. It seems this PR still has the code to update `fbxctest`, so we should probs revert that particular commit?

robbertvanginkel:
Removed the fbxctest commit.

PR #1722, opened 2018-01-23
[TestNG] Log method parameters in tests

jiangty-addepar:
Replacing #1664 now that the other PRs have been merged.

TestNG has features like `@DataProvider` which involve passing in parameters to tests. This changes Buck to log those test parameters.

Test Plan: `TestNGLoggingTest`


src/com/facebook/buck/testrunner/TestNGRunner.java
```diff
@@ -55,9 +63,18 @@ public void run() throws Throwable {
       } else {
         results = new ArrayList<>();
         TestNG testng = new TestNG();
+        testng.setUseDefaultListeners(false);
         testng.setAnnotationTransformer(new FilteringAnnotationTransformer(results));
-        testng.setTestClasses(new Class<?>[] {testClass});
```
ttsugriy: looks like you're breaking formatting here... Please use google-format to fix all formatting issues

jiangty-addepar: I installed the IntelliJ plugin, but I guess it doesn't auto-format :(

jiangty-addepar:
@ttsugriy sorry about that!

Is there an easy way to tell IntelliJ to run formatting on all changed files or something?

ttsugriy:
sorry, I don't know @jiangty-addepar, we have `google-format` tool integrated with arcanist tool so it happens as part of `arc lint` workflow :(

ttsugriy:
@jiangty-addepar, looks like your changes break the build - https://ci.appveyor.com/project/Facebook/buck/build/5912#L2252

jiangty-addepar:
I'm really confused -- I'm running this test locally and it not only compiles, but passes?

jiangty-addepar:
hmm, perhaps it's a semantic merge conflict; I'll rebase on master and try again

src/com/facebook/buck/testrunner/TestNGRunner.java
```diff
@@ -114,6 +131,30 @@ private boolean shouldIncludeTest(String className) {
...
+      return name;
+    }
+
+    StringBuilder builder = new StringBuilder(name).append(" (");
+    builder.append(
+        Arrays.stream(parameters)
+            .map(
+                parameter -> {
+                  try {
+                    return parameter == null ? "null" : parameter.toString();
```
styurin:
```
return String.valueOf(parameter);
```

src/com/facebook/buck/testrunner/TestNGRunner.java
```diff
@@ -114,6 +131,30 @@ private boolean shouldIncludeTest(String className) {
...
+                  } catch (Exception e) {
+                    return "Unstringable object";
```
styurin: Why do we need to catch it and not fail fast?

src/com/facebook/buck/testrunner/TestNGRunner.java
```diff
@@ -247,4 +289,11 @@ private PrintStream streamToPrintStream(ByteArrayOutputStream str, PrintStream f
       }
     }
   }
+
+  private static class JUnitReportReporterImproved extends JUnitReportReporter {
```
styurin: `Improved` is not a great term to use. Could you replace it with a name that explain the purpose of this class.

src/com/facebook/buck/testrunner/TestNGRunner.java
jiangty-addepar: we can fail instead, but I don't think "failure to `toString` an object" should be enough to block a developer from seeing their test results

PR #1723, opened 2018-01-23
[JVM] Remove test args that do not affect the output from rulekey computation

### LINKED ISSUE #1713: java_test rule vm_args affect the rulekey
Seems like https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/JavaTestDescription.java#L202 should not affect the rulekey as they are purely runtime flags. They should not influence the need to recompile the tests jar again

kageiit:

Before this change, changing a runtime related argument like `env` or `vm_args` etc. would cause the test rule to rebuild, which is unnecessary.

Fixes #1713


kageiit:
cc @styurin @jkeljo @dreiss @robbertvanginkel @rmaz

styurin:
Could you split these changes into individual PRs? This way we can revert specific commits if something is broken. Also, we can separate discussion about some of these changes without blocking other changes.

kageiit:
Sure

kageiit:
@styurin made this one for jvm tests. Will open more for the rest

PR #1734, opened 2018-01-31
Fixed mixed kotlin and java integration tests

### LINKED ISSUE #1371: Un-ignore mixed sources tests when kotlin 1.1.3 is released.

thalescm:
This fixes #1371 (also you can close #1399 too after).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,6 @@
 /.idea/uiDesigner.xml
 /.idea/workspace.xml
 /.idea/misc.xml
-/.idea/markdown*
 /.idea/copyright/profiles_settings.xml
 
 # From building Emma with Ant.
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@
 /.idea/uiDesigner.xml
 /.idea/workspace.xml
 /.idea/misc.xml
+/.idea/markdown*
 /.idea/copyright/profiles_settings.xml
 
 # From building Emma with Ant.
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -38,6 +38,7 @@
 /.idea/uiDesigner.xml
 /.idea/workspace.xml
 /.idea/misc.xml
+/.idea/markdown*
```
styurin: doesn't look like this change is related, can you revert it?

thalescm:
@styurin :done:

styurin:
There are some problems with formatting. Could you reformat the files? See https://github.com/facebook/buck/blob/master/CONTRIBUTING.md#code-style to find out how to do it.

thalescm:
Huuum, added the pre commit phase, it apparently didn't change anything. @styurin do you know which files were it complaining? I didn't find any that wasn't on the format. Actually if you check the changes, none of them involved any indentation or spacing in source files. The formatting problem seems unrelated to the changes of this PR.
But if there are, if you could point them out it would be a pleasure to fix.

thalescm:
@styurin think I found it (should be fixed in this last commit) ☝️

thalescm:
@styurin anything else I need to change?

styurin:
Nope, it should be merged in later today

PR #1735, opened 2018-01-31
Update d8 to latest

kageiit:
Summary of changes

- Pass in the `android.jar` to D8 so it can perform more accurate desugaring
- Interface method desugaring is only turned off for default interface methods now. Static interface methos should be desugared correctly. This solves a [lambda desugaring issue on 4.x](https://issuetracker.google.com/issues/72475684)
- Try with resources desugaring can now be turned on
- `DalvikStatsTool` has been enhanced to account for the forwarding methods D8 creates in the companion class due to the presence of static interface method desugaring


kageiit:
cc @dreiss @styurin

styurin:
could you reformat the source code?

kageiit:
Done

Issue #560, opened 2015-12-08
How to use the outputs of genrule()?


spierce7:
I'm using genrule to compile a java library to a directory structure of .m and .h files. I'm trying to find some way to use that with the buck build. 

For instance I could use apple_library(), and I could list the genrule name as a dependency, but seeing as how genrule ends up placing the output in the buck-out folder, it's difficult to access it. How should I be accessing these files?

The apple_library rule requires headers and srcs to be separate. I could create a script to separate out the .h and .m files into separate directories, but how would I link those files to the apple_library srcs and headers array?

Or perhaps apple_library isn't the best tool to use?

spierce7:
In a related question, if I had a .a library compiled from objc files with objc headers, what would be the best way to compile that into the application? The documentation specify the requirement of C / C++ header files only it seems for the `prebuilt_cxx_library()`

Coneko:
It's a big complicated because genrule can output directories, but very few target types that read files can also read directories containing those files.

What we do is end up creating a genrule to generate the files in a directory, then other genrules to export every single file from that directory. You can use some python to make that easy, for example:

```python
# this would be ['src/Foo.java', 'src/Bar.java']
JAVA_SRCS = glob(['src/*.java'])
# this would be ['Foo.m', 'Bar.m'], you can adjust it to also include a header for each source file
# or add a prefix or whatever you want
OBJC_FILENAMES = [get_objc_filename(x) for x in JAVA_SRCS]
  
genrule(
  name = 'generate-objc-sources',
  srcs = JAVA_SRCS,
  cmd = '$(exe :generator) -o $OUT $SRCS',
  out = 'output',
)

# this is will generate a genrule for each source file that will copy the file out of the directory, like:
# genrule(
#   name = 'Foo.m',
#   cmd = 'cp buck-out/gen/.../generate-objc-sources/output/Foo.m buck-out/gen/.../Foo.m/Foo.m'
#   out = 'Foo.m',
# )
for objc_filename in OBJC_FILENAMES:
  genrule(
    name = objc_filename,
    cmd = 'cp $(location generate-objc-sources)/{}'.format(objc_filename) $OUT,
    out = objc_filename,
  )

# this will create a library that uses the sources exported by the above genrules, like:
# apple_library(
#   name = 'library',
#   srcs = [':Foo.m', ':Bar.m'],
#   ...
# )
apple_library(
  name = 'library',
  srcs = [':' + x for x in OBJC_FILENAMES],
  ...
)
```

[...]

Issue #518, opened 2015-11-18
Shorthand for including an output file of a genrule directory as a src


mikekap:
If you have a genrule that creates a directory with several files, it's a bit of a pain to write another `genrule` rule for each (expected) output, just to be able to pass it to *_library (java, python, etc). It would be nice to allow you to somehow specify this in the sources - maybe via flavors (although it is a bit hackish)? e.g.

``` python
genrule(
    name='gen_my_thing',
    out='things',
    cmd='$(exe :my_custom_generator) $SRCDIR $OUT',
    srcs=glob(['File1.java', 'File2.java']),
)

java_library(
    name='lib',
    srcs=[
        ':gen_my_thing#lib_output/File1Out.java',
        ':gen_my_thing#lib_output/File2Out.java',
    ]
)
```

sdwilsh:
If we were to support this, we'd have to declare each and every output.  We have a hacky way we did this internally that basically redefines `genrule` to be something that generates a genrule for each output.

sdwilsh:
Internally, that'd probably create an `export_file` rule addressable by the flavor like you have there.

cc @shs96c and @Coneko for design feedback there, but I think we'd take a PR for this if you wanted to make it official.

Coneko:
We've discussed this before and we want the syntax to be with brackets like `rule[file]`.

We won't get around implementing it anytime soon though.

sdwilsh:
@mikekap, if you want to take a stab at this, I think we've given you a bunch of design direction.  If not, that's okay too. :)

njlr:
Has anyone made any progress on this? 

I am having a similar issue where a legacy build system generates a folder of headers that I need to reference. Unfortunately the legacy build system is very complex, and reimplementing it properly in Buck is probably unrealistic, hence the need to "wrap" it.

Coneko:
Note this is just a UI improvement, it would still be functionally equivalent to creating a genrule for each file, minus the copy.

njlr:
@Coneko Thanks for the response. Is there a recommended approach when there are hundreds of files? I don't think it's possible to read the output directory of the `make` genrule to get the list of header files.

Coneko:
You need to know the names of the files in advance and write them in the build files.

You can do that using a python loop if you which, but you cannot list files in the output directory.

The same limitation exists with the new syntax.

njlr:
Yes, that works. 

For those who stumble on this, try something like: 

```
def extract(x):
  genrule(
    name = x,
[...]

PR #1736, opened 2018-02-01
Expand directories in Go srcs param

linzhp:
As discussed in #518 and #560, to compile the output files of a genrule, we need to make a genrule for each output. This could be problematic when the first genrule produces multiple files and their names are not predictable.

This PR makes it possible to put such genrule target in the `srcs` and compile the all generated files.

### DIFF
--- a/src/com/facebook/buck/go/GoCompile.java
+++ b/src/com/facebook/buck/go/GoCompile.java
@@ -38,7 +38,10 @@ import com.facebook.buck.step.fs.TouchStep;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Optional;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


linzhp:
@styurin could you take a look?

src/com/facebook/buck/go/GoCompile.java
```diff
@@ -108,15 +113,27 @@ public GoCompile(
     ImmutableList.Builder<Path> compileSrcListBuilder = ImmutableList.builder();
     ImmutableList.Builder<Path> headerSrcListBuilder = ImmutableList.builder();
     ImmutableList.Builder<Path> asmSrcListBuilder = ImmutableList.builder();
+    List<Path> srcFiles = new ArrayList<>();
```
styurin: This method is already too big, could you move the logic responsible for collecting source files into a standalone method?

src/com/facebook/buck/go/GoCompile.java
```diff
@@ -108,15 +113,27 @@ public GoCompile(
...
+        try {
+          srcFiles.addAll(Files.list(srcPath).filter(Files::isRegularFile).collect(Collectors.toList()));
+        } catch (IOException e) {
+          throw new RuntimeException("An error occur when listing the files under " + srcPath, e);
+        }
+      } else {
+        srcFiles.add(srcPath);
+      }
+    }
+    for (Path f : srcFiles) {
```
styurin: Please, use full work, `f` is not a well-known reduction

linzhp:
@styurin All requested changes are made. Could you take another look?

src/com/facebook/buck/go/GoCompile.java
```diff
@@ -255,6 +260,23 @@ public GoCompile(
     return steps.build();
   }
 
+  private List<Path> getSourchFiles(BuildContext context) {
```
styurin: Should be `getSourceFiles`

styurin:
Could reformat source using google-java-format? Here are the instructions: https://github.com/facebook/buck/blob/master/CONTRIBUTING.md#code-style

linzhp:
@styurin Let's move the discussion on the directory support here:

We've considered the zipping option before implementing this PR, but felt it might slow down the build. And I couldn't remember if I figured out how to unzip the source files in go_library either. We are open to more discussion on that issue.

styurin:
You can use `UnzipStep` to unzip an archive.

linzhp:
The output of UnzipStep is a directory. So in the end, we still have to deal with source directories anyway. Zipping and unzipping merely add some unnecessary computational cost. Do you see any benefits?

PR #1740, opened 2018-02-02
Support Go Subtests

linzhp:
Go 1.7 introduced subtests, which will print out report like this:

```
=== RUN   TestTime
=== RUN   TestTime/12:31_in_Europe/Zuri
=== RUN   TestTime/12:31_in_America/New_York
=== RUN   TestTime/08:08_in_Australia/Sydney
--- FAIL: TestTime (0.00s)
    --- FAIL: TestTime/12:31_in_Europe/Zuri (0.00s)
    	command_test.go:23: could not load location
    --- PASS: TestTime/12:31_in_America/New_York (0.00s)
    --- PASS: TestTime/08:08_in_Australia/Sydney (0.00s)
```

This PR make Buck capable of consuming such test report.

### DIFF
--- a/src/com/facebook/buck/go/GoTest.java
+++ b/src/com/facebook/buck/go/GoTest.java
@@ -56,8 +56,10 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -69,7 +71,7 @@ public class GoTest extends NoopBuildRuleWithDeclaredAndExtraDeps
   private static final Pattern TEST_START_PATTERN = Pattern.compile("^=== RUN\\s+(?<name>.*)$");
   private static final Pattern TEST_FINISHED_PATTERN =
       Pattern.compile(
-          "^--- (?<status>PASS|FAIL|SKIP): (?<name>.+) \\((?<duration>\\d+\\.\\d+)(?: seconds|s)\\)$");
+          "^\\s*--- (?<status>PASS|FAIL|SKIP): (?<name>.+) \\((?<duration>\\d+\\.\\d+)(?: seconds|s)\\)$");
   // Extra time to wait for the process to exit on top of the test timeout
   private static final int PROCESS_TIMEOUT_EXTRA_MS = 5000;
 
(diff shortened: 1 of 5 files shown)


src/com/facebook/buck/go/GoTest.java
```diff
@@ -207,19 +209,19 @@ public GoTest(
...
-          currentTest = Optional.empty();
+          currentTests.remove(matcher.group("name"));
           stdout.clear();
         } else {
           stdout.add(line);
         }
       }
 
-      if (currentTest.isPresent()) {
+      for (String t : currentTests) {
```
styurin: please, use full word

linzhp:
Older version of Go doesn't have subtests. This new change is compatible with both regular tests and subtests, as verified by the existing integration tests with a lot of regular tests.

linzhp:
@styurin All requested changes are made. Could you take another look?

test/com/facebook/buck/go/testdata/go_test/subtests.go
```diff
@@ -0,0 +1,33 @@
...
+
+// modified from an example here: https://blog.golang.org/subtests
+
+import (
+	"fmt"
+	"testing"
+	"time"
+)
+
+func TestTime(t *testing.T) {
```
styurin: What would happen if Buck runs this tests with go that is lower than 1.7?

linzhp: It will most likely break. Should I skip that test for now and put a comment to enable it only after Buck officially supports Go 1.7?

styurin: The proper solution would be to detect the version of go and enable this test only when it's higher than 1.7. Check `GoAssumptions.assumeGoCompilerAvailable();` that is called in the test to make sure go is available. You'll need to extend it to detect the version of go.

linzhp: Added the check for Go version. Please take another look

test/com/facebook/buck/go/GoAssumptions.java
```diff
@@ -64,4 +71,55 @@ public static void assumeGoCompilerAvailable() throws InterruptedException, IOEx
...
+          processExecutor.launchAndExecute(
+              ProcessExecutorParams.builder().addCommand("go", "version").build(),
+              EnumSet.of(ProcessExecutor.Option.EXPECTING_STD_OUT),
+              /* stdin */ Optional.empty(),
+              /* timeOutMs */ Optional.empty(),
+              /* timeoutHandler */ Optional.empty());
+      if (goToolResult.getExitCode() == 0) {
+        // e.g., "go version go1.9.2 darwin/amd64", "go version go1.9.2 windows/amd64",
+        // "go version go1.2.1 linux/amd64"
+        actualVersion = goToolResult.getStdout().get().substring(13, 18);
```
styurin: That wouldn't work for versions like '1.10.123'. Could you use regex instead? Something like `((\d)+)\.((\d)+)\.((\d)+)` ?

linzhp: Good catch! Updated to use regex.

PR #1741, opened 2018-02-02
Force use of Python 2.7

plinehan:
Buck requires Python 2.7, so force its use.  One one of my machines, "python2" unfortunately resolves to Python 2.6 even though Python 2.7 is on my PATH.

Unfortunately, I can't uninstall Python 2.6 for reasons I won't go into.

### DIFF
--- a/bin/buck
+++ b/bin/buck
@@ -18,5 +18,5 @@ function abs_script_dir_path {
 
 BUCK_DIR=$(abs_script_dir_path "$0")
 
-PYTHON=$(command -v python2.7 python | head -1)
+PYTHON=$(command -v python2.7 python2 python | head -1)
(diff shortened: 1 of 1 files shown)


bin/buck
```diff
@@ -18,5 +18,5 @@ function abs_script_dir_path {
 
 BUCK_DIR=$(abs_script_dir_path "$0")
 
-PYTHON=$(command -v python2 python | head -1)
+PYTHON=$(command -v python2.7 python | head -1)
```
philipjameson: This should be fine, but can you make this python2.7 python2 python in case, for some reason, a person doesn't have a 2.7 link?

plinehan: Done!

PR #1742, opened 2018-02-03
[Go] Remove test args that do not affect the output from rulekeys

kageiit:


kageiit:
cc @linzhp @styurin

PR #1746, opened 2018-02-03
[Cxx] Remove test args that do not affect the output from rulekeys

kageiit:

### DIFF
--- a/src/com/facebook/buck/cxx/CxxTest.java
+++ b/src/com/facebook/buck/cxx/CxxTest.java
@@ -63,9 +63,9 @@ import java.util.stream.Stream;
 public abstract class CxxTest extends AbstractBuildRuleWithDeclaredAndExtraDeps
     implements TestRule, HasRuntimeDeps, BinaryBuildRule {
 
-  @AddToRuleKey private final ImmutableMap<String, Arg> env;
-  @AddToRuleKey private final ImmutableList<Arg> args;
-  @AddToRuleKey private final Tool executable;
+  private final ImmutableMap<String, Arg> env;
+  private final ImmutableList<Arg> args;
+  private final Tool executable;
 
   @AddToRuleKey
   @SuppressWarnings("PMD.UnusedPrivateField")


src/com/facebook/buck/cxx/CxxTest.java
```diff
@@ -83,7 +78,7 @@ public CxxTest(
       Tool executable,
       ImmutableMap<String, String> env,
       Supplier<ImmutableList<String>> args,
-      ImmutableSortedSet<? extends SourcePath> resources,
+      @SuppressWarnings("unused") ImmutableSortedSet<? extends SourcePath> resources,
```
styurin: Could you remove it?

kageiit: Sure

kageiit: ant complains if I remove this though :(

kageiit:
Ping! is this blocked on something? :)

kageiit:
@styurin Ping :)

src/com/facebook/buck/cxx/CxxTest.java
```diff
@@ -63,13 +62,9 @@
 public abstract class CxxTest extends AbstractBuildRuleWithDeclaredAndExtraDeps
     implements TestRule, HasRuntimeDeps, BinaryBuildRule {
 
-  @AddToRuleKey private final ImmutableMap<String, Arg> env;
-  @AddToRuleKey private final ImmutableList<Arg> args;
-  @AddToRuleKey private final Tool executable;
-
-  @AddToRuleKey
-  @SuppressWarnings("PMD.UnusedPrivateField")
-  private final ImmutableSortedSet<? extends SourcePath> resources;
```
styurin: We cannot remove this field right now. Some tests use resources and removing this filed would break situations when a resource has changed but the test's rule key hasn't changed.

The proper handling of these resources would be to provide access to them from test's current folder (by symlinking like in genrule), but right now we don't have this logic.

kageiit: I can add this back and keep it adding to the rulekey

kageiit:
@styurin updated

PR #1748, opened 2018-02-04
[Android] Fix primary dex scenario file usage when progaurd is enabled.

raviagarwal7:
While computing the scenario classes, input scenario classes were being
obfuscated to get the dependent classes, but the dependent classes which
were not being de-obfuscated.

Since `createRequiredInPrimaryZipPredicate` needs to de-obfuscate the 
input classFileName primary dex scenario file wouldn’t work with proguard.


raviagarwal7:
cc: @dreiss

Issue #1750, opened 2018-02-06
disable_implicit_native_rules doesn't work in Skylark parser


linzhp:
[This commit](https://github.com/facebook/buck/commit/6586f3c231b11dc3f4ef0f849b4e81a2fbefafcd) introduced a new parser config called "disable_implicit_native_rules". However, when the parser is set to Skylark, the value of disable_implicit_native_rules is ignored and implicit native rules are always disable in bzl files.

philipjameson:
Well, that definitely shouldn't have happened. Lemme take a look .

philipjameson:
Actually, on a reread, this was intentional after discussing with @ttsugriy. It wasn't actually supposed to work that way to start with, we just did it because it was easier for some of our migrations. @ttsugriy , thoughts on making this configure extension files as well?

linzhp:
This is not very backward compatible. After pulling your commit, all our builds failed in a sudden...

philipjameson:
No it's not, though I believe the skylark parser is marked experimental and cards on the table, I didn't know that anyone really was using it yet other than FB. Mind if we meet and the middle and just toss a config flag to re-export them? I think going forward we'd like new projects to be explicit about using native rules.

linzhp:
To be future-proof, should we replace all go_library with native.go_library and genrule with native.genrule in both BUCK and .bzl files?

philipjameson:
.bzl files should use native.go_library and native.genrule. By default (unless you disable it), go_library and genrule will work in BUCK files without the prefix.

linzhp:
OK. Out of curiosity, could you explain the design decision behind requiring explicit native functions? Or give me a pointer to some existing discussions/documents?

ttsugriy:
@linzhp There are a number of reasons we're doing it this way, but they can be broadly grouped into following:
- we would like to keep the global namespace pollution as much as possible. It makes it easier to reaon about code without having to scan an entire document to check if function call is for a native rule or a macro. Unfortunately it's not uncommon for macro writers to override/shadow native rules, so having an explicit namespace for them makes this distinction easier.
- we'd like to be as close to Bazel API as reasonably [possible](https://docs.bazel.build/versions/master/skylark/lib/native.html). We've met Google, Twitter and Microsoft and would love to work with them on standardizing extension definitions to democratize them.

[...]

PR #1751, opened 2018-02-06
[buck] Allow implicit native functions in skylark extension files

philipjameson:
With skylark, we're trying to enforce that 'native' should be used to access native rules. However, when migrating from python_dsl -> skylark, it can be useful to have the native rules exposed in the global scope. This allows it to be configured, but by default we disallow it to encourage new projects to not follow the globals practice.


philipjameson:
This is one way to address GH-1750

PR #1752, opened 2018-02-07
[WIP] Send a GH status with linter results.

beauby:
Note: problem with this at the moment is that it would potentially leak the GH token.

### DIFF
--- a/scripts/travisci_test_java_file_format
+++ b/scripts/travisci_test_java_file_format
@@ -9,6 +9,16 @@ function finish {
 
 trap finish EXIT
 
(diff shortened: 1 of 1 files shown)


kageiit:
@beauby Why would it leak `GH_LINTER_TOKEN`. Env vars can be obfuscated in the travis settings

beauby:
@kageiit I believe obfuscated env vars are not visible from pull requests from forks, as otherwise anyone could issue a PR modifying the .travis.yml, get the new config to run and add a statement to print the env var in the build logs.

PR #1757, opened 2018-02-08
Add info about prebuilt buck binaries

kageiit:
As discussed in https://github.com/facebook/buck/commit/38d3f04f814b1c42c7b7f6da01f4aa3ec7b46586#commitcomment-27427942


ttsugriy:
thank you, @kageiit!

PR #1768, opened 2018-02-12
Enabling mvn URI for http_file and http_archive downloadables.

romanoid:


philipjameson:
Can we please get both the unit tests and integration tests updated here?

sdwilsh:
Why this change over using `remote_file`?

kageiit:
> Why this change over using remote_file?

We want to use fallback url options for resolving maven artifacts as well as sha256 content hash checks and unarchive functionality for archived binaries hosted on artifactory

romanoid:
@philipjameson thanks for suggestion, added integration tests. 

MavenUrlDecoder is already covered by unit tests. Please advice if any specific unit test you have in mind is desired.

philipjameson:
Thanks for the updates!

PR #1770, opened 2018-02-12
 Add per target environent to apple_test

robbertvanginkel:
Enable target specific environment variables to be set for apple_test
rules. This is for example useful to set variables like
LLVM_PROFILE_FILE and get a per test target file with profile data.

### DIFF
--- a/src/com/facebook/buck/apple/AppleTest.java
+++ b/src/com/facebook/buck/apple/AppleTest.java
@@ -111,6 +111,7 @@ public class AppleTest extends AbstractBuildRuleWithDeclaredAndExtraDeps
   private final String testLogDirectoryEnvironmentVariable;
   private final String testLogLevelEnvironmentVariable;
   private final String testLogLevel;
+  private final Optional<ImmutableMap<String, String>> testSpecificEnvironmentVariables;
 
   /**
    * Absolute path to xcode developer dir.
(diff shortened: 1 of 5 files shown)


src/com/facebook/buck/apple/AppleTest.java
```diff
@@ -265,6 +268,12 @@ public void readOutput(InputStream output) throws IOException {
     Optional<Path> testHostAppPath = extractBundlePathForBundle(testHostApp, buildContext);
     Optional<Path> uiTestTargetAppPath = extractBundlePathForBundle(uiTestTargetApp, buildContext);
 
+    ImmutableMap<String, String> testEnvironmentOverrides =
+        ImmutableMap.<String, String>builder()
+            .putAll(options.getEnvironmentOverrides())
+            .putAll(this.testSpecificEnvironmentVariables.orElse(ImmutableMap.of()))
```
robbertvanginkel: In this case a value `--test-runner-env` could be overwritten by a target specific variable in a test. I think this makes sense as target level is more granular than a setting for a buck test command. However one could also argue that it is better to have a cli value override a value defined from a file. I don't have a strong preference here.

src/com/facebook/buck/apple/AppleTestDescription.java
```diff
@@ -647,6 +648,9 @@ default boolean getIsUiTest() {
     // Bundle related fields.
     ImmutableMap<String, String> getDestinationSpecifier();
 
+    // Bundle related fields.
+    Optional<ImmutableMap<String, String>> getEnvironment();
```
robbertvanginkel: Open to suggestions. `environment` seems clear enough on `apple_test` but open to renaming it to something more specific like `test_environment`

robbertvanginkel: Renaming this to `env`, to have the same naming as `cxx_test` and `python_test`.

src/com/facebook/buck/apple/AppleTest.java
milend: @robbertvanginkel I believe having it in the above order is beneficial as it allows developer control in both local and remote environments. In a local env, a developer can control both `--test-runner-env` and BUCK files. In a remote env, developers can usually only control BUCK files.

milend:
LGTM.

PR #1771, opened 2018-02-12
Make default depfile mode cache

kageiit:
Having the default be cached seems to be a big benefit to consumers of buck who may not have turned this on explicitly


kageiit:
cc @cjhopman @jkeljo @styurin

jkeljo:
@andrewjcg

kageiit:
@andrewjcg can you please approve/import?

PR #1776, opened 2018-02-14
fix typo in documentation

IlyaNaumovich:


ttsugriy:
Thanks for your contribution, @IlyaNaumovich, but it's not really a typo - buck query does indeed support specifying targets without quotes, although some targets have to be quoted in case they contain some special characters.

PR #1778, opened 2018-02-15
Android: fix aapt processing of aars without res dir.

artem-zinnatullin:
### Problem: 

Google Play Services v11.4.2 and above have special "license" artifacts (`aar`) that normal artifacts depend on. These "license" `aar`s  only have few files in them and no folders, specifically no `res` folder:

Example:
```console
play-services-base-license-11.8.0.aar:

./AndroidManifest.xml
./third_party_licenses.json
./third_party_licenses.txt
```

Building Android projects that depend on Google Play Services `11.4.2+` causes following Buck build error:

```console
ERROR: resource directory 'buck-out/bin/prebuilt_aar_without_res/__unpack_play-services-base-license#aar_unzip__/res' does not exist

Build failed: Command failed with exit code 1.
stderr: ERROR: resource directory 'buck-out/bin/prebuilt_aar_without_res/__unpack_play-services-base-license#aar_unzip__/res' does not exist
```
### Fix

This PR fixes problem of non-existing `res` dir by running `MkdirStep` during `aar` unzip which creates empty `res` folder if it doesn't exist.

Originally I've tried to hook into other places like `FilteredResourcesProvider`,  `AndroidBinaryResourcesGraphEnhancer` and so on to filter out those non-existing `res` `SourcePath`s from going into `aapt` steps, but that complicated code much more due to different code paths for `aapt` and `aapt2`.

Creating empty `res` dir during unzip looks consistent to existing steps happening during `unzip` which apply same workaround for `assets`, proguard config and native libs.

I've also added a test that was failing without the fix.

Fix works with both `aapt` and `aapt2`.

cc @kageiit


artem-zinnatullin:
Didn't see the issue, thanks for linking @kageiit!

kageiit:
cc @styurin @dreiss

PR #1780, opened 2018-02-21
Add support for @classpath_abi macro

raviagarwal7:
This is similar to @classpath macro but gives the corresponding 
class-abi jar for the classpath entries.

This is useful for genrules which only depends on the dependencies abi’s.
For eg: android lint where it need not run the lint checks again if 
the dependent’s abi doesn’t change.

Usage: `$(@classpath_abi :<target_path>)`

### DIFF
--- a/src/com/facebook/buck/jvm/core/HasJavaAbi.java
+++ b/src/com/facebook/buck/jvm/core/HasJavaAbi.java
@@ -43,16 +43,16 @@ public interface HasJavaAbi {
   Flavor SOURCE_ONLY_ABI_FLAVOR = InternalFlavor.of("source-only-abi");
   Flavor VERIFIED_SOURCE_ABI_FLAVOR = InternalFlavor.of("verified-source-abi");
 
-  static BuildTarget getClassAbiJar(BuildTarget libraryTarget) {
-    Preconditions.checkArgument(isLibraryTarget(libraryTarget));
-    return libraryTarget.withAppendedFlavors(CLASS_ABI_FLAVOR);
-  }
-
   static boolean isAbiTarget(BuildTarget target) {
     return isClassAbiTarget(target)
         || isSourceAbiTarget(target)
         || isSourceOnlyAbiTarget(target)
         || isVerifiedSourceAbiTarget(target);
+  }
+
+  static BuildTarget getClassAbiJar(BuildTarget libraryTarget) {
+    Preconditions.checkArgument(isLibraryTarget(libraryTarget));
+    return libraryTarget.withAppendedFlavors(CLASS_ABI_FLAVOR);
   }
 
   static boolean isClassAbiTarget(BuildTarget target) {
--- a/src/com/facebook/buck/jvm/core/HasJavaAbi.java
+++ b/src/com/facebook/buck/jvm/core/HasJavaAbi.java
@@ -43,16 +43,16 @@ public interface HasJavaAbi {
   Flavor SOURCE_ONLY_ABI_FLAVOR = InternalFlavor.of("source-only-abi");
   Flavor VERIFIED_SOURCE_ABI_FLAVOR = InternalFlavor.of("verified-source-abi");
 
+  static BuildTarget getClassAbiJar(BuildTarget libraryTarget) {
+    Preconditions.checkArgument(isLibraryTarget(libraryTarget));
+    return libraryTarget.withAppendedFlavors(CLASS_ABI_FLAVOR);
+  }
+
   static boolean isAbiTarget(BuildTarget target) {
     return isClassAbiTarget(target)
         || isSourceAbiTarget(target)
         || isSourceOnlyAbiTarget(target)
         || isVerifiedSourceAbiTarget(target);
-  }
-
-  static BuildTarget getClassAbiJar(BuildTarget libraryTarget) {
-    Preconditions.checkArgument(isLibraryTarget(libraryTarget));
-    return libraryTarget.withAppendedFlavors(CLASS_ABI_FLAVOR);
   }
 
   static boolean isClassAbiTarget(BuildTarget target) {
--- a/src/com/facebook/buck/rules/macros/ClasspathAbiMacroExpander.java
+++ b/src/com/facebook/buck/rules/macros/ClasspathAbiMacroExpander.java
@@ -38,7 +38,7 @@ import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 /**
- * Used to expand the macro {@literal $(abi_jar_path //some:target)} to the transitive abi's jars
+ * Used to expand the macro {@literal $(classpath_abi //some:target)} to the transitive abi's jars
  * path of that target, expanding all paths to be absolute.
  */
 public class ClasspathAbiMacroExpander extends BuildTargetMacroExpander<ClasspathAbiMacro> {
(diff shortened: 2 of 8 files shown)


raviagarwal7:
Added more context in the description and also added to the macros soy.
Have fixed the build errors as well.

src/com/facebook/buck/jvm/core/HasJavaAbi.java
```diff
@@ -43,18 +43,18 @@
   Flavor SOURCE_ONLY_ABI_FLAVOR = InternalFlavor.of("source-only-abi");
   Flavor VERIFIED_SOURCE_ABI_FLAVOR = InternalFlavor.of("verified-source-abi");
 
-  static BuildTarget getClassAbiJar(BuildTarget libraryTarget) {
```
styurin: Any changes in this file? Why do you need to move this method?

src/com/facebook/buck/rules/macros/ClasspathMacroExpander.java
```diff
@@ -85,7 +85,7 @@ protected Arg expand(SourcePathResolver resolver, ClasspathMacro ignored, BuildR
   // The # characters that might be present in classpaths due to flavoring would be read as
   // comments. As a simple workaround, we quote the entire classpath.
   private static class ClassPathWriteToFileArg extends WriteToFileArg {
-    public ClassPathWriteToFileArg(BuildTarget target, String prefix, Arg delegate) {
```
styurin: Changes in this file look unrelated, could you revert them?

src/com/facebook/buck/rules/macros/ClasspathAbiMacroExpander.java
```diff
@@ -0,0 +1,155 @@
...
+
+    BuildRule inputRule = resolve(resolver, input);
+    return expand(resolver, inputRule);
+  }
+
+  protected Arg expand(BuildRuleResolver ruleResolver, BuildRule inputRule) throws MacroException {
+
+    ImmutableList<SourcePath> jarPaths =
+        getHasClasspathEntries(inputRule)
+            .getTransitiveClasspathDeps()
```
raviagarwal7: `getTransitiveClasspathDeps` returns both `#aar_prebuilt_jar,class-abi` and `#class-abi` flavours. I verified that both the output jar's are same, just on different paths.

//cache:com.support-compat-27.0.2.aar#aar_prebuilt_jar,class-abi
buck-out/gen/cache/com.support-27.0.2.aar#aar_prebuilt_jar,class-abi/com.support-27.0.2.aar-abi.jar

//cache:com.support-compat-27.0.2.aar#class-abi
buck-out/gen/cache/com.support-27.0.2.aar#class-abi/com.support-27.0.2.aar-abi.jar

How can I filter and keep only the `#class-abi` deps? Also is there a scenario where the output of the two flavours will differ?

src/com/facebook/buck/jvm/core/HasJavaAbi.java
raviagarwal7: just moved the method to be in accordance with others, but will revert back

src/com/facebook/buck/rules/macros/ClasspathMacroExpander.java
raviagarwal7: Intellij was showing suggestions that this is not needed, will revert back.

src/com/facebook/buck/rules/macros/ClasspathAbiMacroExpander.java
raviagarwal7: added a filter to only include rules which have non-null `sourcePathToOutput`, same behaviour is in `ClasspathMacro`.

src/com/facebook/buck/rules/macros/ClasspathAbiMacroExpander.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  /**
+   * Get the class abi jar if present for the rule otherwise return the rule's output
+   *
+   * @param rule The rule whose jar path needs to be returned
+   * @param ruleResolver BuildRuleResolver
+   * @return class abi jar or output jar if not found
+   */
+  private SourcePath getJarPath(BuildRule rule, BuildRuleResolver ruleResolver) {
```
styurin: This needs to be annotated with `@Nullable`.

test/com/facebook/buck/rules/macros/ClasspathAbiMacroExpanderTest.java
```diff
@@ -0,0 +1,198 @@
...
+            .addSrc(Paths.get("Example.java")) // Force a jar to be created
+            .build(ruleResolver, filesystem);
+
+    assertExpandsTo(
+        rule,
+        ruleResolver,
+        String.format(
+            "%s" + File.separator + "%s",
+            filesystem.getRootPath(),
+            "buck-out/gen/cheese/cake#class-abi/cake-abi.jar"));
```
styurin: These tests are failing on Windows. Our CI doesn't run these tests for some reason (we're investigating), but it looks like it's failing because of the path separators.

src/com/facebook/buck/rules/macros/ClasspathAbiMacroExpander.java
raviagarwal7: done

test/com/facebook/buck/rules/macros/ClasspathAbiMacroExpanderTest.java
raviagarwal7: using `new File.("file-path-string")toPath()` now, hopefully this would work fine on windows.

test/com/facebook/buck/rules/macros/ClasspathAbiMacroExpanderTest.java
```diff
@@ -0,0 +1,198 @@
...
+import org.hamcrest.Matchers;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ClasspathAbiMacroExpanderTest {
+  private ClasspathAbiMacroExpander expander;
+  private ProjectFilesystem filesystem;
+
+  @Before
+  public void createMacroExpander() {
```
styurin: this is usually called `setUp`. It also does more than creating macro expander.

test/com/facebook/buck/rules/macros/ClasspathAbiMacroExpanderTest.java
```diff
@@ -0,0 +1,198 @@
...
+    this.expander = new ClasspathAbiMacroExpander();
+    this.filesystem = FakeProjectFilesystem.createJavaOnlyFilesystem();
+  }
+
+  @Test
+  public void shouldIncludeARuleIfNothingIsGiven() throws Exception {
```
styurin: test methods usually start with `test` so that the method name is a verb, right now it's a statement

test/com/facebook/buck/rules/macros/ClasspathAbiMacroExpanderTest.java
raviagarwal7: done

test/com/facebook/buck/rules/macros/ClasspathAbiMacroExpanderTest.java
raviagarwal7: done

kageiit:
cc @styurin the comments are addressed. Any update on this?

PR #1782, opened 2018-02-23
Filtering platform specific source files

### LINKED ISSUE #1701: Platform sources for Go
Is there a way to have go_binary and go_library to include only the source files compatible with current platform (OS/arch)? The native `go build <package>` command does that automatically, but I don't know how to configure buck to have the same behavior.

For example, the following command shows the list of files included in `go build` when I am building the package:

`go list -f '{{.GoFiles}}' github.com/jessevdk/go-flags`

linzhp:
This PR fixes #1701 

Known issue:

`srcs=['.']` will crash on `go_binary` rules but not on `go_library`. The crash happens before `GoCompile.getBuildSteps` gets called. Temporary workaround is to use `srcs=glob(["*.go"])`. The filtering still works.


styurin:
Could you explain why `srcs=['.']` is preferred to using `glob(["*.go"], exclude=["*_test.go"])` and `glob(["*.go"])`?

linzhp:
I wouldn't say `srcs=['.']` is preferred. I just feel if we are going to filter `glob`, we should support the same filtering for `srcs=['.']` too. But it turned out not that easy to support that latter.

.travis.yml
```diff
@@ -141,7 +141,7 @@ before_install:
     fi
   # Install go 1.8, required for cgo -srcdir flag to work
   - eval "$(gimme 1.8)"
-  - echo -e "[go]\n  root = ${GOROOT}" >> .buckconfig.local
+#  - echo -e "[go]\n  root = $(go env GOROOT)" >> .buckconfig.local
```
ttsugriy: why is this commented out?

.travis.yml
```diff
@@ -154,4 +154,4 @@ notifications:
   webhooks: https://code.facebook.com/travis/webhook/
 
 script:
-  - ./scripts/travisci_run.sh
+  - ./scripts/travisci_run.sh
```
ttsugriy: why is the new line removed?

.travis.yml
linzhp: I was struggling to pass the integration test on Travis CI. GOROOT was set somewhere to `/usr/local/go`, cause `go list` command to throw an error `go: cannot find GOROOT directory: /usr/local/go`.

Do you know how GOROOT gets the value `/usr/local/go` in Travis CI? Running integration tests locally doesn't have the problem.

ttsugriy: @linzhp, afaik we have a default goroot in [test buck file](https://github.com/facebook/buck/blob/2d5e69e901e774a221997bed1580d0828e5a3561/test/com/facebook/buck/go/BUCK#L7):
```
"GOROOT": read_config("go", "root", "/usr/local/go"),
```

so I assume it's the one and as such it should be overridden by the
```
[go]
  root = <some other root>
```

linzhp:
@ttsugriy @styurin I have finished my changes and finally passed all Travis tests. Could you take another look/

linzhp:
Hold on. The CGo test is failing

linzhp:
@ttsugriy @styurin Now all the integration tests passed. Please take another look. Thanks!

styurin:
I don't know why go rules support specifying directories as inputs, but this is definitely not a recommended way to declare inputs. Unless there is a good reason we shouldn't support that at all. Other rules use globs and we have no plans to specify directories as inputs.

linzhp:
When will this PR get merged in Github? It was imported into FB's internal repo yesterday, but it seems taking more time than usual to find its way out to Github.

src/com/facebook/buck/go/GoCompileStep.java
```diff
@@ -69,37 +70,45 @@ public GoCompileStep(
...
-      commandBuilder.add("-importmap", importMap.getKey() + "=" + importMap.getValue());
-    }
+      for (Path dir : includeDirectories) {
+        commandBuilder.add("-I", dir.toString());
+      }
 
-    if (asmHeaderPath.isPresent()) {
-      commandBuilder.add("-asmhdr", asmHeaderPath.get().toString());
-    }
+      for (Map.Entry<Path, Path> importMap : importPathMap.entrySet()) {
```
styurin: Could you rename `importMap` to something that reflects the purpose of this object? It's definitely not a map.

src/com/facebook/buck/go/GoCompileStep.java
```diff
@@ -69,37 +70,45 @@ public GoCompileStep(
...
+        commandBuilder.add("-complete");
+      }
 
-    commandBuilder.addAll(srcs.stream().map(Object::toString).iterator());
+      commandBuilder.addAll(pathStrings);
 
-    return commandBuilder.build();
+      return commandBuilder.build();
+    } else {
+      return ImmutableList.of("echo", "No source files found in " + workingDirectory);
```
styurin: Why do you need to call `echo`?

src/com/facebook/buck/go/GoListStep.java
```diff
@@ -0,0 +1,105 @@
...
+  @Override
+  protected void addOptions(ImmutableSet.Builder<Option> options) {
+    super.addOptions(options);
+    options.add(Option.EXPECTING_STD_OUT);
+  }
+
+  @Override
+  public ImmutableMap<String, String> getEnvironmentVariables(ExecutionContext context) {
+    return ImmutableMap.<String, String>builder()
+        // The GOROOT is set to /usr/local/go somehow on Travis CI, which doesn't exist.
```
styurin: This comment is out of place here. Why do we need to know about Travis CI in this file?

I think it's totally reasonable to set these variables before running go command without explanation, so you can remove this comment or change it to state that the go commands rely on these variables and we set them to override external variables.

test/com/facebook/buck/go/GoAssumptions.java
```diff
@@ -43,9 +43,9 @@
 
 abstract class GoAssumptions {
   // e.g., "go version go1.9.2 darwin/amd64", "go version go1.9.2 windows/amd64",
-  // "go version go1.2.1 linux/amd64"
+  // "go version go1.8 linux/amd64"
   private static final Pattern VERSION_PATTERN =
-      Pattern.compile("^go version go(\\d+)\\.(\\d+)\\.(\\d+).*$");
```
styurin: A better fix would be to allow specifying both variants so that you don't need to change this code in future.

src/com/facebook/buck/go/FilteredSourceFiles.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    sourceFiles.addAll(extraSrcFiles);
+    return sourceFiles.iterator();
+  }
+
+  /**
+   * @param srcFiles Extra source files to be added to the iterator. These files will NOT be subject
+   *     to filtering
+   */
+  public void addExtraSourceFiles(List<Path> srcFiles) {
```
styurin: It looks like this method is called right after the creation of `FilteredSourceFiles`. Why not pass it as a parameter in the constructor and make it immutable?

src/com/facebook/buck/go/GoCompileStep.java
linzhp: The caller of this method will throw an exception if I return an empty list. So I have to return some command. What other command do you suggest?

test/com/facebook/buck/go/GoAssumptions.java
linzhp: The new regex allows both variants. The new regex ignores the patch version as there shouldn't be any new feature added as the patch version changes.

src/com/facebook/buck/go/GoCompileStep.java
styurin: Why not throw an exception?

linzhp: Throwing an exception here will crash buck. As this step doesn't expect any stdout, putting a message in the stdout will result in a warning message in Buck's output (colored as red). If there are things in this package needed by other packages, the build will fail with message saying something is undefined. Combined with the warning message here, it's easy to track down the problem. If the build succeeds regardless, it means this package is not necessary for the current platform, although it might be necessary for other platforms. Buck should tolerate such scenario too.

src/com/facebook/buck/go/FilteredSourceFiles.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public Collection<GoListStep> getFilterSteps() {
+    return filterSteps.values();
+  }
+
+  @Override
+  public Iterator<Path> iterator() {
+    HashSet<Path> sourceFiles = new HashSet<>();
+    for (Path srcFile : rawSrcFiles) {
+      if (filterSteps.get(srcFile.getParent()).getSourceFiles().contains(srcFile)) {
```
styurin: Could you add a check to `filterSteps.get` to make sure it doesn't return a null? Something like this:

     Preconditions.checkNotNull(filterSteps.get())

Our code analysis check is complaining about that

src/com/facebook/buck/go/GoCompileStep.java
linzhp: The compile command will fail if there is no source file, producing an error message prompting the correct format of the command line arguments, which can be used in place of this echo message. I do the due diligence check for lack of input files [here](https://github.com/facebook/buck/pull/1782/files#diff-d4c3760f594f832155f3915c72110e06R171), but the existence of raw source files doesn't guarantee the existence of filtered source files.

styurin: Then if srcs is empty we can skip this step, right?

linzhp: The length of `srcs` can not be determined until GoListStep is executed. So in GoCompile.getBuildSteps, we are not always sure whether to skip GoCompileStep. In this method, when GoCompileStep is about to execute, it's probably too late to yank it out. Do you see a way to skip GoCompileStep after GoListStep is executed?

styurin: This issue shows the problem with supporting the directories as inputs in general. Buck need to know about the inputs before creating these steps, we shouldn't skip steps based on the output from the previous steps.

Going to back to the original issue of supporting directories as inputs in go rules, Java, for example, supports using zip files as inputs. This gives you a way to pass multiple files as a single file. Is there a reason we cannot add support for that in go rules?

linzhp: This issue is really not about supporting directories. The FilteredSourceFiles only takes individual source files, not directories. What it does is run a "go list" command on the parent directory of each file, and decide whether to ignore the source file. Removing the support for directories doesn't affect any code in this filtering logic. One can still pass a Windows specific Go file into go_library, by either using glob or the file name, and try to compile it on Linux. This filtering still has to happen in order to have the compiler skip that file.

styurin: Is running this "go list" command a prerequisite for a build or it's just for the sake of convenience?

The issue as I see it is the steps generated in `FilteredSourceFiles`are accessing the whole directory and it's not clear whether the files in that directory are dependencies of the build rule that produced the steps or not.

It still feels to me that this logic is more complicated than necessary because of the support of directories.

linzhp: Go source files have [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints), which specify whether a source file should be compiled for the current platform. In order for Buck to honor the build constraints, there are two options:

1. Running "go list" command on the directory each go file resides. As "go list" command only returns source files with all build constraints satisfied, we can know whether we should pass a file to the compiler by checking if it's in the output of "go list"

2. Implement a constraints solver ourselves. Then it won't rely on "go list" command, and we don't need GoListStep anymore. Also, we will know early on whether to skip GoCompileStep in GoCompile.getBuildSteps. However, the constraint solver will need to update whenever the definition of Go build constraints changes

In either option, the build constraints are checked on file-to-file basis. All directories passed in from build rules have been expanded before the filtering logic takes place, so the filtering logic only sees files, not directories. Do you see how this logic can be simplified if the directory support is removed?

styurin: How about overriding `execute` to check if `srcs` is empty, return `StepExecutionResult.of(0)` if it is and proceed with compilation if some source files are present?

linzhp: Good idea. Updated now

src/com/facebook/buck/go/GoListStep.java
```diff
@@ -0,0 +1,105 @@
...
+    this.fileTypes = fileTypes;
+  }
+
+  @Override
+  protected ImmutableList<String> getShellCommandInternal(ExecutionContext context) {
+    ImmutableList.Builder<String> commandBuilder =
+        ImmutableList.<String>builder()
+            .add(goToolchain.getGoRoot().resolve("bin").resolve("go").toString())
+            .add("list")
+            .add("-f")
```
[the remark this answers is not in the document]
linzhp: Oh, I wasn't aware of such option before. Thanks for pointing out.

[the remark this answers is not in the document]
styurin: How about backward compatibility? What's general practice around updating to latest go toolchain?

styurin:
> This branch has conflicts that must be resolved

Please, rebase these changes

linzhp:
Done. Hope this is the last rebase I have to do before it gets merged. I've done several times before for this PR, and it got conflicts after hanging here for a few more days:(

src/com/facebook/buck/go/GoListStep.java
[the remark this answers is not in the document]
nemith: -json was there in Go 1.0 so no worries.

https://github.com/golang/go/blob/release-branch.go1/src/cmd/go/list.go#L71

Issue #1730, opened 2018-01-27
NDK r16b has unified headers that buck does not support


sheldonneuberger:
In NDK 16, the sysroot is now /usr/local/share/android-sdk/ndk-bundle/sysroot/usr/include instead of /usr/local/share/android-ndk-r13b/platforms/android-19/arch-arm/usr/include. There are a few other changes that need to be made during compilation as well (noted in the UnifiedHeaders doc here https://android.googlesource.com/platform/ndk/+/ndk-release-r16/docs/UnifiedHeaders.md).

There are other commits in buck to add support for r16b, so I was surprised to see these issues with headers in r16b. Is no one building with buck and r16b yet?

PR #1783, opened 2018-02-23
Add unified headers support

sheldonneuberger:
See #1730. Spec: https://android.googlesource.com/platform/ndk/+/ndk-release-r16/docs/UnifiedHeaders.md

NDK 15 only support android api 14+, so I think there's a number of tests that break because of this. I also haven't added any tests for this PR yet, but it does compile a sample project.

### DIFF
--- a/src/com/facebook/buck/android/toolchain/ndk/impl/NdkCxxPlatforms.java
+++ b/src/com/facebook/buck/android/toolchain/ndk/impl/NdkCxxPlatforms.java
@@ -60,6 +60,7 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableBiMap;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList.Builder;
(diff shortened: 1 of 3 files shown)


src/com/facebook/buck/android/toolchain/ndk/impl/NdkCxxPlatforms.java
```diff
@@ -984,11 +999,29 @@ Path getNdkToolRoot() {
       }
     }
 
+    boolean isUnifiedHeaders() {
+      return unifiedHeaders || ndkMajorVersion >= 16;
```
styurin: I'd rather throw an error rather than silently disable feature especially given that non-unified headers may be wrong and unified headers are generated from the Android platform.

Also, unified headers are supported from v14.

sheldonneuberger:
@styurin reimport this if you can. The __ANDROID_API__ should only be defined when unified headers is on. Forgot to push this a while ago, sorry about that.

styurin:
I'm going to update this PR with a bit more logic that verifies that configuration makes sense with the current NDK version and adding some tests.

PR #1785, opened 2018-02-27
Add tests for modular swift libraries

robbertvanginkel:
These tests currently fail, but should serve as a reference implementation for what is expected to work of modular `apple_libraries`. I'm working towards all the PR's that will make these work and will ping this PR as soon as it can be merged and expected to pass.


robbertvanginkel:
Tests included with code in #1874 #1875

PR #1788, opened 2018-03-01
Add FrameworkPaths from dependencies to SwiftCompile

robbertvanginkel:
The dependency flags passed to SwiftCompile by the `AppleLibraryDescriptionSwiftEnhancer` currently only include "include" flags. This is not sufficient as some rules export frameworks as part of the rule, and the framework search paths for those need to be visible to use the framework.

This change adds the FrameworkPaths to the PreprocessorFlags that are passed when creating the swiftcompile rule. Those are included in the compile command in `getSwiftIncludeArgs`, but unfortunately this prefixes them with `-Xcc` to only make them visible to the clang importer. The frameworks search paths need to be seen visible to both the clang and swift compiler, which is why the paths are also added without `-Xcc`. This duplication is unfortunate but does not cause problems.

### DIFF
--- a/src/com/facebook/buck/apple/AppleLibraryDescriptionSwiftEnhancer.java
+++ b/src/com/facebook/buck/apple/AppleLibraryDescriptionSwiftEnhancer.java
@@ -83,6 +83,7 @@ public class AppleLibraryDescriptionSwiftEnhancer {
 
     PreprocessorFlags.Builder flagsBuilder = PreprocessorFlags.builder();
     inputs.forEach(input -> flagsBuilder.addAllIncludes(input.getIncludes()));
+    inputs.forEach(input -> flagsBuilder.addAllFrameworkPaths(input.getFrameworks()));
     PreprocessorFlags preprocessorFlags = flagsBuilder.build();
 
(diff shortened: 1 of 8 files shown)


test/com/facebook/buck/apple/testdata/prebuilt_apple_framework_static_swift/make_prebuiltlib.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+#
+# Helper file to regenerate the prebuilt framework for different swift versions
```
milend: @robbertvanginkel Can we not create a prebuilt Obj-C framework so that there's no need to recompile as Swift versions change? That would still be able to test whether the framework path is passed to the Swift compiler, right?

robbertvanginkel: Yes, very true and a much better idea. I guess I was so deep in the swift stuff I didn't even consider this. I'll make the change.

PR #1794, opened 2018-03-05
Upgrade NuProcess library.

brettwooldridge:
NuProcess v1.1.3 contains changes for Linux that massively reduce the amount of garbage generated during stream handling of spawned processes.

Before (< v.1.1.3):
```
Iteration   4: 14.460 ops/s
                 ·gc.alloc.rate:      6.760 MB/sec
                 ·gc.alloc.rate.norm: 693246.857 B/op
                 ·gc.count:           ≈ 0 counts
                 ·threads.alive:      11.000 threads
                 ·threads.daemon:     10.000 threads
                 ·threads.started:    11.000 threads
```
After (v1.1.3+):
```
Iteration   4: 21.312 ops/s
                 ·gc.alloc.rate:      0.656 MB/sec
                 ·gc.alloc.rate.norm: 52814.261 B/op
                 ·gc.count:           ≈ 0 counts
                 ·threads.alive:      11.000 threads
                 ·threads.daemon:     10.000 threads
                 ·threads.started:    11.000 threads
```


kageiit:
Need to also update `build.xml`

brettwooldridge:
@kageiit Done.

kageiit:

styurin:
I got some windows-related failures in out production CI and didn't have time to take a closer look. I'll try to take a look next week.

brettwooldridge:
I’m hoping that is an unrelated failure; as far as I know, the Windows code path was completely untouched.

brettwooldridge:
@styurin @kageiit Any update on this pull request status?

styurin:
Still investigating, I hope we'll find the root cause later this week

brettwooldridge:
@styurin Bump.

styurin:
It looks like it's failing because of 505511d41647a04be21bbec00c8dc22fe87628c2. I'll follow up with the author of that commit.

styurin:
https://github.com/brettwooldridge/NuProcess/commit/505511d41647a04be21bbec00c8dc22fe87628c2

brettwooldridge:
@styurin Ok, seeking some advice here.  Here is what I found.  Buck's Windows tasks seem to have been working primarily by luck.

1/ Buck has code in ``ListeningProcessExecutor`` that [clears the environment](https://github.com/facebook/buck/blob/master/src/com/facebook/buck/util/ListeningProcessExecutor.java#L240-L243)  set in ``NuProcessBuilder`` when the ``ProcessExecutorParams`` has its own environment defined.  Then, sets the environment from ``ProcessExecutorParams``.

2/ Prior to the referenced change, NuProcess [*always* injected the environment](https://github.com/brettwooldridge/NuProcess/pull/62/files#diff-196ca43ace2ab31173860ff10435d412L644) into the launch of the spawned task -- even if the environment has been cleared.  This was not by design, and did not match either the documented behavior nor the behavior on Linux or MacOS -- basically a bug.

3/ Separately, and by default, on all platforms, NuProcess [always seeds the environment] when the ``NuProcessBuilder`` [is constructed](https://github.com/brettwooldridge/NuProcess/blob/master/src/main/java/com/zaxxer/nuprocess/NuProcessBuilder.java#L130).  This behavior remains unchanged, and mirrors that of Java's ``ProcessBuilder``.

So, after the fix to NuProcess, removing the injection of the environment, the Buck Windows build started to fail.

This implies to me that some, or possibly all, Windows tasks are not capable of executing based solely on the environment specified in their ``ProcessExecutorParams`` objects.  They were succeeding because NuProcess was re-injecting the environment after it was cleared by Buck.

The question is, what should be done?  NuProcess's behavior is now correct.  So either ``ListeningProcessExecutor`` needs to *not* clear the environment in the case of Windows -- basically preserving *exactly* the previous behavior, or somehow (someone?) needs to determine what the correct environment for Windows is and ensure that the ``ProcessExecutorParams`` instances passed in contain that environment.

The later feels more "correct", but I'm not sure I'd know where to start, since I have no familarity with any of the Windows tasks in Buck.  The former, again, preserves existing behavior exactly, but without correcting what appears to be a larger issue.

Or *is it* a larger issue?  Maybe "merging" the environment *is* the correct behavior in the end.  But that's not my call.

styurin:
@ilya-klyuchnikov

ilya-klyuchnikov:
Basically, I agree with @brettwooldridge that now the behavior of NuProcess on windows is correct and is aligned with mac and linux. I will look into buck on windows problems soon

brettwooldridge:
Updated pull request to NuProcess v1.2.0.

brettwooldridge:
@ilya-klyuchnikov Do you want me to stub-in a Windows-compatible change to this request that you can remove when your Windows changes are finished?  Something in ``ListeningProcessExecutor. launchProcess()`` like:
```Java
...
240  if (params.getEnvironment().isPresent()) {
241     if (Platform.detect().getType() != Platform.WINDOWS) {  // TODO: remove conditional
242        processBuilder.environment().clear();
243     }
244     processBuilder.environment().putAll(params.getEnvironment().get());
245  }
```

ilya-klyuchnikov:
@brettwooldridge - I am leaving it to @styurin.

.idea/libraries/buck_lib.xml
```diff
@@ -22,6 +21,7 @@
       <root url="jar://$PROJECT_DIR$/third-party/java/jackson/jackson-datatype-jdk8-2.7.8.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/android/ddmlib-25.3.0.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/errorprone/error_prone_annotations-2.0.19.jar!/" />
+      <root url="jar://$PROJECT_DIR$/third-party/java/nuprocess/nuprocess-1.1.3.jar!/" />
```
styurin: This needs to be updated to the latest version

.idea/libraries/buck_lib.xml
```diff
@@ -30,14 +30,14 @@
...
       <root url="jar://$PROJECT_DIR$/third-party/java/nailgun/nailgun-server-0.9.3-SNAPSHOT-sources.jar!/" />
-      <root url="jar://$PROJECT_DIR$/third-party/java/nuprocess/nuprocess-1.1.0-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/oshi/oshi-core-3.3-SNAPSHOT-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/thrift/libthrift-0.10.0-patched-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/slf4j/slf4j-jdk14-1.7.5-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/jimfs/jimfs-1.1-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/android/ddmlib-25.3.0-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/errorprone/error_prone_annotations-2.0.19-sources.jar!/" />
       <root url="jar://$PROJECT_DIR$/third-party/java/sqlite/sqlite-jdbc-3.20.0-sources.jar!/" />
+      <root url="jar://$PROJECT_DIR$/third-party/java/nuprocess/nuprocess-1.1.3-sources.jar!/" />
```
styurin: the same

.idea/libraries/buck_lib.xml
```diff
@@ -8,7 +8,6 @@
       <root url="jar://$PROJECT_DIR$/third-party/java/jna/jna-4.4.0.jar!/" />
```
styurin: Please, update this too

brettwooldridge:
Opps, that was part of the original pull request, I must have rolled it back when I rebased onto the head.

brettwooldridge:
Wait, what?  My [master](https://github.com/brettwooldridge/buck/blob/master/build.xml#L124) branch shows this as updated to v4.5.1

.idea/libraries/buck_lib.xml
styurin: This is still pointing to the old version

styurin:
Yeah, but PR still doesn't contain that change for some reason.

brettwooldridge:
@styurin Must have been some kind of delay on github, it now shows correctly.

.idea/libraries/buck_lib.xml
brettwooldridge: I didn’t realize pull requests needed to update IDE files.  I’ll be happy to update it though.

brettwooldridge:
I am curious as well.  Is it related to the [Windows issues](https://github.com/facebook/buck/pull/1794#issuecomment-385242251) that were uncovered?  If this pull request was merged without those being resolved, I can understand.  But if not, again, I am curious as to any issue encountered (and why NuProcess wasn't notified or this pull request annotated).

cjhopman:
@brettwooldridge We found some issues with alignment on some platforms. I think it was an easy fix. @styurin should know more (and I think he reached out to you directly on slack).

brettwooldridge:
@styurin Do you know what the "alignment" issues that @cjhopman refers to were?

@cjhopman When you say, "I think it was an easy fix", what are you referring to?  An easy fix to NuProcess?  An easy fix to Buck?

cjhopman:
@brettwooldridge we got reports that https://github.com/brettwooldridge/NuProcess/commit/4989a9edc2a8f0dbed19bc03b9dfaa6a0963067d#diff-9973b36943d6182ee7db84c55a8cf709R49 broke on aarch64. They said that nuprocess's own tests hang after that change.

The report indicated that the nuprocess epoll_data structure is 8 bytes, but on aarch64 it should be 16 bytes. 

Also, they indicated that changing the alignment back to ALIGN_GNUC resolved the issue.

I can reach out to them if you need more information.

bturner:
Interesting that aarch64 would need a 16 byte structure, because `ALIGN_GNUC` on x86-64 is wrong.

`eventpoll.h` on x86-64, at least the distributions I have handy (Mint 19, Fedora 28, CentOS 7), looks something like this:

```c
/* 
 * On x86-64 make the 64bit structure have the same alignment as the
 * 32bit structure. This makes 32bit emulation easier.
 *
 * UML/x86_64 needs the same packing as x86_64
 */
#ifdef __x86_64__
#define EPOLL_PACKED __attribute__((packed))
#else
#define EPOLL_PACKED
#endif

struct epoll_event {
        __u32 events;
        __u64 data;
} EPOLL_PACKED;
```

The packing means the struct is _12_ bytes, even on 64-bit, and requires `ALIGN_NONE`. Here's a simple test:

```c
#include <stdio.h>
#include <linux/eventpoll.h>

int main(int argc, char *argv[]) {
	printf("sizeof(epoll_event) = %ld\n", sizeof(struct epoll_event));

	return 0;
}
```

With the following output:

```
bturner@auvry:~$ uname -a
Linux auvry 4.15.0-20-generic #21-Ubuntu SMP Tue Apr 24 06:16:15 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
bturner@auvry:~$ file a.out
a.out: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, BuildID[sha1]=40a6f0bfb5eb1904c2c7951faa2068b87b6680cd, not stripped
bturner@auvry:~$ ./a.out
sizeof(epoll_event) = 12
```

(The output is similar on CentOS and Fedora)

I don't have an ARM64 machine, or a cross-compile environment, where I can test. As [discussed here](https://github.com/brettwooldridge/NuProcess/commit/4989a9edc2a8f0dbed19bc03b9dfaa6a0963067d#commitcomment-29710045), though, it may be that the most expeditious way to move forward, so that Buck can get the JNA improvements made in newer versions of NuProcess, is to switch the library back to `ALIGN_GNUC`, at least for the moment, because the library doesn't currently ever pass an array larger than 1 to `epoll_wait`. If it did, though, then `ALIGN_GNUC` fails on x86-64 Linux.

PR #1797, opened 2018-03-08
Convert integer entries to hex decimal entries during RDotTxtEntry initialization.

raviagarwal7:
__Problem:__
Android attributes are being populated to R.java created by mini aapt like `android_font`. This causes robolectric test failures. ([fails to construct Robolectric resource table](https://github.com/robolectric/robolectric/blob/b019ccaad2ab76a632306956c71c4bcdc730e95b/resources/src/main/java/org/robolectric/res/PackageResourceTable.java#L109))
```
java.lang.IllegalArgumentException: Incompatible package for com.uber:attr/android_font with resId 1 to ResourceIndex with packageIdentifier 7
    at org.robolectric.res.PackageResourceTable.addResource(PackageResourceTable.java:109)
    at org.robolectric.res.ResourceTableFactory.addRClassValues(ResourceTableFactory.java:71)
    at org.robolectric.res.ResourceTableFactory.lambda$newResourceTable$1(ResourceTableFactory.java:45)
    at org.robolectric.util.PerfStatsCollector.measure(PerfStatsCollector.java:50)
    at org.robolectric.res.ResourceTableFactory.newResourceTable(ResourceTableFactory.java:38)
    at org.robolectric.res.ResourceMerger.buildResourceTable(ResourceMerger.java:20)
    at org.robolectric.RobolectricTestRunner.getAppResourceTable(RobolectricTestRunner.java:529)
    ...
```    

__RootCause:__
This is happening because a third party library is including `android_font` in its R.txt with an integer value. The attribute is then included in the dependee's R.java generated by miniaapt. 

__Solution:__
Don't include any attributes which starts with `"0x01"`, since those should come from the android framework. Also need make `RDotTxtEntry` initialization convert integers into hexdecimal strings so that the check is effective.

This only affects R.java files generated for robolectric tests and intermediatray libraries.

### DIFF
--- a/src/com/facebook/buck/android/MergeAndroidResourcesStep.java
+++ b/src/com/facebook/buck/android/MergeAndroidResourcesStep.java
@@ -474,38 +474,49 @@ public class MergeAndroidResourcesStep implements Step {
           }
           resource = resource.copyWithNewIdValue(finalIds.get(resource));
 
-        } else if (useOldStyleableFormat && resource.idValue.startsWith("0x7f")) {
-          Preconditions.checkNotNull(enumerator);
-          resource = resource.copyWithNewIdValue(String.format("0x%08x", enumerator.next()));
+        } else if (useOldStyleableFormat) {
+          if (resource.idValue.startsWith("0x7f")) {
+            Preconditions.checkNotNull(enumerator);
+            resource = resource.copyWithNewIdValue(String.format("0x%08x", enumerator.next()));
+          }
+        } else {
+          if (resourceToIdValuesMap.containsKey(resource)) {
+            resource = resourceToIdValuesMap.get(resource);
 
-        } else if (useOldStyleableFormat) { // NOPMD  more readable this way, IMO.
-          // Nothing extra to do in this case.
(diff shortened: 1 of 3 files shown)


kageiit:
@styurin can you PTAL?

styurin:
Could you update the description and put the reason why you need this change first and then list the changes. Right now it's not clear what the phrase that starts with "This is needed to prevent framework resources" is related to.

raviagarwal7:
@styurin -- updated the description with the actual changes, let me know if you have other queries.

styurin:
Looks like we have some miscommunication here.

> Convert integer entries to hex decimal entries during RDotTxtEntry initialization

Why do we need this change at all?

Could you use the following format in the description:

State the problem. Describe the root cause. Describe the solution and provide any details about implementation which are not clear from the source code and which would help with the review.

No need to include source code in the description.

raviagarwal7:
@styurin, updated the description with the suggested format.

styurin:
thanks a lot! It's much better now. Do you mind adding integration tests for robolectric_test?

PR #1445, opened 2017-08-01
[IntelliJ plugin] Buck run configuration

zhan-xiong:
Similar to the "Buck test" configuration in the plugin, this adds a "Buck run" configuration to run buck targets directly in IntelliJ.

PR #1817, opened 2018-03-23
Add run support to buck intellij plugin from running applications from green arrow.

clonetwin26:
This differential adds support for the run button with the buck build system through intellij. With this change you can now run main methods from the green arrow that appears in the intellij gutters or even the debugger.

Running main:
screen shot 2018-03-22 at 9 35 17 pm
screen shot 2018-03-22 at 9 35 24 pm
screen shot 2018-03-22 at 9 35 56 pm

Running with error:
screen shot 2018-03-23 at 9 57 22 am
screen shot 2018-03-23 at 9 57 37 am
screen shot 2018-03-23 at 9 58 07 am
screen shot 2018-04-11 at 11 10 34 am
screen shot 2018-04-11 at 11 10 44 am

### DIFF
--- a/tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
+++ b/tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
@@ -1,0 +1,182 @@
+package com.facebook.buck.intellij.ideabuck.run;
+
+import com.facebook.buck.intellij.ideabuck.build.BuckBuildCommandHandler;
+import com.facebook.buck.intellij.ideabuck.build.BuckBuildManager;
+import com.facebook.buck.intellij.ideabuck.build.BuckCommand;
+import com.facebook.buck.intellij.ideabuck.build.BuckQueryCommandHandler;
+import com.facebook.buck.intellij.ideabuck.config.BuckModule;
+import com.facebook.buck.intellij.ideabuck.file.BuckFileUtil;
+import com.facebook.buck.intellij.ideabuck.icons.BuckIcons;
+import com.google.common.base.Function;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.intellij.execution.application.ApplicationConfiguration;
+import com.intellij.execution.configurations.RunConfiguration;
+import com.intellij.execution.runners.ExecutionEnvironment;
+import com.intellij.openapi.actionSystem.DataContext;
+import com.intellij.openapi.externalSystem.model.ProjectSystemId;
+import com.intellij.openapi.externalSystem.service.execution.ExternalSystemBeforeRunTask;
+import com.intellij.openapi.externalSystem.service.execution.ExternalSystemBeforeRunTaskProvider;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.Key;
+import com.intellij.openapi.vfs.VirtualFile;
+import java.io.FileReader;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.Icon;
+import java.util.List;
+
+public class BuckBuildBeforeRun extends ExternalSystemBeforeRunTaskProvider {
+
+  private static final Key<ExternalSystemBeforeRunTask> ID = Key.create("Buck.BeforeRunTask");
+  private static final ProjectSystemId SYSTEM_ID = new ProjectSystemId("buck");
+  private List<String> targetResults;
+  static private String tempFilePath;
+
+  public BuckBuildBeforeRun(@NotNull Project project) {
+    super(SYSTEM_ID, project, ID);
+  }
+
+  public void setTempFileOutput(String tempFileOutput) {
+    tempFilePath = tempFileOutput;
+  }
+
+  @Override
+  public String getName() {
+    return "Buck Build";
+  }
+
+  @Override
+  public String getDescription(ExternalSystemBeforeRunTask task) {
+    return "Buck Build";
+  }
+
+  @NotNull
+  @Override
+  public ExternalSystemBeforeRunTask createTask(@NotNull RunConfiguration runConfiguration) {
+    return new ExternalSystemBeforeRunTask(ID, SYSTEM_ID);
+  }
+
+  @Nullable
+  @Override
+  public Icon getIcon() {
+    return BuckIcons.BUCK_TOOL_WINDOW_ICON;
+  }
+
+  @Override
+  public boolean canExecuteTask(
+      RunConfiguration configuration, ExternalSystemBeforeRunTask beforeRunTask
+  ) {
+    if (!(configuration instanceof ApplicationConfiguration)) {
+      return false;
+    }
+    VirtualFile containingFile = ((ApplicationConfiguration) configuration).getMainClass()
+        .getContainingFile().getVirtualFile();
+    VirtualFile buckFile = BuckFileUtil.getBuckFile(containingFile);
+
+    return buckFile != null;
+  }
+
+  @Override
+  public boolean executeTask(
+      final DataContext context,
+      RunConfiguration configuration,
+      ExecutionEnvironment env,
+      ExternalSystemBeforeRunTask beforeRunTask
+  ) {
+
+    Project currentProject = configuration.getProject();
+
+    AtomicBoolean successfulBuild = new AtomicBoolean(false);
+    if (configuration instanceof ApplicationConfiguration) {
+      VirtualFile containingFile = ((ApplicationConfiguration) configuration).getMainClass()
+          .getContainingFile().getVirtualFile();
+
+      VirtualFile buckFile = BuckFileUtil.getBuckFile(containingFile);
+
+      if (buckFile == null) {
+        return false;
+      }
+
+      BuckBuildManager buildManager = BuckBuildManager.getInstance(currentProject);
+      BuckQueryCommandHandler buckQueryCommandHandler =
+          getBuckQueryCommandHandler(currentProject, buckFile);
+      buckQueryCommandHandler.command().addParameter("owner(" + containingFile.getPath() + ")");
+      buildManager.runInCurrentThreadPostEnd(
+          buckQueryCommandHandler,
+          () -> {
+            for (String target : targetResults) {
+              buildTarget(currentProject, successfulBuild, buckFile, buildManager, target);
+            }
+          });
+    }
+    return successfulBuild.get();
+  }
+
+  private void buildTarget(Project currentProject, AtomicBoolean successfulBuild,
+      VirtualFile buckFile, BuckBuildManager buildManager, String target) {
+    BuckBuildCommandHandler buckBuildCommandHandler =
+        new BuckBuildCommandHandler(
+            currentProject,
+            buckFile.getParent(),
+            BuckCommand.BUILD,
+            true
+        );
+    buckBuildCommandHandler.command()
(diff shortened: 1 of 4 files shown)


styurin:
This change looks reasonable, but I'm not sure about practical value of this approach:

- How often developers add new `main` methods? In practice, not very often.
- What if there are multiple targets that own this file?
- What if the file is compiled in a `java_library` and there is no `java_binary` or there are multiple `java_binary` targets?

At the same time this feature modifies UI for all users of the Buck plugin and add some run time cost when a new file is opened (or even switched to, don't know how often IntelliJ would call it.) 

My general recommendation would be to ask before spending time on writing a new feature.

clonetwin26:
Multiple targets owning a file is for sure an issue in this approach. Thank you for taking the time to look through this.

I think the value of main and run, is you can get into the debugger and build system from within the ide, without having to write a test and go through the test runner. This is usually my flow when I want to develop a library, and new up instances and test functionality, before I am ready to write a proper unit test.

We get quite a few people who try to build through intellij run buttons, and are put through some pretty jarring experiences, as the default runner is no where close to actually using the buck build system to build.

I will reach out to you first before committing to develop such a feature, and will think more about the multiple owning targets, I suppose I could just build and add all of the classpaths to the program-runner classpath for all owning targets.

I think if the file is part of a Java library, which is consumed by a Java binary, then the end user would just use the main of the Java binary to run and debug, or add their own main to the library.

kageiit:
Also see https://github.com/facebook/buck/pull/1445

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
```diff
@@ -0,0 +1,170 @@
...
+                public Void apply(@javax.annotation.Nullable List<String> strings) {
+                  if (strings == null
+                      || strings.isEmpty()
+                      || strings.get(0) == null
+                      || strings.get(0).isEmpty()) {
+                    return null;
+                  }
+                  String targetResultCandidate = strings.get(0);
+                  for (String candidate : strings) {
+                    if (!candidate.contains("lint")) {
```
styurin: Why?

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
```diff
@@ -0,0 +1,170 @@
...
+            BuckBuildCommandHandler buckBuildCommandHandler =
+                new BuckBuildCommandHandler(
+                    currentProject,
+                    buckFile.getParent(),
+                    BuckCommand.BUILD,
+                    true
+                );
+            if (targetResult != null) {
+              buckBuildCommandHandler.command()
+                  .addParameters("--deep", targetResult, "--build-report", "buck-out/out.json");
```
styurin: why `--deep`?

Also, I'd recommend avoid writing to buck-out, in this case a temporary folder would just fine.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
```diff
@@ -0,0 +1,170 @@
...
+              buckModule.attach(targetResult);
+              if (targetResult == null) {
+                buildManager.showNoTargetMessage(currentProject);
+                return;
+              }
+              buildManager.runBuckCommandWhileConnectedToBuck(
+                  buckBuildCommandHandler,
+                  "Building target: " + targetResult,
+                  buckModule);
+              while (!buckBuildCommandHandler.isStarted()) {
```
styurin: I'd add a sleep here

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
```diff
@@ -0,0 +1,170 @@
...
+      return false;
+    }
+    VirtualFile containingFile = ((ApplicationConfiguration) configuration).getMainClass().getContainingFile().getVirtualFile();
+    VirtualFile buckFile = BuckFileUtil.getBuckFile(containingFile);
+
+    return buckFile != null;
+  }
+
+  @Override
+  public boolean executeTask(
```
styurin: Could you split this method into smaller methods to show what's actually happening here. Right now it requires reading the whole method.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckRunConfigurationExtension.java
```diff
@@ -0,0 +1,171 @@
...
+              public Void apply(@javax.annotation.Nullable List<String> strings) {
+                if (strings == null
+                    || strings.isEmpty()
+                    || strings.get(0) == null
+                    || strings.get(0).isEmpty()) {
+                  return null;
+                }
+                String targetResultCandidate = strings.get(0);
+                for (String candidate : strings) {
+                  if (!candidate.contains("lint")) {
```
styurin: looks like duplicated logic.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckRunConfigurationExtension.java
```diff
@@ -0,0 +1,171 @@
...
+                    buckFile.getParent(),
+                    BuckCommand.AUDIT,
+                    new FutureCallback<String>() {
+                      @Override
+                      public void onSuccess(@Nullable String result) {
+                        if (result == null
+                            || result.isEmpty()) {
+                          return;
+                        }
+                        String lines[] = result.split("\\r?\\n");
```
styurin: `buck query` can output in json format, can you use it instead of parsing new lines?

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
clonetwin26: This was to filter out our own lint targets, and should take all targets instead.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
clonetwin26: All of the resulting jars from all of the transitive dependencies has to be present to be put on the runtime classpath.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckRunConfigurationExtension.java
clonetwin26: Hard not to duplicate here since it has to pass the target results to other buck commands.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
clonetwin26: This made this much cleaner ty :)

clonetwin26:
You have the right understanding. It also puts all the results of deep build onto the run time classpath. I was unable to do unit tests without degrading ide perf since their configurations don't give virtual files and have to be deduced from a FQN.

styurin:
Sorry I didn't get, why we cannot use `buck run` instead of deep build and running the application using IntelliJ capabilities?

kageiit:
> why we cannot use buck run instead of deep build and running the application using IntelliJ capabilities?

When you have a lot of dependencies, changing a single line of code can take a very long time to build the fat jar which makes the incremental run times pretty bad. `buck run` always builds the fat jar and does not offer a good ux since it can be very slow

clonetwin26:
Not every main method has a binary. Some are purely contained in a Java library.

styurin:
> Not every main method has a binary. Some are purely contained in a Java or android library.

Thinking about that from the Buck plugin for IntelliJ perspective, it's questionable why we need some hacks that emulate `buck run` in this plugin. From Buck perspective the target graph contains all the necessary information (except command line args) that is required to run an application. There might be some options in `java_binary` which is crucial to running an application.

At the same time, IntelliJ provides a way to run an application and if you have a working project it should work by creating a configuration that runs `buck build --deep` before running the application.

So I'm not convinced that this is a right solution, but I'll let @mkillianey to take a look.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckBuildBeforeRun.java
```diff
@@ -0,0 +1,182 @@
...
+      VirtualFile buckFile, BuckBuildManager buildManager, String target) {
+    BuckBuildCommandHandler buckBuildCommandHandler =
+        new BuckBuildCommandHandler(
+            currentProject,
+            buckFile.getParent(),
+            BuckCommand.BUILD,
+            true
+        );
+    buckBuildCommandHandler.command()
+        .addParameters("--deep", target, "--build-report", tempFilePath);
```
mkillianey: Observation of a potential edge case:  if you're using a build server, then building `--deep` may not be sufficient to make your code runnable.  IIUC, asking buck to build `--deep` doesn't require it to build all of your dependencies -- it is permitted to fetch ABIs from the build server, instead, if that is sufficient to build the target you've selected.

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckRunConfigurationExtension.java
```diff
@@ -0,0 +1,207 @@
...
+            if (result == null
+                || result.isEmpty()) {
+              return;
+            }
+            JsonArray classpathArray = new JsonParser().parse(result)
+                .getAsJsonObject().getAsJsonArray(target);
+            if (classpathArray != null) {
+              classpathArray.forEach(
+                  jsonElement ->
+                      additionalClasspath.add(jsonElement.getAsString())
```
mkillianey: As per my other comment, I think that some of these elements won't be generated by `--deep`.

If that thinking is correct, then I'm not sure if there are one or two error conditions:

1.  If the dependency is missing, that's because `--deep` only fetched the ABI from the build server.  This seems easy to detect, and could probably be remedied by repeating a more thorough version of the previous build, like `buck build $(buck query 'deps(original-target)')`?
2.  I'm unclear...is it possible for the dependency to exist, but be stale?  This would happen if buck only rebuilt the ABI and not the actual jar.  (I'm not totally sure what buck's guarantee is for `--deep` builds of java_library targets.)

tools/ideabuck/src/com/facebook/buck/intellij/ideabuck/run/BuckRunConfigurationExtension.java
```diff
@@ -0,0 +1,207 @@
...
+          .getContainingFile().getVirtualFile();
+      buckFile = BuckFileUtil.getBuckFile(containingFile);
+    } else {
+      return;
+    }
+    List<String> additionalClasspath = new ArrayList<>();
+    BuckBuildManager buildManager = BuckBuildManager.getInstance(currentProject);
+    BuckQueryCommandHandler buckQueryCommandHandler =
+        getBuckQueryCommandHandler(currentProject, buckFile);
+    buckQueryCommandHandler.command().addParameter("owner(" + containingFile.getPath() + ")");
```
mkillianey: Should this query also restrict the `buck.type` to one of the `java_*` targets, so that it doesn't apply to arbitrary targets like `python_library` or even `genrule`?

PR #1825, opened 2018-03-30
Fix travis CI build

bobyangyf:
Trying to fix travis CI.


ttsugriy:
looks like your test didn't fail @bobyangyf (https://travis-ci.org/facebook/buck/jobs/360100023#L5159-L5160) You can manually restart a [job](https://travis-ci.org/facebook/buck/jobs/360100023#L5159-L5160) a few times to rule out flakiness, but so far it's very promising :)

PR #1826, opened 2018-03-30
[WorkerProcessPool] Replace busy wait w/ `Thread#join()`

davidaurelio:
These tests were using a busy wait approach in order to avoid timeouts
on CI systems like Travis.

Test threads get into different states across different test runs and systems.

Here, we replace all busy waits for thread states with `Thread#join(500)`.

Given the small workloads of the test threads, this should be enough even for Travis CI.


davidaurelio:
Done, it removed unused imports.

Issue #1830, opened 2018-04-02
Buck is inconsistent in dependency resolution order between java tests and binaries, problematic with duplicates


romanoid:
I'm planning to work on fixing this issue, please comment and advice!

We've noticed issue with Buck resolving dependencies in different order for different (but related targets). 

This is especially dangerous in the case if any duplicates are present in classpath. 

Common reason for it is repository including multiple versions of 3rdparty library different modules need (or authors assume different version is needed), then suddenly one of these modules depends on another. (see [example](https://github.com/romanoid/buck-tinkering/blob/master/3rdparty/dep_example/junit/BUCK))

Buck uses different lookup order for tests and binaries, this may make all the tests pass but binary may fail with "Method not found in production".

Behaviors: 
java_binary: lexicographically earlier target selected first (in case of common version naming, lowest version), sorting happens in [AbstractSourcePathResolver](https://github.com/facebook/buck/blob/4c5b71133112f0d5b4e9179b109f80962cf5bc4b/src/com/facebook/buck/rules/AbstractSourcePathResolver.java#L119) 

Java test: I haven't looked into implementation details yet, but [experiments](https://github.com/romanoid/buck-tinkering/blob/master/deps-experiment/buildables/BUCK) shown postorder DFS with direct children sorted. 

Additionally, if any attempt to trace class path is made with buck query, buck query uses BFS, which is different for both.

Demonstrated in experiment: 
https://github.com/romanoid/buck-tinkering/blob/master/deps-experiment/buildables/BUCK

This uses resources instead of java files since they are lighter and easier to trace, but behavior is same for .class files

Proposed changes: I'd like to update both java_test and java_binary to be consistent in dependency traversal order. 

Options we have (please add if I'm missing some):

1.  Always sort lexicographically: (java_binary of today)
  + This is the consistent and easy to trace option. (sort output of deps query)
  - With duplicate deps, this always "downgrades dep".
  + Change of dep order does not change artifact.
  - Gives no control to owners of individual module to override deps.  
2. Always sort reversed lexicographically: 
  + This is the consistent and easy to trace option. (sort output of deps query)
  + With duplicate deps, this always "upgrades dep".
  + Change of dep order does not change artifact.
  - Gives no control to owners of individual module to override deps.  
3. (pre/post order) DFS with children order equal to declared deps:
  - This is hard to exactly trace since there no command to output the order. 
  - With duplicate deps, this will get one that is closer in the order
[...]

PR #1841, opened 2018-04-10
Adding integration test for java dependency resolution order.

romanoid:
This is integration test for https://github.com/facebook/buck/issues/1830.
When issue is fixed, the test will be updated accordingly.

Meanwhile as well as after the fix, test will indicate any regressions in the dependency conflict resolution order.


romanoid:
Replaced with https://github.com/facebook/buck/pull/1844

Issue #1836, opened 2018-04-08
[Bug]Broken support for android ndk build when deal with arm64


gengjiawen:
If you want to build arm64 with armeabv7a, you have to set  `app_platform = android-21`  above.

Repro repo: https://github.com/gengjiawen/android-jsc/tree/feature/abi_support.

Suggest fix: when deal with arm64, automatically use `android-21`, in other case, use `android-16` or something like that.

sdwilsh:
Doesn't this have other implications?  I have concerns if the user specifies that they want `android-16`, but then Buck decides they really want something different.

It's quite possible that we should just bump up the default from `android-16` (which is pretty old), but I think morphing what was requested to something else is an approach that doesn't fit with how Buck generally works.

sdwilsh:
You are still doing what I said, but for a small subset of options.  It would likely be better to provide a descriptive error message to the user telling them they need to be on `android-21` or newer.  Otherwise, if someone builds their application with Buck, they'll ship their native code with different `android` targets, which could introduce other non-obvious problems.

LegNeato:
RN doesn't want to bump wholesale to 21. I think the risk is acceptable as Google is forcing it.

sdwilsh:
I know it says `Owner` next to my name, but that's just because I am an admin to the facebook organization here on GitHub.  My comments here aren't authoritative, and someone who actually works on Buck should make the call here.

gengjiawen:
Let's put it in another way. If without buck, we build native module using Android.mk and Application.mk or cmake and gradle to get arm64 support, do I have to set app_modules to android-21 ?

Nope.

The reason arm64 need android-21 to build is because with 21 we have arm64 support.

An example you can see is [fresco](https://github.com/facebook/fresco). It supports Android 2.3 (Gingerbread) and later with arm64 support.

gengjiawen:
It can be a solution. A fix to this is all I needed.

gengjiawen:
@styurin Any action on this issue ?

styurin:
As I said in #1843 we shouldn't change the configuration and instead let user know that the configuration is incorrect.

gengjiawen:
Did you have looked the discussion in this issue ? I think the configuration is correct. It's buck which handle the build wrong.

styurin:
Again, we are not going to have any kind of magic logic in Buck that would change the configuration from what's defined in `.buckconfig`. If this configuration is not flexible enough you can add additional options, for example, platform level by ABI.

[...]

PR #1843, opened 2018-04-10
Fix android ndk 64bit cpu android platform bug

gengjiawen:
Fix issue https://github.com/facebook/buck/issues/1836.


styurin:
Looking at this change it seems that a binary will be built using different Android platforms depending on ABI and the setting from the configuration will be overridden if it's too low.

Instead of adding such magic to the toolchain configuration logic it's safer to fail fast by showing an error and letting user fix the configuration. This might be a bit challenging though since NDK toolchain is configured eagerly.

gengjiawen:
@styurin Please see this issue page https://github.com/facebook/buck/issues/1836 for detailed discussion, it's not a user config bug. It's a bug in buck.

styurin:
fe15dae213ac0e64886157cd513b043269f68edb added ability to provide target platform by ABI

PR #1844, opened 2018-04-10
Fix java test classpath to have the same ordering as java binary

romanoid:
Fixing https://github.com/facebook/buck/issues/1830 using option # 1 described there. 

This replaces https://github.com/facebook/buck/pull/1841

PR #1845, opened 2018-04-11
Cleaned up multiple copies of junit/hamcrest jars.

romanoid:
Replaced it with buck cell usage, similar to as requested by @styurin in https://github.com/facebook/buck/pull/1844


ttsugriy:
please format java classes using `google-java-format` to fix Travis CI build. Thanks for cleaning this up!

romanoid:
done

PR #1855, opened 2018-04-16
Cleaned up multiple copies of junit/hamcrest jars.

romanoid:
Follow up to https://github.com/facebook/buck/pull/1845
Cleaned up excess jars in more targets


test/com/facebook/buck/cli/DaemonIntegrationTest.java
```diff
@@ -93,7 +93,7 @@ public void whenConcurrentCommandExecutedThenSecondCommandFails()
       throws IOException, InterruptedException, ExecutionException {
 
     ProjectWorkspace workspace =
-        TestDataHelper.createProjectWorkspaceForScenario(this, "exclusive_execution", tmp);
+        TestDataHelper.createProjectWorkspaceForScenario(this, "exclusive_execution", tmp, true);
```
styurin: I wonder if we can make it default.

romanoid: I don't see why not, I can clean this up (since I've already started)

PR #1857, opened 2018-04-18
Kotlinc Step: Don't add generated code to javac sources if we aren't generating code.

cwoodwar6:
When using `language= 'kotlin'` in an `android_library` rule without any kotlin source code, it jumps straight into passing the non-existent generated source zip to javac. This results in `__src_debug_gen_sources__/generated.src.zip (No such file or directory)`.


kageiit:
@styurin any updates on landing this PR?

PR #1859, opened 2018-04-19
Make tracking of javac phase events configurable

### LINKED ISSUE #1839: ChromeTraceBuildListener leaking memory?
We did a Buck build with many targets that starting running out of memory, and a heap dump showed `ChromeTraceBuildListener` holding on to more than 1.5GB of memory.  Looking closer, it seemed that its [`outputExecutor`](https://github.com/facebook/buck/blob/master/src/com/facebook/buck/event/listener/ChromeTraceBuildListener.java#L926) had accumulated a huge number of tasks in its queue that were holding on to memory.  Running the build with `--config log.chrome_trace_generation=false` fixed the issue.  It seems there may be a leak here that should be fixed.

kageiit:
Fixes #1839


src/com/facebook/buck/jvm/java/AbstractCompilerParameters.java
```diff
@@ -69,6 +69,11 @@ public boolean shouldTrackClassUsage() {
     return false;
   }
 
+  @Value.Default
+  public boolean shouldTrackJavacPhaseEvents() {
+    return false;
```
styurin: Where this option is set?

kageiit: in .buckconfig
```
[java]
  track_javac_phase_events = true
```

PR #1870, opened 2018-04-27
Use a non blocking executor for dir artifact cleanup

### LINKED ISSUE #1842: buck daemon hangs trying to cleanup dir artifact cache
After a buck command, the daemon hangs for a period of time before returning control of the shell back to the user. Yourkit says there is a potential deadlock

```
Potential deadlock: frozen threads found

It seems that the following threads have not changed their stack for more than 10 seconds.
These threads are possibly (but not necessarily!) in a deadlock or hung.

NGSession 1: com.facebook.buck.cli.Main (NGCommunicator reader) <--- Frozen for at least 21s
com.martiansoftware.nailgun.NGUnixDomainSocketLibrary.read(int, ByteBuffer, int) NGUnixDomainSocketLibrary.java (native)
com.martiansoftware.nailgun.NGUnixDomainSocket$NGUnixDomainSocketInputStream.doRead(ByteBuffer) NGUnixDomainSocket.java:125
com.martiansoftware.nailgun.NGUnixDomainSocket$NGUnixDomainSocketInputStream.read() NGUnixDomainSocket.java:98
java.io.DataInputStream.readInt() DataInputStream.java:387
com.martiansoftware.nailgun.NGCommunicator.readChunk() NGCommunicator.java:327
com.martiansoftware.nailgun.NGCommunicator.lambda$null$0() NGCommunicator.java:124
com.martiansoftware.nailgun.NGCommunicator$$Lambda$12.call()
java.util.concurrent.FutureTask.run() FutureTask.java:266
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor$Worker) ThreadPoolExecutor.java:1149
java.util.concurrent.ThreadPoolExecutor$Worker.run() ThreadPoolExecutor.java:624
java.lang.Thread.run() Thread.java:748
```

cc @msridhar @sbalabanov @styurin

kageiit:
- Fixes #1842

Verified by having a very large artifact cache dir on local disk that was cleaned without the the daemon killing the executor service prematurely

### DIFF
--- a/src/com/facebook/buck/cli/Main.java
+++ b/src/com/facebook/buck/cli/Main.java
@@ -823,7 +823,16 @@ public final class Main {
               GlobalStateManager.singleton()
                   .setupLoggers(invocationInfo, console.getStdErr(), stdErr, verbosity);
           DefaultBuckEventBus buildEventBus = new DefaultBuckEventBus(clock, buildId);
-          ) {
+          // We use a new executor service beyond client connection lifetime since it can take a
+          // long time
+          // to stat and cleanup large disk artifact cache directories
+          // See https://github.com/facebook/buck/issues/1842
+          ThrowingCloseableWrapper<ExecutorService, InterruptedException>
+              dirArtifactExecutorService =
+                  getExecutorWrapper(
+                      MostExecutors.newSingleThreadExecutor("Dir Artifact"),
+                      "Dir Artifact",
+                      EXECUTOR_SERVICES_TIMEOUT_SECONDS); ) {
 
         try (ThrowingCloseableWrapper<ExecutorService, InterruptedException> diskIoExecutorService =
                 getExecutorWrapper(
@@ -932,7 +941,7 @@ public final class Main {
                     httpWriteExecutorService.get(),
                     httpFetchExecutorService.get(),
                     stampedeSyncBuildHttpFetchExecutorService.get(),
-                    diskIoExecutorService.get());
+                    dirArtifactExecutorService.get());
 
             // Once command completes it should be safe to not wait for executors and other stateful
             // objects to terminate and release semaphore right away. It will help to retry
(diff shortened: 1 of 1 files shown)


kageiit:
cc @sbalabanov

sbalabanov:
I think you can just move diskIO executor service to here: https://github.com/kageiit/buck/blob/49fce8e6b791bdbf76a59b210852c2b67a68e887/src/com/facebook/buck/cli/Main.java#L829
Please do not forget comments explaning why we let it go beyond client connection lifetime.

kageiit:
@sbalabanov Not sure what you mean `MostExecutors.newSingleThreadExecutor("Dir Artifact")` is not closeable. I cannot move it to the try block

```
ExecutorService ne = MostExecutors.newSingleThreadExecutor("Dir Artifact");
    [javac] 	^^^^^^^^^^^^^^^
    [javac] The resource type ExecutorService does not implement java.lang.AutoCloseable
```

kageiit:
@sbalabanov Updated. Please take a look

kageiit:
@sbalabanov Ping

cc @styurin

src/com/facebook/buck/cli/Main.java
```diff
@@ -830,6 +830,12 @@ public ExitCode runMainWithExitCode(
                     MostExecutors.newSingleThreadExecutor("Disk I/O"),
                     "Disk IO",
                     DISK_IO_STATS_TIMEOUT_SECONDS);
+            // We use a new executor service that we do not wait on since it can take a long time
+            // to stat and cleanup large disk artifact cache directories
+            // See https://github.com/facebook/buck/issues/1842
+            ThrowingCloseableWrapper<ExecutorService, InterruptedException> dirArtifactExecutorService =
+                ThrowingCloseableWrapper.of(MostExecutors.newSingleThreadExecutor("Dir Artifact"),
+                  executor -> executor.shutdown());
```
sbalabanov: Use getExecutorWrapper() function to get a safe wrapper that shutdowns safely

styurin:
What about a possibility of a next build trying to access an artifact that is being removed by a cleaner from the previous execution?

src/com/facebook/buck/cli/Main.java
kageiit: Cant use `getExecutorWrapper()` as it does a blocking await. Thats exactly the situation we are trying to avoid. Bumping timeout just makes the buck process hang longer before giving control back to the user on the terminal

PR #1874, opened 2018-05-03
Add modulemap to headers for modular apple_library

robbertvanginkel:
When an `apple_library` is marked as modular, add a modulemap to the generated headersymlinktree so other rules can import this library modularly.

Modules are always platform specific so we need to only have one platform specific headersymlinktree with a modulemap. We cannot forward the metadata to a CxxLibraryDescription as it will create two trees (platform specific and platform agnostic). This PR makes sure that if a library is modular, AppleLibraryDescription doesn't forward the `CxxLibraryDescription.MetadataType.CXX_PREPROCESSOR_INPUT` to a CxxLibraryDelegate but instead creates it itself.

### DIFF
--- a/src/com/facebook/buck/apple/AppleLibraryDescription.java
+++ b/src/com/facebook/buck/apple/AppleLibraryDescription.java
@@ -306,7 +306,7 @@ public class AppleLibraryDescription
                         () ->
                             AppleLibraryDescriptionSwiftEnhancer
                                 .getPreprocessorInputsForAppleLibrary(
-                                    buildTarget, graphBuilder, cxxPlatform));
+                                    buildTarget, graphBuilder, cxxPlatform, args));
 
             return Optional.of(
                 AppleLibraryDescriptionSwiftEnhancer.createSwiftCompileRule(
@@ -616,7 +616,7 @@ public class AppleLibraryDescription
         && headerMode.isPresent()
         && headerMode.get().equals(HeaderMode.SYMLINK_TREE_WITH_MODULEMAP)) {
       return createExportedModuleSymlinkTreeBuildRule(
-          buildTarget, context.getProjectFilesystem(), graphBuilder, args);
+          buildTarget, context.getProjectFilesystem(), graphBuilder, platform.get(), args);
     }
 
     return graphBuilder.computeIfAbsent(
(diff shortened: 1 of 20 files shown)


src/com/facebook/buck/apple/AppleLibraryDescription.java
```diff
@@ -641,13 +687,27 @@ private boolean shouldWrapIntoDebuggableBinary(BuildTarget buildTarget, BuildRul
...
     if (CxxLibraryDescription.METADATA_TYPE.containsAnyOf(buildTarget.getFlavors())) {
-      CxxLibraryDescriptionArg.Builder delegateArg = CxxLibraryDescriptionArg.builder().from(args);
-      AppleDescriptions.populateCxxLibraryDescriptionArg(
-          pathResolver, delegateArg, args, buildTarget);
-      return cxxLibraryMetadataFactory.createMetadata(
-          buildTarget, resolver, cellRoots, delegateArg.build(), metadataClass);
+      // Modules are always platform specific so we need to only have one platform specific
+      // headersymlinktree with a modulemap. We cannot forward the metadata to a cxxlibrary
+      // description as it makes an optimization of having multiple header symlinktrees (platform
+      // specific and general). This also gives us more controll over exposing the correct swift
```
milend: @robbertvanginkel `controll` -> `control`

src/com/facebook/buck/apple/AppleLibraryDescription.java
```diff
@@ -769,6 +828,68 @@ private boolean shouldWrapIntoDebuggableBinary(BuildTarget buildTarget, BuildRul
...
+            .orElseThrow(IllegalArgumentException::new);
+    Entry<Flavor, HeaderVisibility> visibility =
+        CxxLibraryDescription.HEADER_VISIBILITY
+            .getFlavorAndValue(buildTarget)
+            .orElseThrow(IllegalArgumentException::new);
+    BuildTarget baseTarget =
+        buildTarget.withoutFlavors(
+            cxxMetaDataType.getKey(), platform.getKey(), visibility.getKey());
+
+    CxxPreprocessorInput.Builder cxxPreprocessorInputBuilder = CxxPreprocessorInput.builder();
```
milend: @robbertvanginkel Seems like the code from 851-862 is duplicated from `CxxLibraryMetadataFactory`. Can we just extract and expose it instead?

src/com/facebook/buck/apple/AppleLibraryDescription.java
robbertvanginkel: Fixed

src/com/facebook/buck/apple/AppleLibraryDescription.java
robbertvanginkel: Refactored into a static method on `CxxLibraryMetadataFactory`

PR #1875, opened 2018-05-03
Allow modular mixed apple_library

robbertvanginkel:
This pr adds the functionality to support modular mixed objective-c/swift librariesusing the `-import-underlying-module` flag. The new swift metadata type `SWIFT_UNDERLYING_MODULE` refers to a build rule for a headersymlinktree that exposes the public Objective-C headers through a module with the same name, so that the swift compiler can read this using the import-underlying-module flag when compiling the swift sources for the library.

Based on having support for modular libraries in #1874.


shepting:
@ttsugriy Anything holding up this PR?

milend:
@shepting The PR has been on my plate for a while but I've been overwhelmed with other work, I've finally made some time to go through it.

milend:
@robbertvanginkel LGTM, should be good after a rebase together with #1874.

PR #1876, opened 2018-05-04
Tolerate hyphens in go file names

linzhp:
There may be hyphens in Go file names. They need to be replaced before using the file names as variable names, otherwise the hyphens will be treated as minus signs by Go compiler

@ttsugriy @styurin Please take a look

cc @VitalyArbuzov

### DIFF
--- a/test/com/facebook/buck/features/go/testdata/go_test/BUCK.fixture
+++ b/test/com/facebook/buck/features/go/testdata/go_test/BUCK.fixture
@@ -66,3 +66,13 @@ go_test(
     name = "subtests",
     srcs = ["subtests.go"],
 )
+
+go_test(
+    name = "test-hyphen",
+    package_name = "lib",
+    srcs = [
+        "power-driven.go",
+        "power-driven_test.go",
+    ],
+    coverage_mode='set',
+)
--- a/test/com/facebook/buck/features/go/testdata/go_test/BUCK.fixture
+++ b/test/com/facebook/buck/features/go/testdata/go_test/BUCK.fixture
@@ -74,5 +74,5 @@ go_test(
         "power-driven.go",
         "power-driven_test.go",
     ],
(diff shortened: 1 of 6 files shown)


styurin:
> There may be hyphens in Go file names.

What does go guidelines say about such filenames?

What would happen if you specify "file-1.go" and "file_1.go" in `srcs`?

IMO we should fail fast and do not do any magic under inside Buck. Let users decide how to properly name files and don't do that for the,

linzhp:
Go guideline doesn't talk about file name conventions in general, besides a few special prefixes and suffixes treated differently by go toolchain. So people are free to use any file names, and that doesn't cause problems for go toolchain. It only becomes a problem when buck uses the file names as coverage variable names.

Having hyphens in file names are much more common than having two files in the same directory only different by hyphens vs underscores. The latter case is clearly a poor naming conversion, and the former one is not so much. Although this fix doesn't handle the case you mentioned, it at least makes Buck more tolerant to Go file names

linzhp:
I updated the PR to use the hash of the file name as the variable name, similar to what "go test" does. Now the hyphen and underscore problem is avoided completely. @styurin Please take another look

linzhp:
Fixed @ttsugriy

test/com/facebook/buck/features/go/testdata/go_test/BUCK.fixture
```diff
@@ -66,3 +66,13 @@ go_test(
...
 )
+
+go_test(
+    name = "test-hyphen",
+    package_name = "lib",
+    srcs = [
+        "power-driven.go",
+        "power-driven_test.go",
+    ],
+    coverage_mode='set',
```
ttsugriy: this should be `coverage_mode = "set",` according to `buildifier`. Please fix.

PR #1878, opened 2018-05-08
Fixed `buck publish` to stop incorrectly packaging direct dep srcs

romanoid:
The problem was the difference between how TraversedDeps is initialized for sources jar versus regular jar.

For the sources jar java_library Build rule is not available yet and set of all classpath dependencies was used as root and auto-included in sources jar.

This led to sources jar including superset of what it is supposed to contain.


styurin:
Can you add a test for that?

romanoid:
Added unit test, let's see if I haven't broke build on the way.

Please let me know if you think this needs integration test too.

PR #1879, opened 2018-05-08
Cleaned up `buck publish`:

romanoid:
1. Added validation case for where user both passes the repo and toMavenCentral (previously toMavenCentral was simply ignored)
2. Replaced implicit default of maven Central in Publisher with explicit logic.
3. Switched temp folder location from root of the repo to buck-out/tmp/publish to avoid adding clutter to the repo root.


src/com/facebook/buck/cli/PublishCommand.java
```diff
@@ -110,6 +112,16 @@ public ExitCode runWithoutHelp(CommandRunnerParams params)
       throws IOException, InterruptedException {
 
     // Input validation
+    if (remoteRepo != null && toMavenCentral) {
```
styurin: Can you add tests for this case?

romanoid: done

PR #1888, opened 2018-05-14
Remove stdout/stderr headers when there is nothing in it

linzhp:
When a test failed and there nothing in stdout/stderror, buck should not print out headers with empty body like the one below:

```
====STANDARD OUT====

====STANDARD ERR====

```
This PR fixed that.

### DIFF
--- a/src/com/facebook/buck/event/listener/TestResultFormatter.java
+++ b/src/com/facebook/buck/event/listener/TestResultFormatter.java
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.TimeZone;
+import org.testng.util.Strings;
 
 public class TestResultFormatter {
 
@@ -218,16 +219,19 @@ public class TestResultFormatter {
       }
     }
 
-    if (summaryVerbosity.getIncludeStdOut()
-        && testResult.getStdOut() != null
-        && testResult.getStdOut().length() > 0) {
+    if (summaryVerbosity.getIncludeStdOut() && Strings.isNullOrEmpty(testResult.getStdOut())) {
       addTo.add("====STANDARD OUT====", testResult.getStdOut());
     }
 
-    if (summaryVerbosity.getIncludeStdErr()
-        && testResult.getStdErr() != null
-        && testResult.getStdErr().length() > 0) {
+    if (summaryVerbosity.getIncludeStdErr() && Strings.isNullOrEmpty(testResult.getStdErr())) {
       addTo.add("====STANDARD ERR====", testResult.getStdErr());
+    }
+
+    if (Strings.isNullOrEmpty(testResult.getMessage())
+        && Strings.isNullOrEmpty(testResult.getStacktrace())
+        && Strings.isNullOrEmpty(testResult.getStdOut())
+        && Strings.isNullOrEmpty(testResult.getStdErr())) {
+      addTo.add("Test did not produce any output.");
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/com/facebook/buck/event/listener/TestResultFormatter.java
```diff
@@ -218,11 +218,15 @@ public void reportResultSummary(
...
+    if (summaryVerbosity.getIncludeStdOut()
+        && testResult.getStdOut() != null
+        && testResult.getStdOut().length() > 0) {
       addTo.add("====STANDARD OUT====", testResult.getStdOut());
     }
 
-    if (summaryVerbosity.getIncludeStdErr() && testResult.getStdErr() != null) {
+    if (summaryVerbosity.getIncludeStdErr()
+        && testResult.getStdErr() != null
+        && testResult.getStdErr().length() > 0) {
```
styurin: The same

src/com/facebook/buck/event/listener/TestResultFormatter.java
```diff
@@ -218,11 +218,15 @@ public void reportResultSummary(
...
-    if (summaryVerbosity.getIncludeStdOut() && testResult.getStdOut() != null) {
+    if (summaryVerbosity.getIncludeStdOut()
+        && testResult.getStdOut() != null
+        && testResult.getStdOut().length() > 0) {
       addTo.add("====STANDARD OUT====", testResult.getStdOut());
     }
 
-    if (summaryVerbosity.getIncludeStdErr() && testResult.getStdErr() != null) {
+    if (summaryVerbosity.getIncludeStdErr()
+        && testResult.getStdErr() != null
```
ttsugriy: you can simplify by using [Strings.isNullOrEmpty](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/base/Strings.html#isNullOrEmpty-java.lang.String-)

src/com/facebook/buck/event/listener/TestResultFormatter.java
```diff
@@ -218,11 +218,15 @@ public void reportResultSummary(
       }
     }
 
-    if (summaryVerbosity.getIncludeStdOut() && testResult.getStdOut() != null) {
+    if (summaryVerbosity.getIncludeStdOut()
+        && testResult.getStdOut() != null
+        && testResult.getStdOut().length() > 0) {
```
styurin:
```
if (summaryVerbosity.getIncludeStdOut()
    && !Strings.isNullOrEmpty(testResult.getStdOut())) {
```

linzhp:
@ttsugriy Any more concerns about this PR?

src/com/facebook/buck/event/listener/TestResultFormatter.java
```diff
@@ -115,6 +116,7 @@ public void runStarted(
   public void reportResult(ImmutableList.Builder<String> addTo, TestResults results) {
     if (verbosity.shouldPrintBinaryRunInformation() && results.getTotalNumberOfTests() > 1) {
       addTo.add("");
+
```
bobyangyf: no need to add extra line

src/com/facebook/buck/features/go/GoTest.java
```diff
@@ -229,7 +229,7 @@ public GoTest(
                   "",
                   Joiner.on(System.lineSeparator()).join(stackTrace),
                   Joiner.on(System.lineSeparator()).join(stdout),
-                  ""));
+                  null));
```
bobyangyf: I don't think we need to make this null. Leave it as is before

Issue #1890, opened 2018-05-16
BUCK crashes on the latest NDK


hzsweers:
with NDK 17. Some architectures were removed

```
UNKNOWN ERROR: com.google.common.util.concurrent.ExecutionError: java.lang.AssertionError: /Users/zsweers/dev/android/android-sdk/ndk-bundle/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi doesn't exist.
com.google.common.util.concurrent.ExecutionError: com.google.common.util.concurrent.ExecutionError: java.lang.AssertionError: /Users/zsweers/dev/android/android-sdk/ndk-bundle/sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi doesn't exist.
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2216)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4147)
	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:4151)
	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:5140)
	at com.facebook.buck.toolchain.impl.DefaultToolchainProvider.getOrCreate(DefaultToolchainProvider.java:166)
	at com.facebook.buck.toolchain.impl.DefaultToolchainProvider.getByName(DefaultToolchainProvider.java:109)
	at com.facebook.buck.toolchain.BaseToolchainProvider.getByName(BaseToolchainProvider.java:24)
	at com.facebook.buck.jvm.java.JavaBinaryDescription.getCxxPlatform(JavaBinaryDescription.java:79)
	at com.facebook.buck.jvm.java.JavaBinaryDescription.findDepsForTargetFromConstructorArgs(JavaBinaryDescription.java:175)
	at com.facebook.buck.jvm.java.JavaBinaryDescription.findDepsForTargetFromConstructorArgs(JavaBinaryDescription.java:55)
	at com.facebook.buck.rules.TargetNodeFactory.create(TargetNodeFactory.java:115)
	at com.facebook.buck.rules.TargetNodeFactory.createFromObject(TargetNodeFactory.java:65)
	at com.facebook.buck.parser.DefaultParserTargetNodeFactory.createTargetNodeFromObject(DefaultParserTargetNodeFactory.java:233)
	at com.facebook.buck.parser.DefaultParserTargetNodeFactory.createTargetNode(DefaultParserTargetNodeFactory.java:152)
[...]

PR #1889, opened 2018-05-16
Add support for ABI-specific NDK app platforms

LegNeato:
Fixes https://github.com/facebook/buck/issues/1836.

PR #1894, opened 2018-05-23
Drop `arm` as a default CPU ABI

LegNeato:
With the latest version of the NDK, `arm` is not supported.

Fixes https://github.com/facebook/buck/issues/1890.

**This could possibly be a breaking change** for people who have not set CPU ABI and are not using the latest NDK---before their builds would have `arm`, after updating Buck they will no longer and might wonder what happened. Open to strategies to dealing with that, though they can always just add `arm` to `cpu_abis` in their config to fix any bustage. While buck could detect this case, it seems like a bit too much magic to be worth it.

Note that this is somewhat related to https://github.com/facebook/buck/pull/1889 but not exactly.

### DIFF
--- a/docs/concept/buckconfig.soy
+++ b/docs/concept/buckconfig.soy
@@ -2873,10 +2873,10 @@ your <code>.buckjavaargs</code> file</a>:
 {call buckconfig.entry}
   {param section: 'ndk' /}
   {param name: 'cpu_abis' /}
-  {param example_value: 'armv7, x86' /}
+  {param example_value: 'armv7, arm64, x86' /}
   {param description}
     A comma separated list of the CPU ABIs that this repo supports.  Buck will only build NDK code
-    for these ABIs.
+    for these ABIs. The default is <code>armv7, x86</code>.
   {/param}
 {/call}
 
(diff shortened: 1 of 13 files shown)


LegNeato:
@styurin I am not sure I agree about not modifying the tests...if you look at them they are not testing anything `arm` specific, it just happened to be in the default arch set and thus was chosen to check out of simplicity or completeness. Changing them to check `armv7` keeps the intent of the tests the same while allowing people with both with v17 and earlier NDKs to run the tests. It also means we don't have to duplicate a ton of tests or abstract out logic to test two sides of the NDK arch divide.

I of course will add tests to check the new behavior of switching defaults based on NDK versions.

LegNeato:
Ping @styurin

kageiit:

styurin:
I'm working on it, there is another issue with NDK 16 which should be resolved before allowing NDK 17.

LegNeato:
I believe this is actually still on me...I was waiting for my other PR to land before I rebased and fixed comments

styurin:
Implemented in 4ef8e96bae72988f4bd77994a0228a595aa42ac6.

LegNeato:
Sweet, thanks!

PR #1896, opened 2018-05-23
Filtering platform-dependent files before generating test main

linzhp:
While generating test main for Go tests, Go files need to be filtered so test main doesn't contain coverage variables for Go files of other platforms, which are ignored before passing to Go compiler.

cc @ChinmayR


linzhp:
@ttsugriy @styurin Could you take a look?

PR #1908, opened 2018-06-06
Add GoLand support for "buck project"

linzhp:
Right now, we don't need to create any project files for GoLand, but we need to run a script to symlink generated code into GOPATH. This PR allow us to run that script using "buck project".

It may sound overkill now, but using "buck project" to initialize IDE workspace allows us to hide implementation details from users. At some point in the future when we need to create a project file later, user experience will not change.

@philipjameson @styurin @ttsugriy Please take a look

### DIFF
--- a/docs/command/project.soy
+++ b/docs/command/project.soy
@@ -87,6 +87,7 @@
 <ul>
   <li><a href="#intellij">IntelliJ</a></li>
   <li><a href="#xcode">Xcode</a></li>
+  <li>GoLand</li>
 </ul>
 
 <h3 id="intellij">IntelliJ</h3>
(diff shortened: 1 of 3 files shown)


linzhp:
When --ide=GoLand, Buck will skip project file generate, and only runs pre_process and post_process scripts. This is a very generic approach, as some IDEs/Code editors use the source code as the ground truth, and don't need a project file to start with but need some processing to put code into right place. We may also think about defaulting to this behavior when --ide is neither intellij or xcode.

styurin:
> When --ide=GoLand, Buck will skip project file generator, and only run pre_process and post_process scripts.

Are you saying you use Buck to run some scripts? Why not run those scripts directly instead of adding basically no-op logic to Buck?

linzhp:
Isn't this [.buckconfig](https://buckbuild.com/concept/buckconfig.html#project.pre_process) option designed to do so?

Issue #1906, opened 2018-06-04
Ignore warnings for cxx_library headers


sheldonneuberger:
If we copy the source of a third party library into our repo, make a cxx_library target for it, then include its headers in our other targets, we'll see warnings for the third party library's headers.

prebuilt_cxx_library includes its headers with -isystem to fix this, but in our case it's not a prebuilt because we're copying the whole source into the repo.

Having a systemInclude=true/false flag on cxx_library (and similar, e.g., apple_library) targets would let us solve this. Does that sound like the right path? Any reasonable alternative that I'm missing?

philipjameson:
Is there any reason you can't create a prebuilt_cxx_library that has header_dirs set and header_only set to true, and then depend on that in your cxx_library?

sheldonneuberger:
Ah I see, so specify the headers in a prebuilt_cxx_library and specify the source in a cxx_library. We just gave that a try but prebuilt_cxx_library doesn't work with generated xcode projects (https://github.com/facebook/buck/issues/1098).

styurin:
The issue you linked seems to be resolved, do you still have a problem with using `prebuilt_cxx_library` with XCode projects?

sheldonneuberger-sc:
The linked issue was closed because the workaround was to use cxx_library instead of prebuilt_cxx_library.

sheldonneuberger:
The linked issue was closed because the workaround was to use cxx_library instead of prebuilt_cxx_library.

styurin:
> prebuilt_cxx_library doesn't work with generated xcode projects

Maybe it's worth fixing XCode project generator?

styurin:
@williamtwilson can you help with adding support for prebuilt_cxx_library to Xcode project generator?

PR #1909, opened 2018-06-06
Add system_include parameter for cxx_library to use -isystem for includes

sheldonneuberger:
Fixes https://github.com/facebook/buck/issues/1906

Adds a system_include parameter to cxx_library that defaults to false. When set to true, that library's exported headers will be included with "-isystem" instead of "-I" by consumers. The practical effect of this is to ignore warnings from those header files.

prebuilt_cxx_library includes its headers with -isystem, but it's common to copy the entire source of a project into the repo and build it with cxx_library, in which case system_include=True comes in handy when you want to avoid making edits to the source.


sheldonneuberger:
I believe this is ready for merge, if someone could review. Thanks.

sheldonneuberger:
Friendly ping.

sheldonneuberger:
Any update on this?

styurin:
@sheldonneuberger please, check question in the issue.

PR #1910, opened 2018-06-07
Allow non-test function name to start with "Test"

linzhp:
Buck assumed functions with names starting with "Test" are test functions, and tried to verify their signatures, which would crash on regular functions happen to have names starting with "Test".

This PR fixed that.

@philipjameson @styurin @ttsugriy Please review.


linzhp:
@styurin @ttsugriy Could you take a look?

nemith:
Technically we should only should _test files to the testmaingen.go generation script which is how the `go` tool handles this.  

If you use the `go test` and define a function that starts with Test inside a _test.go. It will fail.  It is ok using Test as a prefix outside of _test.go files.

PR #1911, opened 2018-06-07
Remove framework flavor from apple_libray

robbertvanginkel:
There's an undocumented `#framework` flavor on `apple_library` that in theory should generate a framework from that library. This flavor is incomplete (for example, loader paths are not set correctly), and oddly intertwined with framework embedding. There's some reference to this in the code:

https://github.com/facebook/buck/blob/3e12866d3cf233386a471f5592bd1885fb527494/src/com/facebook/buck/apple/AppleDescriptions.java#L626-L628
as well as on previous issues: https://github.com/facebook/buck/pull/818#issue-78882532

At this point the feature is undocumented, the implementation is incomplete, framework generation and embedding tightly coupled and does not allow for frameworks to be embedded correctly without making frameworks of all apple_library dependencies. I propose we remove the framework flavor as it provides little to no value and only complicates correctly fixing other features, such as prebuilt framework embedding. A commit from 2016 already proposed removing this flavor https://github.com/facebook/buck/commit/af4a756e8a982781783f8e643184b25901bf0899:
> (and removing the apple library #framework flavor)

and a https://github.com/facebook/buck/pull/818#issuecomment-235945113 indicates this flavor is not used internally at facebook.

The only breakage I can see from this is that I also removed the `info_plist(_substitutions)` args from `apple_library`, now if a library does specify that after this pr buck will fail to parse. Open to suggestions if we should maybe keep it around, but as it was only ever used for when you wanted to use the framework flavor I think removing it should be okay. 

I've considered if there's a possible way to decouple framework embedding and generation backward compatible way, but the only possibility I see is a new independent flavor for embedding dependent bundles (such as prebuilt frameworks). I think this will overcomplicate the code too much though, and would rather remove the framework flavor considering it does not work out of the box.

@milend @yiding


src/com/facebook/buck/apple/AppleLibraryDescription.java
```diff
@@ -761,8 +688,7 @@ private HeaderSymlinkTree createUnderlyingModuleSymlinkTreeBuildRule(
       }
     }
 
-    if (metadataClass.isAssignableFrom(FrameworkDependencies.class)
-        && buildTarget.getFlavors().contains(AppleDescriptions.FRAMEWORK_FLAVOR)) {
```
milend: @robbertvanginkel Shouldn't we just delete the whole `if` branch, as if `buildTarget.getFlavors().contains(AppleDescriptions.FRAMEWORK_FLAVOR` will always be `false`?

src/com/facebook/buck/apple/AppleLibraryDescription.java
```diff
@@ -771,21 +697,13 @@ private HeaderSymlinkTree createUnderlyingModuleSymlinkTreeBuildRule(
...
-        if (frameworks.isPresent()) {
-          sourcePaths.addAll(frameworks.get().getSourcePaths());
-        }
+                    AppleDescriptions.NO_INCLUDE_FRAMEWORKS_FLAVOR, cxxPlatformFlavor.get()),
+                FrameworkDependencies.class)
+            .ifPresent(s -> sourcePaths.addAll(s.getSourcePaths()));
       }
-      // Not all parts of Buck use require yet, so require the rule here so it's available in the
-      // graphBuilder for the parts that don't.
-      BuildRule buildRule = graphBuilder.requireRule(buildTarget);
```
milend: @robbertvanginkel If we keep the `if` branch, why did we remove this code?

src/com/facebook/buck/apple/AppleLibraryDescription.java
robbertvanginkel: We could delete the whole branch, but the reason I left it is because the code to copy frameworks into the final bundle will be useful for collecting prebuilt frameworks correctly in a later pr.

src/com/facebook/buck/apple/AppleLibraryDescription.java
robbertvanginkel: Because the apple library won't be built as a framework, so it shouldn't add itself as a framework dependency.

PR #1913, opened 2018-06-08
Add an option to compile swiftmodule files separately

rmaz:
This change adds an option to build swift libraries with greater concurrency. It splits the build of a target into a step to generate the `swiftmodule` file and separate bitcode files for each of the source files. The second pass takes these as inputs and compiles each object file separately. This significantly improves the build parallelism as dependent modules do not need to wait for a dependency to completely build before starting, and also helps reduce the time taken to build large modules as the object files are generated concurrently.


src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -211,19 +214,48 @@ private SwiftCompileStep makeCompileStep(SourcePathResolver resolver) {
...
+        compilerCommand.add("-num-threads", "1");
+        this.srcs
+            .stream()
+            .map(
+                input -> getBitcodePath(resolver.getAbsolutePath(input).getFileName().toString())
+            )
+            .forEach(
+                x -> {
+                  compilerCommand.add("-o");
+                  compilerCommand.add(x.toString());
```
rmaz: @robbertvanginkel this means we will have collisions if there are multiple files with the same name in a target. Better to store them with BUCK relative paths?

robbertvanginkel: Not an issue, swift requires unique filenames when compiling a module:
```
$ swiftc temp/test.swift test.swift
<unknown>:0: error: filename "test.swift" used twice: 'temp/test.swift' and 'test.swift'
<unknown>:0: note: filenames are used to distinguish private declarations with the same name
```

src/com/facebook/buck/swift/SwiftBitcodeCompile.java
```diff
@@ -0,0 +1,77 @@
...
+    buildableContext.recordArtifact(outputPath);
+    ImmutableList.Builder<String> compilerCommand = ImmutableList.builder();
+    compilerCommand.addAll(swiftCompiler.getCommandPrefix(context.getSourcePathResolver()));
+    compilerCommand.add(
+        "-module-name",
+        this.moduleName,
+        "-c",
+        this.bitcodePath.toString(),
+        "-o",
+        this.outputPath.toString());
```
rmaz: This is missing llvm flags that need to be passed through, so will default of `O0`

dingjingpisces2015:
Curious why this changes not be merged eventually? We needs to generate the split object file in some case(when use index-store-path), otherwise it will report errors.

PR #1916, opened 2018-06-12
Fixed link from csharp_library to visibility

aventadorm:
The link from the csharp_library rule page to the visibility page is broken because of a capitalization. Fixed that.


styurin:
Thanks for taking a look. We generate website from definitions in `docs` folder in the `master` branch so to fix that problem you would need change that line: https://github.com/facebook/buck/blob/master/docs/__common.soy#L892

aventadorm:
It should be 
page: 'Visibility' -- > page: 'visibility'
Am I right ?

styurin:
Yeah, looks like it.

You can also test your changes locally by running `docs/soyweb-local.sh`, it should start a webserver and you can check your changes by opening `http://localhost:9811/`

PR #1927, opened 2018-06-18
"buck project" to copy generated Go code to vendor

linzhp:
For most Go IDE/tools to work, every imported package has to be in some proper directories under GOPATH. However, Go code generated during a Buck build is not. This PR copies them to the vendor directory as part of "buck project --ide=intellij", although this change not only makes the IntelliJ/GoLand work, but also enable other IDEs and static check tools work with Go projects built with Buck.

In order for "buck project" command to identify Go targets and `IjProjectCommandHelper` to use Go related classes, I moved the Go module from features to core.

@styurin @kageiit  Please review


styurin:
Let's add support for modules in `project` command. Moving Go back to core would be a regression.

I think we can do that by adding a plugin extension point that would provide implementations of `project` command. No need to move existing project commands to this model. I can take a look at this, but it'll take me some time to do that.

linzhp:
This is currently a blocking issue for us. As a compromise, can we keep Go in the core temporarily, and we are committed to move it back to features once you have that plugin extension point ready? JVM and Apple languages are still in the core anyway, and that is why project command is able to create IntelliJ and XCode projects.

We don't like the idea of keeping a private fork for this PR, as it moves many files, and changes from upstream are likely to result in a painful conflict and merge process. It's relatively easier to Go out of core.

linzhp:
Figured out a way to load Go extension in runtime. So now Go module is out of core again. Please take another look @styurin

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -233,6 +251,75 @@ private TargetGraph getProjectGraphForIde(
         buckEventBus, cell, enableParserProfiling, executor, passedInTargets);
   }
 
+  @SuppressWarnings("unchecked")
+  private ExitCode copyGeneratedGoSource(TargetGraphAndTargets targetGraphAndTargets)
```
vitarb: Can you please add some comments on how it works and what assumptions we make.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -233,6 +251,75 @@ private TargetGraph getProjectGraphForIde(
...
+      throws IOException, InterruptedException {
+    HashMap<String, BuildTargetSourcePath> generatedPackages = new HashMap<>();
```
vitarb: s/HashMap/Map in the declaration

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -233,6 +251,75 @@ private TargetGraph getProjectGraphForIde(
...
+          if (packageName.isPresent()) {
+            generatedPackages.putAll(
+                ((HasSrcs) constructorArg)
+                    .getSrcs()
+                    .stream()
+                    .filter(srcPath -> srcPath instanceof BuildTargetSourcePath)
+                    .collect(
+                        Collectors.toMap(
+                            src -> packageName.get(), src -> (BuildTargetSourcePath) src)));
+          }
```
vitarb: else fail the build? Are there any valid cases when go library may have no package name?

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -233,6 +251,75 @@ private TargetGraph getProjectGraphForIde(
...
+      Class<?> goLibraryDescriptionArgClass =
+          Class.forName("com.facebook.buck.features.go.GoLibraryDescriptionArg", true, classLoader);
+      for (TargetNode<?, ?> targetNode : targetGraphAndTargets.getTargetGraph().getNodes()) {
+        Object constructorArg = targetNode.getConstructorArg();
+        if (constructorArg.getClass() == goLibraryDescriptionArgClass) {
+          Method getPackageNameMethod = goLibraryDescriptionArgClass.getMethod("getPackageName");
+          Optional<String> packageName =
+              (Optional<String>) getPackageNameMethod.invoke(constructorArg);
+          if (packageName.isPresent()) {
+            generatedPackages.putAll(
```
vitarb: Should we remove some line breaks from these chaining calls? (same relates to some other statements below)

linzhp: Those line breaks are added by Google Java formatter, in order to pass CI checks

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
linzhp: Yes there are, but we don't. Figuring out a proper default package name will need to load another class from go module and use the same reflection trick to call its methods.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -304,6 +304,7 @@ private void copyGeneratedGoCode(
         DefaultSourcePathResolver.from(new SourcePathRuleFinder(result.getActionGraphBuilder()));
     for (BuildTargetSourcePath sourcePath : generatedPackages.keySet()) {
       Path desiredPath = vendorPath.resolve(generatedPackages.get(sourcePath));
+      fs.deleteRecursivelyIfExists(desiredPath);
```
vitarb: This will not handle sub-packages well.
For example if we have:
github.com/foo/bar
github.com/foo/bar/A
and process the later first then contents of A will be removed while processing root package of the repo.

Let's make sure we cleanup only contents of the current package.

linzhp:
Pinging @styurin @ttsugriy @philipjameson Could you review this PR?

linzhp:
Pinging @styurin @ttsugriy @philipjameson again

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -190,7 +206,10 @@ public ExitCode parseTargetsAndRunProjectGenerator(List<String> arguments)
 
       return ExitCode.SUCCESS;
     }
-
+    ExitCode result = initGoWorkspace(targetGraphAndTargets);
```
styurin: So it will affect all the invocations of `buck project` even if it's not related to Go?

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -233,6 +252,127 @@ private TargetGraph getProjectGraphForIde(
...
+   * expensive and unreliable (e.g., what if there are multiple src directory?).
+   */
+  private void copyGeneratedGoCode(
+      TargetGraphAndTargets targetGraphAndTargets,
+      Map<BuildTargetSourcePath, Path> generatedPackages)
+      throws IOException {
+    Path vendorPath;
+    ProjectFilesystem fs = cell.getFilesystem();
+    if (fs.exists(Paths.get("src"))) {
+      vendorPath = Paths.get("src", "vendor");
```
styurin: What's vendor?

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
```diff
@@ -233,6 +252,127 @@ private TargetGraph getProjectGraphForIde(
...
+  private Map<BuildTargetSourcePath, Path> findCodeGenerationTargets(
+      TargetGraphAndTargets targetGraphAndTargets) {
+    ClassLoader classLoader = pluginManager.getPluginClassLoader("com.facebook.buck.features.go");
+    Map<BuildTargetSourcePath, Path> generatedPackages = new HashMap<>();
+    try {
+      Class<?> goLibraryDescriptionArgClass;
+      if (classLoader != null) {
+        goLibraryDescriptionArgClass =
+            Class.forName(
+                "com.facebook.buck.features.go.GoLibraryDescriptionArg", true, classLoader);
```
styurin: Could you wait with that? I'll try to migrate project command to a plugin model, but it can take some time.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
linzhp: No, `initGoWorkspace` only scans go_library rules. If there is no such rule in the workspace, it is a no-op.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
linzhp: `vendor` is a place to put third-party libraries. Go toolchain and IDEs is able to automatically pick up the Go packages in vendor directory. It is a good place to put Go packages that we don't want to put into source control.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
vitarb: Do you see any difficulties with changing this in future?

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
styurin: We do not have repository-specific details in Buck logic. You need to change it to smarter about detecting this location.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
linzhp: We really need this feature now. Since changes in this PR is very localized, it is easy to change whenever you migrate project command. I am happy to assist you for migrating this piece of code.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
vitarb: It's not really repository-specific. Vendor folder is a golang concept: https://golang.org/cmd/go/#hdr-Vendor_Directories

linzhp:
@mkaczanowski You imported this PR 4 days ago. Could you check why it is still not merged yet?

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
styurin: We don't really want to change Buck when there is a change in where these libs are located. You can put it in a configuration option and set it to whatever you need in your repository.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
styurin: From the way it looks it should implemented as a separate command.

For which IDE this command is implemented? Is that just copying files?

linzhp: This has been tested for GoLand and Visual Studio Code. Yes, it is just copying files. This is sufficient for these IDEs to work. Go is more like "convention over configuration" style. So it has strong conventions about where packages should be put, rather than relying on configuration files to tell the IDEs and Go toolchain where to search for packages.

src/com/facebook/buck/ide/intellij/IjProjectCommandHelper.java
linzhp: Go toolchain and IDEs doesn't take configurations like that, unfortunately. Only packages in the main src directory and `vendor` directories are recognized.

styurin: > Go toolchain and IDEs don't take configurations like that

But what if they change location? You would need to change this code to make it working, then deal with support of the different versions of these tools.

A configuration option makes it much easier to maintain. You can even use a configuration option with default value of `src/vendor` when it's not present.

linzhp: The approach here is very flexible, not repository specific at all. If a project has its own vendor folder at `GOPATH/src/foo/bar/vendor`, and the buck root is at `GOPATH`, then the generated code will be copied into `GOPATH/src/vendor`. If the buck root is at GOPATH/src/foo, the the generated code will be copied to `GOPATH/src/foo/vendor`. All of these vendor folders are recognized by Go toolchain and IDEs at the same time.

Of course, if you want the vendor folder location to be further configurable by users, I can do that too.

Overall, I see this PR a temporary solution to unblock Go developers using Buck. For long term solution, there are discussions on Bazel side for over a year ([1](https://github.com/bazelbuild/rules_go/issues/512), [2](https://github.com/bazelbuild/rules_go/issues/393)). We may want to watch with Bazel community and see what they end up with. Whatever the long term solution ends up, I am happy to assist the migration.

If you have further concerns about this feature, I can also add a configuration that disable this feature by default.

I also notice that @mkaczanowski recently commit [changes](https://github.com/facebook/buck/commit/9fe8c34606e1ee8dc0daa6e558434ec01608dab9) trying to support IDEs (great thanks to @mkaczanowski), but the performance is somehow quite bad (I need to dig a bit deeper to understand why), and I am not quite sure about symlink tree too, as people in Bazel run into problems with symlink trees ([3](https://github.com/bazelbuild/rules_go/issues/393#issuecomment-320710886))

mkaczanowski:
thanks @linzhp this is very useful for us as well!

PR #1945, opened 2018-06-27
Add prebuilt frameworks to tests without bundle loader in xcode projects

robbertvanginkel:
When a test has no bundle loader, it should include all its frameworks in the test bundle and have `@loader_path/Frameworks` set in `LD_RUNPATH_SEARCH_PATHS`.


yiding:
Is this in line with the behavior of test bundles generated with buck during buck build?

robbertvanginkel:
Yes, this was specifically to address an issue that passed in buck test but did not in a generated project.

PR #1957, opened 2018-07-08
Add target base path into cgo include directory

### LINKED ISSUE #1809: Buck expects import paths in cgo files to be relative to the Buck project root instead of the file's current directory
I have an example of this here: https://github.com/monty-uber/cgo-buck-example/tree/master/go-calls-c

All of the files are in the same directory, therefore the cgo_source.go file's import should be ```// #include "call_from_go.h"```. However, Buck complains that it cannot find the header file in that case. Changing it to be relative to the project root like so: ```// #include "go-calls-c/call_from_go.h"``` makes the Buck build succeed.

This makes these cgo files incompatible with the go toolchain as ```go build``` expects the import paths to be relative to the file itself.

This is also apparent in the integration tests in Buck: https://github.com/facebook/buck/tree/master/test/com/facebook/buck/go/testdata/cgo/src/cgo_test. It's the same scenario, the go code that references C is in cgosrc.go, which has the include path of "src/cgo_test/lib.h" even though lib.h is in the same directory.

linzhp:
Fixed #1809 

@styurin @ttsugriy Please review

### DIFF
--- a/src/com/facebook/buck/features/go/CGoCompileStep.java
+++ b/src/com/facebook/buck/features/go/CGoCompileStep.java
@@ -58,9 +58,9 @@ public class CGoCompileStep extends ShellStep {
         .add("-importpath", context.getBuildCellRootPath().toString())
         .add("-srcdir", context.getBuildCellRootPath().toString())
         .add("-objdir", outputDir.toString())
+        .addAll(cgoCompilerFlags)
         .add("--")
         .add("-I" + getBuildTarget().get().getBasePath().toString())
-        .addAll(cgoCompilerFlags)
         .addAll(srcs.stream().map(Object::toString).iterator())
         .build();
   }
--- a/src/com/facebook/buck/features/go/CGoCompileStep.java
+++ b/src/com/facebook/buck/features/go/CGoCompileStep.java
@@ -58,6 +58,8 @@ public class CGoCompileStep extends ShellStep {
         .add("-importpath", context.getBuildCellRootPath().toString())
         .add("-srcdir", context.getBuildCellRootPath().toString())
         .add("-objdir", outputDir.toString())
+        .add("--")
+        .add("-I" + getBuildTarget().get().getBasePath().toString())
         .addAll(cgoCompilerFlags)
         .addAll(srcs.stream().map(Object::toString).iterator())
         .build();
--- a/test/com/facebook/buck/features/go/GoBinaryIntegrationTest.java
+++ b/test/com/facebook/buck/features/go/GoBinaryIntegrationTest.java
@@ -259,4 +259,12 @@ public class GoBinaryIntegrationTest {
     workspace.setUp();
     workspace.runBuckBuild("//:family").assertSuccess();
   }
+
+  @Test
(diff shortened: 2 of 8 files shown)


test/com/facebook/buck/features/go/testdata/cgo/src/mixed_with_c/BUCK.fixture
```diff
@@ -0,0 +1,22 @@
...
+    package_name = "main",
+    srcs = ["cgosrc.go"],
+    deps = [":lib"]
+)
+
+cxx_library(
+    name = "lib",
+    srcs = ["lib.c"],
+    exported_headers = ["lib.h"],
+    header_namespace = "",
```
mkaczanowski: is the header_namespace important here? does it work without it?

linzhp: No, unfortunately. `header_namespace = ""` allows the C compiler running from buck root to compile CGo generated C files with `include "lib.h"`. Otherwise, C compiler would try to find `lib.h` from its current directory (i.e., buck root), resulting a failure.

The `--I` flag in this fix only tells the CGo compiler to find `lib.h` from the target's base directory. It doesn't pass that message to the C compiler.

src/com/facebook/buck/features/go/CGoCompileStep.java
```diff
@@ -58,6 +58,8 @@ public CGoCompileStep(
         .add("-importpath", context.getBuildCellRootPath().toString())
         .add("-srcdir", context.getBuildCellRootPath().toString())
         .add("-objdir", outputDir.toString())
+        .add("--")
+        .add("-I" + getBuildTarget().get().getBasePath().toString())
         .addAll(cgoCompilerFlags)
```
mkaczanowski: I think this should be before "--" otherwise the cgoCompilerFlags will be passed to the gcc

mkaczanowski: I think it's reasonable to add extra arg to CgoLibraryDescription 'cgoCxxCompilerFlags'

linzhp: Fixed.

linzhp:
@mkaczanowski Could you import this PR again? I updated an outdated doc related to cgo too.

test/com/facebook/buck/features/go/GoBinaryIntegrationTest.java
```diff
@@ -259,4 +259,12 @@ public void buildConstraints() throws IOException, InterruptedException {
     workspace.setUp();
     workspace.runBuckBuild("//:family").assertSuccess();
   }
+
+  @Test
```
linzhp: @mkaczanowski Thanks for landing this PR. Looks like this line disappeared during merging, effectively disabling this test. Could you fix it?

mkaczanowski: ah, crap, thanks for spotting this! I'll fix this soon

PR #1966, opened 2018-07-13
Support for generated cgo package

linzhp:
A generated Go package that include Go, Cgo, C and header files can be built with Go toolchain. This PR makes it possible to build with Buck too. The detail of the scenario can be found in the added test. 

@styurin @mkaczanowski Please take a look. Or let me know if you found Buck is already able to build the package generated by `:gen` target in the testdata of this PR, without the code changes in this PR.


mkaczanowski:
I quickly skimmed through the PR. So this enables the genrule generated sources, right?
For generating the .h, .c and .go files by genrule I internally use the "cxx_genrule" and everything works fine.

Can you check if cxx_genrule would work for you?

linzhp:
This PR is not about the generation process, but about the compilation of generated cgo files that include generated .h files. `cxx_genrule` is not capable of compiling cgo files, as I don't see a good way to call the `cgo` command from the configured GOROOT.

mkaczanowski:
Alright, it looks lgtm to me

src/com/facebook/buck/features/go/GoCompile.java
```diff
@@ -278,7 +279,7 @@ public GoCompile(
     List<Path> srcFiles = new ArrayList<>();
     for (SourcePath path : srcPaths) {
       Path srcPath = context.getSourcePathResolver().getAbsolutePath(path);
-      if (Files.isDirectory(srcPath)) {
+      if (path instanceof BuildTargetSourcePath && Files.isDirectory(srcPath)) {
```
mkaczanowski: @linzhp why is this needed? after I removed the instanceof the tests are still passing, if this is really required can you update the test to reflect the case?:
`path instanceof BuildTargetSourcePath`

test/com/facebook/buck/features/go/GoBinaryIntegrationTest.java
```diff
@@ -269,4 +269,12 @@ public void cgoIncludeHeaderFromSamePackage() throws IOException {
     ProcessResult result = workspace.runBuckCommand("run", "//src/mixed_with_c:bin");
     result.assertSuccess();
   }
+
+  @Test
+  public void generatedCgoPackage() throws IOException {
+    ProjectWorkspace workspace = TestDataHelper.createProjectWorkspaceForScenario(this, "cgo", tmp);
```
mkaczanowski: please add:
`GoAssumptions.assumeGoVersionAtLeast("1.10.0");`

src/com/facebook/buck/features/go/GoCompile.java
mkaczanowski: thanks for explanation. @styurin can you check that line? If that is the case likely we might need to update other places as well

linzhp:
Any more concerns with this PR?

mkaczanowski:
@linzhp I had a quick chat with @philipjameson and there is likely larger improvement we can do to the way we deal with header files in general.
I will let you know whats the status soon

linzhp:
That would be great. The test case introduced by this PR capture the scenario that's currently blocking us. It is essentially a generated Go package that has cgo, C, and headers. Ideally, a `cgo_libary` should be able to take all cgo, C source files and headers as input without having to rely on a `cxx_library`.

Could you merge this PR as a temporary fix so that we can be unblocked without having to maintain our fork before your larger improvement comes in?

mkaczanowski:
Solved in:
https://github.com/facebook/buck/commit/b9012d7b8567043a1624ef5e6c3bc0e819af4dc5

PR #1970, opened 2018-07-19
Add support for bazel-like short build target names

burakcoskun:
Bazel omits the part after colon in target names
if that part is equal to the part after last slash.
e.g. bazel converts //foo/bar:bar to //foo/bar
Though both bazel and buck use Skylark, this creates
confusion in some tools like
https://github.com/bazelbuild/buildtools/tree/master/buildifier
There has been a discussion going on
https://github.com/bazelbuild/buildtools/pull/339

### DIFF
--- a/src/com/facebook/buck/parser/BuildTargetParser.java
+++ b/src/com/facebook/buck/parser/BuildTargetParser.java
@@ -92,15 +92,15 @@ public class BuildTargetParser {
     }
 
     List<String> parts = BUILD_RULE_SEPARATOR_SPLITTER.splitToList(targetAfterCell);
-    if (parts.size() != 2) {
+    if (parts.size() > 2) {
       throw new BuildTargetParseException(
           String.format(
-              "%s must contain exactly one colon (found %d)", buildTargetName, parts.size() - 1));
+              "%s must contain at most one colon (found %d)", buildTargetName, parts.size() - 1));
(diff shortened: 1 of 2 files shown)


ttsugriy:
The decision not to shorten labels was [intentional](https://github.com/bazelbuild/buildtools/issues/34) and as I've commented on https://github.com/bazelbuild/buildtools/pull/339 it's not really necessary since shortening can be disabled in buildifier by passing `-buildifier_disable=label` to it.

romanoid:
@ttsugriy 
I do agree with the intention and the fact that a single way to declare target is less confusing.  

On the other hand this decision was likely made before the decision to use skylark and does not 100% hold with skylark anymore, for example even now, when using skylark, in some of the cases short names work: 

`load('//tooling/experimental/genrules/rules.bzl', 'some_rule')` is expanded into `load('//tooling/experimental/genrules/rules.bzl:rules.bzl', 'some_rule')`

Expanding in some but not all contexts may lead to even more confusion.

I think all and all it is worth to discuss it again in the new context (skylark) and re-consider or consciously accept inconsistency tradeoff instead of just deferring to past decision.

ttsugriy:
Excellent point about the short syntax being accepted in `load` functions, @romanoid! I'll make a change to disallow it for consistency :)

PR #1971, opened 2018-07-21
Package prebuilt_apple_framework when embedding in bundle

ghvg1313:
Fix the `prebuilt_apple_framework` with the following:

Strip unused architectures from the binary when embedding
Strip headers, modulemaps, privateheaders etc when embedding
Recursively include dependency prebuilt apple framework when embedding

### DIFF
--- a/test/com/facebook/buck/apple/PrebuiltAppleFrameworkIntegrationTest.java
+++ b/test/com/facebook/buck/apple/PrebuiltAppleFrameworkIntegrationTest.java
@@ -138,15 +138,16 @@ public class PrebuiltAppleFrameworkIntegrationTest {
     assertFalse(Files.isDirectory(includedFramework.resolve("Headers")));
     ProcessExecutor.Result includedLipoVerifyResult =
         workspace.runCommand(
-            "lipo",
-            includedFramework.resolve("BuckTest").toString(),
-            "-verify_arch",
-            "x86_64");
+            "lipo", includedFramework.resolve("BuckTest").toString(), "-verify_arch", "x86_64");
     assertEquals(
         includedLipoVerifyResult.getStderr().orElse(""), 0, includedLipoVerifyResult.getExitCode());
     ProcessExecutor.Result removedLipoVerifyResult =
         workspace.runCommand(
-            "lipo", includedFramework.resolve("BuckTest").toString(), "-verify_arch", "armv7", "arm64");
+            "lipo",
+            includedFramework.resolve("BuckTest").toString(),
+            "-verify_arch",
+            "armv7",
+            "arm64");
     assertEquals(
         removedLipoVerifyResult.getStderr().orElse(""), 1, removedLipoVerifyResult.getExitCode());
   }


yiding:
It seems like the changes here does more than merely strip unused architectures and other misc files. Namely, it's also changing which prebuilt frameworks are embedded. Is this intentional? If so, can you update the message to reflect that?

Stripping unneeded files and arches is good. My main concern is with the recursive traversal for finding frameworks to embed. This takes away some control over how an app bundle is laid out, since previously embedding is controlled directly by making the framework a dependency of the bundle.

Namely, consider this situation of "diamond" framework dependencies:

App -> Framework1
App -> Framework2
Framework1 -> Framework3
Framework2 -> Framework3

In this case I might want to put all three frameworks directly under the App bundle, whereas (correct me if I'm wrong) your changes would cause Framework3 to be embedded in both Framework1 and Framework2.

rmaz:
> In this case I might want to put all three frameworks directly under the App bundle, whereas (correct me if I'm wrong) your changes would cause Framework3 to be embedded in both Framework1 and Framework2.

Are these all prebuilt frameworks? In that case they will be embedded in the first containing bundle (ie App).

ghvg1313:
>Are these all prebuilt frameworks? In that case they will be embedded in the first containing bundle (ie App).
@yiding Does that make sense? Only bundle container would package the framework resources, in your example, only app would package Framework3 instead of Framework 1 and 2 (it does overwrite framework3 twice, but only one copy would stay inside bundle, assuming everything is dylib)

yiding:
If that's how the code works it should be fine. Someone on the buck team should probably import this and test to make sure it doesn't break for our internal apps which relies on this property.

ghvg1313:
> Someone on the buck team should probably import this and test to make sure it doesn't break for our internal apps which relies on this property.
Should I reach out the team on Slack? I actually don't know who's the contact person now

ghvg1313:
@styurin Do you mind give it a review and import this if everything checks out?

styurin:
It's failing on some tests, can you rebase it to trigger CI runs on master?

ghvg1313:
@styurin just checking the test, however I don't think this change would cause the test failure, could it be some master change failing the test? Also included #2016 here since it's solving an issue related to this PR

test/com/facebook/buck/apple/testdata/app_bundle_with_cxx_library/dep/BUCK
```diff
@@ -0,0 +1,9 @@
+cxx_library(
```
styurin: You need to add `.fixture` extension to this file

ghvg1313: Oops, fixed now

styurin:
Some tests are failing in `test/com/facebook/buck/apple`. Travis CI doesn't run them because we don't run tests on OS X. Can you check them locally?

ghvg1313:
@styurin My bad, it was using the wrong arch to test, fixed now

ghvg1313:
@styurin How's the progress on importing this? Is there any issue that I can help on?

styurin:
Sorry for the delay, our internal testing got an error in `PrebuiltAppleFrameworkIntegrationTest#testPrebuiltAppleFrameworkCopiedToBundle`. Could you take a look, please?

styurin:
Here's the error:

```
Looking for a provisioning profile for bundle ID com.example.TestApp
No provisioning profile matching the bundle ID com.example.TestApp was found
Build failed: No valid non-expired provisioning profiles match for *.com.example.TestApp
Looking for a provisioning profile for bundle ID com.example.TestApp
No provisioning profile matching the bundle ID com.example.TestApp was found
    When running <provisioning-profile-copy>.
    When building rule //app:TestAppBundle#dwarf-and-dsym,include-frameworks,iphoneos-arm64,iphoneos-armv7.
```

ghvg1313:
@styurin Fixed the local provisioning file assumption

ghvg1313:
@styurin Sorry fixed linting

PR #1973, opened 2018-07-23
Add additional fields to the kotlin rule key.

cwoodwar6:
It seems reasonable to add these to the rule key generation since they are included in JavacToJarStepFactory.java.

This should essentially invalidate if anything changes to java or its arguments (e.g. annotation processor changes for kapt).


cwoodwar6:
@thalescm but it seems difficult to determine that? Because you may be generating java code via kapt. I'm not sure if there is a way to dynamically include something as part of the rule key.

thalescm:
I agree, that's why I said it's tricky. Probably better to put add them to the rule key then doesn't.
Don't know about dynamically including stuff, but I think it's not possible

styurin:
Tests are failing

PR #1979, opened 2018-07-25
Set isAllowRDotJavaInSecondaryDex to true by default

### LINKED ISSUE #1978: Regression in dex split for non predexed apps
`isAllowRDotJavaInSecondaryDex` in `AndroidGraphEnhancerArgs` needs to be defaulted to `true`.

This commit https://github.com/facebook/buck/commit/8c331339d4484f90c8888db175af805887a01015 broke apps that have a large number of R classes . This is a regression and can affect any apps in the wild using buck. The previous behavior should be the default to prevent this from happening.

cc @dreiss @styurin

kageiit:
Fixes #1978


kageiit:
cc @styurin @asp2insp @dreiss

kageiit:
Fixed in interal PR. Thanks @asp2insp !

PR #1981, opened 2018-07-27
Add support to attach custom headers to buck doctor request

kageiit:
This allows attaching additional headers for auth as an example

### DIFF
--- a/src/com/facebook/buck/doctor/DefaultDefectReporter.java
+++ b/src/com/facebook/buck/doctor/DefaultDefectReporter.java
@@ -203,6 +203,10 @@ public class DefaultDefectReporter implements DefectReporter {
       Request.Builder requestBuilder = new Request.Builder();
       requestBuilder.addHeader(
           REQUEST_PROTOCOL_VERSION, doctorConfig.getProtocolVersion().name().toLowerCase());
+      for (Map.Entry<String, String> entry :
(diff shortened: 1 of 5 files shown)


kageiit:
cc @styurin

kageiit:
> please add a test to verify that this functionality actually works :)

@ttsugriy There are no existing tests that check the validity of args/headers. Whats the best way to test this, given its mostly looking like a smoke test

ttsugriy:
@kageiit, looks like current code was not really written with testability in mind, which is very unfortunate. HttpClient should not be created in that method and instead passed to the method. This way you can create a `FakeHttpClient` that would capture everything that it was passed with. You can see suggestions on how to do this using interceptors in https://github.com/square/okhttp/issues/1096 but I'm ok with a plain old fake :) (just no mocks please).

While at it, please also add documentation for the configuration option you're adding :)

linzhp:
@kageiit You can write integration tests similar to [this](https://github.com/facebook/buck/pull/1986/files#diff-849bfcb0b18649e5c25bc4e88e57ed40R168)

kageiit:
@ttsugriy added integration tests

ttsugriy:
thanks, @kageiit! I no longer work on Buck, so I cannot import your changes, but I'm sure @styurin would be able to help :)

PR #1986, opened 2018-07-30
Publish lone build targets

linzhp:
This PR allows `buck publish` to upload the output of lone build target (whose output is a single file) to an HTTP address.

@styurin Please review.

cc @kageiit @romanoid

### DIFF
--- a/src/com/facebook/buck/cli/PublishCommand.java
+++ b/src/com/facebook/buck/cli/PublishCommand.java
@@ -22,14 +22,18 @@ import static com.facebook.buck.jvm.java.Javadoc.DOC_JAR;
 
 import com.facebook.buck.config.BuckConfig;
 import com.facebook.buck.core.model.BuildTarget;
+import com.facebook.buck.core.rules.ActionGraphBuilder;
 import com.facebook.buck.core.rules.BuildRule;
 import com.facebook.buck.core.rules.SourcePathRuleFinder;
 import com.facebook.buck.core.sourcepath.resolver.impl.DefaultSourcePathResolver;
+import com.facebook.buck.event.BuckEventBus;
 import com.facebook.buck.event.ConsoleEvent;
+import com.facebook.buck.file.OutputFilePublisher;
 import com.facebook.buck.jvm.java.MavenPublishable;
 import com.facebook.buck.maven.Publisher;
 import com.facebook.buck.parser.BuildTargetSpec;
 import com.facebook.buck.parser.TargetNodeSpec;
+import com.facebook.buck.step.ExecutorPool;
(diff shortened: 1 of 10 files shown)


src/com/facebook/buck/cli/PublishCommand.java
```diff
@@ -68,20 +78,25 @@
   private URL remoteRepo = null;
 
   @Option(
-      name = TO_MAVEN_CENTRAL_LONG_ARG,
-      usage = "Same as \"" + REMOTE_REPO_LONG_ARG + " " + Publisher.MAVEN_CENTRAL_URL + "\"")
-  private boolean toMavenCentral = false;
```
romanoid: this changes the logic and breaks existing use cases:

1) it requires `toMaven` when it otherwise would not. 
2) it makes new behavior default.
3) it changes flag names and logic.

romanoid:
Can you please advice why it is done as change in buck code and not as a small script outside of buck? 

I don't see any information that it uses not available otherwise.

linzhp:
- The same reasoning for the support of publishing to Maven repo should apply here
- The document of `buck publish` never mentions that it is dedicated for Java/Maven only. This PR just made it as generic as it is supposed to be

src/com/facebook/buck/cli/PublishCommand.java
linzhp: If backward compatibility is the preference, I can certainly make it happen by keeping all the existing flags and add another flag call `isMaven`. However, this new flag is mutually exclusive of the `to-maven-central`. It feels a bit redundant.

I can also set the default value of `toMaven` to be `true`, to keep the changes to existing use cases minimal.

romanoid:
The reason why maven publish needs to be implemented in buck are more intricated: 
It needs to understand classpath and dependency graph for both pom generation and uberjar creation from multiple targets. 

This is not possible to do outside of Buck. 

Unlike that, current addition seems to be easily implementable. 

According to the approach shared by the BUCK team, they prefer to not include in BUCK itself features that can easily be implemented outside of BUCK source code. 

Will rely on @styurin feedback besides my comment that current change breaks existing workflows.

src/com/facebook/buck/cli/PublishCommand.java
romanoid: Breakage of existing users seems unnecessary here. `isMaven` does not need to be mutually exclusive with `to-maven-central` and `to-maven-central` can imply `isMaven=true` and fail if `isMaven=false`

linzhp:
Closing this. Will send out another PR so it can publish zip_files to Maven repo

PR #1987, opened 2018-08-02
Add annotation_processing_tool (kapt | javac) option to Kotlin modules

artem-zinnatullin:
Original motivation was explained in [discussion in Slack](https://buckbuild.slack.com/archives/C6N81MF55/p1533090203000161):

>Hey guys, after spending couple days on trying to switch us to Buck upstream I came to conclusion that `kapt` support needs to be changed in the upstream, `kapt` should be _optional_.
>
>We have mixed Java/Kotlin modules where we deliberately run annotation processing only against Java sources (our reasoning is build speed & stability in both Buck and Gradle setups).
>
>In our setup `kapt` is **not enabled** on the Gradle side
>
>On the Buck side however `kapt` kicks in even if we don't ask for it, what's worse it becomes primary method of annotation processing for mixed modules.
>
>We have 2 problems because of that:
>
>1) If module doesn't have any `.kt` with annotations that need apt, build fails because `.zip` with generated classes is not produced. Can be "solved" by converting one of the classes with annotations to Kotlin, but we don't need nor want that.
>
>2) Build will behave differently if we convert (see point №1) one of the classes to `.kt`, it'll work in Buck but won't work in Gradle because we don't use `kapt` there
>
>I've dealt with it before by turning `kapt` off in our fork, but we really want to get on to the upstream Buck to move faster, and `kapt` support is last huge blocker I'm dealing with…
>
>I think OkBuck could look for `kapt` configuration and only then enable `kapt` in Buck

@cwoodwar6 mentioned they have same problem:

>To your first point, we face the same issue and have been putting simple stub kotlin files to work around it. But we don’t like this solution either and have looking to fix it

---

Techincal aspects of this PR:

PR adds a separate `annotation_processing_tool = "kapt/javac"` to the DSL of Kotlin modules.

Default value is kept as `kapt` for compatibility.

Tools:

- `kapt` tool: Allows to run Java annotation processors against Kotlin sources while backporting it for Java sources too.
- `javac` tool: Works only against Java sources, Kotlin sources won't have access to generated classes at compile time.

---

If this gets accepted, the next step will be to update OkBuck with respecting Gradle `kapt` vs `annotationProcessor` configurations.

cc @thalescm @cwoodwar6 @kageiit


src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
```diff
@@ -261,20 +262,33 @@ public void createCompileStep(
...
+    switch (annotationProcessingTool) {
+      case KAPT:
+        finalJavacOptions = javacOptions.withAnnotationProcessingParams(AnnotationProcessingParams.EMPTY);
+        break;
+
+      case JAVAC:
+        finalJavacOptions = javacOptions;
+        break;
+
+        default:
```
cwoodwar6: nit: extra indent

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
```diff
@@ -162,8 +162,23 @@ public BuildRule createBuildRule(
...
+    /**
+     * Works only against Java sources, Kotlin sources won't have access to generated
+     * classes at compile time.
+     */
+    JAVAC,
+  }
+
   public interface CoreArg extends JavaLibraryDescription.CoreArg {
     ImmutableList<String> getExtraKotlincArguments();
+    Optional<AnnotationProcessingTool> getAnnotationProcessingTool();
```
cwoodwar6: Be sure to document this in the rules. E.g. `docs/rule/android_library.soy`

src/com/facebook/buck/jvm/kotlin/KotlincToJarStepFactory.java
artem-zinnatullin: fixed

src/com/facebook/buck/jvm/kotlin/KotlinLibraryDescription.java
artem-zinnatullin: added 👍

test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/annotation-processing-tool-javac/AJavaClass.java
```diff
@@ -0,0 +1,10 @@
+package com.example.ap;
+
+class AJavaClass {
+  public void foo() {
+    new JavaClassWithJavaAnnotation_();
+    //new KotlinClassWithJavaAnnotation_();
```
thalescm: ✂️ ?

test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/annotation-processing-tool-javac/AKotlinClass.kt
```diff
@@ -0,0 +1,10 @@
+package com.example.ap
+
+class AKotlinClass {
+    fun foo() {
+        //JavaClassWithJavaAnnotation_()
```
thalescm: ✂️ ?

test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/annotation-processing-tool-javac/AJavaClass.java
artem-zinnatullin: 👍

styurin:
Could you reformat files and make sure CI passes?

artem-zinnatullin:
Should be good now

docs/rule/android_library.soy
```diff
@@ -159,6 +159,18 @@ compiled class files and resources.
   {/param}
 {/call}
 
+{call buck.arg}
+  {param name: 'annotation_processing_tool' /}
+  {param default : 'kapt' /}
+  {param desc}
+  Specifies the tool to use for annotation processing. Possible values: "kapt" or "javac".
+  "kapt" allows to run Java annotation processors against Kotlin sources while backporting
```
philipjameson: s/"kapt" allows to run Java annotation/"kapt" allows running Java annotation/

docs/rule/kotlin_library.soy
```diff
@@ -60,6 +60,18 @@ of the compiled class files and resources.
   {/param}
 {/call}
 
+{call buck.arg}
+  {param name: 'annotation_processing_tool' /}
+  {param default : 'kapt' /}
+  {param desc}
+  Specifies the tool to use for annotation processing. Possible values: "kapt" or "javac".
+  "kapt" allows to run Java annotation processors against Kotlin sources while backporting
```
philipjameson: s/"kapt" allows to run Java annotation/"kapt" allows to running Java annotation/

artem-zinnatullin:
yeah, I was running tests before I opened up the PR

I remember I did some renames and refactorings in tests right before opening up the PR, must have had copied that `kotlin_library` twice into the BUCK file and didn't run tests after that, my bad

artem-zinnatullin:
Thanks for review, everyone 😽

PR #2006, opened 2018-08-24
Add source path to unzipped aar path for android prebuilt aar

kageiit:
`prebuilt_jar` already supports getting to the `binary_jar` via the `location` macro in genrules. Unfortunately, it seems `android_prebuilt_aar` does not do so yet. This behavior is very useful to have to implement features like android lint via genrules in buck. It also seems like the flavored `#aar_unzip` target only outputs paths to the text symbols (Not sure why). To cause minimal disruption to existing functionality, this pr adds the ability to get to the unzipped aar location of an `android_prebuilt_aar` for use in location macro expansion.


kageiit:
cc @styurin

kageiit:
@styurin ping :)

Issue #2007, opened 2018-08-27
BUCK test runner sometimes silently skips tests if it failed to initialize the class


romanoid:
I'll create repro / add more details as I investigate.

I've discovered that if error happens during loading/initialization of test class, BUCK ends up silently ignoring the test instead of reporting this failure.

Also interestingly enough it properly reports the error with connected debugger, so some race condition may be happening too. 
Note: with debugger not connected behavior is 100% consistent, it always skips.

bobyangyf:
Is this for JUnitTest 3/4?

romanoid:
To be more specific, this is Scalatest (which is compatible with both) and is used in junction with Junit4.

romanoid:
I found how the problem manifests itself and will have PR with remedy shortly.

romanoid:
the problem I found only fixes some of the examples (that behave the same with debugger and without). I'm perplexed by one test that fails correctly if (`--debug` or `--filter` are applied, but not otherwise)

romanoid:
PR is UP: https://github.com/facebook/buck/pull/2009

PR #2009, opened 2018-08-28
Fixed BUCK default test runner sometimes skipping test failures if delegate (junit) test runner fails with error.

romanoid:
Fixes https://github.com/facebook/buck/issues/2007

Unfortunately the case where I were able to consistently reproduce the problem involves heavy amount of 3rd-party dependencies and I were unable to condense the problem to a simple reproduction case.

### DIFF
--- a/src/com/facebook/buck/testrunner/DelegateRunnerWithTimeout.java
+++ b/src/com/facebook/buck/testrunner/DelegateRunnerWithTimeout.java
@@ -18,10 +18,11 @@ package com.facebook.buck.testrunner;
 
 import com.facebook.buck.util.Threads;
 import com.facebook.buck.util.concurrent.MostExecutors;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.runner.Description;
 import org.junit.runner.Runner;
 import org.junit.runner.notification.RunNotifier;
@@ -105,7 +106,7 @@ class DelegateRunnerWithTimeout extends Runner {
       return;
     }
 
-    AtomicBoolean testsCompleted = new AtomicBoolean(false);
+    CompletableFuture<Boolean> testCompleted = new CompletableFuture<>();
 
     // We run the Runner in an Executor so that we can tear it down if we need to.
     executor
@@ -114,10 +115,12 @@ class DelegateRunnerWithTimeout extends Runner {
             () -> {
               try {
                 delegate.run(wrapper);
-              } finally {
                 if (!wrapper.hasTestThatExceededTimeout()) {
-                  testsCompleted.set(true);
+                  testCompleted.complete(true);
                 }
+              } catch (Throwable t) {
+                testCompleted.completeExceptionally(t);
+              } finally {
                 completionSemaphore.release();
               }
             });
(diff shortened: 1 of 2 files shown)


bobyangyf:
Few questions:

This isn't regarding any failures in the test itself or even in `@Before` methods right? It seems like JUnit handles that.
Does `RunNotifier` or any other junit listeners provide a way of listening to the initialization/loading failures you mention here?
We had similar bugs in TestNG fixed like this: https://github.com/facebook/buck/commit/0c7d1341af82edddcddd247205caae73b5a3c719

src/com/facebook/buck/testrunner/JUnitRunner.java
```diff
@@ -370,6 +370,17 @@ public void testFinished(Description description) throws Exception {
               stdErr.length() == 0 ? null : stdErr.toString()));
     }
 
+    @Override
+    public void testRunFinished(Result runResult) {
```
romanoid: This fixes when Runner start test and then encounters failure that prevents it from continuing the run.

src/com/facebook/buck/testrunner/DelegateRunnerWithTimeout.java
```diff
@@ -114,10 +115,12 @@ private void runWithBuckManagedTimeout(DelegateRunNotifier wrapper) {
             () -> {
               try {
                 delegate.run(wrapper);
-              } finally {
                 if (!wrapper.hasTestThatExceededTimeout()) {
-                  testsCompleted.set(true);
+                  testCompleted.complete(true);
                 }
+              } catch (Throwable t) {
```
romanoid: This properly wraps runner failing completely. 

Before this change testrunner will behave differently in case of crash depending on if it runs directly (because `--filter=` or `--debug` was passed to run command) or if wrapped by this class.

src/com/facebook/buck/testrunner/JUnitRunner.java
romanoid: Previously `public void testAssumptionFailure(Failure failure) {` would just swallow the error since `resultListener` is not null, but `testFinished` is never called, so this failure is lost.

romanoid:
Added comments inline. 

This does not fix if you run test with default Junit runner, i.e `@RunWith(Junit4.class)`, Junit4 is properly wrapping all the errors. 

On the other hand some other runners may not behave the same. 

This fixes 2 separate issues that I mentioned inline on this PR.

romanoid:
After applying this fix to the repo, discovered single non-scala test (using `@RunWith(PowerMockRunner.class)` ) that had been failing silently before.

src/com/facebook/buck/testrunner/DelegateRunnerWithTimeout.java
```diff
@@ -126,9 +129,22 @@ private void runWithBuckManagedTimeout(DelegateRunNotifier wrapper) {
...
-        // Normal termination: hooray!
-        return;
+      if (testCompleted.isDone()) {
+        try {
+          testCompleted.join();
+          // Normal termination: hooray!
+          return;
+        } catch (CompletionException completionException) {
+          // Unwrap and re-throw original RuntimeException or Error
+          if (completionException.getCause() instanceof RuntimeException) {
```
bobyangyf: nit: Use one of these?
https://google.github.io/guava/releases/21.0/api/docs/com/google/common/base/Throwables.html#throwIfInstanceOf-java.lang.Throwable-java.lang.Class-

src/com/facebook/buck/testrunner/JUnitRunner.java
bobyangyf: would it not make sense to change line 399 then??
We can check whether the failure is actually an assumptionViolationException, or else record as a failure.

romanoid: that's not necessary `testAssumptionFailure`

Here's how rest runner execution works in the bugged example: 

* call `testStarted `
* call `testFailure ` (this is attached to test, so we don't record it as unbounded, since we record all of these in `testFinished`)
* do not call `testFinished` (since this was kind of failure that prevents runner to continue)
* call `testRunFinished` 

since during call of the method `testFailure` we don't yet know if runner will call `testFinished` or not (normally we'd like to record all the failures in `testFinished`) we can not make a decision yet. 

 `testFinished` wraps up the test and makes `resultListener = null`, so check `resultListener != null` detects where some test was started but not finished. In this case we otherwise will lose the error causing runner to terminate the execution.

bobyangyf:
This will be temporarily reverted as it's causing some weird test behaviour internally.

romanoid:
I hope it means it uncovered your test failures were silently swallowed too. 

Please let me know (what you can), once you know anything.

romanoid:
@bobyangyf any updates?

bobyangyf:
Sorry I got sick and didn't get a chance to look at this yet. 
Our test discovery with this change started running tests with method of "null".
I'll look at it soon and re commit your stuff.

romanoid:
Interesting. Test result report (out xml) has method="null" for failures of runners that happened outside of running any method. 

I.e. failures that this change intended to detect will look exactly like this.

bobyangyf:
I think I found the issue.
Our logic in determining whether to run with timeouts fails with certain custom runners, and causes nulls down the line in our infra.
I'll be fixing that and relanding your diff.

PR #2014, opened 2018-08-30
Download statically built python for jitpack builds

kageiit:
Jitpack recently changed their build environment to no longer have python installed (See https://github.com/jitpack/jitpack.io/issues/3200)

This PR downloads a statically built version of python with no system dependencies from https://github.com/kageiit/jitpack-python for use inside jitpack builds. This lets users of buck continue to download and install buck built at any sha from jitpack.

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,5 +1,9 @@
 install:
    - set -exo pipefail
+   - mkdir python_tmp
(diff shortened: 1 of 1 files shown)


kageiit:
cc @ttsugriy @styurin

ttsugriy:
thank you, @kageiit !

PR #2016, opened 2018-08-30
Ignore unhandled metadata in cxx library

ghvg1313:
cxx_library raises if it gets unknow metadata creation requests.  Changing it to ignore them instead.


styurin:
Can you add a test for that? Also, some example would be great too.

ghvg1313:
Close this and merge into #1971 with tests added

PR #2017, opened 2018-08-31
Update d8 to latest

kageiit:
Updates d8 to upstream version that is also now compiled against asm 6.2.1 to be in sync with the asm update in buck recently.


kageiit:
cc @styurin

PR #2018, opened 2018-09-02
Add option to configure custom package exclusions in android binaries

kageiit:
This allows users of buck to configure additional patterns of archive paths to exclude from the apk (Analogous to https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.PackagingOptions.html). This is very useful to remove things like metadata that comes from kotlin standard libraries that just add bloat to the apk

### DIFF
--- a/src/com/facebook/buck/android/AabBuilderStep.java
+++ b/src/com/facebook/buck/android/AabBuilderStep.java
@@ -66,6 +66,7 @@ public class AabBuilderStep implements Step {
   private final ImmutableSet<ModuleInfo> modulesInfo;
   private final AppBuilderBase appBuilderBase;
   private final ImmutableSet<String> moduleNames;
+  private final ImmutableSet<String> packagingExcludePatterns;
 
   /**
    * @param modulesInfo A set of ModuleInfo containing information about modules to be built within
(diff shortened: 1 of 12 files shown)


kageiit:
cc @styurin ping! :)

kageiit:
This is completely broken by https://github.com/facebook/buck/commit/afdf9be370c9cae6fa94c1231395de10c8d77696

cc @styurin @jtorkkola Can you please suggest an alternative?

jtorkkola:
@kageiit Can you prune the jars before passing them into `ApkBuilder`, i.e. do essentially what `SignedJarBuilder::writeZip` does? I understand there'd be a performance hit with that approach, but we'd really like to remove as much of our AOSP fork as possible.

kageiit:
> Can you prune the jars before passing them into `ApkBuilder`, i.e. do essentially what `SignedJarBuilder::writeZip` does? I understand there'd be a performance hit with that approach, but we'd really like to remove as much of our AOSP fork as possible.

@jtorkkola That is not an option. That will require things like buck project to build all the pruned jar rules for IDE to work correctly. We also need to rewrite aar files which is more complex than jars. This seems like something buck should support tbh since the android gradle plugin already does. Im ok with not forking AOSP and open to suggestions of doing it in buck earlier than calling into  `ApkBuilder`. There are also several bits that buck forks from aosp - dx, d8 etc. I am not sure moving them to aosp forks will be easy due to how buck is architected

jtorkkola:
@kageiit I might have been unclear. I was suggesting that you create pruned copies of any jars that you feed into `ApkBuilder` in `ApkBuilderStep`. This should be equivalent to what you're trying to do in the PR, though with a performance hit.

kageiit:
This kind of pruning is mostly useful for release builds only. I think another option would be to do this as a post_process step after proguard etc. have run. I was going to hook in the android resource shrinker there to remove unused resources from the `.ap_` file (similar to what redex does, but those bits are not opensource). I will try to see if this can be hooked in there instead

PR #2030, opened 2018-09-19
Adds kapt apoptions to kotlin_library

mmdango:
This will allow users to pass in kapt annotation processor options to kapt.

Not sure if the way I wrote the docs is okay, but I don't think there are any examples of Hash type params so I wanted to be explicit.

Slack discussion here: https://buckbuild.slack.com/archives/C6N81MF55/p1537224378000100

### DIFF
--- a/docs/rule/android_library.soy
+++ b/docs/rule/android_library.soy
@@ -186,6 +186,20 @@ compiled class files and resources.
   {/param}
 {/call}
 
+{call buck.arg}
+  {param name: 'kapt_apoptions' /}
+  {param default : '{}' /}
+  {param desc}
+  Map of annotation processor options to pass into kapt via the apoptions plugin option. Each entry
+  should be a key value pair of the processor option and its value.
+  E.g.
+  kapt_apoptions = {
+    'someAnnotationOption': 'someValue'
+  }
+  More information here: https://kotlinlang.org/docs/reference/kapt.html
+  {/param}
+{/call}
+
 {call jvm_common.exported_deps}
   {param library: 'android_library' /}
 {/call}
--- a/docs/rule/android_library.soy
+++ b/docs/rule/android_library.soy
@@ -188,10 +188,9 @@ compiled class files and resources.
 
 {call buck.arg}
   {param name: 'kapt_ap_options' /}
-  {param default : '{}' /}
   {param desc}
   Map of annotation processor options to pass into kapt via the apoptions plugin option. Each entry
-  should be a key value pair of the processor option and its value.
+  should be a key value pair of the processor option and its value. Default is an empty map.
   E.g.
   kapt_ap_options = {
     'someAnnotationOption': 'someValue'
--- a/docs/rule/android_library.soy
+++ b/docs/rule/android_library.soy
@@ -187,13 +187,13 @@ compiled class files and resources.
 {/call}
 
 {call buck.arg}
-  {param name: 'kapt_apoptions' /}
+  {param name: 'kapt_ap_options' /}
   {param default : '{}' /}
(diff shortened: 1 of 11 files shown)


mmdango:
cc @thalescm @cwoodwar6

docs/rule/kotlin_library.soy
```diff
@@ -32,7 +32,7 @@ of the compiled class files and resources.
 
 {call buck.arg}
   {param name: 'srcs' /}
-  {param default: '[]' /}
+  {param default: '{}' /}
```
cwoodwar6: Was the intentional? Seems that you want this as part of your `kapt_apoptions`

mmdango: oof, good catch. yeah, i meant it for the kapt_apoptions param. thanks!

docs/rule/android_library.soy
```diff
@@ -171,6 +171,20 @@ compiled class files and resources.
   {/param}
 {/call}
 
+{call buck.arg}
+  {param name: 'kapt_apoptions' /}
```
kageiit: kapt_apoptions -> kapt_ap_options

sbalabanov:
tests are failing, seems like soy template is incorrect

mmdango:
@sbalabanov I believe I fixed the soy tests, but looks like there are unrelated tests that are failing in AppVeyor. I'm also a bit confused what the error message for the Travis CI is: https://travis-ci.org/facebook/buck/jobs/439005432#L7325 it just seems like the script always throws 1...

thalescm:
Another point is. If we open an API just for ap_options. it seems fair to open another one for every new type of thing the compiler accepts (and the old ones too) and not having the `extra_kotlinc_argmunts` at all. Which can lead us to a very confusing and over granular API that is hard to use and hard to maintain.

src/com/facebook/buck/jvm/kotlin/KotlinConfiguredCompilerFactory.java
```diff
@@ -65,6 +65,7 @@ public CompileToJarStepFactory configure(
         kotlinBuckConfig.getKotlinHomeLibraries(),
         kotlinArgs.getExtraKotlincArguments(),
         kotlinArgs.getAnnotationProcessingTool().orElse(AnnotationProcessingTool.KAPT),
+        kotlinArgs.getKaptApOptions(),
```
raviagarwal7: would be good to support macros as well in https://github.com/facebook/buck/commit/3e0b4cc444d86f565124807f729bbe7d9748c83c#diff-8bd16d403f308d9df417653053234adb in kaptApOptions

thalescm:
@mmdango would you mind rebasing it?

thalescm:
@mmdango Seems like there are still formatting issues.
What I normally do (to fix them all automatically) is to call [this script](https://github.com/facebook/buck/blob/master/scripts/travisci_test_java_file_forma)
But first deleting lines:
4-10;
32-33;

Also: Don't forget to replace `${TRAVIS_BRANCH}` on line 21 to whatever sha is the master you're pointing to (e.g `$(git rev-parse master)`.

At the end you'll have all necessary changes, just add them and commit :)

mmdango:
@thalescm done

thalescm:
@sbalabanov Can we get this merged please?

thalescm:
@styurin You imported this more than two weeks ago. Do you have any updates?

styurin:
Could you rebase it? There are some merge conflicts.

mmdango:
@styurin rebased. sorry for the delay, just got back from vacation

PR #2033, opened 2018-09-22
Update jacoco to 0.8.2

kageiit:
- Release notes: https://github.com/jacoco/jacoco/releases/tag/v0.8.2

Major changes include:
- Support for Java 11 and Java 12 class files
- Several kotlin bytecode related fixes
- Jacoco now depends on ASM 6.2.1
- The XML report now has an optional attribute sourcefilename on the class element to allow unambiguously relate classes to source files


ttsugriy:
looks like you forgot to update paths to source jars which is why travisci is broken

kageiit:
> looks like you forgot to update paths to source jars which is why travisci is broken

good catch! updated @ttsugriy

PR #2042, opened 2018-09-27
Respect cache_links in CxxStrip

### LINKED ISSUE #2041: CxxStrip output is cached even when cxx.cache_links = false
When setting the config value `cxx.cache_links = false`, `CxxLink` rules are not cached. The option does not apply to `CxxStrip` rules however, so the stripped binary is still cached. This has caused issues for us as certain changes will result in an app binary being relinked with a new dSYM created, but then the cached stripped binary is fetched and put in the app bundle. This causes symbolication to fail as the dSYM UUID is different from the binary.

The obvious option is to disable caching of `CxxStrip` when `cache_links = false` too, but I'm wondering if its ever safe to cache stripped binary output, at least on apple platforms. Is it possible that a change that would not affect the rulekeys would still cause a binary to be relinked with a different UUID?

rmaz:
Respect `cxx.cache_links` for the stripped binaries too. This ensures that cached output corresponds to the equivalent cached binary so we do not end up with dSYM UUID mismatches.

Fixes #2041


rmaz:
Also fine with just doing `isCacheable( return false; )`, or filtering for Apple platforms or something equivalent.

rmaz:
This requires further changes, in its current state this makes the problem even worse as the dSYM itself is getting cached but none of the binaries are.

PR #2043, opened 2018-10-02
Fix mismatched exception type

ghvg1313:
This function doesn't catch Exceptions that are not throwable (e.g. `NoSuchBuildTargetException`), revert the parameter back to `Exception`

### DIFF
--- a/src/com/facebook/buck/util/MoreExceptions.java
+++ b/src/com/facebook/buck/util/MoreExceptions.java
@@ -24,14 +24,14 @@ public class MoreExceptions {
   private MoreExceptions() {}
 
   /**
-   * @param throwable the Throwable to get error message from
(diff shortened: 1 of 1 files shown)


ghvg1313:
@timmych Let me know if there's anything wrong or a fix is already in progress

timmych:
@ghvg1313 can you explain why this function doesnt work as you expected? Throwable is a [parent class of Exception](https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/Exception.html) so I dont understand why it doesn't work for `NoSuchBuildTargetException`

ghvg1313:
Sorry this might've been an ant issue, looks like the patch was only partially applied when 
I rebase and tested on my end, the issue is fixed after a clean ant, sorry for the confusion

PR #2054, opened 2018-10-10
Allow for thread_core_ratio of above 1.0 if reserved_core isn't set

keithprestonsc:
This changes the logic to make thread_core above 1 work

Previously we unconditionally ran:
if (scaledValue > threadLimit) {
     scaledValue = threadLimit;
}

Which eliminated values over 1.0.

Now we only run it if reserved_core is set

Also with the new logic
"thread_core_ratio_max_cores is larger than thread_core_ratio_reserved_cores allows");
is not necessary.


v-jizhang:
CLA not signed.

PR #2055, opened 2018-10-10
Fix conditional bundle loader

ghvg1313:
It seems like xcode 9.3 has changed the evaluation of conditional bundler and it no longer works if the contents bit resolves to something empty.
Instead of using the maybe available macro, resolve to a full path based on if it should apply to deep or shallow bundles


williamtwilson:
This is great! I'm doing some additional testing on our end, but I'm hoping to merge it soon.

PR #2056, opened 2018-10-10
Add common debug default settings for xcode project

ghvg1313:
Adds some common settings that makes sense for Debug config, including:
- 'DEAD_CODE_STRIPPING': 'NO'
- 'ONLY_ACTIVE_ARCH': 'YES'
- 'GCC_SYMBOLS_PRIVATE_EXTERN': 'NO'

Also fix the issue where default flags are wiped out if user defined custom project configs in the target level.

With the old approach (user defined config override the default configs), Buck have no context which config is the debug version, so now default configs are merged with the user-defined ones, with debug having the extra flags above.

### DIFF
--- a/src/com/facebook/buck/features/apple/project/CxxPlatformXcodeConfigGenerator.java
+++ b/src/com/facebook/buck/features/apple/project/CxxPlatformXcodeConfigGenerator.java
@@ -43,9 +43,23 @@ class CxxPlatformXcodeConfigGenerator {
 
   private CxxPlatformXcodeConfigGenerator() {}
 
-  public static ImmutableMap<String, ImmutableMap<String, String>>
-      getDefaultXcodeBuildConfigurationsFromCxxPlatform(
-          CxxPlatform cxxPlatform, Map<String, String> appendedConfig) {
+  public static ImmutableMap<String, String> getAppleXcodeTargetBuildConfigurationsFromCxxPlatform(
+      CxxPlatform cxxPlatform, Map<String, String> appendedConfig) {
+
+    ArrayList<String> notProcessedCxxFlags = new ArrayList<String>(cxxPlatform.getCxxflags());
+    LinkedHashMap<String, String> notProcessedAppendedConfig =
+        new LinkedHashMap<String, String>(appendedConfig);
+
+    ImmutableMap.Builder<String, String> configBuilder = ImmutableMap.builder();
+    setSdkRootAndDeploymentTargetValues(
+        configBuilder, cxxPlatform, notProcessedCxxFlags, notProcessedAppendedConfig);
+    removeArchsValue(notProcessedCxxFlags, notProcessedAppendedConfig);
+
+    return configBuilder.build();
+  }
+
+  public static ImmutableMap<String, String> getCxxXcodeTargetBuildConfigurationsFromCxxPlatform(
+      CxxPlatform cxxPlatform, Map<String, String> appendedConfig) {
 
     ArrayList<String> notProcessedCxxFlags = new ArrayList<String>(cxxPlatform.getCxxflags());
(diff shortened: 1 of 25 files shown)


ghvg1313:
@williamtwilson this might affect your work

ghvg1313:
Pending #2055 , will rebase and fix the test once that's merged

williamtwilson:
@ghvg1313, could you rebase this and fix the test? Otherwise looks good to me 👍

ghvg1313:
Will rebase once #2055 is merged

ghvg1313:
@williamtwilson Rebased, should be good to go

ghvg1313:
@williamtwilson ping

williamtwilson:
@hutley had concerns about this. Could you enumerate those concerns, @hutley?

hutley:
@ghvg1313  thanks for the pull request. As @williamtwilson mentioned I do have some concerns.  BUCK generally doesn't set default options but it does allow for an individual target to do so pretty easily and seems pretty flexible already. I am not sure that we should prescribe any set of default options since they can be set in the BUCK files pretty simply - for example using the "configs" param of https://github.com/facebook/buck/blob/master/src/com/facebook/buck/apple/AppleNativeTargetDescriptionArg.java#L31 - you can set something like configs = {"Debug": {'DEAD_CODE_STRIPPING': 'NO'}, "Release": {'DEAD_CODE_STRIPPING': 'YES'}}. If you wanted some code reuse across many targets, this can then be wrapped into a higher level macro where you pass your default config on the "configs" param to the native.cxx_library etc.

ghvg1313:
@hutley Thanks for the feedback! There're actually two goals of this PR, setting defaults and merge configs (the first one is more or less depending on the second)

For the defaults, my aim is to eliminate differences with Xcode, which has different defaults for debug config. While `DEAD_CODE_STRIPPING` might be a little bit too much, the other two have different settings for Xcode than default buck-generated project. Doing this might prevent new buck users/developers having wrong flags' expectations, also lowers on-boarding cost for new buck users and encourage them use it IMO.

As for merging defaults, I'm quite surprised that setting custom config would wipe out some very basic c flags (current c++ defaults in project generator), maybe I'm wrong but I'm quite certain this is not the original intention.

I'm happy to keep this open, let me know how do you guys think

hutley:
@ghvg1313  I just did a basic workspace with a new project via Xcode and of the three settings you mention only the ONLY_ACTIVE_ARCH has a value defined in the auto-generated config, the other two have the same default as you propose to set already. I am concerned about setting global defaults mainly because there was no merging previously so you are changing behavior for established users of BUCK. If you can put this behind a configuration option and just set the bare minimum of flags then that would minimize the changes to downstream users. Also, it would be good to have tests that exercise both with the config option and without.

ghvg1313:
@hutley sorry for the confusion, `GCC_SYMBOLS_PRIVATE_EXTERN` should works the same as `ENABLE_TESTABILITY`, only `ENABLE_TESTABILITY` has higher precedence over `GCC_SYMBOLS_PRIVATE_EXTERN`, the later is more lower-level control IMO. Enable this setting in debug avoid the issue test host can't access source code sometimes.

As for merging config, if no `Debug` is predefined, buck has no knowledge of which config is used for debug and might produce the wrong settings, since all the options here are only for debug.

I'll amend the test for other configs if we are all onboard with this.

shepting:
I would have _really_ appreciated some default settings when getting started with Buck. These could still be overridden as you mention, but would really help new engineers to enabled debugging right from the start.

ghvg1313:
@hutley @williamtwilson Concerns? Suggestions?

ghvg1313:
@hutley @williamtwilson The unit tests has been updated to exercise both assigning custom config and not. Should be good to import now

PR #2062, opened 2018-10-16
Reordering extraClasspath dependencies ahead of module dependencies 

naveenOnarayanan:
In order to prevent Intellij from picking the first incorrect `R.class` file from the module dependencies, we need to move its own `extraClasspath` dependencies higher than external modules such that the first lookup of `R.class` will come from the module's generated directory.

This is required for support layout-preview with BUCK in Intellij.


kageiit:
cc @mkillianey

kageiit:
cc @christolliday Could you please import this as well? This is required for android layout preview support

christolliday:
Sure, just one failing test - com.facebook.buck.features.project.intellij.ProjectIntegrationTest testAndroidResourcesInDependencies please fix and I'll merge.

PR #2065, opened 2018-10-16
Fix incorrect regex match check in AbstractLocationMacro

ejdhzopu:
The existing code will throw IllegalStateException if the match was unsuccessful: https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#group(java.lang.String) The check matcher.matches() in line 80 does nothing: if the match succeeded but the supplementary part is absent, matcher.group("output") will return null anyway.

### DIFF
--- a/test/com/facebook/buck/rules/macros/AbstractLocationMacroTest.java
+++ b/test/com/facebook/buck/rules/macros/AbstractLocationMacroTest.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2017-present Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


styurin:
Can you add some tests, please? Also, it may not look pretty when it fails, you probably need to use `HumanReadableExceptino`.

ejdhzopu:
Done! There were not unit tests for this method: it was tested indirectly by a couple of other tests. Added a separate unit test class.

test/com/facebook/buck/rules/macros/AbstractLocationMacroTest.java
```diff
@@ -0,0 +1,35 @@
...
+    assertThat(
+        AbstractLocationMacro.splitSupplementaryOutputPart("target[sup]").supplementaryOutput,
+        Matchers.equalTo(Optional.of("sup")));
+    assertThat(
+        AbstractLocationMacro.splitSupplementaryOutputPart("target").target,
+        Matchers.equalTo("target"));
+    assertThat(
+        AbstractLocationMacro.splitSupplementaryOutputPart("target").supplementaryOutput,
+        Matchers.equalTo(Optional.empty()));
+    try {
```
styurin: Could you move the test case with the exception to a separate method? Other test cases are homogeneous and can stay in one method, but the the test case with an error stands out.

ejdhzopu: Makes sense. Done.

PR #2066, opened 2018-10-17
Update d8 to latest

kageiit:
Fixes a new issue in D8 release builds processed via proguard

The problem is related to buck using D8 to process classes in separate buckets i.e predexing, in which case D8 incorrectly handles naming of lambdas that are expressed as method references when the same lambda expression is used in two (or more) Java classes that end up being placed in different buckets (dex files).

If classes containing lambda expressions are in different packages, D8 will generate code that will append the name of accessing class to the lambda class name, making the lambdas distinct. Since package names can be stripped down by pro-guard, this does not happen on proguarded builds. The fix is to have D8 append accessor name to lambda class name also in the case of an empty package.  Somewhat unfortunately, this change also partially defeats the effects of one of D8's optimizations that allows sharing the same lambda code between different classes in the same dex file which can increase the method count a bit.

Until buck's dexing infrastructure is refactored to fit D8's optimizations, this is the best workaround possible to ensure apk correctness with a very minimal hit to method count increase.


kageiit:
cc @styurin @msridhar

PR #2067, opened 2018-10-17
Adding support for uppercase ide type

naveenOnarayanan:
With this https://github.com/facebook/buck/commit/fb9dae6d8c66ec5c23e7bea25d170335912a646b we lost the ability for ide type to be case insensitive when running **buck project**

This diff allows us to both define case insensitive IDE type in both the **.buckconfig** as CLI through the **--ide** flag.


styurin:
It looks fine, but I'm just curios, is that something that blocks you in some way?

naveenOnarayanan:
> It looks fine, but I'm just curios, is that something that blocks you in some way?

Hey Tyurin, unfortunately we utilized the uppercase approach in our internal plugins when interacting with BUCK and have been passing in **INTELLIJ** whenever we run buck project. Due to the fact that plugin rollouts are slow, we needed to have backwards compatibility support.

styurin:
Could you re-format the source code? You can find instructions here: https://github.com/facebook/buck/blob/master/CONTRIBUTING.md#code-style

naveenOnarayanan:
Reformatted the code using `google-java-format 1.6`

PR #2072, opened 2018-10-19
Fix RES_FOLDERS_RELATIVE_PATH to be actually relative

kageiit:
This changes the behavior of buck project to correctly set the `RES_FOLDERS_RELATIVE_PATH` value to be relative to the module. The android intellij plugin expects a proper relative path and does not work when macros like `$MODULE_DIR` etc. are prepended to the path.

This change is required to support Android Layout Preview in buck project and intellij.

### DIFF
--- a/src/com/facebook/buck/features/project/intellij/IjProjectPaths.java
+++ b/src/com/facebook/buck/features/project/intellij/IjProjectPaths.java
@@ -71,6 +71,21 @@ public class IjProjectPaths {
     return getModuleDir(module).relativize(path);
   }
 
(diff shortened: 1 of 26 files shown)


christolliday:
Can you rebase and make this build? There's a method missing.

kageiit:
> Can you rebase and make this build? There's a method missing.

Will do!

kageiit:
@christolliday Fixed the missing method and rebased

kageiit:
@christolliday Fixed the tests

PR #2077, opened 2018-10-22
Fix default swift version fallback

ghvg1313:
Allow swift targets read and fallback to global Swift version setting, when the version is not specified on the target level.

Originally it will fallback to the compiler default version.

Also fix test with expired swift version


ghvg1313:
@williamtwilson ping

shepting:
This will be great to get integrated!

Issue #2073, opened 2018-10-19
Deprecating #default platform in favor of the inferred host platform


mzlee:
Enabling should_remap_host_platform switches the `#default` platform to be the inferred host platform instead (one of linux-86_64, windows-x86_64, macosx-x86_64, or freebsd-x86_64). If the configuration `cxx.host_platform` is set, then use that instead.

Effect:
* Building on multiple platforms will now see a difference between Linux, Windows, MacOS, and FreeBSD flavors
* Any hard coded `#default` matches for `platform_*` will no longer match

See https://github.com/facebook/buck/pull/1562 for the original pull request and discussion

mzlee:
Just an update here: testing across multiple platforms is made much easier due to this setting, so it's taking longer than originally predicted to switch the default. Sorry for the delay.

PR #2078, opened 2018-10-22
Switch the default for `should_remap_host_platform` from false to true

mzlee:
https://github.com/facebook/buck/issues/2073 announces the deprecation and switch of should_remap_host_platform.

Given the lack of opposition, the current timeline is as follows:
1. Switch the default from false to true next week (10/29)
2. Remove the false codepath in two more weeks unless there is any opposition

Thoughts?


ttsugriy:
@mzlee, the announcements are created using github issues with `announcement` tag :) You can land this change internally but only after creating a github issue :)

mzlee:
@ttsugriy https://github.com/facebook/buck/issues/2073 is the announcement issue. I was hoping they'd be automatically linked.

PR #2079, opened 2018-10-24
Escape paths in the query_paths macro expander

rmaz:
Escape paths output by the `$(query_paths )` macro, otherwise the space separated list of paths is not parseable for paths with spaces in them.

### DIFF
--- a/test/com/facebook/buck/rules/macros/testdata/query-paths/BUCK.fixture
+++ b/test/com/facebook/buck/rules/macros/testdata/query-paths/BUCK.fixture
@@ -7,7 +7,7 @@ export_file(
 # Has no deps
 java_library(
     name = "example",
-    resources = ["beta.txt"],
+    resources = ["beta.txt", "file with spaces.txt"],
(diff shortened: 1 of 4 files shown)


test/com/facebook/buck/rules/macros/testdata/query-paths/BUCK.fixture
```diff
@@ -20,5 +20,5 @@ genrule(
 genrule(
     name = "files",
     out = "files.txt",
-    cmd = "echo $(query_paths 'inputs(deps(:example))') >$OUT",
+    cmd = 'PATHS=($(query_paths "inputs(deps(:example))")); for p in "${PATHS[@]}"; do echo "$p"; done | sort > $OUT',
```
styurin: This is not compatible with Windows

rmaz: Updated to read from the filelist instead

PR #2084, opened 2018-11-02
Make Dummy R class compilation faster

kageiit:
Dummy R java files have no dependencies/annotation processors run on them. Using empty options makes sense to avoid any overhead from javac plugins/errorprone etc. as these files do not need to be analyzed by such tools


kageiit:
cc @styurin

kageiit:

PR #2086, opened 2018-11-02
Don't run resource union in MiniAapt

raviagarwal7:
- Resource union is already handled by `MergeAndroidResourcesStep.java`, which also has information about the entires parents.
- In MiniAapt resource union would load each resource file individually, merge them and put them in R.txt using natural ordering. This breaks styleable generation.


kageiit:
cc @styurin Can we please import this?

styurin:
Could you fix typos in the description?

artem-zinnatullin:
This PR seems to break kotlin compilation for us, can't reference resources from other module even tho there is an explicit dependency on it 🤔

kageiit:
@styurin any update on this?

styurin:
> Could you fix typos in the description?

> This PR seems to break kotlin compilation for us, can't reference resources from other module even tho there is an explicit dependency on it

artem-zinnatullin:
@raviagarwal7 can you try this patch on OkBuck's sample Kotlin project? 

As an alternative, I can try to come up with reproducible sample next week

artem-zinnatullin:
Looks like this change on its own doesn't break Kotlin, I'm updating OkBuck from 0.44.0 to 0.44.3 and got exact that issue again:

```console
instant-features/driver/ride-flow/pending/widgets/src/main/java/com/lyft/android/driver/rideflow/pending/widget/timer/FlatCountdownTimerView.kt:50:68: error: unresolved reference: color
        val defaultStartColor = context.resources.getCompatColor(R.color.design_core_ui_purple60)
                                                                   ^
instant-features/driver/ride-flow/pending/widgets/src/main/java/com/lyft/android/driver/rideflow/pending/widget/timer/FlatCountdownTimerView.kt:51:66: error: unresolved reference: color
        val defaultEndColor = context.resources.getCompatColor(R.color.design_core_ui_pink50)
```

Code was working fine on OkBuck 0.44.0, which is super weird. We're on Buck upstream https://github.com/facebook/buck/commit/a8000262f17495ce095930ffd8acd50bc5842123

Something is wrong in Android Graph enhancer or the way Buck computes cache keys, I think

I'll spawn a separate issue to track this

alexjlockwood:
FYI, this change breaks resource merging for us: https://github.com/facebook/buck/issues/2196

alexjlockwood:
@raviagarwal7 @kageiit I have a branch that reproduces the error we are seeing: https://github.com/alexjlockwood/okbuck/pull/1

PR #2088, opened 2018-11-06
Do not turn off framework detection by default

kageiit:
Buck projects should have the ability to let the IDE decide when certain frameworks (eg. spring) are applied. This changes the default project generation behavior to not disable framework detection anymore.

### DIFF
--- a/src/com/facebook/buck/features/project/intellij/templates/ij-misc.st
+++ b/src/com/facebook/buck/features/project/intellij/templates/ij-misc.st
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="FrameworkDetectionExcludesConfiguration" detection-enabled="false" />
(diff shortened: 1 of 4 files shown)


kageiit:
cc @mkillianey @christolliday

kageiit:
cc @styurin

kageiit:
@mkillianey @christolliday @styurin Can someone import this please?

mkillianey:
@kageiit – This was introduced in 9979531a5755545ca610382eb7c1abebc6234956, and I don't think we want it universally enabled (at least, not for Android facets, and maybe not for other module types that we expect `buck project` to handle for you, like Kotlin or Python).

Note that IntelliJ disables it in their own repo:

https://github.com/JetBrains/intellij-community/blob/06b984c337ee49320629320c721b82fe63f15263/.idea/misc.xml#L6

This repo selectively disables it for android/python, so maybe we could consider that approach, and exclude facets that we think buck should handle for you?

https://github.com/facebook/buck/blob/master/.idea/misc.xml#L38-L41

kageiit:
excluding facets that buck should handle by default is fine. I can make that change to the template

I can add android and python to start. anything else?

Issue #2032, opened 2018-09-21
Project generator fails to generate symlink tree according to `subfolder_glob`


ghvg1313:
While platform sources/headers are included in the project by the update abb6939ce214492dd1fbfd5b44edfc46d47ca3f9, there're couple of issues left with the approach.
- Exported platform headers are not included in the excluded sources
- They are not included in the project source tree
- If headers are included in buck by `subfolder_glob`, since symlink tree is never generated, one can only reference these headers by name instead of full path. This eventually prevent us exporting platform headers since the path/namespace is not preserved.

First twos are relatively easy to fix, to fix the last one, one solution I can think of is to generate symlink trees for each relative architecture, exclude them all and add them for different sdks like we are doing right now.

Wondering if there's any plan/ideas around this.

ghvg1313:
@williamtwilson Is there any progress or thoughts on this?

ghvg1313:
@williamtwilson It's definitely a step forward now that platform headers are exposed, they're not handled similarly to `platform_srcs`, which were excluded first and then added back based on platform regex. 

Current approach will fail if we have header name collision even when they are on different platform. For instance, consider the following target platform header specification:
```
exported_platform_headers = [
       ('iphone', subdir_glob([
            ('include/header/platform/ios', 'exposed.h'),
        ])),
        ('macos', subdir_glob([
            ('include/header/platform/macos', 'exposed.h'),
        ])),
]
```
generation will fail because of duplicated `exposed.h` in symlink tree.

I can think of two solutions:
1. Using an option to specify project generation, maybe `default_platform` somewhere in the buckconfig. However passing the option down runtime might be an issue, and you have to re-generate each time platform is switched.
2. Create symlink tree for different platform and include them per platform the same way as abb6939

PR #2101, opened 2018-11-20
Apple project generator respects default cxx platform

ghvg1313:
Resolves https://github.com/facebook/buck/issues/2032 using a different approach.

This PR essentially solve two issues:
1. `supported_platforms_regex` now won't be included in the undesired platform, originally this could create an empty project target or unsupported source/header being included.
2.   Header collision by `glob` usage in `exported_platform_headers`, this is covered in detail in https://github.com/facebook/buck/issues/2032

@williamtwilson While i don't think this is a very elegant solution it does the job. Since this reverted couple of commits (bd07dcd6b0fe90c938985f330b2c268880b12b56, 4ab71d6a16f1bd3174657300180beed784ab5d3b and abb6939ce214492dd1fbfd5b44edfc46d47ca3f9) and uses a different approach than excluding headers based on platform, I'm up for discussion

### DIFF
--- a/src/com/facebook/buck/apple/AppleBuildRules.java
+++ b/src/com/facebook/buck/apple/AppleBuildRules.java
@@ -16,7 +16,6 @@
 
 package com.facebook.buck.apple;
 
-import com.facebook.buck.apple.toolchain.AppleCxxPlatform;
 import com.facebook.buck.core.description.BaseDescription;
 import com.facebook.buck.core.model.BuildTarget;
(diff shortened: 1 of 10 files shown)


v-jizhang:
Sorry we never got back to this! Thank you for your contributions. At this point, we are closing this because it is quite out of date and it is not clear if there is still interest in this approach. If you are still interested in, please submit a new PR and we promise to look at it in a timely manner. Thank you (and sorry) again.

PR #2102, opened 2018-11-20
Add swift build target version to rulekey

ghvg1313:
Add swift archicture version into rule key.

This allow swift module to be rebuilt when buckconfig architecture version change (e.g. iphoneos_target_sdk_version = 9.3). 

Originally this would potentially cause issue since module is not rebuilt, and linking modules with different target architecture versions is not allowed -> build failure


ghvg1313:
@styurin @williamtwilson for visibility

shepting:
@styurin @williamtwilson @cjhopman Any chance we could land this diff?

PR #2115, opened 2018-11-28
fix swift version query in project gen

ghvg1313:
This fix a problem where project gen ignores the swift version when it's default

### DIFF
--- a/src/com/facebook/buck/features/apple/project/ProjectGenerator.java
+++ b/src/com/facebook/buck/features/apple/project/ProjectGenerator.java
@@ -1051,8 +1051,8 @@ public class ProjectGenerator {
         targetNodeWithSwiftArgs.flatMap(t -> t.getConstructorArg().getSwiftVersion());
     if (!targetExplicitSwiftVersion.isPresent()
         && (targetNode.getDescription() instanceof AppleLibraryDescription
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/features/apple/project/ProjectGenerator.java
```diff
@@ -1047,7 +1047,15 @@ private String generateConfigKey(String key, String platform) {
   private Optional<String> getSwiftVersionForTargetNode(TargetNode<?> targetNode) {
     Optional<TargetNode<SwiftCommonArg>> targetNodeWithSwiftArgs =
         TargetNodes.castArg(targetNode, SwiftCommonArg.class);
-    return targetNodeWithSwiftArgs.flatMap(t -> t.getConstructorArg().getSwiftVersion());
+    Optional<String> targetExplicitSwiftVersion =
+        targetNodeWithSwiftArgs.flatMap(t -> t.getConstructorArg().getSwiftVersion());
+    if (!targetExplicitSwiftVersion.isPresent()
+        && (targetNode.getDescription() instanceof AppleLibraryDescription
```
bobyangyf: Let's move all the instance of checks into a separate method so it looks nicer and we only need to call getDescription() once

PR #2118, opened 2018-12-02
Make proguard buck config support taking source target

raviagarwal7:


styurin:
This change will add proguard target to the deps, but where is it used?

raviagarwal7:
> This change will add proguard target to the deps, but where is it used?

`getProguardJarOverride` is the function which provides the `SourcePath` corresponding to the specified proguard target or progaurd jar path. The target is added to the deps to make sure it is built before the jar from `getProguardJarOverride` is used.

raviagarwal7:
@styurin -- tests seems to have passed. could you please import this. thanks!

kageiit:

kageiit:
Friendly Ping @styurin :)

PR #2120, opened 2018-12-02
Make groovy_home buck config support specifying source path

raviagarwal7:


src/com/facebook/buck/jvm/groovy/GroovyConfiguredCompilerFactory.java
```diff
@@ -16,24 +16,36 @@
...
-  private final GroovyBuckConfig groovyBuckConfig;
+  @Nullable private final GroovyBuckConfig groovyBuckConfig;
+  @Nullable private Tool groovyc;
 
-  public GroovyConfiguredCompilerFactory(GroovyBuckConfig groovyBuckConfig) {
+  public GroovyConfiguredCompilerFactory(@Nullable GroovyBuckConfig groovyBuckConfig) {
     this.groovyBuckConfig = groovyBuckConfig;
   }
 
+  private Tool getGroovyc() {
```
styurin: This needs to be annotated with `@Nullable` since it can return `null` when both `groovyc` and `groovyBuckConfig` is `null`.

src/com/facebook/buck/jvm/groovy/GroovyConfiguredCompilerFactory.java
```diff
@@ -42,8 +45,17 @@ public CompileToJarStepFactory configure(
       ToolchainProvider toolchainProvider) {
     GroovyLibraryDescription.CoreArg groovyArgs = (CoreArg) Objects.requireNonNull(args);
     return new GroovycToJarStepFactory(
-        Objects.requireNonNull(groovyBuckConfig).getGroovyCompiler().get(),
+        groovyBuckConfig.getGroovyc(),
```
styurin: this can trigger NPE

raviagarwal7: added a precondition check since at this step it needs to be non null.

src/com/facebook/buck/jvm/groovy/GroovyLibraryDescription.java
```diff
@@ -26,31 +28,35 @@
...
 
   private final ToolchainProvider toolchainProvider;
-  private final GroovyBuckConfig groovyBuckConfig;
   private final JavaBuckConfig javaBuckConfig;
+  private final GroovyConfiguredCompilerFactory compilerFactory;
 
   public GroovyLibraryDescription(
       ToolchainProvider toolchainProvider,
-      GroovyBuckConfig groovyBuckConfig,
+      @Nullable GroovyBuckConfig groovyBuckConfig,
```
styurin: I'm not really following here, how it can be null?

raviagarwal7: this need not be null, but in tests this called with a null groovy buck config which is why I added `@Nullable`

https://github.com/facebook/buck/blob/3f1c33f5a4fed73559c890383374c6dcc9d9a59e/test/com/facebook/buck/jvm/groovy/GroovyLibraryBuilder.java#L51

raviagarwal7: providing a non null instance now.

raviagarwal7:
also seems like below tests are failing on master and also on this diff. not sure how to fix this.
```
com.facebook.buck.jvm.groovy.GroovyTestIntegrationTest#allTestsPassingMakesTheBuildResultASuccess
com.facebook.buck.jvm.groovy.GroovyTestIntegrationTest#oneTestFailingMakesTheBuildResultAFailure
```

styurin:
Tests are failing to compile

Issue #2108, opened 2018-11-26
Buck now requires all cells to be declared at the top-level?


njlr:
It seems that Buck now requires you to declare all cells at the top-level or you get this error:

> repositories.vendor-bar must exist in the root cell's cell mappings.

This is a hindrance because now you cannot nest cells without figuring out the transitive cell declarations, for example when doing sub-modules. 

I _think_ the old behaviour was to bail on conflicts, but basically allow the cells to be declared anywhere, which made more sense! 

Will this be changed back? 

Demo repo: https://github.com/njlr/buck-cells-test

njlr:
Relavent commit:https://github.com/facebook/buck/commit/0114639193efc4934b43e00536c6dc17a8c300a8

njlr:
Proposed solution: 

 - Starting from the root `repositories` list, we can recursively explore the cells, discovering new ones along the way. 
 - If 2 cell names point to different locations, then we can throw an error (this way cells are still canonical)
 - If a cell is not "reachable" from the root config, then we also throw an error. 

I have made an implementation here: https://github.com/njlr/buck/commit/8c5d3d417d19b5487f05f55fd93fd4d8299f6261

nikhedonia:
Did you guys had a chance to look at this?

bobyangyf:
This is intended behaviour. We want all cells to be declared at the root

PR #2126, opened 2018-12-06
Fixes 2108

njlr:
* Transitive targets are now discovered from the root manifest
 * Added tests

See: https://github.com/facebook/buck/issues/2108


v-jizhang:
Sorry we never got back to this! Thank you for your contributions. At this point, we are closing this because it is intended behavior as commented in https://github.com/facebook/buck/issues/2108. Thank you again.

PR #2133, opened 2018-12-10
use SDK 26 and build tools 26.0.2

dulmandakh:
Starting November 2018, Google requires apps to target at least API 26 for Play Store. And it might be best to drop older APIs from test and use API 26 as a base. So everyone must have upgraded to SDK 26 by now.

See https://support.google.com/googleplay/android-developer/answer/113469#targetsdk

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,6 +96,7 @@ matrix:
         - build-tools-26.0.2
         - android-26
         - addon-google_apis-google-23
+        - addon-google_apis-google-21
         - extra-android-support
     dist: trusty
     addons:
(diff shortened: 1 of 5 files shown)


dulmandakh:
getting error like this in NDK tests, please help

```java
[2018-12-10 16:21:33.461][warn ][tid:1091][com.facebook.buck.event.listener.LoggingBuildListener] Error: Can't read [/usr/local/android-sdk-25.2.3/platforms/android-26/android.jar] (Can't process class [org/apache/http/params/HttpConnectionParams.class] (Unsupported class version number [52.0] (maximum 51.0, Java 1.7)))
```

dulmandakh:
Once landed Circle CI setup will be much simpler.

PR #2146, opened 2018-12-31
Xcode project set LD_RUNPATH_SEARCH_PATHS for binaries with Swift

thelvis4:
With #1543, the project generator sets the `LD_RUNPATH_SEARCH_PATHS` for the targets with Swift dependencies. This expands on that for the case when the binary itself has Swift sources but no Swift dependencies.


thelvis4:
> @williamtwilson has imported this pull request. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.internmc.facebook.com/D13605525).

@williamtwilson ^Does this mean that the change is going to be merged soon? 🙃
Please let me know if the PR needs any modifications. Thanks!

PR #2148, opened 2019-01-03
Remove LcUuidContentsScrubber

rmaz:
The `LcUuidContentsScrubber` class rewrites the LC_UUID load command in the binary based on the contents of the binary. This seems unnecessary and potentially inaccurate, ld64 already generates the UUID based on the contents, with some exceptions for bitcode and STABS entries:

https://opensource.apple.com/source/ld64/ld64-253.6/src/ld/OutputFile.cpp.auto.html

`OutputFile::computeContentUUID()`


rmaz:
Tested by rebuilding the same binary a few times and verifying the UUID was the same each time.

rmaz:
Abandoning in favour of #2150

PR #2150, opened 2019-01-04
Disable the OSO scrubber when not caching links

rmaz:
The Darwin OSO scrubber is used to relativize the paths to object files in the generated symbol table for linked code. This is to aid with debugging when sharing artifacts across machines. If the `cache_links` config option is `false`, we know that these artifacts are not going to be shared so we can disable the scrubbing steps.

This resulted in a > 20% speedup on one of our build jobs.


rmaz:
Any updates on this one?

PR #2163, opened 2019-01-15
[iOS][Umbrella] Do not include Target-Swift.h in umbrella header

dfed:
This PR removes `#import <Target/Target-Swift.h>` from the `Target.h` umbrella file for each module named "Target".

Umbrella files should not have the generated `-Swift.h` file included in them. The generated `-Swift.h` should be used by the module map, not by the umbrella header.

### DIFF
--- a/src/com/facebook/buck/apple/clang/UmbrellaHeader.java
+++ b/src/com/facebook/buck/apple/clang/UmbrellaHeader.java
@@ -29,8 +29,7 @@ public class UmbrellaHeader {
 
   public String render() {
     String swiftGeneratedHeader = String.format("%s-Swift.h", targetName);
-    return headerNames
-        .stream()
+    return headerNames.stream()
         // Remove the Target-Swift.h header file from the list of header names.
         // The umbrella header should not import the generated Objective-C header.
         // It is the job of the module map to make sure that Swift code is accessible
--- a/src/com/facebook/buck/apple/clang/UmbrellaHeader.java
+++ b/src/com/facebook/buck/apple/clang/UmbrellaHeader.java
@@ -28,7 +28,14 @@ public class UmbrellaHeader {
   }
 
   public String render() {
-    return headerNames.stream()
+    String swiftGeneratedHeader = String.format("%s-Swift.h", targetName);
+    return headerNames
+        .stream()
+        // Remove the Target-Swift.h header file from the list of header names.
+        // The umbrella header should not import the generated Objective-C header.
+        // It is the job of the module map to make sure that Swift code is accessible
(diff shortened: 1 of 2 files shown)


dfed:
@rmaz since I'm doing the filtering work in `UmbrellaHeader.java` (which generates the umbrella header), rather than the code that calls into `UmbrellaHeader`, I can get away with only doing the work in one spot.

However, your test is on point, and I've included it in this PR.

dfed:
A friendly bump. Would love to get this reviewed.

dfed:
Just force-pushed to fix conflicts.

dfed:
Looks like CI on this branch is failing in exactly the same spots as [master](https://travis-ci.org/facebook/buck/jobs/523750937). This change should be good to review @styurin and @williamtwilson.

dfed:
Another friendly bump

src/com/facebook/buck/apple/clang/UmbrellaHeader.java
```diff
@@ -28,7 +28,13 @@ public UmbrellaHeader(String targetName, ImmutableList<String> headerNames) {
   }
 
   public String render() {
+    String swiftGeneratedHeader = String.format("%s-Swift.h", targetName);
```
styurin: Let's move this to a constructor and have a set in case we need to add more stuff in future. Something like that:

```
ImmutableSet<String> headersToExclude;

public UmbrellaHeader(...) {
    ...
    headersToExclude = ImmutableSet.of(
        // Target-Swift.h header file should be excluded from the list of header names.
        // The umbrella header should not import the generated Objective-C header.
        // It is the job of the module map to make sure that Swift code is accessible
        // to Objective-C via the Target-Swift.h header file.
      String.format("%s-Swift.h", targetName));
}
```

src/com/facebook/buck/apple/clang/UmbrellaHeader.java
```diff
@@ -28,7 +28,13 @@ public UmbrellaHeader(String targetName, ImmutableList<String> headerNames) {
...
     return headerNames.stream()
+        // Remove the Target-Swift.h header file from the list of header names.
+        // The umbrella header should not import the generated Objective-C header.
+        // It is the job of the module map to make sure that Swift code is accessible
+        // to Objective-C via the Target-Swift.h header file.
+        .filter(x -> !x.equals(swiftGeneratedHeader))
```
styurin: With that change this would be

```
.filter(header -> !headersToExclude.containt(header))
```

dfed:
Thank you for the review @styurin! I've updated the PR to address your feedback.

PR #2174, opened 2019-01-28
Support matching file extensions with a dot

### LINKED ISSUE #2173: Kotlin code in srcs.zip not included in built jar
I'm generating some Kotlin code from a genrule and consuming it as a dependency in another Kotlin library, but the generated classes are not found.

This seems to be because in `KotlincToJarStepFactory`, it only expands the source paths if a srcs zip file exists. The srcs zip path is properly added as the first in the list, but the `PatternMatchers` for SRC_ZIP returns false because it relies on `Files.getFileExtension` to compare against “src.zip”, but the bug is that `getFileExtension` always only return the “zip” part of the file extension, so it’ll never match.

jbarr21:
Resolves #2173

### DIFF
--- a/src/com/facebook/buck/io/filesystem/FileExtensionMatcher.java
+++ b/src/com/facebook/buck/io/filesystem/FileExtensionMatcher.java
@@ -19,7 +19,6 @@ package com.facebook.buck.io.filesystem;
 import com.facebook.buck.io.watchman.Capability;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.io.Files;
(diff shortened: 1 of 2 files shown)


kageiit:
cc @styurin

styurin:
Is that possible to add tests for kotlin use case?

styurin:
ping @jbarr21

kageiit:
It seems this was fixed by something @thalescm did recently. We have a fixture for this already https://github.com/facebook/buck/blob/master/test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/zip/BUCK.fixture and we no longer need this change

PR #2176, opened 2019-01-30
Reusing same classesDir for kapt and javac

### LINKED ISSUE #1897: AutoService generated META-INF files is not getting packages into the kotlin_library's output jar.
https://github.com/google/auto/tree/master/service

Works on a `java_library` jar but fails on `kotlin_library`.

thalescm:
Resolves #1897


styurin:
Tests are broken

thalescm:
Tests failure seems unrelated

styurin:
Is that possible to add an integration test?

thalescm:
@styurin done :)

thalescm:
@styurin any news?

thalescm:
The way of adding an integration for the test, without reimplementing AutoService would be to add it.

Could you provide an actual example (rule name) of a test who access resources on buck itself please, I can try to modify that part to minimize dependencies.

But reimplementing AutoService makes no sense, so would be either keeping the rest of the jars, or reverting those additions

thalescm:
@styurin Should be better now if you could take another look :)

styurin:
Just wanted to clarify, this change modifies an existing test that would fail if that change hadn't been introduced, right?

thalescm:
@styurin Right,
By removing the file at `META-INF/services/` describing which annotation processor should be run, the tests begin to fail (as expected).

Even after adding AutoService (an annotation processor that automatically generates the aforementioned file for you), the tests continue to fail.

The change therefore solves the issue (and the tests that existed are still valid, with the addition of this new validation)

thalescm:
@styurin Do you need any other changes for this PR?

thalescm:
@styurin could you please take another look?

PR #2185, opened 2019-02-05
Use absolute paths for DERIVED_FILE_DIR

dfed:
If a `.mlmodel` is added to an Xcode target where that module's `DERIVED_FILE_DIR` is set to a relative path, Xcode will crash.

This PR sets `DERIVED_FILE_DIR` to be an absolute path, which allows `.mlmodel`s to be included in generated Xcode projects without Xcode crashing.


dfed:
Friendly bump for a review 🙂

styurin:
There are some legitimate errors in Travis CI

dfed:
Looks like I need to update com.facebook.buck.features.apple.project.ProjectGeneratorTest. Will do so and circle back.

dfed:
`ProjectGeneratorTest` is now passing after I updated the unit test to test the new behavior @styurin.

Issue #2182, opened 2019-02-04
android_bundle cannot resolve `tools.proguard` configuration


thalescm:
I’m getting a `Rule for target '//path/to:target' could not be resolved.`
The rule exists and it *is* a SourcePath on it’s declaration.

To be more specific:
When defining a custom proguard jar (tools.proguard) config, when creating `android_bundle`, an error happens when parsing buck files:
```
Rule for target '//path/to:target' could not be resolved.
    When amending proguardJarOverride.
```
This only happens for bundles (doesn't happen for `android_binary`), even though both will try to resolve that rule inside `NonPreDexedDexBuildable`, which in it’s turn is created by `AndroidBinaryGraphEnhancer` (by `AndroidBinaryGraphEnhancerFactory`)
Even providing both rules the same instance of `AndroidBinaryGraphEnhancerFactory` results in the error above (today they are not https://github.com/facebook/buck/blob/master/src/com/facebook/buck/android/AndroidDescriptionsProvider.java#L68)

I’m totally willing to create a pull request fixing this, but finding the actual root source of the problem is the real problem now, as I'm out of ideas (everything seems correct), so posting here seeking for help.

thalescm:
@joelmccall @styurin if you have any idea, any help would be appreciated :)

PR #2193, opened 2019-02-07
Applies #2118 for android app bundles

thalescm:
* Keep consistency between `AndroidBundleDescription` and `AndroidBinaryDescription` (if they are supposed to be equal, they should be equal on the API as well)
* Resolves https://github.com/facebook/buck/issues/2182

### DIFF
--- a/src/com/facebook/buck/android/AndroidBundleDescription.java
+++ b/src/com/facebook/buck/android/AndroidBundleDescription.java
@@ -47,6 +47,7 @@ import com.facebook.buck.jvm.java.JavaOptions;
 import com.facebook.buck.jvm.java.JavacFactory;
 import com.facebook.buck.jvm.java.toolchain.JavaOptionsProvider;
 import com.facebook.buck.rules.macros.StringWithMacros;
+import com.facebook.buck.util.Optionals;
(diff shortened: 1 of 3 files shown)


src/com/facebook/buck/android/AndroidBundleDescription.java
```diff
@@ -243,12 +244,13 @@ public void findDepsForTargetFromConstructorArgs(
       ImmutableCollection.Builder<BuildTarget> targetGraphOnlyDepsBuilder) {
     javacFactory.addParseTimeDeps(targetGraphOnlyDepsBuilder, null);
     javaOptions.get().addParseTimeDeps(targetGraphOnlyDepsBuilder);
+
+    Optionals.addIfPresent(proGuardConfig.getProguardTarget(), extraDepsBuilder);
```
styurin: So this is the only change in this PR that is actually fixing the logic, right? Is the rest of the changes for consistency with AndroidBinary?

thalescm: @styurin right

thalescm:
@styurin any news on the import?

Issue #2201, opened 2019-02-22
Inconsistency between watchman and native globbing 


naveenOnarayanan:
Seems like there is a discrepancy between the watchman globbing and native globbing with regards to handling wild characters. The native globber explicitly requires individual segmented wild characters whereas watchman does not seem to have that requirement. This is causing inconsistencies in building / running targets on environments with/without watchman.

**Example target setup**
## Directory Structure
```
tests/
  | --> test.py 
  | --> foo/bar/data/
       |----> a.txt
       |----> b.txt
       |----> c.txt
```

## Target definition
```
python_test(
    name = "test_main",
    srcs = glob([
        'tests/*.py',
    ]),
    resources = glob([
        'tests/**data/**'
    ])
)
```

## Watchman globbing
* Passes and is able to add `tests/foo/bar/data/*` as resources

## Native globbing
* Fails with the following error:
```
ERROR:..scripts/BUCK:50:17: Traceback (most recent call last):
	File "scripts/BUCK", line 42
		python_test(name = "test_main", srcs = glob(["..."]), <2 more arguments>)
	File "scripts/BUCK", line 50, in python_test
		glob(["tests/**data/**"])
recursive wildcard must be its own segment (in glob pattern 'tests/**data/**')
```

naveenOnarayanan:
Resolved with latest changes

PR #2202, opened 2019-02-22
Fixed Hybrid globber to fail on incorrectly defined glob paths.

romanoid:
This fixes the https://github.com/facebook/buck/issues/2201 .

The problem was caused by watchman globbing accepting the patterns that are not correct from bazellib globbing point of view.

Luckily the path validation is exposed as public API.


romanoid:
cc @styurin can you please take a look?

PR #2210, opened 2019-03-01
throw exception when no source file

linzhp:
Stop the build when there is no source file in a package. This provides better error message to help debugging. If we let the build continue, the final .a file will be empty, and the build only fails when compiling another package that imports the empty package, making the error message misleading.

@styurin @mkaczanowski Could you take a look?


styurin:
Does it fail in some other way if sources are empty? Many folks use libraries with empty sources to combine dependencies into one target, is that supported in go?

v-jizhang:
Thank you for your contributions. We are closing this because it is by design as commented by styurin. Thank you again.

PR #2213, opened 2019-03-04
Fix instrumentation apk desugaring

kageiit:
Instrumentation apks were not being passed the full classpath for interface method desugaring due to the apkUnderTest's classpath being removed ahead of time from the dexing call. This change makes the full classpath available and ensures that instumentation apks are also correctly desugared by d8 when enabled.

### DIFF
--- a/src/com/facebook/buck/android/NonPreDexedDexBuildable.java
+++ b/src/com/facebook/buck/android/NonPreDexedDexBuildable.java
@@ -84,7 +84,7 @@ import javax.annotation.Nullable;
 import org.immutables.value.Value;
 
 class NonPreDexedDexBuildable extends AbstractBuildRule {
-  @AddToRuleKey private final ImmutableSortedSet<SourcePath> additionalJarsForProguard;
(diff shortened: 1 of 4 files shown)


kageiit:
cc @styurin @demon-xxi

src/com/facebook/buck/android/NonPreDexedDexBuildable.java
```diff
@@ -178,7 +178,7 @@
       boolean desugarInterfaceMethods) {
     super(buildTarget, filesystem);
     this.androidPlatformTarget = androidPlatformTarget;
-    this.additionalJarsForProguard = additionalJarsForProguard;
```
demon-xxi: Does `this.additionalJarsForProguard` ever get assigned or we don't use it at all?

kageiit: Its been renamed to make the purpose clear. We use it below

styurin:
Please, rebase on master

kageiit:
@styurin Rebased and fixed conflicts

PR #2215, opened 2019-03-04
Removed unnecessary files from META-INF going to final bundle

thalescm:


kageiit:
cc @styurin

styurin:
I'm not sure this is a reasonable approach to hard the list of entries. Why these entries are there and can they be stripped out before? What if there are more entries that needs to be removed?

kageiit:
I think the ApkBuilder from aosp does this implicitly today for apks in buck. What alternative approach would you suggest @styurin 

Can we follow something like https://github.com/facebook/buck/blob/master/src/com/facebook/buck/android/ApkBuilderStep.java#L178 ?

styurin:
I was thinking about having an argument on a rule that would list entries to remove, but these details may look very low level for top level rule.

kageiit:
> I was thinking about having an argument on a rule that would list entries to remove, but these details may look very low level for top level rule.

I think such a list would be very hard to maintain. I think what we wanted here is an include list rather than an exclude list. But having a list would be too low level as you noticed. This current change mirrors what the apk builder does in buck. Maybe we can add documentation to the code to make it clearer why we choose these entries only?

styurin:
> Maybe we can add documentation to the code to make it clearer why we choose these entries only?

Sounds good.

thalescm:
@styurin Se if you think that is enough please :)

thalescm:
Couldn't find any relationship of the CI failures with this PR, so they seem unrelated to me

styurin:
Sorry for late comments, I forgot to post them

thalescm:
@styurin if you can see if this makes it more readable pls

src/com/facebook/buck/android/AabBuilderStep.java
```diff
@@ -419,6 +418,41 @@ private void packageFile(
...
+  private boolean isEntryPackageable(ZipEntry entry) {
+    return isDirectoryEntryPackageable(entry) || isFileEntryPackageable(entry);
+  }
+
+  private boolean isDirectoryEntryPackageable(ZipEntry entry) {
+    return entry.isDirectory() && ApkBuilder.checkFolderForPackaging(entry.getName());
+  }
+
+  private boolean isFileEntryPackageable(ZipEntry entry) {
+    String entryName = entry.getName();
```
bobyangyf: don't need this. Just refer to entry.getName() in both places

thalescm:
@bobyangyf done 👍 
@styurin how did the import go?

PR #2227, opened 2019-04-02
Include Prebuilt jar's transitiveClasspath when building the javaplugin properties 

raviagarwal7:
PrebuiltJar has `deps` which is exported by default. Include it’s transitiveClasspath when building the javaplugin properties (https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/AbstractJavacPluginProperties.java#L97).

Previously: With rules defined below `dagger` processor will only have `javaDep` and `firstLevelPrebuilt` in its classpath.

After:  With rules defined below `dagger` processor will only have `javaDep`, `firstLevelPrebuilt` and `transitivePrebuilt` as well. Same is the behaviour if the prebuilt is added as a dependency to a java library since deps of prebuilt are exported by default.
```
java_annotation_processor(
    name = 'dagger',
    deps = [
         '//:javaDep',
         '//lib/firstLevelPrebuilt'
     ]
)

prebuilt_jar(
    name = 'firstLevelPrebuilt'
    deps = [
        '//lib:transitivePrebuilt'
    ]
)

prebuilt_jar(
    name = 'transitivePrebuilt'
)
```

Added Tests.

### DIFF
--- a/src/com/facebook/buck/jvm/java/PrebuiltJar.java
+++ b/src/com/facebook/buck/jvm/java/PrebuiltJar.java
@@ -61,7 +61,6 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.function.Supplier;
 
-@BuildsAnnotationProcessor
 public class PrebuiltJar extends AbstractBuildRuleWithDeclaredAndExtraDeps
     implements AndroidPackageable,
         ExportDependencies,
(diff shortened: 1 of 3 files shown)


raviagarwal7:
cc: @thalescm, @styurin

PR #2231, opened 2019-04-03
Added path to bundle config json on android_bundle rule

thalescm:


thalescm:
Tests failures seems unrelated

thalescm:
@styurin mind taking a look? Thanks!

PR #2237, opened 2019-04-13
Small animation on main page button hovers

joerez:


styurin:
Why? How does it look?

PR #2239, opened 2019-04-16
Update end to end test to pass in TravisCI

### LINKED ISSUE #2229: Travis Failure after commit "Generate new UUID if running buck from within buck"
@philipjameson The travis-ci build has been failing for ~1 month due to commit ce56dfe511ad8bac37cdcb30792fcf036b8fee11

Can you take a look at it?

philipjameson:
Fix issue where TravisCI sometimes does not recognize symlinks

This integration test works by checking what builds were invoked, and their UUIDs. For some reason on TravisCI (only), one of the two last_* links in buck-out/log are reporting 'true' for `Files.isSymbolicLink`, and it would cause last_querycommand and the UUID based dir to show up in the logs. Looking at the code in GlobalStateManager, this should work, but unblock travisci for now, since this test isn't failing anywhere else

Fixes #2229


philipjameson:
@styurin is landing #2243 instead to handle this. Abandoning.

PR #2245, opened 2019-04-18
Update d8 to latest

kageiit:
- Update d8 to version 1.5.20-dev
- Also update libraries d8 depends on that need later versions
- Pulls in latest changes for D8 now included in android gradle plugin 3.4 which was released yesterday


kageiit:
cc @styurin

kageiit:
cc @justinjhendrick

styurin:
Could you describe what are the changes?

kageiit:
> Could you describe what are the changes?

Done

justinjhendrick:
I'd prefer to choose a revision from a release branch rather than master. A release branch is less likely to have bugs, in general. Is there something we need from the master branch that isn't available yet on 1.4.88?

kageiit:
> I'd prefer to choose a revision from a release branch rather than master. A release branch is less likely to have bugs, in general. Is there something we need from the master branch that isn't available yet on 1.4.88?

I am unable to find the base commit for 1.4.88. Can you point me to it? I can rebase. I tried 1.4.87 but it does not rebase cleanly and has some very unrelated changes

justinjhendrick:
I really just meant the most recent commit on the 1.4 branch, I don't think 1.4.88 is tagged yet. 1.4.87 would be fine too.

>  I tried 1.4.87 but it does not rebase cleanly and has some very unrelated changes

Okay, so maybe it's not feasible this time. But in general I'd like to be on a release build. Would it be any cleaner the next time we update D8 to do it on a release branch?

kageiit:
> I really just meant the most recent commit on the 1.4 branch, I don't think 1.4.88 is tagged yet. 1.4.87 would be fine too.
> 
> > I tried 1.4.87 but it does not rebase cleanly and has some very unrelated changes
> 
> Okay, so maybe it's not feasible this time. But in general I'd like to be on a release branch. Would it be any cleaner the next time we update D8 to do it on a release branch?

It would definitely be cleaner for sure. Next time I rebase, will do so on a release branch and do some internal testing first before opening a pr :)

kageiit:
cc @styurin can we please import this now?

justinjhendrick:
@kageiit, which parts of this patch are suitable for an upstream push to D8? I ask because smaller patches are easier to maintain for all involved. Also, upstreaming these changes would make D8 more compatible with Buck and Redex by default.

kageiit:
> @kageiit, which parts of this patch are suitable for an upstream push to D8? I ask because smaller patches are easier to maintain for all involved. Also, upstreaming these changes would make D8 more compatible with Buck and Redex by default.

All the patches are due to how buck's unique build needs

- Buck decides to handle dexing i.e not delegating splitting the dexes to d8 itseld and trying to do the split zip and dex each jar separately itself
- Buck computes referenced resources which I tried to upstream before to D8 but they were not able to accept it as there is no need for them to support it for other build systems

PR #2247, opened 2019-04-19
Add support for exported_deps in android_resource

raviagarwal7:
Similar to how `java_library` & `android_library` can have `exported_deps`, this adds support for the same to `android_res`. Since `android_library` can't have resource deps in its `exported_deps`, this achieve that by having exported deps in `android_res` instead.


kageiit:
Cc @styurins

kageiit:
Can we please add tests?

v-jizhang:
Hi @raviagarwal7, 
Here is a comment from FB buck team: Unless there is a good reason from it, let's avoid exported_deps as far as possible.
For that reason, we won't merge this PR. Let us know if you have any questions. Thanks

PR #2250, opened 2019-04-23
Making kapt to run in a single call

thalescm:
Some advantages of this flow:
-  is more optimised then calling kotlinc 3 times
- We stop depending on kotlin-compiler-embeddable libs and kotlin-annotation-processing-gradle for the expected ones (kotlin-compiler and kotlin-annotation-processing)
- Code is now simpler with one call to kotlinc only

Caveats:
- We need to wait until kt `1.3.40` to do that (that's when I'll update tests kotlinhome and they'll start passing), so previous kotlin versions won't be supported. (won't work with kapt)


bolinfest:
I see that 1.3.40 was released recently, so maybe it's time to rebase this and take a look at merging it? https://blog.jetbrains.com/kotlin/2019/06/kotlin-1-3-40-released/

Also, please make sure to incorporate https://github.com/facebook/buck/commit/7b5315d8ed7c141b9576b65aeca97140ee42789c#diff-311c95e6c277e00e1457a5ba290e0739 when you rebase, as this was a critical fix to `KotlincToJarStepFactory.java`.

bolinfest:
Apparently 1.3.40 did not unblock this, as expected. https://youtrack.jetbrains.com/issue/KT-32535 is tracking the issue.

tyvsmith:
This seems to be unblocked with 1.3.50. I have a local workspace with it that passes all buck tests using our internal version of this diff. I'll work on getting that version pushed next week after Droidcon NYC if @thalescm doesn't beat me to it.

IanChilds:
Think this can now be closed since #2306 is landed. Thanks for doing this @thalescm!

v-jizhang:
Continued in #2306

PR #2256, opened 2019-05-07
java plugin arguments

thalescm:
Added support for `pluings_paramenters` for all targets that relies on `JvmLibraryArgs` as they base description.

This is used to add parameters for plugins defined on `plugins` parameter.

### DIFF
--- a/src/com/facebook/buck/jvm/java/AbstractJavacOptions.java
+++ b/src/com/facebook/buck/jvm/java/AbstractJavacOptions.java
@@ -236,10 +236,9 @@ abstract class AbstractJavacOptions implements AddsToRuleKey {
         // Javac plugins should be declared as:
         // '-Xplugin:<pluginName> <arg1> <arg2> <arg3> ...'
         String pluginName = properties.getProcessorNames().first();
-        String pluginArguments = getJavacPluginArguments()
-            .getOrDefault(pluginName, ImmutableList.of())
-            .stream()
-            .reduce("", (s, s2) -> s + " " + s2);
+        String pluginArguments =
+            getJavacPluginArguments().getOrDefault(pluginName, ImmutableList.of()).stream()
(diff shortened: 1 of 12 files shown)


thalescm:
CI error is unrelated...

thalescm:
cc @styurin

thalescm:
Any news here?

thalescm:
cc @jtorkkola

thalescm:
@jtorkkola any news on the import?

src/com/facebook/buck/jvm/java/AbstractJavacOptions.java
```diff
@@ -230,7 +233,13 @@ public void appendOptionsTo(
...
 
       for (ResolvedJavacPluginProperties properties : javacPlugins) {
-        optionsConsumer.addFlag("Xplugin:" + properties.getProcessorNames().first());
+        // Javac plugins should be declared as:
+        // '-Xplugin:<pluginName> <arg1> <arg2> <arg3> ...'
+        String pluginName = properties.getProcessorNames().first();
+        String pluginArguments =
+            getJavacPluginArguments().getOrDefault(pluginName, ImmutableList.of()).stream()
+                .reduce("", (s, s2) -> s + " " + s2);
+        optionsConsumer.addFlag("Xplugin:" + pluginName + pluginArguments);
```
jtorkkola: Nit: This is a bit of a misuse of `addFlag`. You probably want to do `addFlag` followed by `addExtras`. That would also allow you to get rid of the string reducing logic above. Unless you're doing this because the JVM literally expects this whole string as one atomic arg that includes space characters, in which case, add a comment.

v-jizhang:
Thank you for your contribution @thalescm. Could you take a look at jtorkkola's comments and answer his question? Let's get this PR merged and make buck OpenSource uptodate.
Thank you, let me know if you have any questions.
Jinlin

PR #2258, opened 2019-05-13
Shared class loader for javac plugins

thalescm:
Similar to how we do on annotation processors, we can optimize a lot plugins compile time by using shared classloaders for them.


src/com/facebook/buck/jvm/java/PluginLoaderJavaFileManager.java
```diff
@@ -0,0 +1,36 @@
...
+      ImmutableList<JavacPluginJsr199Fields> javacPlugins) {
+    super(fileManager);
+    this.pluginFactory = pluginFactory;
+    this.javacPlugins = javacPlugins;
+  }
+
+  @Override
+  public ClassLoader getClassLoader(Location location) {
+    // We only provide the shared classloader if there are plugins defined.
+    if (StandardLocation.ANNOTATION_PROCESSOR_PATH.equals(location) &&
```
jtorkkola: Under what circumstances is `getClassLoader` called? What happens when we have both annotation processors and plugins? We instantiate all APs in batch and call `CompilationTask::setProcessors` once in `Jsr199JavaInvocation`, but I'm wondering if there might be weird cases where we end up using the wrong classloader for an AP somehow.

src/com/facebook/buck/jvm/java/PluginLoaderJavaFileManager.java
```diff
@@ -0,0 +1,36 @@
+package com.facebook.buck.jvm.java;
```
jtorkkola: Please add copyright headers to all new files.

src/com/facebook/buck/jvm/java/PluginFactory.java
```diff
@@ -0,0 +1,47 @@
+package com.facebook.buck.jvm.java;
+
+import com.facebook.buck.util.ClassLoaderCache;
+import com.google.common.collect.ImmutableList;
+import java.net.URL;
+
+// Counter part of AnnotationProcessorFactory
+public class PluginFactory implements AutoCloseable {
```
jtorkkola: Can we add an integration test?

src/com/facebook/buck/jvm/java/PluginLoaderJavaFileManager.java
thalescm: On my local tests no AP was interfered by this and I could find also no indicators on javac (searching on open JDK code) that loaded processors again even if you provide loaded ones. So in my view, as we do the `setProcessors` call to set the processors, always, both can coexist fine.

src/com/facebook/buck/jvm/java/PluginFactory.java
thalescm: I've found a test on `AnnotationProcessorFactory`. Will replicate it's behaviour then in a test for this class, since they are very similar.
But overall there is no integration test that I could find for AP as well, so I actually have no clue on how to check if classes are being reloaded or not in an integration test.
If you have any suggestion please do tell :)

thalescm:
Test failure doesn't seem related

PR #2260, opened 2019-05-14
Add target chrome trace step event

thalescm:
Reverts https://github.com/facebook/buck/commit/22d997120c5424e026aca1aaca889db9f264a345#diff-bd53cf424d15a06cf508a1341b4c93d1
Add target information to step event on chrome trace file.

This is generally useful when parsing the chrome trace file in order to get information about about steps events and associate them with the given target being run.


thalescm:
cc @jtorkkola @styurin

cjhopman:
Can you make sure that in the trace viewer, when selecting across the row of steps (basically the second row of blocks for a build thread) that they are still categorized by the step category?

kageiit:
@thalescm may need a rebase

PR #2275, opened 2019-06-04
Enable interface desugaring for kotlin libraries by default

kageiit:


kageiit:
cc @styurin @bolinfest

PR #2298, opened 2019-07-30
Enable downloading files with .so, .dylib & .dll extensions.

raviagarwal7:
Sqlite4java resources are needed during tests.
https://mvnrepository.com/artifact/com.almworks.sqlite4java


raviagarwal7:
cc: @styurin

PR #2306, opened 2019-08-28
Bump Kotlin to 1.3.50 and Move KAPT to 1 pass

tyvsmith:
- Bumped Kotlin version to 1.3.50 jars in thirdparty and in test data dir.
- Applied @thalescm diff to move Kapt to 1 pass. (Taking over https://github.com/facebook/buck/pull/2250)

Some advantages of this flow:

- This is more optimised then calling kotlinc 3 times
- We stop depending on kotlin-compiler-embeddable libs and kotlin-annotation-processing-gradle for the expected ones (kotlin-compiler and kotlin-annotation-processing)
- Code is now simpler with one call to kotlinc only

Caveats:
Kotlin versions prior to 1.3.50 won't be supported since they don't support this Kapt configuration.


tyvsmith:
The test failures seem unrelated, except possibly the warning without public visibility:

- Travis quotes an OOM on the server while building unrelated modules.
- Appveyor calls out failed tests related to core classes and platform issues like "DefaultProjectFilesystemTest" and "CxxRawHeadersIntegrationTest" 
- Can't see the warning produced from internal CI as it's gated behind FB infra.

IanChilds:
Internally, this looks good. I'm just updating us to use 1.3.50 and then this will be ready to land, probably this afternoon or more likely on Monday. Thanks for doing it, really appreciate it! 

Do you have any data on the perf improvement from only having to call kotlinc once rather than three times?

tyvsmith:
Our benchmark pipeline that produced https://eng.uber.com/measuring-kotlin-build-performance/ is having some issues I need to look into. But I built that dataset on Kotlin 1.3.50 and saw 5-10% savings across a full clean build across a couple runs (about 45-55 seconds off a 13 min build). Once I get the full bench suite working again, I'll do some more thorough analysis and look specifically into the compilation time step.

IanChilds:
Unfortunately it seems like there is a bug in the `compile` version of KAPT. Here is a really simple example of something that fails to build (you have to make sure that it uses the KAPT-compilation, either by hard-coding or by adding an annotation processor):

```
class Example() {

    private val callback = object : Any() {
        val obj = Object()
    }

    fun call() {
        callback.obj
    }

}
```

We get an error that obj is unresolved in the `call` function. It works fine using normal compilation. 

This is blocking us and I don't have a fix right now so I will probably move back to using separate compilation steps until this is resolved. I think it's probably a bug on the Kotlin side rather than the Buck-implementation side but not certain about that. 

cc @tyvsmith

tyvsmith:
> Unfortunately it seems like there is a bug in the `compile` version of KAPT. Here is a really simple example of something that fails to build (you have to make sure that it uses the KAPT-compilation, either by hard-coding or by adding an annotation processor):
> 
> ```
> class Example() {
> 
>     private val callback = object : Any() {
>         val obj = Object()
>     }
> 
>     fun call() {
>         callback.obj
>     }
> 
> }
> ```
> 
> We get an error that obj is unresolved in the `call` function. It works fine using normal compilation.
> 
> This is blocking us and I don't have a fix right now so I will probably move back to using separate compilation steps until this is resolved. I think it's probably a bug on the Kotlin side rather than the Buck-implementation side but not certain about that.
> 
> cc @tyvsmith

Have you filed a Youtrack ticket yet? We should get this in front of @altsvetkov

tyvsmith:
Here's a patch example of that which fails tests on master.

```
diff --git a/test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/kotlinap/AnnotationProcessorKotlin.kt b/test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/kotlinap/AnnotationProcessorKotlin.kt
index d42f9a0ade..e284ec1035 100644
--- a/test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/kotlinap/AnnotationProcessorKotlin.kt
+++ b/test/com/facebook/buck/jvm/kotlin/testdata/kotlin_library_description/com/example/ap/kotlinap/AnnotationProcessorKotlin.kt
@@ -41,5 +41,22 @@ class AnnotationProcessorKotlin : AbstractProcessor() {
                         .build()
                 ).build()
         file.writeTo(File(genDir))
+
+        val test = """
+            package ${pkg}            
+            class ${name}_BAD_() {
+            
+                private val callback = object : Any() {
+                    val obj = Object()
+                }
+            
+                fun call() {
+                    callback.obj
+                }
+            
+            }
+    """.trimIndent()
+
+        File(genDir + "/${pkg.replace(".", "/")}/${fileName}_BAD_.kt").writeText(test)
     }
 }
```

IanChilds:
Haven't filed a ticket, I'll do that tomorrow unless you do it today

tyvsmith:
> Haven't filed a ticket, I'll do that tomorrow unless you do it today

I'm heading on vacation in a couple hours, so likely won't get to it before then.

IanChilds:
https://youtrack.jetbrains.com/issue/KT-33800

IanChilds:
Looks like jetbrains are targeting a fix for 1.3.60

Issue #2316, opened 2019-09-26
Timezone would effect ZipConstants getFakeTime


weiyanlin117:
When I am using buck http cache I am keep experiencing tons of warnings like:
```
error: XXXX debug map object file 'buck-out/gen/YYY.m.o' has changed (actual time is 1985-02-01 08:00:00.000000000, debug map time is 1985-02-01 00:00:00.000000000) since this executable was linked, file will be ignored
```
for every file I download from buck server.
Thus my build works fine but my breakpoint is keeping failure. 

After investigation, the root cause is here:

https://github.com/facebook/buck/blob/e1bfc4f4bf363d109c4654c784a32faa7ee2a192/src/com/facebook/buck/util/zip/ZipConstants.java#L42

`Calendar` create time would based on its environment local timezone,
for example 
```
import java.util.Calendar;
import java.util.TimeZone;

public class Main {

    public static void main(String[] args) {
        Calendar c = Calendar.getInstance();
        c.set(1985, Calendar.FEBRUARY, 1, 0, 0, 0);
        // my environment time is America/Los_Angeles
        System.out.println(c.getTimeInMillis());
        // 476092800837

        c.setTimeInMillis(476064000000L);
        System.out.println(c.getTimeInMillis());
        // 476064000000 
    }
}
```
 
https://github.com/facebook/buck/blob/e1bfc4f4bf363d109c4654c784a32faa7ee2a192/src/com/facebook/buck/util/zip/ZipConstants.java#L42
has no guarantee to be as same as 

https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/src/com/facebook/buck/cxx/toolchain/objectfile/ObjectFileCommonModificationDate.java#L25

If cache timestamp is not consist, all cache consumer would experience the issue I mentioned at beginning.

This issue introduced from https://github.com/facebook/buck/commit/35c93b88c04ea88d1f65f149a73b7c549f38389a#diff-b12d580cc80ce3c8f0f5d2549fa61f67R4702

compare to `ZipEntry `, `TarArchiveEntry` would not include timezone convert.
https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html#setTime-long-
https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html#setModTime-java.util.Date- 

[...]

PR #2317, opened 2019-09-26
Fix ArtifactUploader.java TarArchiveEntry timezone issue

weiyanlin117:
This fixed the issue at https://github.com/facebook/buck/issues/2316
This issue introduced from https://github.com/facebook/buck/commit/35c93b88c04ea88d1f65f149a73b7c549f38389a#diff-b12d580cc80ce3c8f0f5d2549fa61f67R4702

unlike `ZipEntry `, `TarArchiveEntry` would not include timezone convert.
https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipEntry.html#setTime-long-
https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html#setModTime-java.util.Date- 

Since `Calendar` class is timezone sensitive, we should not depend on Calendar generated unix timestamp in `TarArchiveEntry`,  but use 
https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/src/com/facebook/buck/cxx/toolchain/objectfile/ObjectFileCommonModificationDate.java#L25 directly as same as 
https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/src/com/facebook/buck/cxx/FileLastModifiedDateContentsScrubber.java#L28

since ObjectFileCommonModificationDate is more generic I moved it from `cxx.toolchain` to `util`.

`com.facebook.buck.cxx.toolchain.objectfile.ObjectFileCommonModificationDate` --> `com.facebook.buck.util.ObjectFileCommonModificationDate`


weiyanlin117:
do I need to worry about 
java.lang.OutOfMemoryError: Java heap space 
https://travis-ci.org/facebook/buck/jobs/590281983#L4745
https://travis-ci.org/facebook/buck/jobs/590281985#L4774

for https://travis-ci.org/facebook/buck/jobs/590281986#L6491
FAIL com.facebook.buck.remoteexecution.util.OutputsMaterializerTest
I test it locally it is passed passed

PASS     350ms  2 Passed   0 Skipped   0 Failed   com.facebook.buck.remoteexecution.util.OutputsMaterializerTest

for https://travis-ci.org/facebook/buck/jobs/590281987#L8075
Does it mean I need to use something to format ?

would like to have some ideas from your side @philipjameson 🙏🙏

weiyanlin117:
@gnuthor thanks for import, just resolved merge conflict

gnuthor:
@weiyanlin117 sharing comments from buck team "Sounds right, but why not fix ZipConstants.getFakeTime?"

If you could address that in your change, that'd be great

weiyanlin117:
because https://github.com/facebook/buck/blob/e1bfc4f4bf363d109c4654c784a32faa7ee2a192/src/com/facebook/buck/util/zip/ZipConstants.java#L29 
It is intentional to using time-zone sensitive `Calendar` inside `getFakeTime`, because ZipEntry is Timezone sensitive. From my understanding,  there is a hidden pair relationship between `getFakeTime` and `ZipEntry`

If I changed it, unit test https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/test/com/facebook/buck/util/zip/CustomZipEntryTest.java#L80
would be failed,
ZipEntry and getFakeTime are both Time-Zone Sensitive, if I only changed `getFakeTime` it would introduce more issue.

Thus if I modified `getFakeTime` function, I need to make sure there are no ZipEntry would use it from now and in future in buck code base.
That is a very strong hidden assumption.
Compare introducing this assumption, I chose to make sure TarArchiveEntry would use absolute unix timestamp which keep the changes clear and straightforward.  @gnuthor

PR #2338, opened 2019-11-11
Fix inefficient traversals in Java's getRuntimeDeps

mikekap:
- `PrebuiltJar` attempted to transitively fetch runtime deps. This means that if you had a chain of prebuilt jars depending on each other, e.g. [binary] -> A -> B -> C -> D, we would attempt to traverse the runtime deps of A 1x, B 2x, C 3x and D 4x. This is unncessary as `JavaBinary` does the transitive traversal itself, so just stop doing it.
 - `JavaBinary` was resolving transitive runtime deps via `concat(A, concat(B, concat(C, ...)))`. This is extremely inefficient since `concat` adds another stack frame for every call which can result in `StackOverflowException`. Use `flatMap` instead.

### DIFF
--- a/src/com/facebook/buck/jvm/java/JavaBinary.java
+++ b/src/com/facebook/buck/jvm/java/JavaBinary.java
@@ -242,14 +242,7 @@ public class JavaBinary extends AbstractBuildRuleWithDeclaredAndExtraDeps
 
   @Override
   public Stream<BuildTarget> getRuntimeDeps(BuildRuleResolver buildRuleResolver) {
-    Stream<BuildTarget> transitiveRuntimeDeps = Stream.of();
-    for (JavaLibrary lib : getTransitiveClasspathDeps()) {
-      // Skip ourself to avoid infinite recursion.
-      if (lib == this) continue;
-
-      transitiveRuntimeDeps =
-          Stream.concat(transitiveRuntimeDeps, lib.getRuntimeDeps(buildRuleResolver));
-    }
-    return transitiveRuntimeDeps;
+    return getTransitiveClasspathDeps().stream()
+        .flatMap(lib -> lib.getRuntimeDeps(buildRuleResolver));
   }
 }
--- a/src/com/facebook/buck/jvm/java/PrebuiltJar.java
+++ b/src/com/facebook/buck/jvm/java/PrebuiltJar.java
@@ -275,15 +275,7 @@ public class PrebuiltJar extends AbstractBuildRuleWithDeclaredAndExtraDeps
 
   @Override
   public Stream<BuildTarget> getRuntimeDeps(BuildRuleResolver buildRuleResolver) {
-    Stream<BuildTarget> transitiveRuntimeDeps = Stream.of();
-    for (JavaLibrary lib : getTransitiveClasspathDeps()) {
-      // Skip ourself to avoid infinite recursion.
-      if (lib == this) continue;
-
-      transitiveRuntimeDeps =
-          Stream.concat(transitiveRuntimeDeps, lib.getRuntimeDeps(buildRuleResolver));
-    }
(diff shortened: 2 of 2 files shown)


src/com/facebook/buck/jvm/java/JavaBinary.java
```diff
@@ -242,14 +242,7 @@ public boolean isCacheable() {
...
-    for (JavaLibrary lib : getTransitiveClasspathDeps()) {
-      // Skip ourself to avoid infinite recursion.
-      if (lib == this) continue;
-
-      transitiveRuntimeDeps =
-          Stream.concat(transitiveRuntimeDeps, lib.getRuntimeDeps(buildRuleResolver));
-    }
-    return transitiveRuntimeDeps;
+    return getTransitiveClasspathDeps().stream()
+        .flatMap(lib -> lib.getRuntimeDeps(buildRuleResolver));
```
bobyangyf: We probably still need to have a check that the lib is == this and return an empty stream in that case

src/com/facebook/buck/jvm/java/PrebuiltJar.java
```diff
@@ -275,15 +275,7 @@ public boolean neverMarkAsUnusedDependency() {
...
-    Stream<BuildTarget> transitiveRuntimeDeps = Stream.of();
-    for (JavaLibrary lib : getTransitiveClasspathDeps()) {
-      // Skip ourself to avoid infinite recursion.
-      if (lib == this) continue;
-
-      transitiveRuntimeDeps =
-          Stream.concat(transitiveRuntimeDeps, lib.getRuntimeDeps(buildRuleResolver));
-    }
-    return transitiveRuntimeDeps;
+    return Stream.of();
```
bobyangyf: I don't know that we can do this.
I doubt the PrebuiltJars are only ever dependencies of JavaBinary, so this will be wrong for all the other places that uses it.

src/com/facebook/buck/jvm/java/JavaBinary.java
mikekap: That was actually an impossible check - `getTransitiveClasspathDeps` is a list of `JavaLibrary`s, but `this` is a `JavaBinary` - they don't share any common ancestors so it's impossible to have that equality check ever be true.

src/com/facebook/buck/jvm/java/PrebuiltJar.java
mikekap: This change mirrors what `DefaultJavaLibrary` does - it does not attempt to get the runtime deps of its' deps: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/DefaultJavaLibrary.java#L436 (it's kind of a journey to trace through, but this is where there `Description` gets turned into that runtimeDeps list: https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/java/JavaLibraryDeps.java#L127). Both `prebuilt_jar` and `java_library` are noops when built standalone.

Presumably all other users (including `java_binary`) want to support both 3rd-party (`prebuilt_jar`) and locally-built (`java_library`) - in which case they must do the transitive closure themselves. Anyone who depended on `prebuilt_jar` hoisting runtime deps was likely already broken. The original PR that added these (https://github.com/facebook/buck/commit/b22d68aba5a046a91bac38363a5695e72afdb459) didn't touch anything other than `java_binary` though there's a chance more support has been added in the past 2 months.

mikekap:
@bobyangyf ping. I'm happy to change the java_binary traversal type, but afaict `getTransitiveClasspathDeps` is specifically pre-computed as an optimization. Let me know what you'd prefer to do. Without this change existing projects with 7-10 levels of `prebuilt_jar`s hang for minutes before starting the build. Do you think that's a behavior worth keeping?

PR #2348, opened 2019-11-18
Create schemes for APP_EXTENSION subtypes

bogo:
Buck will not auto-create schemes for app extensions that are subtypes of the `APP_EXTENSION` (`com.apple.product-type.app-extension `) type. Those are, among others, the iMessage apps. (`com.apple.product-type.app-extension.messages`)

To avoid adding all possible extension types to the `ProductTypes`, this PR modifies scheme creation to check if the product type BEGINS with the parent extension type. Open to other suggestions on how to fix this if this is deemed too hacky!

### DIFF
--- a/src/com/facebook/buck/features/apple/project/WorkspaceAndProjectGenerator.java
+++ b/src/com/facebook/buck/features/apple/project/WorkspaceAndProjectGenerator.java
@@ -858,9 +858,9 @@ public class WorkspaceAndProjectGenerator {
       return false;
     }
 
-    // Only create schemes for APP_EXTENSION.
+    // Only create schemes for APP_EXTENSION or its subtypes. (e.g. iMesage apps)
     ProductType productType = ProductType.of(bundleArg.getXcodeProductType().get());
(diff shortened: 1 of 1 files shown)


swgillespie:
From internal review:

> Our current system works fine for creating these. We would rather not complicate our Xcode projects with more schemes.

Let's open an issue if there's discussion to be had here.

bogo:
Got it. Opened #2353, since this is a confusing user experience.

FYI, this patch only adds the extension scheme if the app has an iMessage extension dependency, there is no impact otherwise.

v-jizhang:
Thank you for your contributions. At this point, we are closing this because as commented by swgillespie. Thank you again.

digitalbuddha:
Hi folks, how does this complicate your build? We love buck at dropbox and were hoping to help others with similar configurations. I don't feel like this use case is that out of the norm.

v-jizhang:
@digitalbuddha Thank you, I'll reopen this PR for more discussion.

PR #2354, opened 2019-12-04
External Kotlin bug fix, issue 2349

v-jizhang:
Summary
- When buck using external kotlinc, it encounters an internal error: "AddToRuleKey can only be applied to methods of Immutables." That's because ExternalKotlinc is not Immutableand AddToRuleKey can't be applied to one of its methods.
- The fix is to apply AddToRuleKey to a private field instead.

### DIFF
--- a/src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
+++ b/src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
@@ -43,11 +43,12 @@ import java.util.function.Supplier;
 
 /** kotlinc implemented as a separate binary. */
 public class ExternalKotlinc implements Kotlinc, AddsToRuleKey {
-  @AddToRuleKey private final String kotlinCompilerVersion;
+
   private static final KotlincVersion DEFAULT_VERSION = KotlincVersion.of("unknown version");
 
   private final Path pathToKotlinc;
   private final Supplier<KotlincVersion> version;
+  @AddToRuleKey private final String kotlinCompilerVersion;
 
   public ExternalKotlinc(Path pathToKotlinc) {
     this.pathToKotlinc = pathToKotlinc;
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
```diff
@@ -73,11 +73,11 @@ public ExternalKotlinc(Path pathToKotlinc) {
                 return KotlincVersion.of(output);
               }
             });
+    this.kotlinCompilerVersion = getKotlinCompilerVersion();
   }
 
   /** Returns the Kotlin version, or the path if version is unknown */
-  @AddToRuleKey
-  public String getKotlinCompilerVersion() {
+  private String getKotlinCompilerVersion() {
```
philipjameson: If you're running this in the constructor anyways to save to the field, it's probably best to just make this return the field. If it changes during runtime, it's already going to cause issues since we memoize version.get().

src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
```diff
@@ -43,7 +43,7 @@
 
 /** kotlinc implemented as a separate binary. */
 public class ExternalKotlinc implements Kotlinc, AddsToRuleKey {
-
+  @AddToRuleKey private final String kotlinCompilerVersion;
```
philipjameson: Mind dropping this down with pathToKotlinc?

v-jizhang: dropped it down. Thanks

src/com/facebook/buck/jvm/kotlin/ExternalKotlinc.java
v-jizhang: I made it a private helper function that is only called in the constructor and the function uses only "final" fields to produce a compiler version string. That means there won't be any changes during runtime.

PR #2364, opened 2020-01-02
Added timeout for ZipStepTest.handlesLargeFiles

v-jizhang:
**Summary**
Currently buck tests have a default timeout of 3 minutes. That's not enough for some of the tests that take a long time to run on Travis CI, handlesLargeFiles is one of them.
The solution is to add a timeout to the individual tests (@Test(timeout=milliseconds)). This way the buck testrunner will give up timeout handling and let Junit to handle it for us.


v-jizhang:
Tested Travis CI with unit timeout and it works. I'll submit another PR.

PR #2375, opened 2020-01-28
Add support for OpenJDK in JAVA_HOME detection

mikekap:
adoptopenjdk installs as /Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk & /Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk for example.

### DIFF
--- a/programs/java_lookup.py
+++ b/programs/java_lookup.py
@@ -22,70 +22,79 @@ from programs.buck_tool import BuckToolException
 from programs.subprocutils import which
 
 
-JDK_8_AND_UNDER_PATH_VERSION_REGEX_STRING = r"jdk(1\.(\d+)(\.\d+(_\d+)?)?)(\.jdk)?"
-JDK_9_AND_OVER_PATH_VERSION_REGEX_STRING = r"jdk-((\d+)(\.\d+(\.\d+(_\d+)?)?)?)(\.jdk)?"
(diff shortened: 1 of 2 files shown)


programs/java_lookup.py
```diff
@@ -22,70 +22,79 @@
...
+                if match:
+                    major_version = int(match.group('major'))
+                    if major_version == desired_major_version:
+                        version_string = match.group('full')
+                        version = tuple(
+                            map(int, version_string.replace("_", ".").split("."))
+                        )
+                        if not max_version or version > max_version:
+                            max_version = version
+                            max_dir = dir
```
philipjameson: Only concern here is that dirs from os.walk is non-deterministic afaik. I'm pretty sure it uses readdir under the hood. We should probably sort that list, and also ensure that we pick one of openjdk or oracle deterministically if both are present. The determinism problem was already present, but is probably more of an issue now, as it's not horribly unlikely that people have two jdks installed.

mikekap: Updated. I also removed the usage of `os.walk` which was technically wrong for this use case. Since it recursively reads the directory tree, it could have seen a subdirectory of the JDK and then tried to use that directory name as JAVA_HOME without including the parent path - see `return os.path.join(base_path, max_dir) if max_dir else None`.

mikekap:
@philipjameson ping

philipjameson:
Looks good, thanks!

PR #2376, opened 2020-01-29
Adding console output when waiting for debugger to attach

tyvsmith:
We've had a number of developers start working with buck and not grasping that it's suspended and waiting for the debugger to attach.

This added output to inform the developer to attach the debugger.

### DIFF
--- a/programs/buck_tool.py
+++ b/programs/buck_tool.py
@@ -1090,10 +1090,13 @@ class BuckTool(object):
                 and os.environ.get("BUCK_DEBUG_MODE") != "0"
             ):
                 suspend = "n" if os.environ.get("BUCK_DEBUG_MODE") == "2" else "y"
+                port = "8888"
                 java_args.append(
                     "-agentlib:jdwp=transport=dt_socket,"
-                    "server=y,suspend=" + suspend + ",quiet=y,address=8888"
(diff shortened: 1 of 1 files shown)


programs/buck_tool.py
```diff
@@ -1090,10 +1090,13 @@ def _get_java_args(self, version_uid, extra_default_options=None):
...
             ):
                 suspend = "n" if os.environ.get("BUCK_DEBUG_MODE") == "2" else "y"
+                port = "8888"
                 java_args.append(
                     "-agentlib:jdwp=transport=dt_socket,"
-                    "server=y,suspend=" + suspend + ",quiet=y,address=8888"
+                    "server=y,suspend=" + suspend + ",quiet=y,address=" + port
                 )
+                if (suspend=="y"):
+                    print("Waiting for debugger on port {}...".format(port))
```
philipjameson: This should use logging.info/warning, not print to stdout.

tyvsmith: Done. 

There's another random use of print in that file, probably worth swapping out in a separate diff.

philipjameson:
Awesome, thanks for making the change! Sorry for the lag in review.

tyvsmith:
👍

PR #2377, opened 2020-01-31
Added CircleCI macos jobs

v-jizhang:
As the same as the ported linux jobs in https://github.com/facebook/buck/pull/2371, added CircleCI jobs run on macos:

- mac_build_openjdk8
- macos_unit_groovy
- macos_build_ant
- macos_test_integration
- macos_test_heavy_integration
- macos_test_android_ndk_15
- macos_test_android_ndk_16
- macos_test_android_ndk_17
- macos_test_android_ndk_18


.circleci/config.yml
```diff
@@ -1,45 +1,73 @@
-version: 2
+version: 2.1
 
 install_openjdk8: &install_openjdk8
   name: Install OpenJDK8
   command: |
-    sudo apt-get update && sudo apt-get install openjdk-8-jdk
-    sudo update-java-alternatives -s java-1.8.0-openjdk-amd64
+    if [ "x${PLATFORM}" == "xlinux" ]
+    then
```
zpao: I don't love having shell conditionals like this all over the place. Would it make more sense to just have separate install steps for mac & linux, rather than forcing re-use? You do that for some of the jobs, so it's not clear why not here as well.

v-jizhang: I did have that thought in my mind, but there are some steps are very similar between linux and mac. For example, install_android_sdk. separating them will have duplicate codes.

PR #2378, opened 2020-02-02
Allow 'Apple Distribution' code sign identities for iOS signing

jasonnam:
Summary:
When buck searches for code signing identities it fails to process
Apple Distribution certificates new to Xcode 11, which can sign
both iOS and Mac binaries.


v-jizhang:
@jasonnam Thank you for your contributions. Your PR has been merged.

PR #2421, opened 2020-03-18
make buck Python3 compatible

v-jizhang:
The function attributes named func_X have been renamed to use the X form in Python3, https://docs.python.org/3/whatsnew/3.0.html#operators-and-special-methods and also built-in changes https://wiki.python.org/moin/Python3.0#Built-In_Changes.
The old names can't be used in Python3.
This PR changes the old names to new names to make buck Python3 compatible.
Python2 was EOL on 1/1/2020, Homebrew/homebrew-core#49796.


v-jizhang:
Replaces #2391.

PR #2422, opened 2020-03-18
Hg cmdline test fix

v-jizhang:
There is a test passes internally but fails in CircleCI

Failed target: //test/com/facebook/buck/util/versioncontrol:versioncontrol
FAIL com.facebook.buck.util.versioncontrol.HgCmdLineInterfaceIntegrationTest
That is caused by the differences between FB hg and open source hg. In this particular case, FB hg output message is longer than that of open source hg. The fix is to assert the expected message contains the actual result, that should . be true both for FB hg and open source hg.

The following test should pass:
buck test //test/com/facebook/buck/util/versioncontrol:versioncontrol


v-jizhang:
Replaces #2409

Issue #2398, opened 2020-02-21
java.lang.NullPointerException if python shebang is None


plmnikulin:
Please, consider handling of java exceptions in the cases similar to the following. Currently there is no hint concerning particular BUCK file and line number in the error message. It is really hard to realize what has happened in the case of a large project.

```
/home/ubuntu/buck/buck-out/gen/ce9b6f2e/programs/buck.pex build //:custom-interpreter-test
Not using buckd because watchman isn't installed.
Parsing buck files: finished in 0.6 sec
Buck encountered an internal error
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at com.facebook.buck.parser.TargetSpecResolver.collectTargets(TargetSpecResolver.java:360)
...
```

See below for the full error message.

Minimal example could be distilled further, but I suppose in the current form it gives some notion how it can happen in real life.

`BUCK` file

```python
python_binary(
    name = 'custom-interpreter-test',
    main_module = 'just_test_module',
    platform = 'platform',
    build_args = [
        # Intention:
        # '--python-shebang', read_config('python#platform', 'interpreter') + ' -Estt',
        # Accidentally [python].platform section was missed in .buckconfig.local
        '--python-shebang', read_config('python#NOTplatform', 'interpreter'),
    ],
    deps = [
        ':just_test_library',
    ],
)

python_library(
    name = 'just_test_library',
    srcs = ['just_test_module.py'],
)
```

`.buckconfig`

```
[python#platform]
# Should be in .buckconfig.local, for CentOS-8:
# interpreter = /usr/libexec/platform-python
interpreter = /usr/bin/python3
```

`just_test_module.py`

```python
if __name__ == '__main__':
    print('Hi')
else:
    raise RuntimeError('Must be called as main module')
```

BUCK version:
```
commit b2f0616214198619163c194b4c7aba0d563d1c18 (HEAD -> master, origin/master, origin/HEAD)
Author: Milen Dzhumerov <mgd@fb.com>
Date:   Mon Feb 17 09:24:32 2020 -0800
```

OS: Ubuntu-18.04

```
openjdk-8-jdk-headless:
  Installed: 8u242-b08-0ubuntu3~18.04
```

```
Buck encountered an internal error
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.NullPointerException
	at com.facebook.buck.parser.TargetSpecResolver.collectTargets(TargetSpecResolver.java:360)
[...]

PR #2431, opened 2020-03-25
Fixed NPE while parsing buildfile

v-jizhang:
Fixed issue #2398. 
Test: The repro in #2398 should print a meaningful message. No NPE was thrown.


src/com/facebook/buck/json/BuildFilePythonResultDeserializer.java
```diff
@@ -110,7 +110,21 @@ public BuildFilePythonResult deserialize(JsonParser jp, DeserializationContext c
...
+      else {
+        // null elements can't be added to ImmutableList, an NPE will be thrown.
+        // Throw a meaningful exception here instead.
+        StringBuilder message = new StringBuilder();
+        message.append("null value can't be added to [");
+        for (Object element: builder.build()) {
+          message.append("'").append(element).append("'").append(", ");
+        }
+        message.append("]");
+        throw new JsonParseException(null, message.toString());
```
zpao, line 126: This isn't really a JSON parse exception, is it? It sounds like the JSON is valid, just that we can't actually do anything with null.

v-jizhang:
After the fix, the output will be
```
Buck wasn't able to parse <path to the build file>/BUCK:
null value can't be added to ['--python-shebang', ]
```

src/com/facebook/buck/json/BuildFilePythonResultDeserializer.java
v-jizhang: Changed to IllegalArgumentException.

PR #2452, opened 2020-04-16
Stop ci on dev branch

v-jizhang:
Stop ci on dev branch


v-jizhang:
> Can we use references to reduce the copy&paste? It looks like it's not possible to filter on workflows so this might be the best we can do :(
> 
> https://discuss.circleci.com/t/can-you-filter-on-a-workflow-level/30624

Modified.

PR #2454, opened 2020-04-22
'--build-only' test command for buck test

navkast:
When someone runs `buck test --build-only`, this builds all of the test targets. The goal is to allow building all test targets without actually running the tests. 

Is this functionality duplicated elsewhere?

### DIFF
--- a/src/com/facebook/buck/cli/TestCommand.java
+++ b/src/com/facebook/buck/cli/TestCommand.java
@@ -149,6 +149,9 @@ public class TestCommand extends BuildCommand {
   @Nullable
   private Boolean isBuildFiltered = null;
 
+  @Option(name = "--build-only", usage = "Only build test targets without running tests.")
+  private boolean isBuildOnly = false;
+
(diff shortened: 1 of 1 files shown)


v-jizhang:
@navkast, Is your change the same as `buck build test/...`? Thank you.

navkast:
It's not quite the same, but I think you have a point. For the code I use buck on, this is the same as:

```
buck build `buck query "filter(':test', '//...')"`
```

We can just go with using a buck query to find all of the test targets!

navkast:
Thanks for looking - sorry for the trouble.

philipjameson:
No worries, a few other things you can do:
- Use `kind('.*_test'` rather than `filter()`, and you don't have to enforce a naming scheme if you don't want to (https://buck.build/command/query.html#kind)
- You can also just use a quick script to extract test information out with the `test.external_runner` config: https://buck.build/files-and-dirs/buckconfig.html#test.external_runner

Issue #2407, opened 2020-03-05
[Kotlin Class ABI Jars] Failed to resolve public top level functions


tyvsmith:
When consuming a third party dep that has a public top level function, it is not kept in the class abi jar. I've now seen this in a handful of jars as I've explored enabling on the Uber monorepo including

Repro steps:

1 )Set compile_against_abis=true in .buckconfig for Kotlin.
2) Add [Square wire-runtime artifact](wire-runtime-3.1.0.jar) or Android UAST jar as a prebuilt jar.
3) Add prebuilt jar as dependency
4) add `import com.squareup.wire.internal.missingRequiredFields` to kotlin file
5) Build

```
my_target/tester.kt:9:35 error: unresolved reference: missingRequiredFields
import com.squareup.wire.internal.missingRequiredFields
                                  ^

    When running <kotlinc>.
    When building rule //my_target:src_main.
```

tyvsmith:
FYI @IanChilds

PR #2458, opened 2020-05-05
Fixed abi for kotlin top level functions

v-jizhang:
Fix for #2407 
Make buck to generated ABIs for synthetic classes in Kotlin modules.


zpao:
Let's add a test here :)

src/com/facebook/buck/jvm/java/abi/StubJarClassEntry.java
```diff
@@ -152,11 +152,14 @@ private InputStream openInputStream() {
     return new ByteArrayInputStream(writer.toByteArray());
   }
 
-  private static boolean isAnonymousOrLocalOrSyntheticClass(ClassNode node) {
+  private static boolean isSyntheticClass(ClassNode node) {
     if ((node.access & Opcodes.ACC_SYNTHETIC) == Opcodes.ACC_SYNTHETIC) {
       return true;
     }
+    return false;
```
zpao, line 159:
```suggestion
    return (node.access & Opcodes.ACC_SYNTHETIC) == Opcodes.ACC_SYNTHETIC);
```

src/com/facebook/buck/jvm/java/abi/StubJarClassEntry.java
```diff
@@ -106,7 +106,7 @@ public static StubJarClassEntry of(
 
     // The synthetic package-info class is how package annotations are recorded; that one is
     // actually used by the compiler
-    if (!isAnonymousOrLocalOrSyntheticClass(stub)
+    if (!((isSyntheticClass(stub) && !isKotlinModule) || isAnonymousOrLocalClass(stub))
```
zpao, line 109: If we can simplify this logic and match formatting, that'd be great.

v-jizhang: Simplified.

PR #2459, opened 2020-05-06
[Fix] Strip symlink root from headers module map

tjwio:
**Background**
When using `modulemap_mode = HEADERS`, buck was successfully building the module map but it would include the header symlink root in the header path which would cause it to not be found in the xcode header search paths. ie. it would look like
`header "gen/out/folder/anotherfolder/module_name/HeaderTest.h"`
when it should actually be
`header "HeaderTest.h"`

**Change**
Strip out symlink root along with module name from headers generation

### DIFF
--- a/src/com/facebook/buck/apple/clang/ModuleMapFactory.java
+++ b/src/com/facebook/buck/apple/clang/ModuleMapFactory.java
@@ -43,10 +43,11 @@ public class ModuleMapFactory {
       String moduleName,
       ModuleMapMode moduleMapMode,
       UmbrellaHeaderModuleMap.SwiftMode swiftMode,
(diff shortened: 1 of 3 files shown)


tjwio:
@v-jizhang could you ptal whenever you get a chance? should be a pretty simple fix

tjwio:
Could I get a review on this?

v-jizhang:
Sure, give me a couple days. Thanks

tjwio:
friendly bump on this cc @v-jizhang

v-jizhang:
@tjwio lgtm. Could add a test for your change? Thank you.

v-jizhang:
This PR is closed because of no updates for over 4 months.

PR #2464, opened 2020-05-13
Make release process work

v-jizhang:
Make release process work.


zpao:
You've done more than just passing args to ant in here. Can we split or broaden the goal of this PR (which seems to be more about making the release process actually work)

PR #2465, opened 2020-05-21
Fixed bugs for Windows release process

v-jizhang:
Fixed bugs for Windows release process


tools/release/platforms/chocolatey/build.py
```diff
@@ -88,7 +88,7 @@ def write_license_file(original_license):
 
 def write_verification_txt(original_verification_txt, version, timestamp):
     dest = "VERIFICATION.txt"
-    with open(original_verification_txt, "r") as fin, open(dest, "w") as fout:
+    with open(original_verification_txt, "rb") as fin, open(dest, "w") as fout:
```
zpao, line 91: Why do we need binary? Should we instead be specifying the encoding to be utf-8?

tools/release/platforms/chocolatey/Dockerfile
```diff
@@ -5,12 +5,17 @@ SHELL ["powershell", "-command"]
...
 # Install chocolatey
 RUN Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
 
 # Download enough to bootstrap a build
-RUN choco install -y python3 git jdk8
+RUN choco install -y python --version=3.6.2
+RUN choco install -y git jdk8
+
+# Choco installs python3 and named it to python, buck expects python3 so make a copy of python3
+RUN $pythonpath=(Split-Path((Get-Command python).Path)); copy ${pythonpath}\python.exe ${pythonpath}\python3.exe
```
zpao, line 18: Does Windows have symlinks or aliasing?

tools/release/platforms/chocolatey/build.py
v-jizhang: Yes, That's a better fix. I used "rb" because I tried to fix it on the fly while running docker builder. I'll change this to use endcoding instead. Thanks.

tools/release/platforms/chocolatey/Dockerfile
v-jizhang: Yes, I'll change to symlink

PR #2469, opened 2020-06-02
Improve Swift build performance by avoiding unnecessary recompilation of Swift modules

xianwen:
## TL;DR;
Update `SwiftCompile` to support `SupportsInputBasedRuleKey` interface, so that we could avoid unnecessary Swift module recompilation if its input Swift modules' `.swiftmodule` artifacts weren't changed.

## Details
Suppose Swift moduleA depends on Swift moduleB.

The current behavior of Buck is that whenever you touch the source code of moduleB, moduleA will always be rebuilt, even if you didn't change the public interface of moduleB. This has a heavy cost and unnecessarily slows incremental builds a lot.

This PR avoids the recompilation of moduleA in this given example by evaluating the md5sum of `moduleB.swiftmodule`, which is a binary superset of moduleB's all public interfaces. As long as a Swift module's `.swiftmodule` remains unchanged after its compilation, it is safe to not recompile its dependents.

I verified this PR is working in an Airbnb demo project Buck Sample here:
https://github.com/airbnb/BuckSample/pull/132

By applying this PR internally at the Airbnb's fork of Buck here, https://github.com/airbnb/buck/pull/23/, our p50 of local iOS build time has reduced more than 30% in the last 2 weeks.

### DIFF
--- a/src/com/facebook/buck/swift/SwiftCompile.java
+++ b/src/com/facebook/buck/swift/SwiftCompile.java
@@ -21,11 +21,13 @@ import com.facebook.buck.core.build.context.BuildContext;
 import com.facebook.buck.core.build.execution.context.ExecutionContext;
 import com.facebook.buck.core.filesystems.AbsPath;
 import com.facebook.buck.core.model.BuildTarget;
+import com.facebook.buck.core.model.Flavor;
 import com.facebook.buck.core.model.impl.BuildTargetPaths;
 import com.facebook.buck.core.rulekey.AddToRuleKey;
 import com.facebook.buck.core.rules.ActionGraphBuilder;
 import com.facebook.buck.core.rules.BuildRule;
 import com.facebook.buck.core.rules.BuildRuleResolver;
+import com.facebook.buck.core.rules.attr.SupportsInputBasedRuleKey;
 import com.facebook.buck.core.rules.common.BuildableSupport;
 import com.facebook.buck.core.rules.impl.AbstractBuildRule;
 import com.facebook.buck.core.sourcepath.ExplicitBuildTargetSourcePath;
@@ -34,19 +36,19 @@ import com.facebook.buck.core.sourcepath.resolver.SourcePathResolverAdapter;
 import com.facebook.buck.core.toolchain.tool.Tool;
 import com.facebook.buck.cxx.CxxDescriptionEnhancer;
 import com.facebook.buck.cxx.PreprocessorFlags;
-import com.facebook.buck.cxx.toolchain.CxxPlatform;
 import com.facebook.buck.cxx.toolchain.HeaderVisibility;
+import com.facebook.buck.rules.coercer.FrameworkPath;
 import com.facebook.buck.cxx.toolchain.LinkerMapMode;
 import com.facebook.buck.cxx.toolchain.PathShortener;
 import com.facebook.buck.cxx.toolchain.Preprocessor;
 import com.facebook.buck.io.BuildCellRelativePath;
 import com.facebook.buck.io.file.MostFiles;
 import com.facebook.buck.io.filesystem.ProjectFilesystem;
+import com.facebook.buck.rules.args.AddsToRuleKeyFunction;
 import com.facebook.buck.rules.args.Arg;
 import com.facebook.buck.rules.args.FileListableLinkerInputArg;
 import com.facebook.buck.rules.args.SourcePathArg;
 import com.facebook.buck.rules.args.StringArg;
-import com.facebook.buck.rules.coercer.FrameworkPath;
 import com.facebook.buck.step.Step;
 import com.facebook.buck.step.StepExecutionResult;
 import com.facebook.buck.step.StepExecutionResults;
@@ -68,10 +70,9 @@ import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.SortedSet;
-import java.util.function.Function;
 
 /** A build rule which compiles one or more Swift sources into a Swift module. */
-public class SwiftCompile extends AbstractBuildRule {
+public class SwiftCompile extends AbstractBuildRule implements SupportsInputBasedRuleKey {
 
   private static final String INCLUDE_FLAG = "-I";
 
@@ -82,13 +83,20 @@ public class SwiftCompile extends AbstractBuildRule {
   @AddToRuleKey(stringify = true)
   private final Path outputPath;
 
+  @AddToRuleKey(stringify = true)
   private final Path objectFilePath;
+  @AddToRuleKey(stringify = true)
   private final Path modulePath;
+  @AddToRuleKey(stringify = true)
   private final Path moduleObjectPath;
+  @AddToRuleKey(stringify = true)
   private final ImmutableList<Path> objectPaths;
   private final Optional<AbsPath> swiftFileListPath;
 
   @AddToRuleKey private final boolean shouldEmitSwiftdocs;
+  @AddToRuleKey private final boolean useModulewrap;
+  @AddToRuleKey private final boolean compileForceCache;
+  @AddToRuleKey(stringify = true)
   private final Path swiftdocPath;
 
   @AddToRuleKey private final ImmutableSortedSet<SourcePath> srcs;
@@ -96,14 +104,14 @@ public class SwiftCompile extends AbstractBuildRule {
   @AddToRuleKey private final Optional<String> version;
   @AddToRuleKey private final ImmutableList<? extends Arg> compilerFlags;
 
+  @AddToRuleKey(stringify = true)
   private final Path headerPath;
-  private final CxxPlatform cxxPlatform;
-  private final ImmutableSet<FrameworkPath> frameworks;
+  @AddToRuleKey private final ImmutableSet<FrameworkPath> frameworks;
+  @AddToRuleKey private final AddsToRuleKeyFunction<FrameworkPath, Path> frameworkPathToSearchPath;
+  @AddToRuleKey private final Flavor flavor;
 
-  private final boolean enableObjcInterop;
+  @AddToRuleKey private final boolean enableObjcInterop;
   @AddToRuleKey private final Optional<SourcePath> bridgingHeader;
-
-  private final SwiftBuckConfig swiftBuckConfig;
 
   @AddToRuleKey private final Preprocessor cPreprocessor;
 
@@ -114,7 +122,6 @@ public class SwiftCompile extends AbstractBuildRule {
   private BuildableSupport.DepsSupplier depsSupplier;
 
   SwiftCompile(
-      CxxPlatform cxxPlatform,
       SwiftBuckConfig swiftBuckConfig,
       BuildTarget buildTarget,
       SwiftTargetTriple swiftTarget,
@@ -122,6 +129,8 @@ public class SwiftCompile extends AbstractBuildRule {
       ActionGraphBuilder graphBuilder,
       Tool swiftCompiler,
       ImmutableSet<FrameworkPath> frameworks,
+      AddsToRuleKeyFunction<FrameworkPath, Path> frameworkPathToSearchPath,
+      Flavor flavor,
       String moduleName,
       Path outputPath,
       Iterable<SourcePath> srcs,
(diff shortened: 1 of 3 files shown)


src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -475,4 +477,18 @@ public SourcePath getObjCGeneratedHeaderPath() {
...
+   * swiftmodule file contains the public interface for a module, and is basically a binary file
+   * format equivalent to header files for a C framework or library.
+   *
+   * A swiftmodule file contains serialized ASTs (and possibly SIL), it conforms to
+   * Swift Binary Serialization Format, more details about this binary format can be found here:
+   * https://github.com/apple/swift/blob/7e6d62dae4bae4eb3737a6f76c0e51534c1bcca3/docs/Serialization.rst.
+   */
+  public SourcePath getSwiftModuleOutputPath() {
+    return ExplicitBuildTargetSourcePath.of(getBuildTarget(),
+      Paths.get(outputPath.toString(), moduleName + ".swiftmodule"));
```
milend, line 492: This can produce a different module name than what's actually produced by the compilation in the case of sanitisation. Instead, it should just return the precomputed `modulePath` like so:

    return ExplicitBuildTargetSourcePath.of(getBuildTarget(), modulePath);

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -65,13 +66,14 @@
...
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.SortedSet;
 import java.util.function.Function;
 
 /** A build rule which compiles one or more Swift sources into a Swift module. */
-public class SwiftCompile extends AbstractBuildRule {
+public class SwiftCompile extends AbstractBuildRule implements SupportsInputBasedRuleKey {
```
milend, line 76: This change can potentially break things because there's still a bunch of fields which are not part of the rulekey. This becomes problematic if any of those unannotated keys contains data computed by some transitive dependency and that data is part of the transitive rule's rulekey.

I'd need to audit all of the remaining fields to ensure that this is the case but I haven't had the time to do that yet.

On the correctness note, a quick look reveals that some stuff is already broken: e.g., `enableObjcInterop` should have an `@AddToRuleKey` annotation.

src/com/facebook/buck/swift/SwiftCompile.java
xianwen: 👍

src/com/facebook/buck/swift/SwiftCompile.java
xianwen: I see, thx for the review.

Would it be enough for us to review all the "private final" fields in `SwiftCompile` class? Or do we need to review the fields in its super class as well?

If former, then the fields we shall review would be the following:
```
  private final Path objectFilePath;
  private final Path modulePath;
  private final Path moduleObjectPath;
  private final ImmutableList<Path> objectPaths;
  private final Optional<AbsPath> swiftFileListPath;
  private final Path headerPath;
  private final CxxPlatform cxxPlatform;
  private final ImmutableSet<FrameworkPath> frameworks;
  private final boolean enableObjcInterop;
  private final SwiftBuckConfig swiftBuckConfig;
  private BuildableSupport.DepsSupplier depsSupplier;
```

Which means in order to safely apply this change, **we should review these fields and make sure all of them whenever necessary should have that `@AddToRuleKey` annotation.**

Is my understanding above correct? Thanks!

src/com/facebook/buck/cxx/FrameworkPathArg.java
```diff
@@ -23,8 +23,8 @@
 import java.util.Objects;
 
 /** A base class for {@link Arg}s which wrap a {@link FrameworkPath}. */
-abstract class FrameworkPathArg implements Arg {
-  @AddToRuleKey final ImmutableSortedSet<FrameworkPath> frameworkPaths;
+public abstract class FrameworkPathArg implements Arg {
```
xianwen, line 26: Made it `public` so that `com/facebook/buck/swift/SwiftLibraryDescription` can create a subclass of it.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -82,29 +82,36 @@
...
+  @AddToRuleKey(stringify = true)
   private final Path modulePath;
+  @AddToRuleKey(stringify = true)
   private final Path moduleObjectPath;
+  @AddToRuleKey(stringify = true)
   private final ImmutableList<Path> objectPaths;
   private final Optional<AbsPath> swiftFileListPath;
 
   @AddToRuleKey private final boolean shouldEmitSwiftdocs;
+  @AddToRuleKey private final boolean useModulewrap;
```
xianwen, line 96: From `swiftBuckConfig`

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -82,29 +82,36 @@
...
+  @AddToRuleKey private final boolean compileForceCache;
```
xianwen, line 97: From `swiftBuckConfig`

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -287,8 +292,15 @@ public BuildRule createBuildRule(
           cxxPlatform.getCpp().resolve(graphBuilder, buildTarget.getTargetConfiguration());
 
       BuildTarget buildTargetCopy = buildTarget;
+
+      ImmutableSortedSet.Builder<FrameworkPath> builder = ImmutableSortedSet.naturalOrder();
```
xianwen, line 296: This is duplicated code that should be refactored into a method if we want to move forward with this design.
I'm not doing it now because I prefer to adopt a different design.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -203,15 +213,7 @@ private SwiftCompileStep makeCompileStep(SourcePathResolverAdapter resolver) {
...
-    Function<FrameworkPath, Path> frameworkPathToSearchPath =
-        CxxDescriptionEnhancer.frameworkPathToSearchPath(cxxPlatform, resolver);
-
-    compilerCommand.addAll(
-        Streams.concat(frameworks.stream(), cxxDeps.getFrameworkPaths().stream())
-            .filter(x -> !x.isSDKROOTFrameworkPath())
-            .map(frameworkPathToSearchPath)
-            .flatMap(searchPath -> ImmutableSet.of("-F", searchPath.toString()).stream())
-            .iterator());
+    frameworksArg.appendToCommandLine(compilerCommand::add, resolver);
```
xianwen, line 216: @milend , I'm not sure if I understand you correctly regarding using a `frameworksArg` to replace both `cxxPlatform` and `frameworks`.

This usage of `frameworksArg.appendToCommandLine` isn't quite consistent with the other `compilerCommand.addAll` calls nearby. I would prefer to still use the same old `compilerCommand.addAll` here for this `frameworksArg` as well.

My proposal is:
Instead of passing in one single `frameworksArg`, we pass in `Function<FrameworkPath, Path> frameworkPathToSearchPath` and `ImmutableSortedSet<FrameworkPath> frameworkPaths` into `SwiftCompile`, and mark both as `@AddToRuleKey`.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -402,7 +404,7 @@ public SourcePath getSourcePathToOutput() {
             .toToolFlags(
                 resolver,
                 PathShortener.byRelativizingToWorkingDir(getProjectFilesystem().getRootPath()),
-                CxxDescriptionEnhancer.frameworkPathToSearchPath(cxxPlatform, resolver),
+                frameworkPathToSearchPath,
```
xianwen, line 407: According to the comment above:
>  Arg list can't simply be passed in since the current implementation of toToolFlags drops the
> dependency information.

I don't like the idea of making `frameworkArgs.frameworkPathToSearchPath` public, so I chose to pass it in directly instead and marked it as `@AddToRuleKey`.

However, this diminishes the benefit of having `frameworkArgs`. And it's the other reason why I think it might make sense to not use that new `frameworkArgs`, but to use `Function<FrameworkPath, Path> frameworkPathToSearchPath` and `ImmutableSortedSet<FrameworkPath> frameworkPaths` directly instead.

milend: @xianwen:
> I don't like the idea of making frameworkArgs.frameworkPathToSearchPath public, so I chose to pass it in directly instead and marked it as @AddToRuleKey

Agreed, it should not be exposed.

> Arg list can't simply be passed in since the current implementation of toToolFlags drops the dependency information.

I have a suspicion that this comment might be out of date (it was added back in 2017) and `CxxToolFlags` is annotated with `AddsToRuleKey`.

If we wanted to do this properly, we would pass `argsFromDeps` to the constructor, together with a single `Arg` for the frameworks. But it's safer to keep the current behaviour as changing it would require further validation.

src/com/facebook/buck/swift/SwiftCompile.java
milend: @xianwen:

> Instead of passing in one single frameworksArg, we pass in Function<FrameworkPath, Path> frameworkPathToSearchPath and ImmutableSortedSet<FrameworkPath> frameworkPaths into SwiftCompile, and mark both as @AddToRuleKey.

The type of `frameworkPathToSearchFunction` should be `AddsToRuleKeyFunction<FrameworkPath, Path>`, agreed about the rest.

We basically have to go down that route because of the usage of `frameworkPathToSearchPath` in the `toToolFlags()` call.

src/com/facebook/buck/swift/SwiftLibraryDescription.java
```diff
@@ -530,4 +554,32 @@ public static boolean isSwiftTarget(BuildTarget buildTarget) {
      */
     Optional<String> getTargetSdkVersion();
   }
+
+  private static class FrameworkLinkerArgs extends FrameworkPathArg {
+    @AddToRuleKey final AddsToRuleKeyFunction<FrameworkPath, Path> frameworkPathToSearchPath;
+
+    public FrameworkLinkerArgs(
```
milend, line 561: @xianwen: We do not need this anymore if we just pass the list of frameworks + function.

milend:
@xianwen: Once you address the feedback and take the revised approach, can you please squash all changes into a single commit?

src/com/facebook/buck/swift/SwiftLibraryDescription.java
xianwen: No, we don't.
I'll remove this and also update a few other places accordingly.

xianwen:
> @xianwen: Once you address the feedback and take the revised approach, can you please squash all changes into a single commit?

Of course, will do.

xianwen:
@milend, all code review feedbacks are addressed and all commits are squashed into one, PTAL at your convenience, thanks!

milend:
@xianwen: all looks good, will merge.

src/com/facebook/buck/swift/SwiftCompile.java
```diff
@@ -82,29 +83,36 @@
...
   @AddToRuleKey private final Optional<String> version;
   @AddToRuleKey private final ImmutableList<? extends Arg> compilerFlags;
 
+  @AddToRuleKey(stringify = true)
   private final Path headerPath;
-  private final CxxPlatform cxxPlatform;
-  private final ImmutableSet<FrameworkPath> frameworks;
+  @AddToRuleKey private final ImmutableSet<FrameworkPath> frameworks;
+  @AddToRuleKey private final AddsToRuleKeyFunction<FrameworkPath, Path> frameworkPathToSearchPath;
+  @AddToRuleKey private final Flavor flavor;
```
milend, line 111: @xianwen: This requires `@AddToRuleKey(stringify = true)` as `Flavor` is not a supported type for rulekey computation.

xianwen: Oops, sorry about that, I'll send you another PR fixing it up.

xianwen: https://github.com/facebook/buck/pull/2475

milend: Merged, thanks.

PR #2473, opened 2020-06-05
Windows release on CircleCI

v-jizhang:
Windows release on CircleCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1179,13 +1179,43 @@ jobs:
           name: pip install python-dateutil
           command: python3 -m pip install python-dateutil
       - run:
-          name: pip install python-magic-bin
+          name: pip install python-magic
           command: python3 -m pip install python-magic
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1179,14 +1179,40 @@ jobs:
...
+          command: python3 -m pip install --upgrade pip
+      - run:
+          name: pip install requests
+          command: python3 -m pip install requests
+      - run:
+          name: pip install python-dateutil
+          command: python3 -m pip install python-dateutil
+      - run:
+          name: pip install python-magic-bin
+          command: python3 -m pip install python-magic-bin
```
zpao, line 1208: These use `python-magic-bin` but the other step uses `python-magic`?

v-jizhang: The name of the module on Linux is python-magic and and on Mac/Windows is python-magic-bin.

PR #2479, opened 2020-06-16
Update jitpack python 3 url to be compatible with older OS

### LINKED ISSUE #2478: jitpack.io builds appear broken with upgrade to py3
I noticed that the latest release of Buck in OSS (2020.05.14.01) doesn't have a working jitpack.io build. It looks like the move to python3 broke jitpack.

Logs from the latest release builds on jitpack.io:

https://jitpack.io/com/github/facebook/buck/v2020.05.14.01/build.log

kageiit:
- Fixes #2478
- Also added build instructions at https://github.com/kageiit/jitpack-python/blob/master/README.md if anyone wants to build their own version in the future

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,7 +1,7 @@
 install:
    - set -exo pipefail
    - mkdir python_tmp
-   - wget https://github.com/kageiit/jitpack-python/releases/download/3.8/python-3.8.tar.gz -O python_tmp/python.tar.gz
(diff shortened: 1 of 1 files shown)


kageiit:
I used the same base image as jitpack base for better compatibility - https://github.com/jitpack/dockerfiles/blob/master/base/Dockerfile

v-jizhang:
@kageiit Thank you very much.

PR #2484, opened 2020-06-22
pyenv upgraded, invalidate the cache

v-jizhang:
pyenv upgraded, invalidate the cache

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -210,34 +210,34 @@ jobs:
           <<: *install_openjdk8
       - restore_cache:
           # Change this key when upgrading sdk.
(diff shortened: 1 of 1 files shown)


zpao:
Is it possible to do this in a way that doesn't require configs be manually changed? eg https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache

v-jizhang:
> Is it possible to do this in a way that doesn't require configs be manually changed? eg https://support.circleci.com/hc/en-us/articles/115015426888-Clear-project-dependency-cache

Yes, checksum on version# can be used. I'll submit another PR to make all cache based on their versions.

PR #2486, opened 2020-06-23
Automate publishing docs on CirclrCI

v-jizhang:
Automate publishing docs on CirclrCI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1228,10 +1228,10 @@ jobs:
 
   publish_docs:
     environment:
-      <<: *macos_environment
-    working_directory: "/Users/distiller/buck"
-    macos:
-      xcode: "11.3.1"
+      <<: *linux_environment
+    working_directory: "/home/circleci/buck"
+    machine:
+      image: ubuntu-1604:201903-01
     steps:
       - checkout
       - run: 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,6 +195,13 @@ windows_environment: &windows_environment
   BUCK_NUM_THREADS: 3
   BUCK_PEX_LOCATION: "./new_buck.pex"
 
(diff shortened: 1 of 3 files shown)


zpao:
Let's do this on a Linux machine.

.circleci/config.yml
```diff
@@ -1226,6 +1226,36 @@ jobs:
             python3 ./tools/release/publish_release.py --no-build-deb --no-build-homebrew --github-token ${GITHUB_TOKEN} --chocolatey-token ${CHOCO_TOKEN} --use-existing-release --version ${VERSION} --output-dir artifacts
           shell: bash.exe
 
+  publish_docs:
+    environment:
+      <<: *linux_environment
+    working_directory: "/home/circleci/buck"
```
zpao, line 1232: Why do we need to set the working directory? Aren't we just working from the root of the clone?

.circleci/config.yml
```diff
@@ -1277,3 +1307,9 @@ workflows:
               only: /^v20.*/
             branches:
               ignore: /.*/
+      - publish_docs:
+          filters:
+            tags:
+              only: /^v20.*/
+            branches:
+              ignore: /.*/
```
zpao, line 1315: Lot of copy & paste of these filters… might make sense to share them and `<<` include them.

docs/publish.sh
```diff
@@ -101,6 +103,8 @@ cd -
 
 # Commit the new version of the docs:
 cd $STATIC_FILES_DIR
+echo "${CNAME}" > CNAME
```
zpao, line 106: In an ideal world this wouldn't be something that happens magically in bash, but is actually a file that lives in the website folder. This is fine for now though

.circleci/config.yml
v-jizhang: This instructs CircleCI where to clone the code to.

PR #2501, opened 2020-07-22
Do not override SWIFT_OBJC_INTERFACE_HEADER_NAME

pomozoff:
Sometimes the name of `*-Swift.h` header should be customized.
However, Buck always overrides user's value that fails to build a project in Xcode.
This patch intends to avoid automatically setting the value if a user has already configured one.

### DIFF
--- a/src/com/facebook/buck/features/apple/project/ProjectGenerator.java
+++ b/src/com/facebook/buck/features/apple/project/ProjectGenerator.java
@@ -1721,8 +1721,11 @@ public class ProjectGenerator {
           s -> extraSettingsBuilder.put("PRODUCT_MODULE_NAME", getModuleName(targetNode)));
 
       if (hasSwiftVersionArg && containsSwiftCode && isFocusedOnTarget) {
-        extraSettingsBuilder.put(
-            "SWIFT_OBJC_INTERFACE_HEADER_NAME", getSwiftObjCGeneratedHeaderName(buildTargetNode));
+        // Do not override user's value
+        if (!extraSettingsBuilder.build().containsKey("SWIFT_OBJC_INTERFACE_HEADER_NAME")) {
+          extraSettingsBuilder.put(
+              "SWIFT_OBJC_INTERFACE_HEADER_NAME", getSwiftObjCGeneratedHeaderName(buildTargetNode));
+        }
 
         if (swiftBuckConfig.getProjectWMO()) {
           // We must disable "Index While Building" as there's a bug in the LLVM infra which


pomozoff:
Seems like something is broken on the CI runner:
```
Error: An exception occurred within a child process:
 
  FormulaUnavailableError: No available formula with the name "/usr/local/opt/python@2/.brew/python@2.rb" 
```

v-jizhang:
> Seems like something is broken on the CI runner:
> 
> ```
> Error: An exception occurred within a child process:
>  
>   FormulaUnavailableError: No available formula with the name "/usr/local/opt/python@2/.brew/python@2.rb" 
> ```

Yeah, that is unrelated. #2495 fixed it but not merged yet.

src/com/facebook/buck/features/apple/project/ProjectGenerator.java
```diff
@@ -1721,8 +1721,11 @@ private PBXNativeTarget generateBinaryTarget(
           s -> extraSettingsBuilder.put("PRODUCT_MODULE_NAME", getModuleName(targetNode)));
 
       if (hasSwiftVersionArg && containsSwiftCode && isFocusedOnTarget) {
-        extraSettingsBuilder.put(
-            "SWIFT_OBJC_INTERFACE_HEADER_NAME", getSwiftObjCGeneratedHeaderName(buildTargetNode));
```
v-jizhang, line 1725: Does the override really happen? If you put two entries with the same key, eventually an IllegalArgumentException will be thrown when build() is called.

pomozoff: Sorry, seems like I wasn't clear enough. Look, I figured out, that Buck sets incorrect value for this parameter in xconfig files. I set this parameter in "configs" dictionary of "apple_binary" object to one I need. However, these two lines of code override my config during a build. So, I think this behaviour should be changed, because user's settings should have high priority over anything set by Buck itself.
Does it make sense?

v-jizhang: I meant extraSettingsBuilder is an ImmutableMap, and it does not allow set the same key twice. In the case here, setting  "SWIFT_OBJC_INTERFACE_HEADER_NAME" the second time is an error (extraSettingsBuilder.build() throws exception). So my understanding is overriding won't happen even before your change.

v-jizhang:
@pomozoff 
Thanks for your contribution.
I'm closing this PR because seems it does not solve the problem.

Issue #2498, opened 2020-07-16
Buck fails to build non predex apps which include jars containing module-info.class files


kageiit:
When an android app depends on gson 2.8.6 for example, buck fails to create a release apk (non predex) with the exception
```
Buck encountered an internal error
java.lang.RuntimeException: Error calculating size for module-info.class.
	at com.facebook.buck.android.dalvik.DalvikStatsCache.getStats(DalvikStatsCache.java:54)
	at com.facebook.buck.android.dalvik.DalvikAwareOutputStreamHelper.isEntryTooBig(DalvikAwareOutputStreamHelper.java:62)
	at com.facebook.buck.android.dalvik.DalvikAwareOutputStreamHelper.canPutEntry(DalvikAwareOutputStreamHelper.java:76)
	at com.facebook.buck.android.dalvik.DalvikAwareZipSplitter$2.visit(DalvikAwareZipSplitter.java:269)
	at com.facebook.buck.jvm.java.classes.ClasspathTraversal$ZipFileTraversalAdapter$1.visit(ClasspathTraversal.java:90)
	at com.facebook.buck.util.ZipFileTraversal.traverse(ZipFileTraversal.java:39)
	at com.facebook.buck.jvm.java.classes.ClasspathTraversal$ZipFileTraversalAdapter.traverse(ClasspathTraversal.java:93)
	at com.facebook.buck.jvm.java.classes.ClasspathTraversal.traverse(ClasspathTraversal.java:58)
	at com.facebook.buck.jvm.java.classes.DefaultClasspathTraverser.traverse(DefaultClasspathTraverser.java:28)
	at com.facebook.buck.android.dalvik.DalvikAwareZipSplitter.execute(DalvikAwareZipSplitter.java:250)
	at com.facebook.buck.android.SplitZipStep.execute(SplitZipStep.java:209)
	at com.facebook.buck.step.DefaultStepRunner.runStepForBuildTarget(DefaultStepRunner.java:46)
	... 16 more
Caused by: java.lang.NullPointerException
	at java.util.regex.Matcher.getTextLength(Matcher.java:1283)
	at java.util.regex.Matcher.reset(Matcher.java:309)
	at java.util.regex.Matcher.<init>(Matcher.java:229)
	at java.util.regex.Pattern.matcher(Pattern.java:1093)
	at com.facebook.buck.android.dalvik.DalvikStatsTool$StatsClassVisitor.visit(DalvikStatsTool.java:222)
[...]

PR #2504, opened 2020-07-23
Fix for #2498

surapuramakhil:
Buck fails to build non predex apps which include jars containing module-info.class files` #2498

### DIFF
--- a/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
+++ b/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
@@ -33,7 +33,7 @@ class DalvikStatsCache {
 
   DalvikStatsTool.Stats getStats(FileLike entry) {
     String name = entry.getRelativePath();
-    if (!name.endsWith(".class") || name.endswith("module-info.class")) {
+    if (!name.endsWith(".class") || name.endsWith("module-info.class")) {
       // Probably something like a pom.properties file in a JAR: this does not contribute
       // to the linear alloc size, so return zero.
       return DalvikStatsTool.Stats.ZERO;
--- a/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
+++ b/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
@@ -33,7 +33,7 @@ class DalvikStatsCache {
 
   DalvikStatsTool.Stats getStats(FileLike entry) {
     String name = entry.getRelativePath();
-    if (!name.endsWith(".class")) {
+    if (!name.endsWith(".class") || name.endswith("module-info.class")) {
       // Probably something like a pom.properties file in a JAR: this does not contribute
       // to the linear alloc size, so return zero.
       return DalvikStatsTool.Stats.ZERO;
--- a/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
+++ b/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
@@ -35,7 +35,8 @@ class DalvikStatsCache {
     String name = entry.getRelativePath();
     if (!name.endsWith(".class") || name.endsWith("module-info.class")) {
       // Probably something like a pom.properties file in a JAR: this does not contribute
-      //skipping special class files like module descriptor - here no classes will be declared and class visitor thows error
+      
+       //skipping special class files like module descriptor - here no classes will be declared and class visitor thows error
       // to the linear alloc size, so return zero.
       return DalvikStatsTool.Stats.ZERO;
     }
--- a/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
+++ b/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
@@ -36,9 +36,9 @@ class DalvikStatsCache {
     String name = pathParts[pathParts.length-1];
     if (!name.endsWith(".class") || name.equals("module-info.class")) {
       // Probably something like a pom.properties file in a JAR: this does not contribute
-      
-       //skipping special class files like module descriptor - here no classes will be declared and class visitor thows error
       // to the linear alloc size, so return zero.
+      // skipping special class files like module descriptor - here no classes will be
+      // declared and class visitor throws error.
       return DalvikStatsTool.Stats.ZERO;
     }
 
--- a/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
+++ b/src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
@@ -32,8 +32,9 @@ class DalvikStatsCache {
   }
 
   DalvikStatsTool.Stats getStats(FileLike entry) {
-    String name = entry.getRelativePath();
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
```diff
@@ -33,7 +33,7 @@
 
   DalvikStatsTool.Stats getStats(FileLike entry) {
     String name = entry.getRelativePath();
-    if (!name.endsWith(".class")) {
+    if (!name.endsWith(".class") || name.endswith("module-info.class")) {
```
v-jizhang, line 36: Please fix the typo "endswith"

v-jizhang: Similar issue is https://stackoverflow.com/questions/45311295/error-scanning-entry-module-info-class-when-starting-jetty-server.
Should upgrading ClassReader.class work? we use ClassReader to classes and it can't read the new format.

kageiit: Ignoring this class should work. Even if we read it, it would be a waste as we don't need to use any of this info for estimating dex size.

surapuramakhil: @kageiit makes sense.

v-jizhang: Just don't want to add to many special cases. If we do, @shubhraprakash please comment why.

kageiit: Speaking of special cases, we would have to rework this pipeline a bit to handle multi release jars eventually as duplicate class files will be inside META-INF for different versions of the jvm. Thats a separate issue however

surapuramakhil: @v-jizhang  who is  @shubhraprakash, FYI this is @surapuramakhil. And added comment for special case

v-jizhang: Sorry :)

src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
```diff
@@ -33,8 +33,9 @@
 
   DalvikStatsTool.Stats getStats(FileLike entry) {
     String name = entry.getRelativePath();
-    if (!name.endsWith(".class")) {
+    if (!name.endsWith(".class") || name.endsWith("module-info.class")) {
       // Probably something like a pom.properties file in a JAR: this does not contribute
+      //skipping special class files like module descriptor - here no classes will be declared and class visitor thows error
```
v-jizhang, line 38: Could you move your comment 1 line down and add 1 space after "//"? Thanks @surapuramakhil

surapuramakhil: okay :)

v-jizhang: The problem is that you inserted your comment in the middle of another comment and a wrong spelling word "throws". The comment should look like:
```
      // Probably something like a pom.properties file in a JAR: this does not contribute
      // to the linear alloc size, so return zero.
      // skipping special class files like module descriptor - here no classes will be
      // declared and class visitor throws error
```

v-jizhang: Also, should we use name.equals() instead of name.endsWith()?

surapuramakhil: No @v-jizhang  , name is taken as relative path.

`String name = entry.getRelativePath();`

surapuramakhil:
@v-jizhang I had a look at failure test cases - it's clear that nothing related to pull request changes.
Are the testcases are outdated? And expected string message in testcase needs to changed or vice versa

v-jizhang:
> @v-jizhang I had a look at failure test cases - it's clear that nothing related to pull request changes.
> Are the testcases are outdated? And expected string message in testcase needs to changed or vice versa

Don't worry, The test failed tests are unrelated, I'll fix them. Please take a look at my previous messages and fix the comments. Thanks

src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
v-jizhang: What if there is such a class FooBar-module-info.class we don't want to exclude?

surapuramakhil: Done

src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
```diff
@@ -32,9 +32,12 @@
...
   DalvikStatsTool.Stats getStats(FileLike entry) {
-    String name = entry.getRelativePath();
-    if (!name.endsWith(".class")) {
+    String[] pathParts = entry.getRelativePath().split("/");
+    String name = pathParts[pathParts.length-1];
+    if (!name.endsWith(".class") || name.equals("module-info.class")) {
       // Probably something like a pom.properties file in a JAR: this does not contribute
+      
+       //skipping special class files like module descriptor - here no classes will be declared and class visitor thows error
       // to the linear alloc size, so return zero.
```
v-jizhang, line 41: As I mentioned in the previous conversations, these comment should look like:
```
      // Probably something like a pom.properties file in a JAR: this does not contribute
      // to the linear alloc size, so return zero.
      // skipping special class files like module descriptor - here no classes will be
      // declared and class visitor throws error.
```

surapuramakhil: sorry. Updated it.

tyvsmith:
@kageiit @v-jizhang @surapuramakhil with Kotlin 1.4.0 artifacts kotlin-stdlib-jdk7.jar, kotlin-stdlib-jdk8.jar, and kotlin-stdlib.jar they contain "META-INF/versions/9/module-info.class" which causes the same issue, even with this fixed version of Buck. Probably because of the versioned dir.

src/com/facebook/buck/android/dalvik/DalvikStatsCache.java
```diff
@@ -32,10 +32,13 @@
   }
 
   DalvikStatsTool.Stats getStats(FileLike entry) {
-    String name = entry.getRelativePath();
-    if (!name.endsWith(".class")) {
+    String[] pathParts = entry.getRelativePath().split("/");
+    String name = pathParts[pathParts.length-1];
+    if (!name.endsWith(".class") || name.equals("module-info.class")) {
```
kageiit, line 37: to address https://github.com/facebook/buck/pull/2504#issuecomment-691348177 . this could be changed to 

```
name. endsWith("/module-info.class")
```

v-jizhang: That's weird, name is the last part of `"META-INF/versions/9/module-info.class"`, which should e exactly `"module-info.class"`.

PR #2509, opened 2020-07-30
Fixed a phthon installation bug on CircleCI

v-jizhang:
On CircleCI, upgrading pyenv breaks the cached python installation. fix it by re-installing python.


zpao:
Should we just purge the caches?

v-jizhang:
> Should we just purge the caches? Or do whatever it was that bumps the version on the cache?

I could not find a way to purge the caches except for bump version, but the CACHE_VERSION environment variable does not work for forks.

zpao:
Ah because the env vars aren't available to forks… that's unfortunate.

PR #2512, opened 2020-08-03
Make Python3 the first choice of python interpreter

v-jizhang:
Make Python3 the first choice of python interpreter


test/com/facebook/buck/cli/testdata/buck_run/pwdbase.py
```diff
@@ -2,6 +2,6 @@
 import sys
 
 cwdbase = os.path.basename(os.getcwd())
-print cwdbase
+print (cwdbase)
 sys.exit(0 if cwdbase == 'subdir' else 1)
```
v-jizhang, line 7: Python3 has changed its import precedence and using same name as a python module gives you an error "ImportError: No code object available for ...". Since `pwd` is a python module, we can no longer use the file name `pwd.py`.

zpao:
How does this compare with what's happening on dev branch?

v-jizhang:
> How does this compare with what's happening on dev branch?

Dev branch still uses python2 as default interpreter.

ajrpeggio:
Ya'll are awesome. Thank you for looking in to this!

test/com/facebook/buck/features/python/toolchain/impl/PythonInterpreterFromConfigTest.java
```diff
@@ -98,9 +98,9 @@ public void whenPythonPlusExtensionOnPathIsExecutableFileThenItIsUsed() throws I
   }
 
   @Test
-  public void whenPython2OnPathThenItIsUsed() throws IOException {
+  public void whenPython3OnPathThenItIsUsed() throws IOException {
```
zpao, line 101: I think you actually want both versions of this test. The intent seems to be supporting executables that aren't `python`

v-jizhang: I checked the 2 to 3 migration guideline, it highlighted "PYTHON 2 IS DEAD H2 2020." and "Tell linter to check Python 3 only". So we should support Python3 only.

PR #2518, opened 2020-08-08
Update SignedJarBuilder.java

isfaaghyth:
Fix `kotlinx-coroutines-android` services not included in final APK META-INF/services


v-jizhang:
@isfaaghyth Thank you for your contributions.
We had a fix in #2291 but it was reverted. You also closed the same fix in #2290. Do you know the reason? Thanks

PR #2152, opened 2019-01-08
Add full kotlin configurations to kotlin rule

ZacSweers:
These match the configurable JVM arguments in other build systems and better match naming conventions.

Note that much of this logic is borrowed from OkBuck, where we already extract these arguments from gradle configurations

CC @thalescm @kageiit @raviagarwal7

PR #2526, opened 2020-08-21
Add full kotlin configurations to kotlin rule (v2)

tyvsmith:
Reviving https://github.com/facebook/buck/pull/2152 and addressing previous feedback (adding a test). Was unable to push to that, because Zac's original fork is deleted.

> These match the configurable JVM arguments in other build systems and better match naming conventions.
> Note that much of this logic is borrowed from OkBuck, where we already extract these arguments from gradle configurations 

CC @ZacSweers @kageiit @raviagarwal7 @styurin


src/com/facebook/buck/jvm/kotlin/KotlincStep.java
```diff
@@ -168,13 +159,8 @@ public String getDescription(ExecutionContext context) {
...
     }
 
-    if (context.getVerbosity().shouldUseVerbosityFlagIfAvailable()
-        && !verboseModeOnlyExtraArguments.isEmpty()) {
-      for (String extraArgument : verboseModeOnlyExtraArguments) {
-        if (!extraArgument.isEmpty()) {
-          builder.add(extraArgument);
-        }
-      }
+    if (context.getVerbosity().shouldUseVerbosityFlagIfAvailable()) {
```
tyvsmith, line 162: Kotlin verbosity will outputted if either it's set via buck's Context, or if it's set in the Kotlin config. Felt a bit odd, but better than alternatives (not providing kotlin config version, ignoring buck verbosity, or making both required)

tyvsmith:
This will set Buck up to generate Kotlin Facets in .iml files. When hacked in, they seem to provide an Intellij performance benefit for large projects.

PR #2536, opened 2020-09-02
Skip flaky tests on Windows platform

v-jizhang:
Skip flaky tests on Windows platform


zpao:
Does it make sense to do this via the test selector functionality as opposed to in the test itself? https://github.com/facebook/buck/blob/cd9fbc819522796c7545e66dfb2ecbccdc8431e2/.circleci/config.yml#L1466-L1471

v-jizhang:
> Does it make sense to do this via the test selector functionality as opposed to in the test itself?
> 
> https://github.com/facebook/buck/blob/cd9fbc819522796c7545e66dfb2ecbccdc8431e2/.circleci/config.yml#L1466-L1471

Done. Thanks

zpao (approved):

PR #2540, opened 2020-09-11
Added a rust tutorial

v-jizhang:
Added a rust tutorial


zpao (approved):
Let's add the empty case back in now since it can happen (eg MacOS + Linux)

PR #2547, opened 2020-09-25
Add support for async worker processes

mikekap:
The idea is that you might be able to parallelize internally in the worker process, with the benefit of shared memory to cache details about dependencies. Async worker processes can receive & run multiple commands concurrently. They respond with results asynchronously in any order.

PR Notes:
 - The DSL arg is `solo_async=True` because `async` is a python keyword.
 - I'm sort of lukewarm on the idea of `ListenableFuture`s - open to suggestions. The only real benefit is that it makes testing concurrent execution/cancellation easier.
 - In theory the `Sync` worker pool can be cleaned up with the new interface (less of a need for a `BurrowedWorkerProcess` & lifecycle). I did not do this to keep this PR manageable.


v-jizhang:
@mikekap Please fix this test error: Check license of: src/com/facebook/buck/worker/WorkerProcessPool.java.
Thanks

mikekap:
@v-jizhang fixed the missing license issue.

v-jizhang:
@mikekap I discussed this PR with the Buck team and really for this huge change, it's hard for me to get it reviewed.
Here are some questions and concerns from the buck team:
Is this modularized (when without solo_async option, Buck never go through the code this person added)? 
Does this change really improve the performance and how much?
And Please open an issue and discuss with us before dropping a massive PR to the repo.

Thanks

mikekap:
Let me try to answer these. In terms of reviewing, it looks more daunting than it is because github doesn't have rename detection - `WorkerProcessPoolSync` is a renamed version of `WorkerProcessPool` with a 10 line method. The main logic changes that are here are in `WorkerProcess`. A lot of the other files are plumbing through a new boolean.

> Is this modularized (when without solo_async option, Buck never go through the code this person added)?
This PR changes the way `WorkerProcess` works internally so not entirely. Specifically - there is a thread to read the output of the worker process as opposed to assuming the output will be returned synchronously. There's some machinery to deal with the work of matching the worker response to the request and completing futures.

As built, `WorkerProcess` doesn't know if it's used with `async` at all - it's just a question of whether multiple jobs are submitted concurrently or not.

Beyond that - no code is shared between sync/async worker processes.

> Does this change really improve the performance and how much?
For us, it improves performance significantly. We compile Clojure via a JVM-based process. Clojure is Java based, but uses a lot of static initializers even after compilation, which makes dependency loading during compilation a bottleneck. This PR lets us use a single process to cache dependencies that we have already loaded and parallelize via regular futures in the worker process.

> And Please open an issue and discuss with us before dropping a massive PR to the repo.
Sorry, my intent wasn't to burden. Apologies. We are running this in our CI - so it helps in our use case. Unfortunately my previous work around adding persistent processes didn't yield as much gains for our use case. I'm happy to restructure this as needed - not tied to the current implementation at all. I'm hoping to get this upstreamed so various `genrule` changes and the like don't require constant upkeep. Would you like me to file an issue to get the conversation started or should we continue the thread here?

v-jizhang:
> Let me try to answer these. In terms of reviewing, it looks more daunting than it is because github doesn't have rename detection - `WorkerProcessPoolSync` is a renamed version of `WorkerProcessPool` with a 10 line method. The main logic changes that are here are in `WorkerProcess`. A lot of the other files are plumbing through a new boolean.
> 
> > Is this modularized (when without solo_async option, Buck never go through the code this person added)?
> > This PR changes the way `WorkerProcess` works internally so not entirely. Specifically - there is a thread to read the output of the worker process as opposed to assuming the output will be returned synchronously. There's some machinery to deal with the work of matching the worker response to the request and completing futures.
> 
> As built, `WorkerProcess` doesn't know if it's used with `async` at all - it's just a question of whether multiple jobs are submitted concurrently or not.
> 
> Beyond that - no code is shared between sync/async worker processes.
> 
> > Does this change really improve the performance and how much?
> 
> For us, it improves performance significantly. We compile Clojure via a JVM-based process. Clojure is Java based, but uses a lot of static initializers even after compilation, which makes dependency loading during compilation a bottleneck. This PR lets us use a single process to cache dependencies that we have already loaded and parallelize via regular futures in the worker process.
> 
> > And Please open an issue and discuss with us before dropping a massive PR to the repo.
> 
> Sorry, my intent wasn't to burden. Apologies. We are running this in our CI - so it helps in our use case. Unfortunately my previous work around adding persistent processes didn't yield as much gains for our use case. I'm happy to restructure this as needed - not tied to the current implementation at all. I'm hoping to get this upstreamed so various `genrule` changes and the like don't require constant upkeep. Would you like me to file an issue to get the conversation started or should we continue the thread here?

No apologies, my previous comment is just some thoughts from buck team. 
No need to file an issue, we can post any conversations in this PR and I'll get this reviewed. Thank you very much.

src/com/facebook/buck/worker/WorkerProcess.java
```diff
@@ -45,8 +55,14 @@
...
   private final AtomicInteger currentMessageID = new AtomicInteger();
   private boolean handshakePerformed = false;
+  private final ConcurrentHashMap<Integer, SettableFuture<Integer>> commandExitCodes =
+      new ConcurrentHashMap<>();
   @Nullable private WorkerProcessProtocol.CommandSender protocol;
   @Nullable private ProcessExecutor.LaunchedProcess launchedProcess;
+  private final Thread readerThread;
+  private final CountDownLatch readerShutdownSignal = new CountDownLatch(1);
+  private final Thread watchdogThread;
+  private final CountDownLatch watchdogShutdownSignal = new CountDownLatch(1);
```
v-jizhang, line 65: The variables readerShutdownSignal and watchdogShutdownSignal are just flags, should we make them volatile boolean? That way we will have 2 less objects.

src/com/facebook/buck/worker/WorkerProcess.java
```diff
@@ -117,31 +143,60 @@ public synchronized WorkerJobResult submitAndWaitForJob(String jobArgs) throws I
...
+        MoreExecutors.directExecutor());
   }
 
   @Override
   public synchronized void close() {
     LOG.debug("Closing process %d", this.hashCode());
     try {
+      readerShutdownSignal.countDown();
+      synchronized (commandExitCodes) {
+        commandExitCodes.notifyAll();
```
v-jizhang, line 194: Why notifyAll? we have only one thread in waiting.

src/com/facebook/buck/worker/WorkerProcess.java
```diff
@@ -166,11 +221,91 @@ public synchronized void close() {
...
+  }
+
+  private void processNextCommandResponse() throws Throwable {
+    Preconditions.checkState(
+        protocol != null,
+        "Tried to submit a job to the worker process before the handshake was performed.");
+
+    WorkerProcessProtocol.CommandResponse commandResponse =
+        this.protocol.receiveNextCommandResponse();
+    SettableFuture<Integer> result = commandExitCodes.remove(commandResponse.getCommandId());
```
v-jizhang, line 244: This method is called in thread readerLoop, a different thread from submitJob, and both thread access commandExitCodes. Should we protect it?

src/com/facebook/buck/worker/WorkerProcessPoolSync.java
```diff
@@ -0,0 +1,244 @@
...
+    return Optional.ofNullable(availableWorkers.poll(timeout, unit))
+        .map(BorrowedWorkerProcess::new);
+  }
+
+  @Override
+  public synchronized void close() {
+    Throwable caughtWhileClosing = null;
+
+    // remove all available workers
+    int numAvailableWorkers = availableWorkers.drainTo(new ArrayList<>(capacity));
```
v-jizhang, line 110: What is use of numAvailableWorkers?

src/com/facebook/buck/worker/WorkerProcess.java
mikekap: Not quite - the latches are `.await`-able. Realistically these objects probably aren't memory-intensive for buck - there's a `WorkerProcess` object per long-lived process. You would rarely have more than ncpus of these processes running -- though I guess it depends how many worker tools you use!

src/com/facebook/buck/worker/WorkerProcess.java
mikekap: `commandExitCodes` is a `ConcurrentHashMap` so there's no need. Though honestly the threading here is a bit complex because of the watchdog. I updated the implementation to get rid of the watchdog (it was mostly for paranoia) and add a volatile boolean for the shutdown signal as you suggested. Also added a test for worker death.

The locking is still a bit more complex than I'd prefer, so I'm open to suggestions here - there's a weird internal vs external locking of data - externally `this` is used to lock `submitJob` & `close`, but the reader thread uses `readerThread` to be woken up when there is something to do.

src/com/facebook/buck/worker/WorkerProcess.java
```diff
@@ -221,90 +213,88 @@ public synchronized void close() {
...
 
-  private void watchdogLoop() {
-    try {
-      while (!watchdogShutdownSignal.await(5, TimeUnit.SECONDS)) {
-        if (!readerThread.isAlive()) {
-          failAllFutures();
-        }
+      if (!readerAlive) {
+        failAllFutures();
+        continue;
```
v-jizhang, line 269: Once readerAlive becomes false, it stays false forever. Should it be set to true here?

v-jizhang: I'm confused with the readerAlive logic. Could you explain?

src/com/facebook/buck/worker/WorkerProcess.java
```diff
@@ -170,7 +217,85 @@ public synchronized void close() {
...
+    readerThread.start();
+  }
+
+  private void processNextCommandResponse() throws Throwable {
+    Preconditions.checkState(
+        protocol != null,
+        "Tried to submit a job to the worker process before the handshake was performed.");
+
+    Preconditions.checkState(
+        !Thread.holdsLock(this),
```
v-jizhang, line 232: This can't be false, why need we check it?

src/com/facebook/buck/worker/WorkerProcess.java
v-jizhang: I thin that's Ok.

src/com/facebook/buck/worker/WorkerProcess.java
v-jizhang: Thanks you

src/com/facebook/buck/worker/WorkerProcess.java
mikekap: Yep that's the intent. `readerAlive` is set to `false` only if the worker returns a malformed response. In this context, malformed responses are signs something terrible happened - even build step failures are well formed responses. This could only happen on e.g. malformed json or if the process just died. In that case, we stop attempting to communicate with the process at all and just fail all incoming requests until the build eventually `close()`s the process.

src/com/facebook/buck/worker/WorkerProcess.java
mikekap: I think as written, this is definitely true. This is mostly guarding to future changes in the code - if e.g. the caller is `synchronized` by accident.

src/com/facebook/buck/worker/WorkerProcess.java
v-jizhang: I c, thank you.

v-jizhang:
lgtm. Thanks

v-jizhang (approved):

PR #2549, opened 2020-09-29
Fix Appstore Connect publication

pomozoff:
There should not be a **Framewrok** directory inside an app plugin to publish the IPA file to the Appstore Connect:

> ERROR ITMS-90206: "Invalid Bundle. The bundle at 'XXX.app/PlugIns/NotificationService.appex' contains disallowed file 'Frameworks'."

### DIFF
--- a/src/com/facebook/buck/features/apple/project/ProjectGenerator.java
+++ b/src/com/facebook/buck/features/apple/project/ProjectGenerator.java
@@ -2351,7 +2351,6 @@ public class ProjectGenerator {
                         bundleExtension -> {
                           switch (bundleExtension) {
                             case APP:
-                            case APPEX:
                             case PLUGIN:
                             case BUNDLE:
                             case XCTEST:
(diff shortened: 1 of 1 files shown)


src/com/facebook/buck/features/apple/project/ProjectGenerator.java
```diff
@@ -2362,6 +2361,7 @@ private boolean shouldEmbedSwiftRuntimeInBundleTarget(
                               // a Swift runtime, so it must get bundled to ensure they run.
                               return true;
 
+                            case APPEX:
```
v-jizhang, line 2364: This essentially prevent buck from setting ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES to YES. And it does more things other than just not creating the Frameworks directory. I don't feel very confident about this PR. Do we have any other solutions? Thanks
A similar issue: https://github.com/CocoaPods/CocoaPods/issues/4203

v-jizhang:
This PR is closed because no updates for 2 weeks.

PR #2552, opened 2020-10-09
[Apple] Add support to AppClips

Lcsmarcal:
Context:
In iOS 14 Apple launched AppClips, which is like a quick view from a specific flow from a whole app.
AppClips are a normal application that follows the normal application bundle structure and are shipped embedded in the main application bundle at the "AppClips/" folder.


v-jizhang (approved):
lgtm, Thanks

Lcsmarcal:
Nice!

Lcsmarcal:
Could you add the "hacktoberfest-accepted" tag at this PR? :)

PR #2564, opened 2020-11-17
[Apple] Fix Remove copy standard swift libraries step phase from AppClips bundle.

Lcsmarcal:
Since AppClips only runs on iOS 14+ devices, those swift std libraries are not needed and it increases AppClips size to 10.8MB which is more than the AppClips size limit.


Lcsmarcal:
is anyone there?

rajyengi:
Sorry for the delay on this @Lcsmarcal - looks like tests pass and this change seems fine. I'll go ahead and merge

PR #2590, opened 2021-02-22
[1/n Dynamic Feature Support] Create Android Optimiser layer and segregate responsibility of AndroidBinaryBuildable and AndroidBinaryOptimizer

uKetki:
The commit addresses the below changes:
- Creating a light AndroidBinaryBuildable by extracting optimizer flow from buildable
- Refactored Android Buildable layer and segregated responsibility of AndroidBinaryBuildable
- Defining AndroidBinaryPathUtility class with all possible paths for aab and apk
- Adding BinaryType enum to segregate apk and aab flow
---
Looking at the buck code for generating apk and aab, it is evident that `AndroidBinaryBuildable` is bloated with a lot of responsibilities ranging from apk/aab generation, signing to optimization. 

The overview of current buck architecture:
Existing Binary generation logic

The suggestion is to refactor existing code and create two separate layers: Optimizers and Buildable which can re-use the common logic for binary generation and offload the specific implementation to the respective layers: 
Buildable
Optimizer

---
This approach will later help to add changes needed to support dynamic feature delivery.


uKetki:
`windows_tests` seem unrelated.

PR #2595, opened 2021-03-04
Support NDK r21e and r22

R1kk3r:
This PR add support for NDK r21e and r22. The NDK r22 comes with some changes related to include paths. The target parameter need to contain the platform level version in order for Clang to be able to locate the includes.


rajyengi:
@R1kk3r looks like a lot of the CI is still broken, they look like tests related to your change.

PR #2596, opened 2021-03-09
Fixing Kotlin ABI issues

tyvsmith:
This addresses two bugs when dealing with ABI classes for Kotlin.

1. If you're using a mixed source set or a kt target without any kt files, the folder isn't generated to hold the abi class files. When an upstream target tries to consume this target, it fails because of a missing folder. This creates the temp folder if it's supporting source abi generation.

2. For Class ABIs, methods with synthetic parameters were stripped. The means that overrides become problematic across targets. This preserves more signatures in the case of Kotlin by keeping synthetic methods for Kotlin.

This has been tested and deployed on the internal Uber Buck fork.

### DIFF
--- a/src/com/facebook/buck/jvm/kotlin/KotlinConfiguredCompilerFactory.java
+++ b/src/com/facebook/buck/jvm/kotlin/KotlinConfiguredCompilerFactory.java
@@ -87,7 +87,7 @@ public class KotlinConfiguredCompilerFactory extends ConfiguredCompilerFactory {
       ToolchainProvider toolchainProvider) {
     CoreArg kotlinArgs = Objects.requireNonNull((CoreArg) args);
     Path pathToAbiGenerationPluginJar =
-        shouldGenerateSourceAbi() ? kotlinBuckConfig.getPathToAbiGenerationPluginJar() : null;
+        shouldGenerateSourceAbi(kotlinArgs, kotlinBuckConfig) ? kotlinBuckConfig.getPathToAbiGenerationPluginJar() : null;
     return new KotlincToJarStepFactory(
         kotlinBuckConfig.getKotlinc(),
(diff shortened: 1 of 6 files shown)


IanChilds:
Thanks for doing this! Can you explain a bit more about why (2) is required? What is different here that means that we need these for Kotlin but not for Java?

tyvsmith:
> Thanks for doing this! Can you explain a bit more about why (2) is required? What is different here that means that we need these for Kotlin but not for Java?

Kotlin makes different use of synthetic methods than Java. In Kotlin, the use of default parameters on methods causes multiple synthetic method signatures to be generated in bytecode.  This was prominent for any external class using referencing those methods via a direct reference or an override. For Java, nothing needs to change and removing this restriction would just be extra noise in the jars and default the original intention of filtering those out.

IanChilds:
Sorry for the delay, my github notifications are not set up properly I guess :( 

Thanks for the explanation, makes sense!

src/com/facebook/buck/jvm/kotlin/KotlinConfiguredCompilerFactory.java
```diff
@@ -142,6 +142,10 @@ public boolean sourceAbiCopiesFromLibraryTargetOutput() {
     return true;
   }
 
+  private static boolean shouldGenerateSourceAbi(CoreArg kotlinArgs, KotlinBuckConfig kotlinBuckConfig) {
+    return kotlinArgs.getAbiGenerationMode().orElse(kotlinBuckConfig.getAbiGenerationMode()).isSourceAbi();
```
IanChilds, line 146: Why are we doing this? Should the default really be `source`?

IanChilds (approved):

KapJI:
Thanks!

PR #2600, opened 2021-03-29
[apple] Stop propagating underlying libraries when using resources_from_deps.

yanks:
We observed that the `resources_from_deps` would propagate linkables for the transitive dependencies into a resultant binary. With this change, `resources_from_deps` will only include resources, not the actual linkable results.


KapJI:
For some reason CircleCI jobs haven't been triggered. Can you try to push something to make them run?

KapJI:
@yanks can you please follow these steps to make CI jobs running?

https://github.com/facebook/buck/pull/2611#issuecomment-819784681

PR #2604, opened 2021-03-30
[apple]: expose apple.use_short_names_for_targets_in_xcode.

yanks:
This is a convenience feature -- we prefer seeing the entire target name as xcode groups, so this small change exposes it as a configuration setting.


KapJI:
For some reason the CI is not triggered on your PRs. Maybe this is because of your email in commit not linked to your github account.

PR #2605, opened 2021-04-02
[apple] Add "umbrella" suffix to default umbrella header name

alekseevpg:
In our project we have some both internal and external dependencies in Obj-C with manually created "umbrella" files. They are interfere with BUCK generated "ModuleName.h" umbrella files. One example of such dependency is https://github.com/yapstudios/YapDatabase/blob/master/YapDatabase/YapDatabase.h

This PR will add *ModuleName*"-umbrella.h" suffix to avoid this issue. This behaviour will be similar to CocoaPods


KapJI:
Looks like this breaks tests.

PR #2611, opened 2021-04-13
[docs] Fix instructions for Java install with Homebrew 

narlei:
The current doc use `homebrew/cask-cask` instructions, and cask was changed.

The current instructions to install java with Homebrew doesn't work, **and mysteriously it requests your github email and password 🤔🏴‍☠️.**
This PR change the docs to install java using https://github.com/AdoptOpenJDK/homebrew-openjdk with correct cask command in homebrew.

PR #2616, opened 2021-04-28
Add friend path ABI jars as required deps of Kotlin libraries

### LINKED ISSUE #2614: Kotlin library ABI jar sometimes not available for friend_paths config
In some cases, the Kotlin ABI jar [will not be available](https://github.com/facebook/buck/blob/master/src/com/facebook/buck/jvm/kotlin/KotlinConfiguredCompilerFactory.java#L160) in order to be used to configure the friend_paths.

If the ABI jar target exists (which it always does for Kotlin), then shouldn't we just add a required rule so that we can wait for it to be generated instead of falling back to the original non-ABI source jar?

jbarr21:
In some cases, the Kotlin ABI jar for a dependency may not be already generated in the task graph, so when it is used as a friend path, it does not get picked up. This ensures that if a module is used as a friend path, that we add an explicit dependency on calculating its ABI jar so that we can use its absolute path in the friend path Kotlin compiler arg.

Resolves #2614

### DIFF
--- a/src/com/facebook/buck/jvm/java/DefaultJavaLibraryRules.java
+++ b/src/com/facebook/buck/jvm/java/DefaultJavaLibraryRules.java
@@ -563,6 +563,7 @@ public abstract class DefaultJavaLibraryRules {
             getArgs(),
             getJavacOptions(),
             getActionGraphBuilder(),
+            getActionGraphBuilder(),
             getInitialBuildTarget().getTargetConfiguration(),
             getToolchainProvider());
   }
@@ -573,6 +574,7 @@ public abstract class DefaultJavaLibraryRules {
         .configure(
             getArgs(),
             getJavacOptionsForSourceOnlyAbi(),
+            getActionGraphBuilder(),
             getActionGraphBuilder(),
             getInitialBuildTarget().getTargetConfiguration(),
             getToolchainProvider());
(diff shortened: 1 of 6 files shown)


KapJI:
Please follow instruction in https://github.com/facebook/buck/pull/2611#issuecomment-819784681 to enable CircleCI testing.

src/com/facebook/buck/jvm/java/DefaultJavaLibraryRules.java
```diff
@@ -563,8 +563,8 @@ CompileToJarStepFactory getConfiguredCompiler() {
             getArgs(),
             getJavacOptions(),
             getActionGraphBuilder(),
-            getInitialBuildTarget().getTargetConfiguration(),
-            getToolchainProvider());
+            getActionGraphBuilder(),
+            getInitialBuildTarget().getTargetConfiguration(), getToolchainProvider());
```
KapJI, line 567:
```suggestion
            getInitialBuildTarget().getTargetConfiguration(),
            getToolchainProvider());
```

src/com/facebook/buck/jvm/java/DefaultJavaLibraryRules.java
```diff
@@ -574,8 +574,8 @@ CompileToJarStepFactory getConfiguredCompilerForSourceOnlyAbi() {
             getArgs(),
             getJavacOptionsForSourceOnlyAbi(),
             getActionGraphBuilder(),
-            getInitialBuildTarget().getTargetConfiguration(),
-            getToolchainProvider());
+            getActionGraphBuilder(),
+            getInitialBuildTarget().getTargetConfiguration(), getToolchainProvider());
```
KapJI, line 578:
```suggestion
            getInitialBuildTarget().getTargetConfiguration(),
            getToolchainProvider());
```

KapJI:
Thanks! 🎉

PR #2622, opened 2021-05-18
Fix building with Xcode 12.5

pomozoff:
1. Use the correct path for searching swift libraries
2. Update compatibility library to 51


KapJI (approved):
Thanks!

Issue #2623, opened 2021-05-19
Original test class name is lost when running a JUnit test suite


swarren12:
When running `buck test` with the `--xml` argument, the resulting XML file contains the name of the test class that was run in a `<test>` element and the name of each individual method within that class in child `<testresult>` elements. However, when running a JUnit test suite, the name of the test class is replaced with the name of the test suite in the `<test>` element. This can make it difficult (/impossible) to differentiate tests that are in different classes but have the same method name.

**Expected behaviour**: The original test class name should be recorded when using a test suite
**Actual behaviour**: The original test class name is replaced by the name of the test suite
**Steps to reproduce**: Run `buck test --xml` on a JUnit test suite

For example, given a contrived example with the following test classes / methods: 

- `TestCase1` => `test1()`, `test2()`, `test3()` 
- `TestCase2` => `test1()`, `test2()`, `test3()` 
- `TestCase3` => `test1()`, `test2()`, `test3()` 
- `TestCase4` => `testA()`

... running these as a suite might generate an XML such as [the one attached](https://github.com/facebook/buck/files/6506981/output.xml.txt). As you can see, the report generated by buck makes it impossible to test which version of `test1()`, `test2()` and `test3()` failed. It isn't even clear if they all came from the same test class, although one might assume that they did given that they are together.

The [original JUnit XML](https://github.com/facebook/buck/files/6507005/tests.JavaTestSuite.xml.txt) that was parsed in order to create the `output.xml` file does contain the both the suite name and the class name, so I don't see any particular reason why the class name shouldn't make its way across to the buck-generated file also?

However, I'm not sure what the ultimate desired behaviour should be when using a test suite - logically I would expect it to mimic the JUnit output of nesting the test classes under the test suite, but I don't know if this structure lends itself well to the other types of tests that buck supports. The alternatives would be to either:

1. drop the suite information completely and just parse the JUnit XML as if it was the individual tests that had run (this would be my preference) _or_
2. use an approach similar to the JUnit XML where the original class name is included as an extra attribute on the tag (less desirable to me as that's a change to the XML schema)

I have knocked together a change that implements option 1 and I'm happy to raise this as a PR if there is agreement that using a suite should work in this way. Alternatively, happy to look at making a different change if there's some consensus that it should work in a different way.

KapJI:
Option 1 should be enough, please submit a PR.

Issue #2401, opened 2020-02-24
[FB] Moving Buck to a stable/development branch publication model


gnuthor:
In order to allow us to experiment on fundamental changes to the core of Buck, future changes made by Facebook’s Build Infra team to the Buck code base will be published to a development branch. The current master branch is moving to a stable role that doesn’t automatically get these changes.

For this stable master branch we will continue to merge PRs and help with issue tickets as we do now, we are also open to pulling interesting commits from the dev branch on demand.

We will make sure that changes to the new dev branch continue to pass Buck’s test suites inside FB’s internal CI system while we change features and APIs, potentially breaking backwards compatibility. Our goal is to be able to put the buck dev code base into a state that allows easier reuse and evolution of interesting parts quickly without putting non-FB users of buck at risk that might be dependent on some of these code paths.

Longer term this ability to move fast and experiment will lead to a vastly improved build experience. We're exploring changes that will make for a more dynamic, more incremental, and much, much faster build system. We will clearly separate the core from the language rules providing the community a more stable API to depend on.

The Facebook Build Infra Team

shepting:
What are the plans for pushing the development changes back to `master`? Will that happen on a semi-regular schedule? Just for major changes when they are done? Or is that even part of the plan?

Daij-Djan:
how do PRs make it into the DEV Branch

gnuthor:
We only plan to cherry-pick changes from the development branch to master. We expect the changes in development to break existing setups, so merging them back to master too soon would be harmful. It is likely the development branch will eventually become a future version of Buck. We welcome patches to both development and master, but encourage users to stick with master for now.

popeye8777:
> What are the plans for pushing the development changes back to `master`? Will that happen on a semi-regular schedule? Just for major changes when they are done? Or is that even part of the plan?

I don't know I'm know to these codes

dulmandakh:
There is an update https://developers.facebook.com/blog/post/2021/07/01/future-of-buck/

vultix:
@dulmandakh  Is there any news on the buck rewrite? I'm having trouble finding an update. Last I can find is the blog post saying this: 

> We are planning to have this publicly available by January 2022

cjhopman:
The buck rewrite has been available for a bit now at https://github.com/facebook/buck2

We've just publicly announced it today, you can read more here: https://engineering.fb.com/2023/04/06/open-source/buck2-open-source-large-scale-build-system/

[...]

PR #2624, opened 2021-05-19
Use original JUnit test class name instead of suite class in test XML report

swarren12:
As discussed in https://github.com/facebook/buck/issues/2623, this change will use the original test class name instead of the test suite name when generating the test output XML when run against a JUnit suite. I've based this change on the `dev` branch as it doesn't seem the kind of thing that needs to hit the stable branch urgently.

I've tried to keep the change as small as possible by overloading the `TestCaseSummary` constructor to default the new `testSuite` parameter to `false`, so that all the other usages of this aren't affected, but if it's better to push this default up to the places where this class is instantiated then I can do that instead!

**One thing to note:** I originally built this change on top of `master`, but figured that it was probably better suited to `dev`. It all seemed to work for me on the stable build, however since rebasing the change onto the `dev` branch I can no longer build the project? I don't know if this is because I'm trying to build using the latest Buck release, but it gives me an error parsing `buck/programs/defs.bzl` (`name 'name' is not defined`).

### DIFF
--- a/src/com/facebook/buck/jvm/java/JavaTest.java
+++ b/src/com/facebook/buck/jvm/java/JavaTest.java
@@ -85,6 +85,7 @@ import java.util.Set;
 import java.util.SortedSet;
 import java.util.concurrent.Callable;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


KapJI:
Sorry, `dev` is our internal branch, I don't think we accept PRs there. Can you please submit it for `master`?

swarren12:
Oh, sure. That makes my life a bit easier as well as I can actually build on that branch!

(FWIW, https://github.com/facebook/buck/issues/2401 says "We welcome patches to both development and master, but encourage users to stick with master for now." so I assumed either was fine)

PR #2658, opened 2021-09-24
CI: remove old NDK 15, 16, 17, 18, 19

dulmandakh:


dulmandakh:
@KapJI could you please review and help with macos_test_integration.

cortinico:
I don't think this is strictly needed, but I'll leave the decision to @rajyengi on how to proceed.

dulmandakh:
These NDK versions are deprecated a long time ago, and no longer supported. Main concern is it's making CI slow.

cortinico:
Agree that those versions are old. NDK 17 is still used internally though I believe.

> Main concern is it's making CI slow.

All the NDK versions tests are executed in parallel so I don't believe is a major problem in terms of CI speed. The concern is more on maintenance costs

PR #2672, opened 2021-12-15
Update log4j to 2.15

scottcao:
Summary: Updates log4j to avoid a 0day RCE vulnerability. See https://www.lunasec.io/docs/blog/log4j-zero-day/

Reviewed By: suitingtseng, mykola-semko, rajyengi

fbshipit-source-id: 71a46dad938206cdd34139298e71ea096f84965e


shepting:
I think that 2.2.17 is out now to fix an infinite recursion DoS. We likely would want that.

PR #2680, opened 2022-02-09
Don't use `parser` module if `symbol` doesn't exist as well.

aadsm:
Python 3.10 removed the `parser` and the `symbol` modules that were deprecated for a while now.
There's a fallback implementation with markerlib for `parser` but we were not checking for the lack of the `symbol` module as well.


bigfootjon:
This has been superseded by https://github.com/facebook/buck/pull/2697 / https://github.com/facebook/buck/pull/2696

Issue #2678, opened 2022-01-29
Python 3.10:  third-party python modules need update (setuptools: No module named 'symbol', pex: cannot import name 'iterable' from 'collections')


rpuntaie:
I just wanted to install Buck on Archlinux and got the mentioned error.
See also https://stackoverflow.com/questions/70836444/no-module-named-symbol

Buck has its own setuptools and would need to do without "symbol" to support systems with Python 3.10:
https://github.com/facebook/buck/blob/0e72e40074e1731dd3c9068e5ed23ee10bcc75dd/third-party/py/setuptools/pkg_resources/__init__.py#L32

Dgarc359:
+1 experiencing this issue as well

aadsm:
This pr might fix it, I didn't actually tried it though, I only read the code: https://github.com/facebook/buck/pull/2680/commits/1e57b5c4ba1812807a90c7b15bcacb9e10fe9ef5

rpuntaie:
It fixes "No module named 'symbol'", but then
```
third-party/py/pex/pex/base.py", line 6
ImportError: cannot import name 'Iterable' from 'collections'
```

I made the description a bit more general.

bigfootjon:
I think the real solution might be to upgrade the vendored PEX dep

jc65536:
> It fixes "No module named 'symbol'", but then
> 
> ```
> third-party/py/pex/pex/base.py", line 6
> ImportError: cannot import name 'Iterable' from 'collections'
> ```
> 
> I made the title a bit more general.

Could be related to this issue: https://github.com/Azure/azure-cli/issues/20444#issuecomment-1009648464

bigfootjon:
> > It fixes "No module named 'symbol'", but then
> > 
> > ```
> > third-party/py/pex/pex/base.py", line 6
> > ImportError: cannot import name 'Iterable' from 'collections'
> > ```
> > 
> > I made the title a bit more general.
> 
> Could be related to this issue: https://github.com/Azure/azure-cli/issues/20444#issuecomment-1009648464

I believe `collections` no longer exports `Iterable` in Python3.10 right? 

Ref: https://docs.python.org/3/whatsnew/3.10.html which says:

[...]

PR #2696, opened 2022-04-27
Get Buck working for 3.10

lisroach:
Summary:
On 3.10 some `collections` classes have been moved to
`collections.abc` and it causes Buck 1 to crash. This diff fixes those
instances.

Really pex should be upgraded, but I made an attempt and it was very difficult
due to our custom changes to pex.

Fixes: https://github.com/facebook/buck/issues/2678

### DIFF
--- a/scripts/slice_trace.py
+++ b/scripts/slice_trace.py
@@ -16,7 +16,6 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import argparse
 import codecs
-import collections
 import math
 import os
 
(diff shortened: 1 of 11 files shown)


bigfootjon (approved):
For some reason the `dev` branch (synced from Meta) won't build for me, so I cherry-picked this PR onto `main` and built that.

It works for me! I can now `buck run` my python3.10-only program!

I'll approve the linked Diff on Monday/after you publish it. Thanks for fixing this!

PR #2721, opened 2022-08-09
Update sqlite to include ability to build buck on M1 chips. Currently, BUCK fails because sqlite-jdbc-3.20.0 doesn't support aarch64 but 3.39.2.0 does.

Solace-Studios:
Update sqlite to include ability to build buck on M1 chips. Currently, BUCK fails because sqlite-jdbc-3.20.0 doesn't support aarch64 but 3.39.2.0 does.


Solace-Studios (commented):
LFG!

